PR #23, opened 2020-03-05
Add Retip page

davidyuk:
Callbacks names inspired by http://x-callback-url.com/specifications/


petbaik:
How this Retip page can be accessed?

davidyuk:
Only through deep links:
```
superhero:retip?id=26
https://mobile.z52da5wt.xyz/retip?id=26 (after merging of #28)
```
it is needed for aggregator

PR #27, opened 2020-03-06
Trim spaces in seed and claiming .chain name

### LINKED ISSUE #26: remove space after seed and on claiming .chain name
Space after seed phrase still makes it invalid. This can be a bit of a scare for users later on if they have a lot of tokens and suddenly cannot retrieve their account because of a space.

petbaik:
Fixes #26

### DIFF
--- a/src/popup/router/pages/ImportAccount.vue
+++ b/src/popup/router/pages/ImportAccount.vue
@@ -36,6 +36,7 @@ export default {
     async importAccount() {
       if (this.mnemonic) {
         this.loading = true;
+        this.mnemonic = this.mnemonic.trim();
         const mnemonic = this.mnemonic.split(' ');
         if (mnemonic.length >= 12 && mnemonic.length <= 24 && validateMnemonic(this.mnemonic)) {
           this.errorMsg = null;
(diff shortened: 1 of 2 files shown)


src/popup/router/pages/ImportAccount.vue
```diff
@@ -36,6 +36,7 @@ export default {
     async importAccount() {
       if (this.mnemonic) {
         this.loading = true;
+        this.mnemonic = this.mnemonic.trim();
```
davidyuk: In the Base app, we have a bit more complex formatting: https://github.com/aeternity/aepp-base/blob/3fb6a7c3264aab8ad22c1ae65685910858951e9b/src/components/AeInputMnemonic.vue#L23-L27
It handles additional cases:
- the first capital letter (actual for mobile)
- multiple spaces between words
- removing of not allowed symbols

src/popup/router/pages/ImportAccount.vue
```diff
@@ -36,6 +36,7 @@ export default {
...
         const mnemonic = this.mnemonic.split(' ');
```
davidyuk: Ok, I see that this solution works with multiple spaces in the same way.

PR #47, opened 2020-03-13
Remove import which is causing build error

petbaik:


mradkov:
Is this used anywhere?

petbaik:
> Is this used anywhere?

The problem is only in this file, because there was conflicts yesterday in this file and it seems I did't resolve them properly https://github.com/aeternity/superhero-wallet/pull/45/commits/a31a03529b209a6164b0915fa771002509da8dee

PR #64, opened 2020-03-19
Feature connect different node

### LINKED ISSUE #60: add the option to connect to different middleware/Fullnode server

petbaik:
Fixes #60


src/popup/router/components/CheckBox.vue
```diff
@@ -28,6 +25,9 @@ export default {
         this.checkedProxy = val;
       },
     },
+    getType() {
+      return !this.type ? 'checkbox' : this.type;
```
davidyuk: Other option would be to set a default value for `type` property: https://vuejs.org/v2/guide/components-props.html#Prop-Validation

PR #65, opened 2020-03-19
Explicit assets import

davidyuk:
`browser.runtime.getURL` can't be used in css, so I proposing to avoid it by using relative paths (they works both in extensions and Cordova).

Importing of svg by default as Vue components is not works in css, so I proposing to enable importing as a component explicitly by adding `?vue-component`.

Also, these changes is needed for web version of the wallet, because there we should avoid the `popup` folder in `dist` and use absolute paths.

### DIFF
--- a/src/popup/router/components/TransactionItem.vue
+++ b/src/popup/router/components/TransactionItem.vue
@@ -10,7 +10,9 @@
       </div>
       <div class="holder">
         <span class="url" @click="visitTipUrl">{{ tipUrl }}</span>
-        <span class="seeTransaction" @click="seeTx(transactionData.hash)"><img :src="eye"/></span>
+        <span class="seeTransaction" @click="seeTx(transactionData.hash)">
+          <img src="../../../icons/eye.png" />
+        </span>
       </div>
     </ae-list-item>
   </div>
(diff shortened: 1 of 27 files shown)


src/popup/router/components/TransactionItem.vue
```diff
@@ -10,7 +10,9 @@
       </div>
       <div class="holder">
         <span class="url" @click="visitTipUrl">{{ tipUrl }}</span>
-        <span class="seeTransaction" @click="seeTx(transactionData.hash)"><img :src="eye"/></span>
+        <span class="seeTransaction" @click="seeTx(transactionData.hash)">
+          <img src="../../../icons/eye.png" />
```
mradkov: Shouldn't this also be imported as the rest of the resources?

src/popup/router/pages/Notifications.vue
```diff
@@ -2,7 +2,7 @@
   <div class="popup">
     <ae-list class="noti-list">
       <ae-list-item fill="neutral" v-for="(noti, i) in notifications" :key="i" class="noti">
-        <img :src="icon" />
+        <img src="../../../icons/icon_48.png" />
```
mradkov: Same question goes here?

> 

Shouldn't this also be imported as the rest of the resources?

src/popup/router/components/TransactionItem.vue
davidyuk: Vue-loader handles src attribute of img well, I don't see reasons to don't use that 🤷‍♀️, it is just shorter

mradkov:
Can you resolve conflicts and merge?

PR #69, opened 2020-03-19
travis: Combine into single job to speedup testing

davidyuk:
Actually, it doesn't make a significant difference (~2 minutes faster). Just Travis won't be needed to run `npm ci` three times.


davidyuk:
Or the idea is that jobs can run in parallel?

PR #82, opened 2020-03-23
Network page test cases

petbaik:


src/popup/router/pages/Networks.vue
```diff
@@ -105,6 +97,8 @@ export default {
     async addNetwork() {
       try {
         this.network.error = false;
+        if (!this.network.name) throw new Error('Enter network name');
+        if (this.network.name === DEFAULT_NETWORK) throw new Error('Network with this name exist');
```
davidyuk: Shouldn't we go through all network names to check that? Also, why it is important to have all names unique?

tests/e2e/integration/networks.js
```diff
@@ -0,0 +1,104 @@
+
+import { defaultNetworks, networks, DEFAULT_NETWORK } from '../../../src/popup/utils/constants'
+const network = "Testnet";
```
davidyuk: I propose to make e2e test files comply with the current linter settings.

src/popup/router/pages/Networks.vue
petbaik: It's checked for all networks here https://github.com/aeternity/superhero-wallet/blob/6cbb47f55c31db492a0365caa7683ce6e4d3dd4b/src/popup/router/pages/Networks.vue#L109. 
We need unique names because of the way networks object is currently structured 
https://github.com/aeternity/superhero-wallet/blob/master/src/popup/utils/constants.js#L88

davidyuk: Why don't combine these two checks?

I understood the structure that you are using. In the Base app we were storing user-defined networks in an [array](https://github.com/aeternity/aepp-base/blob/3ebf172a25bcfe2444cd037583a33e130b27b4ac/src/store/modules/root.js#L17), it was working fine, no additional checks were needed🤷‍♀️

tests/e2e/integration/networks.js
petbaik: I will run linter on `networks.js` after merging https://github.com/aeternity/superhero-wallet/pull/83

PR #95, opened 2020-03-25
Refactor for getting the right active network

vmangelovv:


src/notifications.js
```diff
@@ -13,16 +13,18 @@ export default class Notification {
...
+    this.network = (await getNodes()).network;
     setInterval(() => {
       this.checkTxReady();
     }, 2000);
     browser.notifications.onButtonClicked.addListener(id => {
       browser.tabs.create({ url: id.split('?')[1], active: true });
     });
   }
-
+
```
petbaik: I think this will break in linter check

PR #101, opened 2020-03-27
Fix failing tests

petbaik:


mradkov:
I guess there is problem with the whole testing setup ATM, and solely increasing the timeout won't do the job this time.

A new enclosed test env is in the process and we'll update the travis setup to test against it asap.

PR #105, opened 2020-03-30
Use Number.isNaN instead of isNaN, enable "no-restricted-globals" eslint rule

davidyuk:
Airbnb's eslint disallows `window.isNaN` because of its confusing behavior: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#Description

### DIFF
--- a/src/popup/router/components/Input.vue
+++ b/src/popup/router/components/Input.vue
@@ -20,11 +20,7 @@ export default {
   created() {},
   watch: {
     value(val) {
-      if ((this.type === 'number' && isNaN(val)) || parseFloat(val) === 0) {
-        this.err = true;
-      } else {
-        this.err = false;
-      }
+      this.err = this.type === 'number' && Number.isNaN(+val);
     },
   },
   computed: {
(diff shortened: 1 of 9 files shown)


src/popup/router/components/Input.vue
```diff
@@ -20,11 +20,7 @@ export default {
   created() {},
   watch: {
     value(val) {
-      if ((this.type === 'number' && isNaN(val)) || parseFloat(val) === 0) {
```
davidyuk: I don't see reasons to check is `parseFloat(val) === 0` if input type is not `number`.

tests/e2e/integration/tip.js
```diff
@@ -40,7 +40,7 @@ describe('Test cases for tip page', () => {
       .enterTipDetails({ amount: tip.amount })
       .buttonShouldBeDisabled('[data-cy=send-tip]')
       .enterTipDetails({ note: tip.note })
-      .buttonShouldNotBeDisabled('[data-cy=send-tip]')
+      .buttonShouldBeDisabled('[data-cy=send-tip]')
```
davidyuk: I think it should be disabled because user before was changing amount and left it invalid (empty) 🤷‍♀️

mradkov:
Can you resolve the conflicts?

PR #106, opened 2020-03-30
Avoid polling of reactive variables

davidyuk:


mradkov:
Can you resolve conflicts?

PR #108, opened 2020-03-31
Sync browser storage with vuex

petbaik:
I'm not sure if I covered all cases, but tests are passing (except send tip). I think we should test it and when we are sure that everything is ok then merge it.


vmangelovv:
On first open after build, it throws errors:
image
Оtherwise everything looks good.

mradkov:
@petbaik can you resolve conflicts?

petbaik:
@davidyuk Can you check if `persistState` plugin is integrated properly

mradkov:
@petbaik can you resolve conflicts?

src/store/plugins/persistState.js
```diff
@@ -1,7 +1,9 @@
+import { cloneDeep } from 'lodash-es';
+
 const KEY = 'state';
 
 const setState = async state => {
-  browser.storage.local.set({ [KEY]: state });
+  browser.storage.local.set({ [KEY]: cloneDeep(state) });
```
davidyuk, line 6: what was the problem there?

petbaik: There was problem with saving the state in Firefox and `cloneDeep` fixed the problem. Not sure what was the exact error.

davidyuk: Putting to local storage is already a kind of cloning, this part is super unobvious.

PR #110, opened 2020-03-31
Fix phishing check for local services and fix empty phishing blank page

### LINKED ISSUE #107: Wallet phishing check blocks running local service (i.e. Plex)
popular home streaming Plex (https://www.plex.tv/) is running a local web server `https://127.0.0.1:32400` which is not accessible while wallet extension is active, probably because phishing check does not redirect properly. There is a blank page.

petbaik:
Fixes #107


venimus:
Seems ok now. A block screen appears, and the link "continue" works. Though, the link for "see the list of domains" did not work for me.

PR #136, opened 2020-04-06
Tour and UI polishing

petbaik:


src/popup/router/components/Header.vue
```diff
@@ -95,16 +92,13 @@ export default {
       }
 
       .title {
-        position: absolute;
-        left: 0;
-        right: 0;
-        text-align: center;
```
davidyuk, line 101: Actually, it makes header misaligned

PR #133, opened 2020-04-03
Twitter bio parser

petbaik:

PR #138, opened 2020-04-06
Check if IS_EXTENSION when send payfortx on send tip

petbaik:


src/popup/router/pages/SuccessTip.vue
```diff
@@ -70,10 +70,11 @@ export default {
     },
   },
   async created() {
-    const { addresses, tab } = await this.$store.dispatch('getWebPageAddresses');
-    console.log(addresses);
-    if (addresses.length) {
-      await axios.post(`${TIP_SERVICE}`, { url: tab.url, address: addresses[0] });
+    if (process.env.IS_EXTENSION) {
+      const { addresses, tab } = await this.$store.dispatch('getWebPageAddresses');
```
petbaik, line 74: If there is no check here error will be thrown on mobile, because there is no content script on mobile. Not sure what has to happen on mobile? cc @mradkov

mradkov: On mobile we should still invoke the payfortx when sending tip.

mradkov:
Actually I'm not sure we should invoke this on mobile, as the tipping there is done either manually or via deeplink and we can not get and pass the `address` (if there is one) from the URL for the payfortx claim.

petbaik:
> Actually I'm not sure we should invoke this on mobile, as the tipping there is done either manually or via deeplink and we can not get and pass the `address` (if there is one) from the URL for the payfortx claim.

If I pass my public key won't the same thing happen as here https://github.com/aeternity/superhero-wallet/pull/133, the service will search for address in the page for us like the twitter profile ?

mradkov:
> > Actually I'm not sure we should invoke this on mobile, as the tipping there is done either manually or via deeplink and we can not get and pass the `address` (if there is one) from the URL for the payfortx claim.
> 
> If I pass my public key won't the same thing happen as here #133, the service will search for address in the page for us like the twitter profile ?

Yes, but also passing your address does not make much difference (just the oracles are looking for it on the page with a higher priority), as the oracles will find another address on the page if not yours and the payfortx will invoke the payment on their behalf.

mradkov:
Putting this on hodl-to be merged after a new endpoint is implemented in the backend - for requesting claims without explicitly providing address.

mradkov:
Closing this for now. We can reopen later when its unblocked.

PR #147, opened 2020-04-08
Add identicon as default avatar

petbaik:
I think backend endpoint for serving avatars should be changed either to return default identicon when no avatar, or return response by which extension can decide whether to use identicon

### DIFF
--- a/src/popup/router/components/UserAvatar.vue
+++ b/src/popup/router/components/UserAvatar.vue
@@ -5,12 +5,7 @@
 
 <script>
 import jdenticon from 'jdenticon';
-import { BACKEND_URL } from '../../utils/constants';
-
-const identiconSizes = {
-  normal: 38,
-  lg: 64,
-};
+import { BACKEND_URL, IDENTICON_CONFIG, IDENTICON_SIZES } from '../../utils/constants';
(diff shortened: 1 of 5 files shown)


src/popup/router/components/UserAvatar.vue
```diff
@@ -1,31 +1,66 @@
...
   },
+  data: () => ({
+    error: false,
+  }),
   computed: {
     avatarUrl() {
       return `${BACKEND_URL}/profile/image/${this.address}`;
     },
+    identicon() {
+      jdenticon.config = {
```
mradkov, line 31: Can we extract the config in the constants file?

mradkov:
> I think backend endpoint for serving avatars should be changed either to return default identicon when no avatar, or return response by which extension can decide whether to use identicon

Yes, I agree.

mradkov:
> > I think backend endpoint for serving avatars should be changed either to return default identicon when no avatar, or return response by which extension can decide whether to use identicon
> 
> Yes, I agree.

This is now implemented - the backend does not serve default avatar now, and displays 404 for non existent user avatars - so we can display the jdenticon.

PR #154, opened 2020-04-09
Apply new styles for tour start button

petbaik:
Ref: #148


src/popup/router/components/RecentTransactions.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="recent-transactions">
+  <div class="recent-transactions" v-bind:class="{ 'tour-bar': tourStartBar }">
```
davidyuk, line 2:
```suggestion
  <div class="recent-transactions" :class="{ 'tour-bar': tourStartBar }">
```

src/popup/router/components/Tour.vue
```diff
@@ -202,56 +224,51 @@ export default {
...
+}
 .v-step {
-  background-color: #12121b !important;
+  background-color: $tour-bg-color !important;
   border-radius: 5px !important;
   border: 1px solid $secondary-color;
   padding: 20px 15px 25px 15px !important;
-
   min-width: 345px;
-
```
davidyuk, line 212: For me personally to have an empty line before selector looks more readable

src/store/getters.js
```diff
@@ -117,4 +117,7 @@ export default {
   tourRunning({ tourRunning }) {
     return tourRunning;
   },
+  tourStartBar({ tourStartBar }) {
+    return tourStartBar;
+  },
```
davidyuk, line 122: I know this kind of agreement in this repo, but getters that just wrapping a single state variable have no sense

petbaik: Shoud I use directly $store.getters

davidyuk: I propose to use `$store.state.tourStartBar` instead of wrapping it (`$store.getters. tourStartBar`)
e.g. map it to components using `mapState`

petbaik: Yeah my mistake I wanted to say `$store.state` not `$storage.getters`

PR #168, opened 2020-04-15
Replace tipping contract with aens name

petbaik:


src/manifest.js
```diff
@@ -17,6 +17,7 @@ module.exports = (isProd, browser) => ({
     'activeTab',
     'clipboardWrite',
     'notifications',
+    'tabs',
```
mradkov, line 20: Do we need this `tabs` permission?

petbaik: If removed we cannot get active tab url in background script where we are making payfortx request

mradkov: > What activeTab allows
```
While the activeTab permission is enabled for a tab, an extension can:

    Call tabs.executeScript or tabs.insertCSS on that tab.
    Get the URL, title, and favicon for that tab via an API that returns a tabs.Tab object (essentially, activeTab grants the tabs permission temporarily).
    Intercept network requests in the tab to the tab's main frame origin using the webRequest API. The extension temporarily gets host permissions for the tab's main frame origin. 
```

Ref: https://developer.chrome.com/extensions/activeTab#what-activeTab-allows

mradkov: So, eventually we still need this `tabs` permissions because of the background script.

mradkov:
Fix the contract version check as we do not need to compare it after this PR is merged.

Screenshot 2020-04-16 at 14 53 31

PR #180, opened 2020-04-23
Allow tips with value > 0.01 USD

### LINKED ISSUE #179: Allow tips with value > 0.01 USD

petbaik:
Fixes #179


mradkov:
Looks good. We should add some text to explain the user that they need to provide more than 0.01 USD in tokens in order to tip.

Screenshot 2020-04-23 at 18 06 45
Screenshot 2020-04-23 at 18 06 55

PR #197, opened 2020-04-27
Fix activity history tx in TopUps section

petbaik:


mradkov:
I still dont see anything in the TopUps section of the activity history? 🤔 

Screenshot 2020-04-27 at 13 42 40

Issue #203, opened 2020-04-27
Remove change.ly


mradkov:
Remove `changelly` from the TopUp page.

mradkov:
ref. #225

PR #214, opened 2020-04-28
Set active AEname, add Buy AE button in Top-Up page

vmangelovv:


mradkov:
Looks good. We should think of how to sync it with the aggregator.

src/popup/router/pages/Names.vue
```diff
@@ -47,6 +50,12 @@
                 :class="name.addPointer ? 'danger' : ''"
                 >{{ $t('pages.namingSystemPage.pointer') }}</Button
               >
+              <Button
+                :small="!name.addPointer"
+                @click="setActiveName(name, key)"
+                :class="activeAccountName == name.name ? 'active disabled' : ''"
```
petbaik, line 56:
```suggestion
                :disabled="activeAccountName == name.name"
```

src/store/actions.js
```diff
@@ -95,13 +99,7 @@ export default {
           names = flatten(names);
           names = uniqBy(names, 'name');
           if (!process.env.RUNNING_IN_TESTS) {
-            if (names.length) {
-              commit(types.SET_ACCOUNT_AENS, {
-                account: index,
-                aename: names[0].name,
-                pending: !!names[0].pending,
-              });
-            } else {
```
petbaik, line 104: Maybe we should leave this, because when imported my account and have registered .chain names I saw message `Claim your .chain name`

mradkov: > Maybe we should leave this, because when imported my account and have registered .chain names I saw message `Claim your .chain name`

Agree.

vmangelovv: This code does not work properly. It always takes the first name. 
I suggest if the user has registered name, the `Claim your .chain name` message to be changed to `Set your active name` or something like that. Or we can make the user to activate his name after registration.

mradkov: > This code does not work properly. It always takes the first name.
> I suggest if the user has registered name, the `Claim your .chain name` message to be changed to `Set your active name` or something like that. Or we can make the user to activate his name after registration.

This is also a good proposal. Please do it.

src/popup/router/components/AccountInfo.vue
```diff
@@ -21,18 +26,24 @@
...
+    availNames: false,
   }),
   computed: {
     ...mapGetters(['account', 'activeAccountName']),
     chainName() {
       return this.activeAccountName.includes('.chain');
     },
   },
+  async created() {
+    const { names } = await getState();
```
davidyuk, line 44: I think `getState` is a too low-level method, can we get the same from `this.$store.state`?

vmangelovv: You are right

src/popup/router/pages/Receive.vue
```diff
@@ -35,7 +28,7 @@ export default {
   data() {
     return {
       jellySwapUrl: 'https://app.jelly.market',
-      changellyUrl: 'https://changelly.com/buy',
+      buyAe: '',
```
mradkov, line 31:
```suggestion
      buyAe: 'https://shop.aeternityuniverse.com',
```

mradkov: This will resolve #203

PR #216, opened 2020-04-28
Add ClaimTips page

### LINKED ISSUE #199: Claim tip on mobile and webwallet
Add new view for the claim tip where user can put the URL they want to claim manually in an input field and send a `claim` request.

davidyuk:
closes #199
I did not find any designs, so I decided to make it as simple as possible:


venimus:
I'll ask designers to prepare a design
Can you also fix that text in the scope of this: 
`No zero amount claims`
It should be something like "No new tips to claim" or "Nothing to claim yet"

mradkov:
> I'll ask designers to prepare a design
> Can you also fix that text in the scope of this:
> `No zero amount claims`
> It should be something like "No new tips to claim" or "Nothing to claim yet"

This can be fixed in a separate PR, when adjusting texts. Lets not hold the functionality further just because of text string.

PR #217, opened 2020-04-28
Do not allow tips to blacklisted websites

### LINKED ISSUE #210: Do not allow tips to blacklisted websites
https://raendom-backend.z52da5wt.xyz/static/wallet/graylist
Ref: https://github.com/aeternity/tipping-community-backend/pull/95

petbaik:
* Fixes #210
* New statuses for tip urls
* Confirm when tip unsuported url
* UrlStatus component
* Icons for url status
* Url status modal info refactoring
* Add new colors for body, navigation, inputs according to new design resources


src/popup/router/pages/TipPage.vue
```diff
@@ -256,45 +241,71 @@ export default {
 };
 </script>
 
-<style lang="scss" scoped>
+<style lang="scss">
```
davidyuk, line 244: Can you use `/deep/` instead of this to point what selector changes styles of child components? In general, I would suggest avoiding changing styles of nested components.

src/mixins/tipUrls.js
```diff
@@ -0,0 +1,41 @@
...
+    verifiedUrls: [],
+    blacklistedUrls: [],
+  }),
+  async created() {
+    try {
+      this.verifiedUrls = (await axios.get(`${BACKEND_URL}/verified`)).data;
+      this.blacklistedUrls = (await axios.get(`${BACKEND_URL}/static/wallet/graylist`)).data;
+    } catch (e) {
+      this.$logError({ e, action: 'fetch-verified' });
+    }
```
davidyuk, line 16:
```suggestion
    const [{ data: verified }, { data: graylist }] = await Promise.all([
      axios.get(`${BACKEND_URL}/verified`),
      axios.get(`${BACKEND_URL}/static/wallet/graylist`),
    ]);
    this.verifiedUrls = verified;
    this.blacklistedUrls = graylist;
```
I propose to:
- make requests simultaneously
- drop specific error logging because anyway it should be caught by global handlers

src/mixins/tipUrls.js
```diff
@@ -0,0 +1,41 @@
+import axios from 'axios';
+import { BACKEND_URL } from '../popup/utils/constants';
+import { getTwitterAccountUrl } from '../popup/utils/helper';
+
+export default {
+  data: () => ({
+    verifiedUrls: [],
+    blacklistedUrls: [],
```
davidyuk, line 8: No need to have them reactive because they are used only in `getStatus` method

src/mixins/tipUrls.js
```diff
@@ -0,0 +1,41 @@
...
+  async created() {
+    try {
+      this.verifiedUrls = (await axios.get(`${BACKEND_URL}/verified`)).data;
+      this.blacklistedUrls = (await axios.get(`${BACKEND_URL}/static/wallet/graylist`)).data;
+    } catch (e) {
+      this.$logError({ e, action: 'fetch-verified' });
+    }
+  },
+  methods: {
+    getStatus(tipUrl) {
```
davidyuk, line 19: Can we move this method to helpers with variables defined outside of method for storing fetched lists? Like:
```js
let verifiedUrls;
let blacklistedUrls;

export async function getStatus(tipUrl) {
  if (!verifiedUrls || !blacklistedUrls) {
    const [{ data: verified }, { data: graylist }] = await Promise.all([
      axios.get(`${BACKEND_URL}/verified`),
      axios.get(`${BACKEND_URL}/static/wallet/graylist`),
    ]);
    verifiedUrls = verified;
    blacklistedUrls = graylist;
  }
  // ...
}
```
To use helper function seems clearer that to use mixin.

src/popup/router/components/Modals/TipBadgeModal.vue
```diff
@@ -1,58 +1,40 @@
...
   <Modal @close="resolve">
-    <div class="badge-modal-header" slot="header">
-      <UrlBadge :type="verifiedStatus" />
-    </div>
+    <template slot="header">
+      <template v-if="status === 'verified'">{{ $t('modals.verified.title') }}</template>
+      <template v-else-if="status === 'not-supported'">{{
+        $t('modals.not-supported.title')
+      }}</template>
+      <template v-else-if="status === 'blacklisted'">{{ $t('modals.blacklisted.title') }}</template>
```
davidyuk, line 8:
```suggestion
      {{ content.title }}      
```

src/popup/router/components/Modals/TipBadgeModal.vue
```diff
@@ -1,58 +1,40 @@
...
-      {{
-        verifiedStatus == 'verified'
-          ? $t('pages.tipPage.verifiedUrlInfo')
-          : $t('pages.tipPage.notVerifiedUrlInfo')
-      }}
+      <template v-if="status === 'verified'">{{ $t('modals.verified.msg') }}</template>
+      <template v-else-if="status === 'not-supported'">{{
+        $t('modals.not-supported.msg')
+      }}</template>
+      <template v-else-if="status === 'blacklisted'">{{ $t('modals.blacklisted.msg') }}</template>
```
davidyuk, line 15:
```suggestion
      {{ content.msg }}
```

src/popup/router/components/Modals/TipBadgeModal.vue
```diff
@@ -1,58 +1,40 @@
...
-import UrlBadge from '../UrlBadge';
 
 export default {
   props: {
     resolve: { type: Function, required: true },
-    verifiedStatus: { type: String, required: true },
+    status: { type: String, required: true },
   },
-  components: { Modal, Button, UrlBadge },
+  components: { Modal, Button },
```
davidyuk, line 32:
```suggestion
  components: { Modal, Button },
  computed: {
    content() {
      switch (this.status) {
      case 'verified': return this.$i18n('modals.verified');
      case 'not-supported': return this.$i18n('modals.not-supported');
      case 'blacklisted': return this.$i18n('modals.blacklisted');
      default: throw new Error(`Unknown url status: ${this.status}`);
      }
    },
  },
```

src/popup/router/components/UrlStatus.vue
```diff
@@ -0,0 +1,31 @@
...
+    type: {
+      type: String,
+      required: true,
+    },
+  },
+  components: {
+    Verified,
+    Blacklisted,
+    NotSupported,
+  },
```
davidyuk, line 23:
```suggestion
  computed: {
    Icon() {
      switch(this.type) {
      case 'verified': return Verified;
      case 'blacklisted': return Blacklisted;
      case 'not-supported': return NotSupported;
      default: throw new Error(`Unknown url status: ${this.type}`);
      }
    },
  },
```

src/popup/router/components/UrlStatus.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <Blacklisted v-if="type === 'blacklisted'" class="url-status" />
+  <Verified v-else-if="type === 'verified'" class="url-status" />
+  <NotSupported v-else-if="type === 'not-supported'" class="url-status" />
```
davidyuk, line 4:
```suggestion
  <Component :is="Icon" class="url-status" />
```
Otherwise, this will get messy after adding a click handler.

src/mixins/tipUrls.js
davidyuk: Actually, I'm not sure that it is better to move it to helpers because to use async getStatus is trickier than the current sync version.

davidyuk: Can we make a vuex plugin to handle this? Like we did in [Base aepp](https://github.com/aeternity/aepp-base/blob/1acbb5c448111dacddedc2efe035b754a78ce82f/src/store/plugins/ui/appsMetadata.js). So, to have a getter in vuex that will return some default value immediately and then replace it (because of getter reactivity) with actual value after fetching of corresponding data.

venimus:
I guess you see the recently updated design (that include Input) instead of "edit" button

petbaik:
> I guess you see the recently updated design (that include Input) instead of "edit" button

Yes I saw them

PR #231, opened 2020-04-30
Refactor logging

davidyuk:
I have not reviewed #208 in time, but I see some issues there:
- would be better to use global error handling in vue components than explicitly call $logError
- not necessary to preserve action name, because it is easily can be extracted from stack trace + map files like in this tutorial: https://medium.com/taskworld-tech/tracking-down-an-error-from-a-minified-production-stack-trace-7a7c7b812d55
- reduce the amount of Logger.write calls
- proposing to save platform as it is

and some other minor changes.

### DIFF
--- a/src/lib/background-utils.js
+++ b/src/lib/background-utils.js
@@ -61,7 +61,7 @@ export const getSDK = async () => {
         compilerUrl: network.compilerUrl,
       });
     } catch (e) {
-      Logger.write({ e, action: 'init-sdk-background' });
+      Logger.write(e);
     }
   }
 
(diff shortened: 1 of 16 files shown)


src/lib/background-utils.js
```diff
@@ -96,40 +96,22 @@ export const getTippingContractAddress = async address => {
...
+    });
+    if (call) return call;
+    const error = new Error('Contract call failed');
+    error.payload = { tx };
+    throw error;
+  }
+  if (!controller.isLoggedIn() && typeof callType !== 'undefined' && callType === 'static') {
+    throw new Error('You need to unlock the wallet first');
+  }
+  throw new Error('No data to return');
```
davidyuk, line 116: Actually, I'm not sure about this case, it was not defined in the previous version.

src/lib/logger.js
```diff
@@ -48,23 +46,16 @@ export default class Logger {
...
-    const { e = {}, ...err } = error;
-    const log = {
-      error: {
-        stacktrace: e.stack,
-        msg: e.message,
-        name: e.name,
-        ...err,
-      },
+    const logEntry = {
+      ...pick(error, ['name', ...Object.getOwnPropertyNames(error)]),
```
davidyuk, line 52: Inspired by https://stackoverflow.com/a/27525073/6176994

src/popup/router/pages/SuccessTip.vue
```diff
@@ -77,18 +77,12 @@ export default {
...
-    }
+    this.verifiedUrls = (
+      await axios.get(`${BACKEND_URL}/verified`).catch(() => ({ data: [] }))
+    ).data;
     const { addresses, tab } = await this.$store.dispatch('getWebPageAddresses');
     if (addresses.length) {
-      try {
-        await axios.post(`${TIP_SERVICE}`, { url: tab.url, address: addresses[0] });
-      } catch (e) {
-        this.$logError({ e, url: tab.url, action: 'claim' });
```
davidyuk, line 90: I assume that it is not necessary to handle error here because of the global axios handler in `src/lib/request.js`.

petbaik: I think `src/lib/request.js` currently is not used. My idea was to to replace `axios.` with `request.` and this way use the error handler but I was not sure if we can extract the action name this way, so I left it

src/lib/background-utils.js
petbaik: I think it's ok.

PR #265, opened 2020-05-05
Handle tip widget click and open popup

petbaik:
* Remove lsroute in Tip page when tip is from popup
* If extension installed set click event of tip link to open tip popup
* Send message form content to backround for tip popup
* Extract method in backgoround script for opening tip popup for both context tip and tip widget


mradkov:
@petbaik is this ready for review?

PR #266, opened 2020-05-05
Speed up Redirect/no redirect tests

petbaik:


tests/e2e/integration/other.js
```diff
@@ -24,12 +24,9 @@ describe('Tests cases not connected to specific page', () => {
     { path: '/transactions', redirect: true },
   ].forEach(({ path, redirect }) => {
     it(`${redirect ? '' : 'no '}redirect to last visited route ${path}`, () => {
-      cy.login()
-        .wait(2000)
-        .visit(`chrome/popup/popup#${path}`)
-        .wait(2000)
+      cy.login({}, path)
+        .wait(500)
```
davidyuk, line 28: Would be great to use some condition instead: https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting But these changes are also good.

PR #277, opened 2020-05-08
Don't log errors when claim

petbaik:


mradkov:
I think we should still log errors, but only if they are not present in the list of known messages received from the contract.

davidyuk:
I"m completely agree with Milen. In the Base app we were doing it like this: https://github.com/aeternity/aepp-base/blob/de884d8bb97053319793cf862b06eea6739ed981/src/pages/mobile/NameNew.vue#L89-L91

PR #279, opened 2020-05-08
Sync active name with aggregator

petbaik:


src/store/actions.js
```diff
@@ -64,12 +65,27 @@ export default {
...
+        const signedChallenge = Buffer.from(
+          await state.sdk.signMessage(response.challenge),
+        ).toString('hex');
+        const respondChallenge = {
+          challenge: response.challenge,
+          signature: signedChallenge,
+        };
+        await Backend.sendProfileData(respondChallenge);
+      } catch (e) {
+        console.warn(`Cannot set preferred name ${e}`);
```
mradkov, line 86: Maybe add error popup here?

davidyuk: Isn't it a rare case? Maybe it would be better to make a global error popup? (to show it after any uncatched exception)

davidyuk: This warning can be used only by developers, I don't think that it is more useful than the usual exception stack trace. So, I proposing to drop try/catch at all.

PR #312, opened 2020-05-20
Expose current address through deeplink

davidyuk:
The template syntax is inspired by https://en.wikipedia.org/wiki/URL_Template

### DIFF
--- a/src/popup/locales/en.json
+++ b/src/popup/locales/en.json
@@ -169,7 +169,8 @@
       "networks": "Networks",
       "comment-new": "Add comment",
       "home": "Superhero Home",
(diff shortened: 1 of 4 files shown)


mradkov:
Lets rename this to `login`. As this is the intended behavior of the page - sign a login request and expose user's address via the deeplink.

davidyuk:
I propose to keep it as it is, because:
- users won't use these deeplinks directly, so their names shouldn't be obvious for them
- sdk has a method with the same name that does the same thing
- `/login` deeplink may be used later if we introduce some more sophisticated process with signing and so

PR #316, opened 2020-05-21
connection: Remove unnecessary try/catch

davidyuk:
As I know, it works in the same way under the hood

### DIFF
--- a/src/popup/utils/connection.js
+++ b/src/popup/utils/connection.js
@@ -47,11 +47,7 @@ export const ensureContentScriptInitialized = async () => {
     });
     contentPostMessage = async (message, tab) => {
       const id = genUuid();
-      try {
-        await browser.tabs.sendMessage(tab, { data: { ...message, uuid: id } });
-      } catch (e) {
-        return Promise.resolve(e);
-      }
+      await browser.tabs.sendMessage(tab, { data: { ...message, uuid: id } });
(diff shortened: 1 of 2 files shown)


src/popup/utils/connection.js
```diff
@@ -47,11 +47,7 @@ export const ensureContentScriptInitialized = async () => {
     });
     contentPostMessage = async (message, tab) => {
       const id = genUuid();
-      try {
-        await browser.tabs.sendMessage(tab, { data: { ...message, uuid: id } });
-      } catch (e) {
-        return Promise.resolve(e);
```
davidyuk, line 53: Well, I did not noticed `resolve` instead of `reject`. @petbaik Is it intentional to return an exception as a result? If the answer is yes then let's close this PR, but I think there should be a more obvious solution.

petbaik: Actually it's not necessary to return exception just promise need to be resolved in case of exception so code can continue without errors https://github.com/aeternity/superhero-wallet/blob/develop/src/store/actions.js#L265

davidyuk: Can we catch this error there instead?

petbaik: I think we can

davidyuk: Well, I did that, can we merge it now?

PR #344, opened 2020-06-04
Fix initial network switch not triggering onNetworkChange

petbaik:


kenodressel:
Not fixed for me. Still the same issue. I should note that I am running the code with the SDK 7.4.1 because my boilerplate will throw other errors if I don't.

mradkov:
> Not fixed for me. Still the same issue. I should note that I am running the code with the SDK 7.4.1 because my boilerplate will throw other errors if I don't.

Where can this be reproduced?

kenodressel:
I had a talk with peter and it looks like the issues I had were due to me not building the new extensions with the fixes properly. I will check again soon.

PR #347, opened 2020-06-05
Adjust selector of superhero button to changes

petbaik:
Ref: https://github.com/aeternity/superhero-button/pull/10


mradkov:
Is this backwards compatible?

petbaik:
> Is this backwards compatible? Would previously integrated buttons work? cc @petbaik

Actually no. Will make it backwards compatible now.

PR #349, opened 2020-06-05
Refactoring and improvements in names page

petbaik:


src/popup/router/pages/Names/AuctionList.vue
```diff
@@ -0,0 +1,74 @@
...
+    topBlockHeight: 0,
+  }),
+  computed: {
+    ...mapGetters(['sdk']),
+    auctions() {
+      if (this.filter === 'soonest') return this.activeAuctions;
+      if (this.filter === 'length')
+        return this.activeAuctions.map(a => a).sort((a, b) => a.name.length - b.name.length);
+      if (this.filter === 'bid')
+        return this.activeAuctions.map(a => a).sort((a, b) => a.winning_bid - b.winning_bid);
```
davidyuk, line 54: It looks like a good place for `switch` statement

src/store/plugins/names.js
```diff
@@ -0,0 +1,137 @@
...
+      async setDefault({ rootState: { sdk }, commit, dispatch }, { name, address, modal = true }) {
+        commit('setDefault', { name, address, networkId: sdk.getNetworkId() });
+        try {
+          const response = await Backend.sendProfileData({
+            author: address,
+            preferredChainName: name,
+          });
+          console.log(response.challenge);
+          console.log(await sdk.signMessage('test'));
+          console.log('here');
```
davidyuk, line 112: 🤷‍♀️

src/popup/router/components/Button.vue
```diff
@@ -21,6 +21,7 @@ export default {
     onboarding: Boolean,
     dark: Boolean,
     inline: Boolean,
+    inactive: Boolean,
```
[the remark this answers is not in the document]
petbaik: `disabled` will set `pointers-event: none` and button will become unclickable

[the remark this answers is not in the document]
davidyuk: yes, but why `inactive` button should be clickable?

src/popup/router/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -71,23 +74,41 @@ import { aettosToAe } from '../../../utils/helper';
...
+      'pointers',
+    ],
+  }),
   computed: {
     ...mapGetters(['account', 'activeAccountName']),
     receiver() {
       return this.transaction.contractId || this.transaction.recipientId || '';
     },
     showAddress() {
       return (
```
davidyuk, line 107:
```suggestion
      return [
```

src/popup/router/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -71,23 +74,41 @@ import { aettosToAe } from '../../../utils/helper';
...
-        this.txType ===
-        (TX_TYPE.contractCreate || TX_TYPE.namePreClaim || TX_TYPE.nameClaim || TX_TYPE.nameUpdate)
+        this.txType === TX_TYPE.contractCreate ||
+        this.txType === TX_TYPE.namePreClaim ||
+        this.txType === TX_TYPE.nameClaim ||
+        this.txType === TX_TYPE.nameUpdate
       );
```
davidyuk, line 112:
```suggestion
        TX_TYPE.contractCreate,
        TX_TYPE.namePreClaim,
        TX_TYPE.nameClaim,
        TX_TYPE.nameUpdate,
      ].includes(this.txType)
```

src/popup/router/components/Button.vue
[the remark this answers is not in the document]
petbaik: yes you are right, the word `inactive` is inappropriate here. Maybe it should be `opacity` prop with value `fade` or something like that or just apply the desired opacity in the component where the button is used?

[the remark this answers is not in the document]
davidyuk: I saw its usage in `NameListHeader.vue` and understand the purpose. I'm not sure about the naming, it is ok to keep it as it is.

src/popup/router/components/NameListHeader.vue
```diff
@@ -0,0 +1,34 @@
...
+
+<style lang="scss" scoped>
+.name-list-header {
+  display: flex;
+  margin-top: 20px;
+  margin-bottom: 30px;
+  justify-content: space-between;
+
+  a {
+    line-height: 36px !important;
```
davidyuk, line 31: Can you make the selector more specific to avoid usage of `important`?

src/store/index.js
```diff
@@ -9,6 +9,7 @@ import modals from './plugins/modals';
 import tipUrl from './plugins/tipUrl';
 import accounts from './plugins/account';
 import tokens from './plugins/tokens';
+import namesModule from './plugins/names';
```
davidyuk, line 12:
```suggestion
import names from './plugins/names';
```
actually, it is a plugin, not a module

src/store/plugins/names.js
```diff
@@ -0,0 +1,134 @@
...
+                { name: 'default', title: 'Backend error', msg: e.message },
+                { root: true },
+              );
+            else throw e;
+          }
+        }
+      },
+      async getHeight({ rootState: { sdk } }) {
+        return (await sdk.topBlock()).height;
+      },
```
davidyuk, line 132: It doesn't seem to be related to this plugin.

src/store/index.js
```diff
@@ -20,8 +21,6 @@ const initialState = {
   account: {},
   mnemonic: null,
   activeAccount: 0,
-  names: [],
```
davidyuk, line 23: Don't we need to create a migration to don't break the state of existing users?

src/popup/router/pages/Names/List.vue
```diff
@@ -0,0 +1,56 @@
...
+  },
+  created() {
+    this.$store.dispatch('names/fetchOwned');
+    const id = setInterval(() => this.$store.dispatch('names/fetchOwned'), 10000);
+    this.$once('hook:destroyed', () => clearInterval(id));
+  },
+  methods: {
+    address(name) {
+      return getAddressByNameEntry(name);
+    },
```
davidyuk, line 47:
```suggestion
    address: getAddressByNameEntry,
```

src/popup/router/pages/Names/AuctionBid.vue
```diff
@@ -0,0 +1,114 @@
...
+        {{ $t('pages.names.auctions.current-heighest-bid') }}
+      </span>
+      <span> {{ highestBid }} {{ $t('pages.appVUE.aeid') }} </span>
+    </div>
+    <div class="details-row">
+      <span>
+        {{ $t('pages.names.auctions.remaining-time') }}
+      </span>
+      <span>
+        {{ blocksToRelativeTime(expiration - topBlockHeight) }}
```
davidyuk, line 25: Can we use it as a filter, like:
```suggestion
        {{ expiration - topBlockHeight | blocksToRelativeTime }}
```
?

src/store/index.js
petbaik: Only migration for default name will be needed i think. Names will be fetched as soon as application is started

src/popup/router/pages/Names/List.vue
petbaik: Actually this method was not used at all. Removing it

src/store/migrations/update-default-name.js
```diff
@@ -0,0 +1,22 @@
...
+      } = store;
+      const networks = {
+        ...network,
+        ...state.userNetworks.reduce((p, n) => ({ ...p, [n.name]: { ...n } }), {}),
+      };
+
+      const { networkId } = networks[state.current.network];
+      setTimeout(() => {
+        state.subaccounts.forEach(({ aename, publicKey }) => {
+          if (aename) commit('names/setDefault', { address: publicKey, name: aename, networkId });
```
davidyuk: Can't this be done manually? To make this migration sync and to don't depend on vuex commit that can be changed later.

src/store/migrations/update-default-name.js
```diff
@@ -0,0 +1,22 @@
+export default {
+  migrate(state, store) {
+    return new Promise(async resolve => {
```
davidyuk, line 3:
```suggestion
  async migrate(state, store) {
```
but actually, I think this migration should run in sync

src/store/plugins/names.js
```diff
@@ -25,8 +25,9 @@ export default store =>
       set(state, names) {
         state.owned = names;
       },
-      setDefault({ defaults }, { address, networkId, name }) {
-        Vue.set(defaults, `${address}-${networkId}`, name);
+      setDefault(state, { address, networkId, name }) {
+        const nameDefault = { [`${address}-${networkId}`]: name };
+        state.defaults = { ...(state.defaults || {}), ...nameDefault };
```
davidyuk, line 29: the previous version was clearer, why do we need this change?

src/store/migrations/update-default-name.js
```diff
@@ -0,0 +1,24 @@
...
+      const {
+        commit,
+        state: { network },
+      } = store;
+      const networks = {
+        ...network,
+        ...(state.userNetworks || []).reduce((p, n) => ({ ...p, [n.name]: { ...n } }), {}),
+      };
+      const net = process.env.RUNNING_IN_TESTS ? 'Testnet' : state.current.network;
```
davidyuk, line 12: I propose to avoid migration running in e2e tests

src/store/plugins/names.js
petbaik: for some reason first time when app is opened `defaults` props is `undefined` and `Vue.set(defaults...)` throws an error. When reopen this problem is gone

davidyuk: Well, I propose to investigate this issue instead of making a workaround

petbaik: After making `update-default-name.js` migration run in sync and remove usage of `commit` this problem is not happening anymore. Returning this mutation to previous version

PR #354, opened 2020-06-06
Put extensions to release, add artifacts

davidyuk:


davidyuk:
Aha, it is done, here is a test release: https://github.com/aeternity/superhero-wallet/releases/tag/test

PR #370, opened 2020-06-12
Disable buttons in aex2 popups until data is initialized

petbaik:


src/lib/rpcWallet.js
```diff
@@ -135,7 +135,7 @@ const rpcWallet = {
       }
       this.tipContractAddress = this.tipContractAddress.includes('.chain')
         ? getAddressByNameEntry(
-            await this.sdk.api.getNameEntryByName(this.tipContractAddress),
+            await this.sdk.api.getNameEntryByName(this.tipContractAddress).catch(() => false),
```
mradkov, line 138: Eventually this can be later changed to sending to name hash directly.

PR #400, opened 2020-06-26
User invites referal

petbaik:


src/popup/router/pages/Invite.vue
```diff
@@ -0,0 +1,180 @@
...
+        Buffer.from(secretKey, 'hex'),
+      )}`;
+
+      this.$store.commit('invites/add', {
+        link,
+        publicKey,
+        secretKey,
+        balance: 0,
+        date: Date.now(),
+        networkId: this.sdk.getNetworkId(),
```
davidyuk, line 95: Does it have the sense to track network id for each link? Without tracking, if the user on a different network invites will just look claimed. Also, I'm not sure switching between different networks (not nodes) is a case for the end-user.

src/popup/router/pages/Invite.vue
```diff
@@ -0,0 +1,180 @@
...
+    },
+    async claim(referral) {
+      this.loading = true;
+      await this.$store.dispatch('invites/claim', referral);
+      this.loading = false;
+      this.$store.dispatch('invites/getBalances');
+    },
+    async topUp({ link, topUpAmount, publicKey }) {
+      this.loading = true;
+      const address = publicKey || this.$store.getters['invites/keypair'](link);
```
davidyuk, line 108: `invites/keypair` returns an object with `publicKey`
```suggestion
      const address = publicKey || this.$store.getters['invites/keypair'](link). publicKey;
```

src/store/plugins/invites.js
```diff
@@ -0,0 +1,113 @@
...
+        return referrals.filter(r => r.networkId === networkId);
+      },
+    },
+    mutations: {
+      add(state, link) {
+        state.referrals.unshift(link);
+      },
+      reset(state) {
+        state.referrals = [];
+      },
```
davidyuk, line 44:
```suggestion
```
this mutation looks unused

src/store/plugins/invites.js
```diff
@@ -0,0 +1,113 @@
...
+      setClient(state, client) {
+        state.client = client;
+      },
+      setBalance(state, { i, balance }) {
+        Vue.set(state.referrals, i, { ...state.referrals[i], balance });
+      },
+    },
+    actions: {
+      async claim({ rootState: { account }, dispatch, getters: { get, keypair } }, referral) {
+        const { publicKey, secretKey } = get(referral);
```
davidyuk, line 54: Can we claim by idx in referrals array? Then we don't need `get` getter

src/store/plugins/invites.js
```diff
@@ -0,0 +1,113 @@
...
+        const node = await Node({
+          url: internalUrl,
+          internalUrl,
+        });
+        const accounts = MemoryAccount({ keypair });
+        const sdkInstance = await Ae({
+          compilerUrl,
+          nodes: [{ name: current.network, instance: node }],
+          accounts: [accounts],
+        });
```
davidyuk, line 91: No need to create sdk instance because you can set keypair to `onAccount` option (all sdk methods supports that?) and it will be used instead of the current account.
https://github.com/aeternity/aepp-sdk-js/blob/2ae757cd4c5701f3fedf3f0b5280d5862a582c19/es/account/selector.js#L48-L59

src/store/plugins/invites.js
```diff
@@ -0,0 +1,113 @@
...
+            ...sdkInstance.Ae.defaults,
+          }).min,
+        );
+        const maxSpend = balance - fee;
+        if (maxSpend <= 0) {
+          return false;
+        }
+        try {
+          const res = await sdkInstance.spend(maxSpend, account.publicKey, { payload: 'referral' });
+          return res;
```
davidyuk, line 74:
```suggestion
        try {
          return sdkInstance.transferFunds(1, account.publicKey, { payload: 'referral' });
```
https://github.com/aeternity/aepp-sdk-js/blob/289598c55ff6a9d9eef8b85030e0d1f901447123/es/ae/index.js#L85

src/store/plugins/invites.js
petbaik: If main sdk is used and pass `onAccount` option this line here(https://github.com/aeternity/aepp-sdk-js/blob/289598c55ff6a9d9eef8b85030e0d1f901447123/es/ae/index.js#L90) returns the address of the main account, because no `onAccount` option is used https://github.com/aeternity/aepp-sdk-js/blob/2ae757cd4c5701f3fedf3f0b5280d5862a582c19/es/account/selector.js#L64

davidyuk: You are right, I have created an issue there: https://github.com/aeternity/aepp-sdk-js/issues/1046
Well, as a workaround, can we add the account by key pair, call `transferFunds` and then remove the added account?

src/store/plugins/invites.js
```diff
@@ -0,0 +1,113 @@
...
+      async getBalances({ state: { referrals, client }, dispatch, getters: { keypair }, commit }) {
+        await Promise.all(
+          referrals.map(async ({ link, publicKey, secretKey }, i) => {
+            const address = publicKey || keypair(link).publicKey;
+            const secret = secretKey || keypair(link).secretKey;
+            const sdk =
+              client || (await dispatch('getClient', { publicKey: address, secretKey: secret }));
+
+            const balance = parseFloat(
+              +aettosToAe(await sdk.balance(address).catch(() => 0)),
```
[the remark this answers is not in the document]
petbaik: It didn't worked for `balance`

src/popup/router/pages/Invite.vue
```diff
@@ -0,0 +1,180 @@
...
+      try {
+        await this.sdk.spend(+aeToAettos(topUpAmount), address, {
+          payload: 'referral',
+        });
+        this.$store.dispatch('invites/getBalances');
+      } catch (e) {
+        this.$store.dispatch('modals/open', { name: 'default', msg: e.message });
+      }
+      this.loading = false;
```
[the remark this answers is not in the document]
petbaik: Global error handler will show modal with donate error functionality, not sure if this is the desired outcome when there is error with sending the transaction

src/popup/router/pages/Invite.vue
```diff
@@ -0,0 +1,180 @@
...
+    this.$store.dispatch('invites/getBalances');
+  },
+  methods: {
+    async generate() {
+      this.loading = true;
+      await this.$watchUntilTruly(() => this.sdk);
+      const { publicKey, secretKey } = Crypto.generateKeyPair();
+      if (this.amount) {
+        try {
+          await this.sdk.spend(+aeToAettos(this.amount), publicKey, {
```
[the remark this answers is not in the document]
petbaik: It worked for spend

src/popup/router/pages/Invite.vue
petbaik: Yes, you are right no need of `networkId`

src/store/plugins/invites.js
petbaik: I changed the account but still the `address` method returns the main account `publicKey` (https://github.com/aeternity/superhero-wallet/blob/develop/src/lib/wallet.js#L90). Not sure how to handle this

src/store/plugins/invites.js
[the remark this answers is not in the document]
petbaik: Actually it worked but the option name is format (https://github.com/aeternity/aepp-sdk-js/blob/a14d13adddec8fb1dd2a2ef10e8bf795e3c913f0/es/chain/node.js#L90)

src/popup/router/pages/Invite.vue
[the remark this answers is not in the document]
davidyuk: It is not the desired behavior and I think the user should be able to report about it. I'm not sure that here can be thrown only node-related exceptions. At least I'm proposing to define a list of exception's messages that will be caught here (instead of all exceptions).

src/store/modules/invites.js
```diff
@@ -0,0 +1,61 @@
...
+import MemoryAccount from '@aeternity/aepp-sdk/es/account/memory';
+import { AE_AMOUNT_FORMATS } from '@aeternity/aepp-sdk/es/utils/amount-formatter';
+
+export default {
+  namespaced: true,
+  state: {
+    referrals: [],
+  },
+  getters: {
+    get: state => link => state.referrals.find(r => r.link === link),
```
davidyuk, line 12: now, this getter is unused
```suggestion
```

src/store/modules/invites.js
```diff
@@ -0,0 +1,61 @@
...
+      }
+    },
+    async getBalances({ rootState: { sdk }, state: { referrals }, commit }) {
+      await Promise.all(
+        referrals.map(async ({ publicKey }, idx) => {
+          const balance = parseFloat(
+            await sdk.balance(publicKey, { format: AE_AMOUNT_FORMATS.AE }).catch(() => 0),
+          ).toFixed(2);
+          commit('setBalance', { idx, balance });
+          return balance;
```
davidyuk, line 56: no need to return balance
```suggestion
```

src/store/plugins/invites.js
davidyuk: Ok, I agree to have a separate sdk instance as a workaround because of all of this.

src/store/modules/invites.js
```diff
@@ -0,0 +1,61 @@
...
+import { AE_AMOUNT_FORMATS } from '@aeternity/aepp-sdk/es/utils/amount-formatter';
+
+export default {
+  namespaced: true,
+  state: {
+    referrals: [],
+  },
+  getters: {
+    get: state => link => state.referrals.find(r => r.link === link),
+    keypair: () => referral => {
```
davidyuk, line 13: Is this getter used somewhere?

petbaik: In the beginning it was used, but currently is not. I decided to leave it because it can be used for extracting keypair from url 🤷‍♀️ 
> Opening such link (deeplink)
> open a new view which will decode the url - and claim the balance to your account

src/store/modules/invites.js
```diff
@@ -0,0 +1,61 @@
...
+          payload: 'referral',
+        });
+      } catch (e) {
+        dispatch('modals/open', { name: 'default', msg: e.message }, { root: true });
+        return false;
+      } finally {
+        sdk.selectAccount(account.publicKey);
+      }
+    },
+    async getBalances({ rootState: { sdk }, state: { referrals }, commit }) {
```
davidyuk, line 48: Actually, it doesn't return any balances, it updates them in the state. So:
```suggestion
    async updateBalances({ rootState: { sdk }, state: { referrals }, commit }) {
```

src/store/modules/invites.js
davidyuk: After removing we will be able to get it back (when needed) from git history✂️

src/popup/router/pages/Invite.vue
```diff
@@ -0,0 +1,90 @@
...
+                name: 'default',
+                msg: this.$t('pages.invite.insufficient-balance'),
+              });
+              return;
+            }
+            throw e;
+          }
+        } finally {
+          this.loading = false;
+        }
```
davidyuk, line 63: I can't get why do you need a nested try/catch
```suggestion
        try {
          await this.sdk.spend(this.amount, publicKey, {
            payload: 'referral',
            denomination: AE_AMOUNT_FORMATS.AE,
          });
        } catch (e) {
          if (e.message.includes('is not enough to execute')) {
            this.$store.dispatch('modals/open', {
              name: 'default',
              msg: this.$t('pages.invite.insufficient-balance'),
            });
            return;
          }
          throw e;
        } finally {
          this.loading = false;
        }
```

src/store/modules/invites.js
```diff
@@ -0,0 +1,67 @@
...
+    async claim({ rootState: { account }, state: { links }, dispatch }, idx) {
+      const { publicKey, secretKey } = links[idx];
+      const sdk = await dispatch('getClient', { publicKey, secretKey });
+      try {
+        sdk.transferFunds(1, account.publicKey, {
+          payload: 'referral',
+        });
+      } catch (e) {
+        if (e.message.includes('is not enough to execute')) {
+          dispatch('modals/open', { name: 'default', msg: e.message }, { root: true });
```
davidyuk, line 34: Shouldn't we use here the same message as before?
```suggestion
          dispatch('modals/open', { name: 'default', msg: this.$t('pages.invite.insufficient-balance') }, { root: true });
```

src/store/modules/invites.js
```diff
@@ -0,0 +1,67 @@
...
+      });
+      const accounts = MemoryAccount({ keypair });
+      const sdkInstance = await Ae({
+        compilerUrl,
+        nodes: [{ name: current.network, instance: node }],
+        accounts: [accounts],
+      });
+      commit('setClient', sdkInstance);
+
+      return sdkInstance;
```
davidyuk, line 54: I think no need to put it into state
```suggestion
      return Ae({
        compilerUrl,
        nodes: [{ name: current.network, instance: node }],
        accounts: [accounts],
      });
```

src/store/modules/invites.js
```diff
@@ -0,0 +1,67 @@
...
+import Ae from '@aeternity/aepp-sdk/es/ae/universal';
+import Node from '@aeternity/aepp-sdk/es/node';
+import MemoryAccount from '@aeternity/aepp-sdk/es/account/memory';
+import { AE_AMOUNT_FORMATS } from '@aeternity/aepp-sdk/es/utils/amount-formatter';
+
+export default {
+  namespaced: true,
+  state: {
+    links: [],
+    client: null,
```
davidyuk, line 11:
```suggestion
```

src/store/modules/invites.js
```diff
@@ -0,0 +1,67 @@
...
+  },
+  mutations: {
+    add(state, link) {
+      state.links.unshift(link);
+    },
+    setClient(state, client) {
+      state.client = client;
+    },
```
davidyuk, line 19:
```suggestion
```

src/store/modules/invites.js
```diff
@@ -0,0 +1,67 @@
...
+    async getClient({ rootState: { network, current }, commit }, keypair) {
+      const { internalUrl, compilerUrl } = network[current.network];
+      const node = await Node({
+        url: internalUrl,
+        internalUrl,
+      });
+      const accounts = MemoryAccount({ keypair });
+      const sdkInstance = await Ae({
+        compilerUrl,
+        nodes: [{ name: current.network, instance: node }],
```
davidyuk, line 49: is it possible to reuse the node from the existing sdk instance? 🤔

src/store/index.js
```diff
@@ -108,6 +109,7 @@ export default new Vuex.Store({
         saveErrorLog,
         tourStartBar,
         tokens: { all },
+        invites: { links },
```
davidyuk, line 112: After removing sdk from state:
```suggestion
        invites,
```

src/store/index.js
```diff
@@ -128,6 +130,7 @@ export default new Vuex.Store({
...
```
davidyuk, line 133:
```suggestion
        invites,
```

src/store/modules/invites.js
```diff
@@ -0,0 +1,67 @@
+import Vue from 'vue';
+import Ae from '@aeternity/aepp-sdk/es/ae/universal';
+import Node from '@aeternity/aepp-sdk/es/node';
+import MemoryAccount from '@aeternity/aepp-sdk/es/account/memory';
```
davidyuk, line 4: I think we should prefer importing from the index file:
```suggestion
import { Universal as Ae, Node, MemoryAccount } from '@aeternity/aepp-sdk/es';
```
this will make it easier to refactor files in sdk 🙂

src/popup/router/pages/Invite.vue
petbaik: No reason, just mistake from my side :)

src/store/modules/invites.js
```diff
@@ -25,22 +25,23 @@ export default {
...
     },
-    async getClient({ rootState: { network, current }, commit }, keypair) {
-      const { internalUrl, compilerUrl } = network[current.network];
-      const node = await Node({
-        url: internalUrl,
-        internalUrl,
-      });
+    async getClient({ rootState: { network, current, sdk } }, keypair) {
+      const { compilerUrl } = network[current.network];
+      const { instance } = sdk.pool.get(current.network);
```
petbaik, line 40: @davidyuk Is this the right way to reuse the node instance?

PR #417, opened 2020-07-08
twitter.js: Match only tweet id in tweet link

### LINKED ISSUE #411: Twitter tip button uses wrong URL
**Describe the bug**
When using the tip button for a tweet the button uses an URL for one of the links that are included at the bottom of the tweet, e.g. `/likes` or `/analytics`. Instead it should use the URL of the tweet. Not the likes, analytics or whatever else.

In my case I didn't check the URL and was logged in. I tipped 1 AE to an URL that isn't publicly visible and therefore not claimable.

This is the tip:
- https://superhero.com/tip/1534

etharner:
Potentially closes #411.
I can't reproduce this issue, but in the case @marc0olo mentioned we should cut part after `/status/numbers/...`, I think this should fix this issue.

### DIFF
--- a/src/content-scripts/twitter.js
+++ b/src/content-scripts/twitter.js
@@ -34,7 +34,7 @@ const getTweetId = tweet => {
     return null;
   }
 
-  return status.href.split('/retweets')[0];
+  return status.href.match('.*/status/[0-9]+')[0];
 };
 
 const createSuperheroTipAction = (tweet, tweetId, numActions) => {


mradkov:
> Potentially closes #411.
> I can't reproduce this issue, but in the case @marc0olo mentioned we should cut part after `/status/numbers/...`, I think this should fix this issue.

I think this is only reproduce-able when manually sending tips from the extension instead of using the injected tip button. But this can only be done on tweets that were created by you as the `analytics` is only accessible to the author of the tweet.

marc0olo:
I can definitely say that I used the injected tip button which had a misbehaviour in that regards. two days ago I tested again with the injected button and didn't have the problem anymore.

might another commit have fixed this?!

so maybe this cannot be reproduced anymore or I had an outdated version when testing it. but the behaviour I mentioned was definitely using the injected tip button. and this occured when I was the author of the tweet.

PR #420, opened 2020-07-21
Revert "Merge branch 'master' into tests"

davidyuk:
From #44 I can't get why my refactoring was reverted there, I'm planning to use the same code in another project, so proposing to revert them back here.

### DIFF
--- a/src/popup/router/components/SidebarMenu.vue
+++ b/src/popup/router/components/SidebarMenu.vue
@@ -1,5 +1,5 @@
 <template>
-  <ul class="sidebar-menu" @click="menuClickHandler" data-cy="sidebar-menu">
+  <ul class="sidebar-menu" data-cy="sidebar-menu">
     <li class="menu-close">
       <Close @click="closeMenu" data-cy="close-menu" />
     </li>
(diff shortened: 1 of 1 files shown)


src/popup/router/components/SidebarMenu.vue
```diff
@@ -92,27 +92,20 @@ export default {
   mixins: [removeAccountMixin],
   components: { Close, Arrow, UserAvatar },
   computed: mapGetters(['account', 'activeAccountName']),
-  data: () => ({ showSettingsDropdown: false, showTokensDropdown: false, balances: null }),
-  watch: {
-    showTokensDropdown(val) {
```
mradkov, line 97: This should be present for the FT functionality I think.

davidyuk: Well, I'm proposing to introduce it in the same time as fungible tokens support.

davidyuk:
For some reason it breaks menu clicking in e2e tests, I will investigate it later

davidyuk:
This one is too old, I don't think that I finish this fix

PR #433, opened 2020-07-27
Inject Youtube tip button

### LINKED ISSUE #429: Inject Youtube tip button
Similar functionality to the twitter tip button for youtube.

etharner:
Closes #429.


src/content-scripts/tipButton.js
```diff
@@ -0,0 +1,89 @@
...
+    e.stopPropagation();
+  };
+
+  buttonContent.onmouseenter = e => {
+    hoverEnter();
+    e.stopPropagation();
+  };
+  buttonContent.onmouseleave = e => {
+    hoverLeave();
+    e.stopPropagation();
```
davidyuk, line 82: @mradkov why do we need `stopPropagation` here?

davidyuk: In general, I would suggest to define hover styles using css.

src/content-scripts/tipButton.scss
```diff
@@ -0,0 +1,38 @@
...
+      position: relative;
+    }
+
+    &:hover {
+      span {
+        color: #2a9cff;
+      }
+
+      img {
+        filter: invert(52%) sepia(46%) saturate(3166%) hue-rotate(186deg) brightness(100%) contrast(104%);
```
etharner, line 43: I propose to use that hack (https://codepen.io/sosuke/pen/Pjoqqp) instead of loading other icon

mradkov (changes requested):
The button I see on YT does not have the appropriate style.

Screenshot 2020-08-20 at 10 40 03

src/content-scripts/tipButton.js
```diff
@@ -1,8 +1,13 @@
 import icon from '../icons/twitter-icon-tip.svg';
-import './tipButton.scss';
 
 global.browser = require('webextension-polyfill');
 
+const style = document.createElement('link');
```
etharner, line 5: I used solution proposed here [https://stackoverflow.com/a/9721569]()

mradkov (approved):
Looks good now, thanks @etharner.

Issue #437, opened 2020-07-29
Logger ignores exceptions shown in vue event handlers


davidyuk:
Check this reproduction: 8365df63bed7c04849f324a9ec282a8f8db76752

https://reproduction-logger.wallet.z52da5wt.xyz/

I assume that the button on click should show a modal proposing to send the exception to us.

mradkov:
> I assume that the button on click should show a modal proposing to send the exception to us.

Yes, this is the intended behavior.

Issue #378, opened 2020-06-17
Make wallet to use forwarding with url param


shapkarin:
Hello. Maybe it will be nice if wallet can read url param like `callback` and forward user to that callback after user retrieve or create new account.
The reason why I think we need that is because when you want to use deeplinks and web wallet to to get the JWT token:
1. first you need to "auth" with web wallet
2. after you "auth" it's forward user to the http://wallet.superhero.com
3. And user see http://wallet.superhero.com instead of `share account page`

I hope it's clear. Please, notice me if I'm wrong or something is not clear.

shapkarin:
I'm wrong a little at point 3. We need to forward not to the jitsi client or superhero but the url was defined as a `callback`.

mradkov:
https://github.com/aeternity/superhero-wallet/blob/26020376989317d32f6d2fbd954d3514e13a6432/tests/pages/index.html#L44-L50

shapkarin:
@mradkov I'm about the different - before we call address method.

PR #438, opened 2020-07-29
Allow to claim tokens by invite link

davidyuk:
closes https://github.com/aeternity/superhero-ui/issues/724

will work much better after implementing of #378 
some exceptions are not displayed because of #437

It has a regression: I proposing to don't persist balances because it complicates state and I think it is enough to have this fancy feature only on the main screen. Otherwise, I propose to implement a general mechanism of balance caching instead of doing the same in different views.

Also, it has a security/privacy issue: shared private key will be exposed to owners of a web server. It may be solved by putting it in a hash part of the URL, but it needed to investigate how to handle this in Cordova because we are using hash navigation there.

### DIFF
--- a/src/store/modules/invites.js
+++ b/src/store/modules/invites.js
@@ -10,8 +10,7 @@ export default {
     add: ({ invites }, secretKey) => invites.unshift({ secretKey, createdAt: Date.now() }),
   },
   actions: {
-    async claim({ rootState: { account, current, sdk }, state: { invites } }, idx) {
-      const { secretKey } = invites[idx];
+    async claim({ rootState: { account, current, sdk } }, secretKey) {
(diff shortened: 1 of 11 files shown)


src/store/modules/invites.js
```diff
@@ -17,7 +17,7 @@ export default {
         nodes: [sdk.pool.get(current.network)],
         accounts: [MemoryAccount({ keypair: { publicKey, secretKey } })],
       });
-      await s.transferFunds(1, account.publicKey, { payload: 'referral' });
+      await s.transferFunds(1, account.publicKey, { payload: 'referral', verify: false });
```
davidyuk, line 20: This looks like a bug on sdk side

PR #439, opened 2020-08-04
twitter tip button: Avoid extra eslint-disable

davidyuk:
I came to this while reviewing #433. The changes look obvious, but I have not tested them.

### DIFF
--- a/src/content-scripts/twitter.js
+++ b/src/content-scripts/twitter.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-continue */
 import iconHover from '../icons/twitter-icon-tip-hover.svg';
 import icon from '../icons/twitter-icon-tip.svg';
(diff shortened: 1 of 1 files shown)


src/content-scripts/twitter.js
```diff
@@ -105,12 +104,11 @@ const configureSuperheroTipAction = async () => {
 
   let bigTweetSkipped = !document.querySelectorAll('div[aria-label="Timeline: Conversation"]')
     .length;
-  // eslint-disable-next-line no-plusplus
-  for (let i = 0; i < tweets.length; ++i) {
```
davidyuk, line 109: Also, it was possible to write this in AirBnb compatible way:
```js
for (let i = 0; i < tweets.length; i += 1) {
```

PR #445, opened 2020-08-06
Refactor backend

davidyuk:


src/background.js
```diff
@@ -6,15 +6,13 @@ import { PopupConnections } from './lib/popup-connection';
 import RedirectChainNames from './lib/redirect-chain-names';
 import rpcWallet from './lib/rpcWallet';
 import TipClaimRelay from './lib/tip-claim-relay';
-import Notification from './notifications';
```
mradkov, line 9: Why are the Notifications removed?

PR #456, opened 2020-08-10
Remove unnecessary migrations stuff

### LINKED ISSUE #360: Refactor update-state migration
https://github.com/aeternity/superhero-wallet/blob/e6ac3e2ffa147b30f3253ed8f02006eac363af89/src/store/migrations/update-state.js

The issues that I see here:
- it has unclear name
- should be used `async migrate` instead of returning a promise
- would be better to have it not depending on existing mutations and actions because they can be changed later
- it shouldn't have a code aware of e2e tests

Actually, can we remove this migration assuming that the most of our users have applied it?

davidyuk:
closes #360


davidyuk:
But it completely breaks tests, I will fix them later.

src/popup/utils/index.js
```diff
@@ -9,33 +10,29 @@ export const formatDate = time =>
...
+    privateKey: mnemonicToSeed(mnemonic).toString('hex'),
+  };
+  return {
+    ...(await runMigrations()),
+    account,
+    subaccounts: [
+      {
+        name: 'Main Account',
+        publicKey: account.publicKey,
+        balance: 10,
```
mradkov, line 26: What does `balance 10` mean?

davidyuk: I have just copied it from the previous version. It should be a balance of a specific subaccount, but I didn't check that. Anyway I'm proposing to drop subaccounts in #463

PR #463, opened 2020-08-11
Remove unnecessary subaccount structures

davidyuk:
depends on #456


davidyuk:
I agree that subaccounts are a necessary feature, but I'm not sure that we should keep the current implementation. For several reasons:
- this code increases the complexity
- it is unclear how to test this code
- when you implementing something new is it unclear should it support subaccounts or not
- for me, it doesn't look easy to bring subaccounts back using the current implementation

I'm proposing to continue cleaning up the background part and in the future, I see reasonable to introduce a store instance in the backend synced with the store in the popup part. Reverse iframe feature requires the popup to have the same functionality like the background so we can reuse a lot of code by that. This approach should be more simple than the current explicit synchronization between the background and the popup. Also, it can be reused for future synchronization between web version and extension.

PR #466, opened 2020-08-13
Invite links restyle

stanislav-slavov:
The "invite" functionality is not included in the tutorial - labels are needed.

Issue #470, opened 2020-08-17
Locale strings should match the official abbreviations


stanislav-slavov:
We should change the locale strings to match its original representations. For example:

- Our locale strings: en, cn
- Official abbreviations: en-US, zh-CN

davidyuk:
I agree regarding `cn` => `zn`, but can't we assume `en` a superset of `en-GB`, `en-US` and so? Of course, until we won't have any difference between these regions.

PR #471, opened 2020-08-17
Use currency symbols instead of abbreviations

davidyuk:
inspired by the changes in #466 

#466 introduces currency symbols in several places, having `currentCurrency` getter that is just an uppercased current currency, it is quite easy to introduce currency symbols in the whole application.

I'm proposing to merge this PR firstly and to rebase #466 on the top of it.

### DIFF
--- a/src/popup/router/components/BalanceInfo.vue
+++ b/src/popup/router/components/BalanceInfo.vue
@@ -11,8 +11,8 @@
         <span class="approx-sign">~</span>
         <li data-cy="currency-dropdown" class="dropdown-container" :class="dropdown ? 'show' : ''">
           <ae-button data-cy="toggle-currency-dropdown" @click="dropdown = !dropdown">
+            <span class="currency">{{ currentCurrencySymbol }}</span>
             {{ balanceCurrency }}
-            <span class="currency">{{ currentCurrency }}</span>
             <ExpandedAngleArrow />
           </ae-button>
           <ul class="sub-dropdown">
@@ -42,7 +42,7 @@ export default {
   data: () => ({ dropdown: false }),
   computed: {
     ...mapState(['current', 'currencies']),
-    ...mapGetters(['tokenBalance', 'balanceCurrency', 'currentCurrency']),
+    ...mapGetters(['tokenBalance', 'balanceCurrency', 'currentCurrencySymbol']),
   },
   methods: {
     async switchCurrency(currency) {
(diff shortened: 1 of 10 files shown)


mradkov:
This only covers $ and €. What about other currencies?

davidyuk:
I didn't notice that the currency list is scrollable. I think the best is to fix that is to use `Intl.NumberFormat`.

src/popup/router/components/BalanceInfo.vue
```diff
@@ -11,8 +11,7 @@
         <span class="approx-sign">~</span>
         <li data-cy="currency-dropdown" class="dropdown-container" :class="dropdown ? 'show' : ''">
           <ae-button data-cy="toggle-currency-dropdown" @click="dropdown = !dropdown">
-            {{ balanceCurrency }}
-            <span class="currency">{{ currentCurrency }}</span>
+            {{ formatCurrency(balanceCurrency) }}
```
davidyuk, line 14: It would be tricky to keep this behavior using `Intl.NumberFormat`, so I propose to drop that.

src/store/getters.js
```diff
@@ -14,9 +14,8 @@ export default {
   balanceCurrency({ current, balance }) {
     return (current.currencyRate * balance).toFixed(2);
   },
-  currentCurrency({ current }) {
-    return current.currency.toUpperCase();
-  },
+  formatCurrency: ({ current: { currency } }) => value =>
+    new Intl.NumberFormat('en', { style: 'currency', currency }).format(value),
```
davidyuk, line 18: It even shows a ₽ sign only if Russian like locale is chosen. I don't think that we should implement it by ourselves.

mradkov: Does `en` should be hardcoded or we should change it dynamically based on the current locale settings?

davidyuk: > should we change it dynamically based on the current locale settings?

I think so, is blocked by #470 

> As per specification Intl.NumberFormat uses locale strings as 'en-US', 'de-DE'

A Unicode BCP 47 locale identifier consists of
- a language code,
- (optionally) a script code,
- (optionally) a region (or country) code,
...

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#Locale_identification_and_negotiation

So, `en` is a correct BCP 47 locale identifier, that is used in `Intl.NumberFormat`

davidyuk:
@mradkov can we merge this?

mradkov:
> @mradkov can we merge this?

Yes

mradkov (approved):

PR #475, opened 2020-08-18
Normalise vuex state

davidyuk:
Depends on #471 

The idea is that the app state should be as small as possible. If we have some data that is calculated on the top of another data that is in our state, then we should extract it calculation to a getter.

cc: @petbaik


mradkov (approved):
The code changes look good.

PR #485, opened 2020-08-28
Handle deeplinks on extension side

davidyuk:
needed for https://github.com/aeternity/superhero-button/pull/24

In general it may be used as a light way to integrate superhero wallet (without dependency on sdk). But I'm not sure is it a good UX to open extension page in a usual tab (at it is implemented now). Technically we can avoid that but with an additional effort.

This may be tested by opening links from https://github.com/aeternity/superhero-wallet/blob/develop/tests/pages/index.html

Also, it makes the extension to support invite links 🤔

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -1,6 +1,7 @@
 import { isEmpty } from 'lodash-es';
-import { switchNode } from './lib/background-utils';
 import './lib/initPolyfills';
+import initDeeplinkHandler from './lib/deeplink-handler';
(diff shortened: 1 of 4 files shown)


mradkov (commented):
Generally, code changes look good to me. 

The question is - is this going to preserve backward compatibility?

davidyuk:
I don't see any breaking changes here.

mradkov (approved):

PR #486, opened 2020-08-28
Remove superhero-button click listener

davidyuk:
The idea is to use deeplinks instead, it may have a worse UI but this is much flexible. This would break the existing superhero-button installations, so I'm not sure when it the best time to merge it.

### DIFF
--- a/src/content-scripts/inject.js
+++ b/src/content-scripts/inject.js
@@ -112,20 +112,6 @@ const runContentScript = () => {
     }
   });
 
-  const setSuperheroButtonClickListener = () =>
-    document.addEventListener('click', event => {
-      if (!event.target.closest('.superhero-button')) return;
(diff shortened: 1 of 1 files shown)


mradkov (commented):
We should not break backward compatibility as we cannot know where the button is already integrated. Maybe we can keep this and announce its deprecation in some time?

davidyuk:
I have resolved this issue by using a different class name on the superhero-button side (https://github.com/aeternity/superhero-button/pull/24/commits/ba7c41010554fafd8881d878a4a11f62d6f19670). But still, I think we need to remove this handler later.

davidyuk:
The lastest backend changes broke amount displaying of the existing superhero-button installations, let's broke the click handler as well 😄

mradkov (approved):

PR #487, opened 2020-08-28
Add firefox and chrome extensions to Index page

davidyuk:
Using deeplinks in desktop (#485) makes it reasonable to add links to extensions to the Intro page. But, I was not able to find good badges images. Also, we can somehow highlight a platform that is preferred (by detection of current browser/mobile os), but it needs a design.

Screenshot 2020-08-28 at 15 16 56


mradkov:
Good thought! @onvisions any input from your side?

onvisions:
@davidyuk  Sorry, I've missed this one. @mradkov I think this can be a good design task for Artem to work on at the moment so we don't lose more time. He will need OP task with the description given above. Another issue I see here is that the logo is not updated with Superhero.com. Although generally I don't like the .com version (suggested by Leah) for the browser extension I think we should be consistent across the platform.

onvisions:
@davidyuk design is ready. I'll get in touch with you on wire to send you the links to source files.

davidyuk:
Designs received, I will implement them soon.

src/popup/locales/en.json
```diff
@@ -206,10 +206,7 @@
       "send": "Send",
       "ever": "Ever.",
       "wellcome": "Wellcome on board!",
-      "understand": "I understand that this is a cryptocurrency wallet and i am responsible for backing up my account recovery seed phrase.",
-      "extension": "Browser extension",
-      "mobileApp": "Mobile app"
-
+      "understand": "I understand that this is a cryptocurrency wallet and i am responsible for backing up my account recovery seed phrase."
```
mradkov, line 209:
```suggestion
      "understand": "I understand that this is a cryptocurrency wallet and I am responsible for backing up my account recovery seed phrase."
```

mradkov (approved):
Looks good to me. This is not to be rendered on the mobile devices though right?

CedrikNikita:
> Looks good to me. This is not to be rendered on the mobile devices though right?
If this [condition](https://github.com/aeternity/superhero-wallet/pull/487/files#diff-edfbc5e5989420b3f2a71c85eab05a37R51) was working before, then yes.

onvisions:
> 
> 
> Looks good to me. This is not to be rendered on the mobile devices though right?

Why? If the user opens the web wallet in a browser on a mobile device... The design is meant to serve also this case. So it shows the proper highlighting of download links according to browser and device OS.

onvisions:
https://feature-improve-landing.wallet.z52da5wt.xyz/   
May be I'm looking at the wrong link. However the link above doesn't match design at all. Please check Figma interactive prototype for reference. Neither colors, nor spaces, sizes, typography match. Interactions (hover, pressed states of elements, links and buttons) are also missing. In the design we have new size of the layout, bordered frame, etc. Generally there is no point of pixel perfect designs and prototyping micro interactions if we are not implementing it. 
**Am I looking at not the right link?** @mradkov @CedrikNikita

PR #488, opened 2020-08-28
Remove unused AirGap utils

davidyuk:


mradkov (approved):
We can still think of integrating airgap in the future

PR #494, opened 2020-09-04
Add auto-extending of names

### LINKED ISSUE #491: Auto extend names
Display a checkbox next to each name to auto-extend when possible.

Next to each name user should be able to mark for auto-extend.

etharner:
Closes #491.

### DIFF
--- a/src/popup/utils/constants.js
+++ b/src/popup/utils/constants.js
@@ -107,6 +107,7 @@ export const MAX_AMOUNT_WITHOUT_CONFIRM = 10;
 export const AENS_DOMAIN = '.chain';
 export const MAX_AUCTION_NAME_LENGTH = 12 + AENS_DOMAIN.length;
 export const MIN_NAME_LENGTH = 14;
+export const AUTO_EXTEND_NAME_BLOCKS_INTERVAL = 100;
 export const TIPPING_CONTRACT = `@compiler >= 4
 
 include "List.aes"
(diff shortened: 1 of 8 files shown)


src/store/plugins/names.js
```diff
@@ -72,8 +76,12 @@ export default store =>
                 { root: true },
               );
             }
+            if (name.expiresAt <= height && oldName.autoExtend) {
```
mradkov, line 79: I think the logic here is wrong. 

If the `height` is gt than `name.expiresAt` then the name would have been already expired. 🤔 

Maybe just auto extend without checking the height?

mradkov, line 79:
```suggestion
            if (name.expiresAt >= height && oldName.autoExtend) {
```

This way we will ensure we dispatch updatePointer only in the case when the current block height has not reached the name's expiration date.

mradkov (changes requested):
Also while testing - the name extend popup appears every few seconds when I'm on the `Names` page.

Let's make it fire only once at opening the extension.

src/popup/utils/constants.js
```diff
@@ -107,6 +107,7 @@ export const MAX_AMOUNT_WITHOUT_CONFIRM = 10;
 export const AENS_DOMAIN = '.chain';
 export const MAX_AUCTION_NAME_LENGTH = 12 + AENS_DOMAIN.length;
 export const MIN_NAME_LENGTH = 14;
+export const AUTO_EXTEND_NAME_BLOCKS_INTERVAL = 100;
```
etharner, line 110: Let's maybe use some interval between blocks when we will be asking for extending, and I also call it only on extension opening

mradkov (approved):

PR #495, opened 2020-09-07
Always use window.open on links inside iframe web wallet

### LINKED ISSUE #493: Mainnet view on SH Web wallet (UX/UI)
I tried to check the transaction on my web wallet and was redirected to the mainnet directly in my wallet view. But now, I don't know how to return to my wallet? 
I logged with Web wallet and want to join Jitsi Meet via the invite link.

My steps:
1. Click on invite Link: https://superhero.com/meet/AdequateMarriagesCorrectEqually
2. Click on Join meeting
3. Redirect to the Room
4. Choose random message
5. Click on Tip Button (Blue Diamant)
6. Enter amount. The amount should be >0.1 ae token
7. Click on Tip
8. Click on Confirm on Pop up
9. Check transaction on superhero.com feed or https://mainnet.aeternal.io/transactions/
10. Click on "eye" on my wallet

Знімок екрана 2020-09-03 о 2 18 21 пп

I think that the user will be confused if there no option to return in the wallet. I have not seen such a button.
￼
* OS: Mac OS 10.13.6
* Browser: chrome
* Version: 85.0.4183.83

P.S. User should use Go back button on the browser to return the previous view on the wallet.

Знімок екрана 2020-09-03 о 2 38 37 пп

etharner:
This quick fix potentially closes #493 as we won't open any links inside iframe web-wallet.

### DIFF
--- a/src/popup/utils/openUrl.js
+++ b/src/popup/utils/openUrl.js
@@ -12,7 +12,7 @@ export default (url, newTab) => {
       });
       break;
     case 'web':
-      if (newTab) {
+      if (newTab || window.parent) {
         window.open(url);
       } else {
         window.location = url;


mradkov (approved):
Seems like a good quickfix.

PR #500, opened 2020-09-11
Fix displaying of transactions

CedrikNikita:
All the transactions were shown with sign `Received`, no `sender` or `receiver` address provided, also did not sort.
Weird, because in README file https://github.com/aeternity/ae_mdw all the receiver or sender type have `_id` in the end.


mradkov (approved):
Changes look good. Working as intended now. 

Maybe it was confusion between different mdw versions previously.

PR #501, opened 2020-09-11
Drop mutation-types

CedrikNikita:


mradkov:
Why should we drop the mutation types?

Practically it introduces more modularity and saves us from making more typos, while also some editors' IntelliSense is recognizing the constants on import. 🤔 

ref: https://stackoverflow.com/questions/51268001/practical-usage-of-mutation-types-in-vuex

CedrikNikita:
> Why should we drop the mutation types?
> 
> Practically it introduces more modularity and saves us from making more typos, while also some editors' IntelliSense is recognizing the constants on import. 🤔
> 
> ref: https://stackoverflow.com/questions/51268001/practical-usage-of-mutation-types-in-vuex

Actually for the same reason we droppedthem in the superhero-ui https://github.com/aeternity/superhero-ui/issues/601#issue-630571948

mradkov:
> > Why should we drop the mutation types?
> > Practically it introduces more modularity and saves us from making more typos, while also some editors' IntelliSense is recognizing the constants on import. 🤔
> > ref: https://stackoverflow.com/questions/51268001/practical-usage-of-mutation-types-in-vuex
> 
> Actually for the same reason we dropped them in the superhero-ui [aeternity/superhero-ui#601 (comment)](https://github.com/aeternity/superhero-ui/issues/601#issue-630571948)

Let's keep it consistent then. My concern was that with the time the project is becoming bigger and maybe we will need to re-introduce those again. But let's see - time will tell.

mradkov (approved):

PR #507, opened 2020-09-14
Replaced middlewareUrl with explorerUrl

### LINKED ISSUE #504: Wrong redirection to the mainnet from the wallet
When I click on the eye in the wallet activity field to check my transactions on the mainnet, the wallet generates an incorrect link:

https://mainnet.aeternity.io/transactions/th_2XKCA9RCHkA6oXMcVDaQPC6B1kHQpZtje4Lv5eP8BwgJH4hWYq

HTTP ERROR 404

 - OS: iOS
 - Browser Chrome 
 - Version of wallet: Superhero 8a93b01 / v.0.3.3

mradkov:
Tx preview when clicking the eye icon was leading to the middleware API URL
Instead of the explorer URL (middleware UI) specified in the constants file

Fixes #504

### DIFF
--- a/src/popup/router/components/TransactionItem.vue
+++ b/src/popup/router/components/TransactionItem.vue
@@ -26,7 +26,7 @@
       </span>
       <span
         class="seeTransaction"
-        @click="openUrl(`${activeNetwork.middlewareUrl}/transactions/${transaction.hash}`)"
+        @click="openUrl(`${activeNetwork.explorerUrl}/transactions/${transaction.hash}`)"
       >
         <img src="../../../icons/eye.png" />
       </span>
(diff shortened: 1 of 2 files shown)


kenodressel (approved):
lgtm

kenodressel (changes requested):
what about this: https://github.com/aeternity/superhero-wallet/blob/master/src/popup/router/pages/Send.vue#L279

mradkov:
> what about this: https://github.com/aeternity/superhero-wallet/blob/master/src/popup/router/pages/Send.vue#L279

Hm, yeah. The issue was reported to be only present at the Activity history screen, so I did miss that. I will revisit it now. 👍

kenodressel (approved):

PR #531, opened 2020-09-25
Fix logout/login related issues

### LINKED ISSUE #523: New wallet generation process hung after deleting an existing account
The wallet hung and did not allow me to create a new account after I left the old one.

Extension on Chrome, Firefox. 

Steps to reproduce the behaviour:
1. Log out from extension.
2. Open extension
3. Click on 'I agree"
4. Click on Generate new wallet
5. Click on Proceed to your wallet
6. See an error on Pop-up:

Знімок екрана 2020-09-18 о 7 50 13 пп

 - OS:  iOS
 - Browser: chrome, firefox
 - Version: 0.3.3

CedrikNikita:
first commit fixes `Error: Attempting to use a disconnected port object`,
second fixes #523

### DIFF
--- a/src/lib/rpcWallet.js
+++ b/src/lib/rpcWallet.js
@@ -307,10 +307,12 @@ export default {
   async [AEX2_METHODS.LOGOUT]() {
     const { clients: aepps } = this.sdk.getClients();
     Array.from(aepps.values()).forEach(aepp => {
-      aepp.sendMessage(
-        { method: 'connection.close', params: { reason: 'bye' }, jsonrpc: '2.0' },
-        true,
-      );
+      if (aepp.info.status !== 'DISCONNECTED') {
+        aepp.sendMessage(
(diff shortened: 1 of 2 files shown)


src/lib/rpcWallet.js
```diff
@@ -307,10 +307,12 @@ export default {
   async [AEX2_METHODS.LOGOUT]() {
     const { clients: aepps } = this.sdk.getClients();
     Array.from(aepps.values()).forEach(aepp => {
-      aepp.sendMessage(
-        { method: 'connection.close', params: { reason: 'bye' }, jsonrpc: '2.0' },
-        true,
-      );
+      if (aepp.info.status !== 'DISCONNECTED') {
```
mradkov, line 310: In this case, we are sending `connection.close` if the status **is not** `DISCONNECTED`. Is that the desired logic?

CedrikNikita: yes

mradkov (approved):

PR #534, opened 2020-09-29
Dynamically configurable backend url

stanislav-slavov:


src/popup/router/pages/Networks.vue
```diff
@@ -59,6 +59,16 @@
         v-model="network.compilerUrl"
         data-cy="compiler"
       />
+      <div class="text-left expand" @click="expanded = !expanded">
+        <img :class="expanded ? 'open' : ''" src="../../../icons/carret-down.svg" />
+        <span>{{ expanded ? $t('pages.network.showLess') : $t('pages.network.showMore') }}</span>
```
mradkov, line 64: I suggest we have something in the way of `Tipping configuration` or something like that instead of the `Show more/Show less` as this does not bring much information to the user e.g. what is this field for.

src/lib/backend.js
```diff
@@ -1,7 +1,5 @@
 // copied from https://github.com/aeternity/superhero-ui/blob/178bc2d63cb362ddc3b2163fed58deb2e253ec00/src/utils/backend.js
```
mradkov, line 1: We should remove this as its no longer the case this is matching the code at the provided URL.

mradkov (commented):
Looks good to me. @thepiwo any opinion on this?

src/popup/router/pages/Networks.vue
```diff
@@ -94,6 +107,7 @@ const networkProps = {
   url: null,
   middlewareUrl: null,
   compilerUrl: null,
+  backendUrl: 'https://raendom-backend.z52da5wt.xyz',
```
thepiwo, line 110: why is it hardcoded here?

src/lib/backend.js
```diff
@@ -25,15 +21,20 @@ const wrapTry = async promise => {
...
 
-const backendFetch = (path, ...args) => wrapTry(fetch(`${BACKEND_URL}/${path}`, ...args));
-
 export default class Backend {
-  static getTipComments = async tipId =>
-    backendFetch(`comment/api/tip/${encodeURIComponent(tipId)}`);
+  backendUrl = 'https://raendom-backend.z52da5wt.xyz';
+
+  constructor(url) {
+    this.backendUrl = url || this.backendUrl;
```
thepiwo, line 28: why hardcoded fallback?

thepiwo:
still says tip/claim only works on mainnet

src/popup/router/pages/Networks.vue
stanislav-slavov: I could export it in a constant and import it here. This is our default backend url, the user has the option to change it. Do you think I should leave the field blank?

src/lib/backend.js
stanislav-slavov: Same as above comment: "default backend" option.

stanislav-slavov:
> still says tip/claim only works on mainnet


Now we have a default configuration for test/main net => so we could remove the limitation for tipping only on mainnet.
I would like to discuss few thing before moving on:
- The user would be able to configure a backend url with this PR, shouldn't we remove the tipping limitation at all(not only for test/mainnet)?
- Should I add the ability to configure AGGREGATOR_URL with the network settings in this PR?
@davidyuk @mradkov @thepiwo

mradkov:
> > still says tip/claim only works on mainnet
> 
> 
> 
> 
> 
> Now we have a default configuration for test/main net => so we could remove the limitation for tipping only on mainnet.
> 
> I would like to discuss few thing before moving on:
> 
> - The user would be able to configure a backend url with this PR, shouldn't we remove the tipping limitation at all(not only for test/mainnet)?
> 
> - Should I add the ability to configure AGGREGATOR_URL with the network settings in this PR?
> 
> @davidyuk @mradkov @thepiwo 

Hmm, well it is possible to have it configurable, but what more will we achieve besides having the redirect button after tip leading to the custom aggregator?

stanislav-slavov:
> > > still says tip/claim only works on mainnet
> > 
> > 
> > Now we have a default configuration for test/main net => so we could remove the limitation for tipping only on mainnet.
> > I would like to discuss few thing before moving on:
> > 
> > * The user would be able to configure a backend url with this PR, shouldn't we remove the tipping limitation at all(not only for test/mainnet)?
> > * Should I add the ability to configure AGGREGATOR_URL with the network settings in this PR?
> > 
> > @davidyuk @mradkov @thepiwo
> 
> * Could be that they (the users) add custom network that does not have any tipping contracts deployed and running so I think the check should be there. Just the testnet is not relevant anymore.
> * Hmm, well it is possible to have it configurable, but what more will we achieve besides having the redirect button after tip leading to the custom aggregator?

- ok so I will update the check with the testnet
- Yes, the tips could lead to a custom aggregator. I guess this is not such a priority.

src/lib/backend.js
davidyuk: This is a reminder of a duplicate code that would be better to extract somehow later.

src/lib/wallet.js
```diff
@@ -55,7 +55,7 @@ async function getKeyPair() {
 }
 
 async function initContractInstances() {
-  if (!store.getters.mainnet && !process.env.RUNNING_IN_TESTS) return;
+  if (!store.getters.allowTipping && !process.env.RUNNING_IN_TESTS) return;
```
davidyuk, line 58: `allow` associated with some kind permission, I think `tippingSupported`/`tippingAvailable` suit better

src/lib/tip-claim-relay.js
```diff
@@ -1,13 +1,13 @@
...
-import { defaultNetwork, TIPPING_CONTRACT, TIP_SERVICE } from '../popup/utils/constants';
+import { defaultNetwork, TIPPING_CONTRACT } from '../popup/utils/constants';
 import {
   contractCallStatic,
   getActiveAccount,
   getAddressFromChainName,
   getTippingContractAddress,
 } from './background-utils';
 import Logger from './logger';
+import getters from '../store/getters';
```
davidyuk, line 10: Usage getters separate from the store looks strange.

src/popup/locales/en.json
```diff
@@ -108,7 +108,7 @@
     },
     "tip-mainnet-warning": {
       "title": "Warning",
-      "msg": "Tip/Claim functionality works only on Mainnet."
+      "msg": "Tip/Claim functionality works on Mainnet/Testnet."
```
davidyuk, line 111: "supported networks"? or maybe "predefined networks"? 🤔 🤷‍♀️

src/lib/tip-claim-relay.js
```diff
@@ -47,8 +47,7 @@ export default {
 
         if (this.checkAddressMatch(account.publicKey, uniq(addresses))) {
           await this.abortIfZeroClaim(url);
-
-          await axios.post(`${TIP_SERVICE}`, { url, address: account.publicKey });
+          await getters.backendInstance.claimTips({ url, address: account.publicKey });
```
davidyuk, line 50: You have imported a kind of 'raw' getters, so `getters.backendInstance` is a function, not an object with backend methods.

src/popup/router/pages/Networks.vue
```diff
@@ -59,6 +59,18 @@
         v-model="network.compilerUrl"
         data-cy="compiler"
       />
+      <div class="text-left expand" @click="expanded = !expanded">
+        <img :class="expanded ? 'open' : ''" src="../../../icons/carret-down.svg" />
```
davidyuk, line 63:
```suggestion
        <img :class="{ 'open' : expanded }" src="../../../icons/carret-down.svg" />
```
or even to rename class to `expanded`, because they mean the same here

src/popup/router/pages/Networks.vue
```diff
@@ -59,6 +59,18 @@
...
+        <span>{{
+          expanded ? $t('pages.network.hideTippingConfig') : $t('pages.network.showTippingConfig')
+        }}</span>
+      </div>
+      <Input
+        v-show="expanded"
```
davidyuk, line 69: why don't use `v-if`?

src/popup/router/pages/Networks.vue
```diff
@@ -108,16 +122,18 @@ export default {
     return {
       mode: 'list',
       network: networkProps,
+      expanded: false,
```
davidyuk, line 125: would be better to name that exactly is expanded or not

src/popup/router/pages/Networks.vue
```diff
@@ -59,6 +59,18 @@
         v-model="network.compilerUrl"
         data-cy="compiler"
       />
+      <div class="text-left expand" @click="expanded = !expanded">
```
davidyuk, line 62: would be more semantic to make buttons (not divs) clickable

src/popup/router/pages/Networks.vue
```diff
@@ -201,4 +219,30 @@ export default {
...
+  span {
+    display: inline-block;
+  }
+
+  img {
+    transform: rotate(-90deg);
+    vertical-align: middle;
+
+    & ~ span {
+      margin-bottom: 20px;
```
davidyuk, line 236: Instead, you can apply this style to the buttons below to don't switch them on opening

src/popup/utils/constants.js
```diff
@@ -64,6 +64,8 @@ export const calculateFee = (type, params) => {
   };
 };
 
+export const DEFAULT_BACKEND_URL = 'https://raendom-backend.z52da5wt.xyz';
```
davidyuk, line 67: Proposing to use `defaultNetwork.backendUrl` instead

davidyuk (commented):
I think, the backend methods should be Vuex actions (extracted into the separate module), that depends on the state (in particular the currently chosen backend) similar to how we are doing it in superhero-ui. This is not significantly more complicated, because we anyway have to update all usages, but this more looks like a way that Vuex is supposed to be used.

src/popup/router/pages/Networks.vue
stanislav-slavov: It is part of the form, the value would be taken into consideration no matter if it is expanded or not - this is the reasoning behind v-show.

stanislav-slavov:
All backend calls go through Vuex actions. backend.js is removed as there will be no further need for it. There were some merge conflicts regarding "notifications" and "chainnames" functionalities - I have refactored the calls so @CedrikNikita please have a look at it. @davidyuk please let me know if things are done as per our sync.

src/popup/router/components/UserAvatar.vue
```diff
@@ -1,7 +1,10 @@
 <template>
   <img
-    v-if="!error"
-    :src="typeof profileImage === 'string' && profileImage.length > 0 ? profileImage : avatar.src"
+    :src="
+      typeof profileImage === 'string' && profileImage.length > 0 && !error
+        ? profileImage
+        : avatar.src
+    "
```
mradkov, line 7: Can't we use https://github.com/aeternity/superhero-avatars for this? There is an instance up and running on https://avatars.z52da5wt.xyz/.

Check it out: `https://avatars.z52da5wt.xyz/{CHAIN_NAME || AK_ADDRESS}`

mradkov:
There is a failing test: https://travis-ci.com/github/aeternity/superhero-wallet/builds/188849765#L923

stanislav-slavov:
> There is a failing test: https://travis-ci.com/github/aeternity/superhero-wallet/builds/188849765#L923

It is fixed in the latest commit, all checks are passing.

src/popup/router/components/UserAvatar.vue
stanislav-slavov: Thanks for sharing. I would rather use the libraries we have included though. In my opinion it will serve better as a fallback if the call for profileImage fails for some reason.

src/store/getters.js
```diff
@@ -28,14 +28,17 @@ export default {
...
+    `${activeNetwork.backendUrl}/profile/image/${address}`,
   activeAccountName({ account }, getters) {
     return getters['names/getDefault'](get(account, 'publicKey')) || 'Main account';
   },
-  allowTipping(state, { mainnet }) {
-    return mainnet || process.env.RUNNING_IN_TESTS;
+  tippingSupported(state, { activeNetwork }) {
+    return (
+      activeNetwork.networkId === 'ae_mainnet' ||
+      activeNetwork.networkId === 'ae_uat' ||
```
davidyuk, line 39:
```suggestion
      ['ae_mainnet', 'ae_uat'].includes(activeNetwork.networkId) ||
```

src/store/plugins/observables.js
```diff
@@ -33,10 +33,27 @@ export default store => {
         : `https://superhero.com/tip/${sourceId}/comment/${entityId}`,
   });
 
+  const getAllNotifications = async (address, signCb) => {
```
davidyuk, line 36: Can we make this method action, and use `state.sdk` instead `signCb`?

src/store/actions.js
```diff
@@ -137,4 +137,48 @@ export default {
...
+    const signedChallenge = await signCb(responseChallenge.challenge);
+    const respondChallenge = {
+      challenge: responseChallenge.challenge,
+      signature: signedChallenge,
+      parentId,
+    };
+    return sendComment(respondChallenge);
+  },
+  async modifyNotification({ getters: { activeNetwork } }, [notifId, status, author, signCb]) {
+    const modifyNotif = async postParam =>
```
davidyuk, line 169: I think we can name this local variable `backendMethod` instead of naming it the same as the whole method, but short.

src/store/actions.js
```diff
@@ -24,25 +24,27 @@ export default {
...
-          await axios
-            .get(
-              `${BACKEND_URL}/cache/events/?address=${publicKey}&event=TipWithdrawn${
-                recent ? `&limit=${limit}` : ''
-              }`,
-            )
-            .catch(() => ({ data: [] }))
-        ).data.map(({ address, amount, ...t }) => ({
+      (async () => {
+        const transactionsFromEvents = await fetchJson
```
davidyuk, line 33: Why don't do it without the intermediate variable as it was before? Probably, using `then` would be more readable

src/popup/router/pages/Networks.vue
```diff
@@ -197,8 +219,36 @@ export default {
...
-  margin-right: 0;
+.network {
+  .edit-btn {
+    margin-left: 5px;
+    margin-right: 0;
+  }
+
+  .expand {
+    font-size: 14px;
+    cursor: pointer;
```
davidyuk, line 230: it is not necessary if .expland is a button

src/popup/router/pages/CommentNew.vue
```diff
@@ -42,14 +41,15 @@ export default {
...
+      ];
       try {
-        await Backend.sendTipComment(
-          this.id,
-          this.text,
-          await this.sdk.address(),
-          async data => Buffer.from(await this.sdk.signMessage(data)).toString('hex'),
-          this.parentId,
-        );
+        this.$store.dispatch('sendTipComment', postData);
```
davidyuk, line 52:
```suggestion
        await this.$store.dispatch('sendTipComment', [
         this.id,
         this.text,
         await this.sdk.address(),
         async data => Buffer.from(await this.sdk.signMessage(data)).toString('hex'),
         this.parentId,
       ]);
```
- await is missing
- don't think that is needed to have intermediate variable (also `postData` name is so unmeaningful in this context)

src/popup/router/components/UserAvatar.vue
davidyuk: Was it needed to refactor the avatars component in this PR? Actually, I like to extract avatars to a service, because packages for generation are too heavy.

davidyuk (commented):
Please, squash the overlapping changes, the most obvious commits to squash is 7f3553ce37f5c046c759389d7292a59c709f7713 and e10086c0bcb635b9942c303b10db169076f91715, but it is not the lest one.

src/popup/router/pages/Networks.vue
davidyuk: The absence of Input won't break `backendUrl` value. `v-if` is a bit more performant and clear, so use it unless you have a strong reason to use `v-show`.

package-lock.json
```diff
@@ -5,7 +5,7 @@
   "requires": true,
   "dependencies": {
     "@aeternity/aepp-components": {
-      "version": "git+https://github.com/aeternity/aepp-components.git#dedf56a5bd08465846b06945cf05a6a65fb94212",
+      "version": "git+https://github.com/aeternity/aepp-components.git#791b2b7d9dde531691778826cfc539db97c296a9",
```
davidyuk, line 8: This change should not be needed.

src/popup/router/components/UserAvatar.vue
stanislav-slavov: I just combined two  and one <div> tags into one . The needed refactoring was regarding the call for profileImage. My proposition is not to overcomplicate the already big PR.

src/lib/tip-claim-relay.js
```diff
@@ -47,8 +47,9 @@ export default {
 
         if (this.checkAddressMatch(account.publicKey, uniq(addresses))) {
           await this.abortIfZeroClaim(url);
-
-          await axios.post(`${TIP_SERVICE}`, { url, address: account.publicKey });
+          // This check is only used on mainnet
+          const { backendUrl } = defaultNetwork;
+          await fetchJson.post(`${backendUrl}/claim/submit`, { url, address: account.publicKey });
```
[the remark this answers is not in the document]
stanislav-slavov: I propose to leave it as it is. It is called only on mainnet currently so we could create a separate issue for handling it with the switching of networks.

src/popup/router/pages/Networks.vue
stanislav-slavov: I am not sure what is the issue here. 
Screenshot 2020-10-09 at 17 26 38
This is the section, which the user is expanding and collapsing.

src/popup/router/components/UserAvatar.vue
stanislav-slavov: https://github.com/aeternity/superhero-wallet/issues/551

src/lib/tip-claim-relay.js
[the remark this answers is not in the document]
stanislav-slavov: https://github.com/aeternity/superhero-wallet/issues/552

src/popup/router/pages/Networks.vue
davidyuk: `cursor: pointer` is the default behavior of `<button>`

davidyuk (commented):
I would introduce `postJson` function like:
```
export const postJson = (url, options) => fetchJson(url, {
  method: 'post',
  headers: { 'Content-Type': 'application/json' },
  ...options,
  body: options.body && JSON.stringify(options.body),
});
```
To reduce code duplication. I think it is better than to use third-party packages like axios because of the bundle size.

Please, squash the overlapping changes, the history should contain only the one implementation of working with the backend, we don't need all intermediate solutions in our history.

davidyuk (approved):

PR #535, opened 2020-09-29
cordova ios: Fix compilation error by choosing a different swift version

davidyuk:
Before there was used Swift 4 and the error was:

> 'openSettingsURLString' has been renamed to 'UIApplicationOpenSettingsURLString'

in https://github.com/bitpay/cordova-plugin-qrscanner/blob/b62efb64801ba0b830f0aac556935c0978d2fb98/src/ios/QRScanner.swift#L471

I have faced this error trying to build 0.3.3 release using Xcode 12.0.1


mradkov (approved):
I faced the same issue the other day when building it locally. Thanks.

PR #536, opened 2020-09-29
Add confirmation settings for popups

### LINKED ISSUE #506: Configurable popup messages
Allow users to select whether they want to confirm popups for signing messages and transactions.

This should be part of the settings menu.

The user should be presented with several checkboxes:
- Ask for login requests (on/off)
- Ask for message signing confirmations (on/off)
- Ask for transactions confirmations (on/off)

All of those should respectively do the action after their name. If those are toggled off the confirmation popups should be automatically confirmed for the user without showing up.

related to #492

etharner:
Closes #506.

### DIFF
--- a/src/lib/rpcWallet.js
+++ b/src/lib/rpcWallet.js
@@ -4,27 +4,19 @@ import Node from '@aeternity/aepp-sdk/es/node';
 import BrowserRuntimeConnection from '@aeternity/aepp-sdk/es/utils/aepp-wallet-communication/connection/browser-runtime';
 import { isEmpty } from 'lodash-es';
 import uuid from 'uuid';
+import { AEX2_METHODS, defaultNetwork, NO_POPUP_AEPPS } from '../popup/utils/constants';
 import {
-  AEX2_METHODS,
-  defaultNetwork,
-  MAX_AMOUNT_WITHOUT_CONFIRM,
-  NO_POPUP_AEPPS,
-} from '../popup/utils/constants';
-import {
-  addTipAmount,
   extractHostName,
   getAllNetworks,
   getAddressByNameEntry,
   getAeppAccountPermission,
-  getContractCallInfo,
-  getTippedAmount,
   parseFromStorage,
-  resetTippedAmount,
   stringifyForStorage,
 } from '../popup/utils/helper';
 import { getState } from '../store/plugins/persistState';
 import popups from './popup-connection';
 import walletController from '../wallet-controller';
+import { checkPermissions } from '../store/modules/permissions';
 
 global.browser = require('webextension-polyfill');
 
@@ -126,23 +118,9 @@ export default {
     return extractHostName(url);
   },
   async shouldOpenPopup(aepp, action) {
-    const { isTip, amount } = getContractCallInfo(action.params.tx, this.tipContractAddress);
-    const origin = this.getAeppOrigin(aepp);
-    if (NO_POPUP_AEPPS.includes(origin)) {
-      if (isTip) {
-        const tippedAmount = await getTippedAmount();
-        if (tippedAmount >= MAX_AMOUNT_WITHOUT_CONFIRM) {
-          resetTippedAmount();
-          return true;
-        }
-        action.accept();
-        await addTipAmount(amount);
-      } else {
-        action.accept();
-      }
-    } else {
+    if (!NO_POPUP_AEPPS.includes(this.getAeppOrigin(aepp)) || (await checkPermissions(action)))
       return true;
-    }
+    action.accept();
     return false;
   },
   async checkAeppPermissions(aepp, action, caller, cb) {
@@ -261,7 +239,7 @@ export default {
         },
       } = client;
       const isConnected = await getAeppAccountPermission(extractHostName(url), address);
-      if (!isConnected) {
+      if ((await checkPermissions({ method: 'address.subscribe' })) && !isConnected) {
         const accept = await this.showPopup({ action: {}, aepp: client, type: 'connectConfirm' });
         if (accept) {
           this.sdk.selectAccount(address);
--- a/src/popup/router/pages/PermissionsSettings.vue
+++ b/src/popup/router/pages/PermissionsSettings.vue
@@ -1,0 +1,187 @@
+<template>
+  <div class="popup">
+    <ListItem class="permission-row">
+      <CheckBox
+        :value="!permissions.address"
+        @click.native="changePermission('address', !permissions.address)"
+        prevent
+      />
+      <p :class="{ checked: !permissions.address }">{{ $t('pages.permissions.login') }}</p>
+    </ListItem>
+    <ListItem class="permission-row">
+      <CheckBox
+        :value="!permissions.messageSign"
+        @click.native="changePermission('messageSign', !permissions.messageSign)"
+        prevent
+      />
+      <p :class="{ checked: !permissions.messageSign }">
+        {{ $t('pages.permissions.message-sign') }}
+      </p>
+    </ListItem>
+    <ListItem class="permission-row">
+      <p class="text-left">{{ $t('pages.permissions.transaction-sign') }}</p>
+      <Input
+        :value="permissions.transactionSignLimit || ''"
+        :error="error"
+        placeholder="no limit"
+        @input="value => changePermission('transactionSignLimit', value)"
+      />
+    </ListItem>
+    <div>
+      <RangeInput
+        :value="permissions.transactionSignLimit"
+        min="0"
+        :max="tokenBalance"
+        step="0.1"
+        @input="value => changePermission('transactionSignLimit', value)"
+      />
+    </div>
+    <ListItem class="permission-row">
+      <p class="text-left grow">{{ $t('pages.permissions.spent-today') }}</p>
+      <span class="tokens">
+        {{ (permissions.transactionSignLimit - limitLeft).toFixed(2) }}
+        <span class="secondary-text">&nbsp;{{ $t('pages.appVUE.aeid') }}</span>
+      </span>
+      <span class="currency">{{
+        `(${formatCurrency(convertToCurrency(permissions.transactionSignLimit - limitLeft))})`
+      }}</span>
+    </ListItem>
+    <ListItem class="permission-row">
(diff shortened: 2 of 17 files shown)


etharner:
I'm proposing to add limit per day in first version. This way we prevent user from scam sites which can spam for small tx signing.

mradkov:
Strange, the CI and tests are not automatically running after the force push. 🤔

etharner:
> Strange, the CI and tests are not automatically running after the force push. 🤔

I've forgot to change hash of commit

etharner:
@onvisions For tx signing option we should probably have not only range for "maximum per day" (and we need to discuss range of values @mradkov) but current day limit "spent" amount as well.

davidyuk:
https://github.com/aeternity/superhero-wallet/blob/e843ca2d7a667e1d1f5d7c090ef47be5ea3ac6ce/src/popup/router/components/Modals/ConfirmTransactionSign.vue
https://github.com/aeternity/superhero-wallet/blob/8e31afc8c23ab08d01de259c38114b64b3c9522c/src/popup/router/pages/Popups/SignTx.vue

As I see, these two modals are not edited in this PR, but just be aware of them.

onvisions:
Design for **Confirmation Settings:**
https://nx3594.your-storageshare.de/apps/files/?dir=/P1/Superhero/Superhero%20Wallet%20Designs/10.%20Confirmation%20settings&fileid=819682

Except the sliders we should add custom value to be manually entered in the inout field with the reason not to limit the user to set **max. value** equal to their current balance.
Otherwise the user may have low balance at the specific moment (for example 1.5AE) and won't be able to set more than this number for **"Limit for spending without asking per day".**

Now as I write this it comes to my mind that **"Limit for amount spent per day"** may not be in its right place on this screen as it has nothing to do with any **confirmation settings**. Or? @mradkov 

1  Superhero Wallet - Confirmation settings

onvisions:
After syncing with @etharner some design changes have been done to the Confirmation Settings. The updated files can be seen in the respective directory.

davidyuk:
Don't you have a plan to implement these permissions per domain? This would be much safer than the current implementation.

src/store/plugins/confirmations.js
```diff
@@ -0,0 +1,33 @@
+import Vue from 'vue';
+import { i18n } from '../../popup/utils/i18nHelper';
+
+const confirmationsText = {
+  login: i18n.t('pages.confirmations.login'),
+  messageSign: i18n.t('pages.confirmations.message-sign'),
+  transactionSign: i18n.t('pages.confirmations.transaction-sign'),
```
davidyuk, line 7: To manage wording from Vuex looks like overengineering, can't we move it to the component? Also, are you sure that these translations remain reactive?

src/store/plugins/confirmations.js
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export default store =>
+  store.registerModule('confirmations', {
+    namespaced: true,
+    state: {
+      undeniable: {
+        login: false,
```
davidyuk, line 15: I would name it `address` because proper logging in requires some proof like message signing.

src/store/plugins/confirmations.js
```diff
@@ -0,0 +1,33 @@
...
+import { i18n } from '../../popup/utils/i18nHelper';
+
+const confirmationsText = {
+  login: i18n.t('pages.confirmations.login'),
+  messageSign: i18n.t('pages.confirmations.message-sign'),
+  transactionSign: i18n.t('pages.confirmations.transaction-sign'),
+};
+
+export default store =>
+  store.registerModule('confirmations', {
```
davidyuk, line 11:
```suggestion
  store.registerModule('permissions', {
```
?

src/store/plugins/confirmations.js
```diff
@@ -0,0 +1,33 @@
...
+    namespaced: true,
+    state: {
+      undeniable: {
```
davidyuk, line 14: is this nesting level necessary? I think it can be just `state.permissions.address`, `state.permissions.transactionSign`, `state.permissions.messageSign`

src/store/plugins/confirmations.js
```diff
@@ -0,0 +1,33 @@
...
+      list: ({ undeniable }) =>
+        Object.entries(confirmationsText).map(([name, text]) => ({
+          name,
+          text,
+          value: undeniable[name],
+        })),
+    },
+    mutations: {
+      setConfirmationValue(state, { name, value }) {
+        Vue.set(state.undeniable, name, value);
```
davidyuk, line 30: why `Vue.set` is necessary? I think it doesn't because all properties of `undeniable` are predefined.

src/popup/router/pages/ConfirmationSettings.vue
```diff
@@ -0,0 +1,58 @@
...
+      <CheckBox
+        v-if="typeof value === 'boolean'"
+        :value="value"
+        type="radio"
+        name="confirmationValue"
+        @click.native="changeConfirmation(name, !value)"
+        prevent
+      />
+      <div class="mr-auto text-left">
+        <p class="f-16" data-cy="network-name">{{ text }}</p>
```
davidyuk, line 13:
```suggestion
        <p class="f-16">{{ text }}</p>
```
also, is it possible to combine `div` and `p`?

src/popup/router/pages/ConfirmationSettings.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+  <div class="popup">
+    <ListItem v-for="{ name, text, value } in list" :key="name" class="confirmation-row">
+      <CheckBox
+        v-if="typeof value === 'boolean'"
+        :value="value"
+        type="radio"
+        name="confirmationValue"
```
davidyuk, line 8: is this used somehow?

src/lib/rpcWallet.js
```diff
@@ -126,23 +118,42 @@ export default {
     return extractHostName(url);
   },
   async shouldOpenPopup(aepp, action) {
```
davidyuk, line 120: The most part of this function is checkConfirmations, I'm proposing to inline it.

mradkov:
> Don't you have a plan to implement these permissions per domain? This would be much safer than the current implementation. Also, would be better to combine it with an [existing permission model](https://github.com/aeternity/superhero-wallet/blob/291f73720a1ded9727e553276fa562023b2a06ba/src/store/actions.js#L99-L106).
> 
> Currently is handles only popups of browser extensions. But there are also [iframe wallet requests](https://github.com/aeternity/superhero-wallet/blob/38fdb4a13115a30e94a87cc1a261006a82a9f42b/src/lib/wallet.js#L118-L146) and [deeplinks](https://github.com/aeternity/superhero-wallet/blob/1a8578bdf67c66c34393637dcc5e9ad866c17ba7/src/popup/router/pages/SignMessage.vue).

I agree. We should handle all possible wallet requests with this feature. Also good suggestion for extending the permissions model with the per domain permissions.

mradkov:
It's not possible to adjust the slider if you do not currently have positive balance. Shouldn't it be possible to do so, so you can predefine the limits when you setup your wallet, and then when you receive balance you do not have to do that again?

etharner:
> It's not possible to adjust the slider if you do not currently have positive balance. Shouldn't it be possible to do so, so you can predefine the limits when you setup your wallet, and then when you receive balance you do not have to do that again?

If you have 0 balance you can type any number in input on the right. Maybe we can hide slider if you have 0 balance and show it when you change the input. But if you have non-zero balance and enter big number in input what number should be the maximum for slider: your balance or typed number in input?

mradkov:
> > It's not possible to adjust the slider if you do not currently have positive balance. Shouldn't it be possible to do so, so you can predefine the limits when you setup your wallet, and then when you receive balance you do not have to do that again?
> 
> If you have 0 balance you can type any number in input on the right. Maybe we can hide slider if you have 0 balance and show it when you change the input. But if you have non-zero balance and enter big number in input what number should be the maximum for slider: your balance or typed number in input?

I think it will be good to have the slider always visible and allowing the user to specify the spend limit no matter if they have a balance or not. @onvisions any opinion on this?

etharner:
> > > It's not possible to adjust the slider if you do not currently have positive balance. Shouldn't it be possible to do so, so you can predefine the limits when you setup your wallet, and then when you receive balance you do not have to do that again?
> > 
> > 
> > If you have 0 balance you can type any number in input on the right. Maybe we can hide slider if you have 0 balance and show it when you change the input. But if you have non-zero balance and enter big number in input what number should be the maximum for slider: your balance or typed number in input?
> 
> I think it will be good to have the slider always visible and allowing the user to specify the spend limit no matter if they have a balance or not. @onvisions any opinion on this?

You can do that through input field
image

src/store/plugins/account.js
```diff
@@ -55,9 +56,14 @@ export default store =>
...
-        const fee = modal
-          ? aeToAettos(await dispatch('modals/open', confirmProps, { root: true }))
-          : txObject.fee;
+        const fee =
+          modal &&
+          (await checkPermissions({
+            method: 'transaction.sign',
+            params: { txObject: { params: { amount: txObject.amount } } },
+          }))
+            ? aeToAettos(await dispatch('modals/open', confirmProps, { root: true }))
```
davidyuk, line 65: Won't it be showing two modals one by one? Is it intended behaviour?

etharner: `checkPermissions` function doesn't show any modals, it's just one more check

src/popup/router/pages/PermissionsSettings.vue
```diff
@@ -0,0 +1,187 @@
...
+  },
+  async mounted() {
+    const storageLimitLeft = (await getLimitLeft())?.limitLeft;
+    this.limitLeft =
+      storageLimitLeft === undefined ? this.permissions.transactionSignLimit : storageLimitLeft;
+  },
+  methods: {
+    async changePermission(name, value) {
+      // eslint-disable-next-line no-restricted-globals
+      if (isNaN(value) || !isFinite(value) || value < 0) {
```
davidyuk, line 101: Proposing to avoid `eslint-disable` by using Number.isNaN (but it works a bit differently)

davidyuk (approved):

mradkov (approved):

PR #540, opened 2020-10-02
hotfix: Poll aepps in reverse iframe

davidyuk:
original issue: https://github.com/aeternity/jitsi-meet/issues/240

Before this change wallet was trying to connect to a new frame two times: immediately and after 3 seconds for the case, when aepp has no enough time to initialize after opening. For some reason, our fork of Jitsi Meet in some cases can't initialize in time.

I assume that initialization time heavily depends on the performance and internet connection quality of the end-user. So, instead of just to increase the timeout I'm proposing a more durable connection approach to a new frame: try to connect immediately and then poll frame until it disappears from DOM or sends any response.

### DIFF
--- a/src/lib/wallet.js
+++ b/src/lib/wallet.js
@@ -154,30 +154,44 @@ export default {
       });
 
       if (IN_FRAME) {
-        const connectedFrames = new Set();
-        const connectToFrame = target => {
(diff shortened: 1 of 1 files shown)


mradkov (approved):
The polling approach seems better. LGTM

PR #541, opened 2020-10-02
hotfix: Open deeplink callbacks in current tab even in iframe use case

davidyuk:
reverts #495 

The thing is that window.parent is always truthy:

> If a window does not have a parent, its parent property is a reference to itself.

https://developer.mozilla.org/en-US/docs/Web/API/Window/parent

Would be better to use IN_FRAME the next time. Forcing openUrl to use a new tab in the web version of wallet breaks deeplink flow. To address #493 I'm proposing explicitly specify that the explorer and ui should be opened in a new tab, but keeping the ability to open deeplink in the same tab that is used in Meet.

closes https://github.com/aeternity/jitsi-meet/issues/259

### DIFF
--- a/src/popup/router/components/ClaimTipButton.vue
+++ b/src/popup/router/components/ClaimTipButton.vue
@@ -20,7 +20,7 @@ export default {
   },
   methods: {
     openHowToClaimURL() {
-      openUrl(`${AGGREGATOR_URL}create-profile`);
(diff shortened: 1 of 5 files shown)


etharner (approved):
Agreed to that solution. At first I'd wanted to do that way but chosed `window.parent` as a shorter solution. But I haven't checked deeplinks though.

mradkov (approved):

PR #542, opened 2020-10-05
Realign develop

mradkov:


davidyuk (approved):

davidyuk:
As I know, we can't solve conflicts in this PR, because we shouldn't change the master branch.

PR #545, opened 2020-10-06
Fix names/setDefault

### LINKED ISSUE #517: Preferred `.chain` name is not in sync
After setting a preferred `.chain` name it is updated on the `superhero.com` (according to the response from the API), but when I get back to the home screen of the wallet - another name that I own is presented as the default.

CedrikNikita:
fixes #517


mradkov (approved):
LGTM. Thanks!

PR #555, opened 2020-10-19
Jdenticons and avatars are extracted as a service

stanislav-slavov:


davidyuk (commented):
I have made some futher refactoring in e64fd58397d0d21dedc93911c35b6d853f837388, consider applying changes from there

davidyuk (approved):

PR #519, opened 2020-09-17
Display fungible tokens

stanislav-slavov:
These changes provide visualization for the available fungible tokens on the network(test in this case), the balances. The user can choose select a token on the account page(for tipping) which would be introduced in a separate PR.

Screenshot 2020-09-18 at 10 48 02
Screenshot 2020-09-18 at 10 48 14
Screenshot 2020-09-18 at 10 48 33

PR #562, opened 2020-10-21
Refactoring

CedrikNikita:


mradkov:
I would like https://github.com/aeternity/superhero-wallet/pull/519 to be merged first and then apply these changes on-top.

src/store/plugins/observables.js
```diff
@@ -19,13 +20,34 @@ export default store => {
...
+            .catch(error => {
+              if (!isNotFoundError(error)) {
+                handleUnknownError(error);
+              }
+              return 0;
+            })
+            .then(result => convertToAE(result))
+        : Promise.resolve(0),
+    ),
+    multicast(new BehaviorSubject(0)),
```
davidyuk, line 38: The initial value of the previous balance implementation was persisted in localStorage (as I understand, to don't show 0 at startup). Shouldn't we consider this feature as important and keep it in this implementation?

src/store/plugins/observables.js
```diff
@@ -41,5 +63,17 @@ export default store => {
...
+    balance: balance$,
+    topBlockHeight: topBlockHeight$,
+    tokenBalance: balance$.pipe(
+      map(bal =>
+        store.state.current.token !== 0
+          ? store.state.tokens[store.state.current.token].balance
+          : bal,
+      ),
+    ),
+    balanceCurrency: balance$.pipe(map(bal => store.getters.currentCurrencyRate * bal)),
```
davidyuk, line 77: The `balanceCurrency` won't be updated if `getters.currentCurrencyRate` changed. Would be more accurate to subscribe to `currentCurrencyRate` using `watchAsObservable`

src/store/plugins/observables.js
```diff
@@ -41,5 +63,17 @@ export default store => {
...
   // eslint-disable-next-line no-param-reassign
-  store.state.observables = { notifications: notifications$ };
+  store.state.observables = {
+    notifications: notifications$,
+    balance: balance$,
+    topBlockHeight: topBlockHeight$,
+    tokenBalance: balance$.pipe(
+      map(bal =>
+        store.state.current.token !== 0
+          ? store.state.tokens[store.state.current.token].balance
```
davidyuk, line 73: The balance will remain polled even if the balance of the tokens is shown.

davidyuk (changes requested):
Proposing to extract obvious commits to a separate PR to merge them earlier.

src/store/plugins/observables.js
```diff
@@ -25,7 +25,7 @@ export default store => {
     entityType,
     ...other,
     sender,
-    chainName: store.state.chainNames[sender],
+    chainName: store.state.chainNames && store.state.chainNames[sender],
```
davidyuk, line 28:
```suggestion
    chainName: store.state.chainNames?.[sender],
```
this is not the same as https://github.com/aeternity/superhero-wallet/pull/562#discussion_r526400981

src/popup/router/pages/Send.vue
```diff
@@ -178,15 +159,22 @@ export default {
...
         to: '',
+        from: '',
         hash: '',
       },
+      wording: {
+        sender: this.$t('pages.send.sender'),
+        recipient: this.$t('pages.send.recipient'),
+        to: this.$t('pages.send.to'),
+        from: this.$t('pages.send.from'),
+        hash: this.$t('pages.send.hash'),
```
davidyuk, line 171: This approach makes it more difficult to track is wording actually used or not

src/popup/router/pages/Send.vue
```diff
@@ -50,23 +50,15 @@
...
+          <div
+            v-for="(value, key) in { sender: account.publicKey, recipient: form.address }"
+            :key="key"
+            class="info-group"
+          >
+            <label class="info-label">{{ wording[key] }}</label>
+            <span :data-cy="`review-${key}`" class="info-span" @click="openTxExplorer(value)">{{
+              value
+            }}</span>
           </div>
```
davidyuk, line 62: This looks ok but a more component-oriented approach would be to extract InfoGroup component and exactly specify fields without a cycle:
```suggestion
          <InfoGroup
            :value="account.publicKey"
            :label="$t('pages.send.sender')"
          />
          <InfoGroup
            :value="form.address"
            :label="$t('pages.send.recipient')"
          />
```
the necessity to pass `data-cy` is just crap, e2e tests shouldn't require that

davidyuk (approved):
Despite remarks, I think we can merge that

src/popup/router/pages/Send.vue
davidyuk: Actually, we may apply `data-cy` on InfoGroup and to make selectors on e2e side a bit more complex.

davidyuk (approved):

PR #565, opened 2020-10-21
Switch middleware to ae_mdw

etharner:


src/lib/wallet.js
```diff
@@ -16,26 +15,21 @@ import Logger from './logger';
...
-        {
-          in: 'path',
-          name: 'name',
-          required: true,
-          type: 'string',
-        },
-      ],
-    },
+
+  const fetchMiddleware = async (url, field) => (await fetchJson(url))[field];
```
davidyuk, line 19: let's add middleware url if it has `middleware` in the name:
```suggestion
  const fetchMiddleware = async (path, field) => (await fetchJson(`${middlewareUrl}/${path}`))[field];
```

src/lib/wallet.js
```diff
@@ -16,26 +15,21 @@ import Logger from './logger';
...
+  const middleware = {
```
davidyuk, line 20: Doesn't the new middleware have a swagger file? The convention of field names case was made by Swagger stamp and I would rather write the swagger file by myself than to rewrite all to snake_kase.

etharner: They has actually https://mainnet.aeternity.io/mdw/swagger/swagger.json, I'll rewrite this part

etharner:
A few thoughts about new mdw:
1) It will be good if they will include name hash in `getActiveNames` request, because I do another call to `sdk.api` to get the hash
2) I don't see the point of having `data` field in response if it's the only field in response

src/popup/router/pages/Names/Details.vue
```diff
@@ -73,11 +74,13 @@ export default {
...
     hasPointer() {
-      return getAddressByNameEntry(this.nameEntry);
+      return this.nameEntry.pointers.accountPubkey;
     },
   },
-  created() {
+  async created() {
+    await this.$watchUntilTruly(() => this.sdk);
     this.$store.dispatch('names/fetchOwned');
+    this.nameHash = (await this.sdk.api.getNameEntryByName(this.name)).id;
```
davidyuk, line 83: Would be more accurate to refetch the hash each time `this.name` changes, proposing to define this in the `watch` section.

src/store/plugins/names.js
```diff
@@ -50,37 +50,45 @@ export default store =>
...
-            };
-          });
+          names = names
+            .map(({ owner, createdAtHeight, expiresAt, pointers, info, name }) => ({
+              createdAtHeight: createdAtHeight || info.activeFrom,
+              expiresAt: expiresAt || info.expireHeight,
+              owner: owner || info.ownership.current,
+              pointers: pointers || info.pointers,
+              name,
+            }))
```
davidyuk, line 66: I'm not a fan of mixing the old and the new structures but do as you want.

skkw:
> 1. It will be good if they will include name hash in `getActiveNames` request, because I do another call to `sdk.api` to get the hash

I can squeeze it in this planning session, it's small fix.

src/lib/wallet.js
```diff
@@ -36,7 +82,11 @@ async function initMiddleware() {
       axiosError: () => '',
     },
   })({ swag });
-  store.commit('setMiddleware', middleware);
+
+  store.commit(
+    'setMiddleware',
+    Object.entries(middleware).reduce((m, [k, v]) => ({ ...m, [camelCase(k)]: v }), {}),
```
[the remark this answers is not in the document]
etharner: Actually this functionality implemented for parameters here https://github.com/aeternity/aepp-sdk-js/blob/develop/es/utils/swagger.js#L241 but this method is private and not added to `statics` in stamp

etharner:
I also think it would be great if mdw will return object with the same fields as the node does, because I'm doing syncing of fields in `fetchOwned` method.

src/lib/wallet.js
```diff
@@ -37,7 +70,12 @@ async function initMiddleware() {
...
+  );
+  console.log(Object.entries(middleware).reduce((m, [k, v]) => ({ ...m, [camelCase(k)]: v }), {}));
```
mradkov, line 78:
```suggestion
```

Let's cleanup the `console.log`?

mradkov (approved):
Looks good to me now. Thanks

PR #566, opened 2020-10-22
Improve tests

CedrikNikita:
Was using https://docs.cypress.io/guides/references/best-practices.html

### DIFF
--- a/tests/e2e/integration/account.js
+++ b/tests/e2e/integration/account.js
@@ -3,22 +3,23 @@ describe('Test cases for Account Page', () => {
     cy.login();
   });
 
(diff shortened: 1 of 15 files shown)


mradkov (approved):
The test improvements look good. What is the overal speed increase improvement?

davidyuk (commented):
I think more important to rewrite some e2e tests as unit tests. Here a lot of too specific e2e tests that may be written as unit tests that run much faster.

CedrikNikita:
> What is the overal speed increase improvement?

by 1.5 or 2 minutes

davidyuk (approved):

PR #571, opened 2020-10-27
Fix commenting of tips with id as string

davidyuk:


CedrikNikita (approved):

CedrikNikita:
It is also fixes https://github.com/aeternity/superhero-ui/issues/801

mradkov (approved):

PR #584, opened 2020-11-05
Add per-domain permissions for applications

etharner:


src/lib/rpcWallet.js
```diff
@@ -107,31 +105,20 @@ export default {
...
-      },
-    } = aepp;
-    return extractHostName(url);
+  getAeppHost(aepp) {
+    return new URL(aepp.connection.port.sender.url).host;
   },
   async shouldOpenPopup(aepp, action) {
-    if (!NO_POPUP_AEPPS.includes(this.getAeppOrigin(aepp)) || (await checkPermissions(action)))
+    if (
+      !NO_POPUP_AEPPS.includes(this.getAeppHost(aepp)) ||
```
davidyuk, line 113: Let's move `NO_POPUP_AEPPS` check to the side of permissions module.

src/lib/rpcWallet.js
```diff
@@ -107,31 +105,20 @@ export default {
...
-        port: {
-          sender: { url },
-        },
-      },
-    } = aepp;
-    return extractHostName(url);
+  getAeppHost(aepp) {
+    return new URL(aepp.connection.port.sender.url).host;
   },
   async shouldOpenPopup(aepp, action) {
```
davidyuk, line 111: Every method like `onSign` calls firstly `shouldOpenPopup` (wich calls `checkPermissions` from permissions module) and then calls `checkAeppPermissions` (wich calls `getAeppAccountPermission` from permissions module). Can we combine `shouldOpenPopup` and `checkAeppPermissions` into a single function extracted to permissions module?

davidyuk: The general idea: if we created a permissions module let's move all related stuff to there.

src/store/modules/permissions.js
```diff
@@ -41,18 +41,45 @@ export const checkPermissions = async ({ method, params }) => {
   return limitLeft === 0;
 };
 
+export const getAeppAccountPermission = async (host, account) => {
+  const { permissions } = await getState();
+  if (!Object.keys(permissions).length || !permissions[host]) return false;
```
davidyuk, line 46:
```suggestion
  if (!permissions?.[host]) return false;
```

davidyuk:
I have seen that you are using `.host` instead of `.hostname`, but for some reason, `localhost:8080` was shown as `localhost`, I think would be nice to preserve port as well.

src/store/migrations/02-set-permissions.js
```diff
@@ -0,0 +1,7 @@
+export default state => {
+  const newState = { ...state };
+
+  newState.permissions = {};
```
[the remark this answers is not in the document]
etharner: The problem is when user has previous permissions (he logined in 0.3.3) which had `address`, 'messageSign', 'transactionSignLimit` they'll have this in new hosts list, because it just takes all state of `permissions` module. I've decided to specially reset their permissions state to avoid that.

davidyuk:
One more idea is to drop `NO_POPUP_AEPPS` and have predefined permissions for these domains instead.

src/lib/wallet.js
```diff
@@ -138,43 +130,35 @@ export default {
         nativeMode: true,
         compilerUrl,
         name: 'Superhero',
-        async onConnection({ info: { icons, name } }, action, origin) {
-          const originUrl = toURL(origin);
+        async onConnection(aepp, action) {
+          // DO SMTH WITH URL
```
davidyuk, line 134: What is the difference between `origin` and aepp's url?

src/lib/wallet.js
```diff
@@ -138,43 +130,35 @@ export default {
...
-          } catch (error) {
-            action.deny();
-            if (error.message !== 'Rejected by user') throw error;
-          }
+          else action.deny();
         },
         onSubscription: acceptCb,
-        onSign: acceptCb,
+        onSign: (aepp, { accept }) => {
+          store.commit('addTxToHost', {
```
davidyuk, line 154: I can't find where it is used later

src/store/modules/permissions.js
```diff
@@ -37,18 +46,63 @@ export const checkPermissions = async (method, params = {}) => {
...
+    if (!permissionsHost) await addNewHost(host, address);
+  } catch {
+    return false;
+  }
+
+  return true;
+};
+
 export default {
   namespaced: true,
```
davidyuk, line 67:
```suggestion

  state: {
    'superhero.com': {
      // allow connection and transaction/message signing without confirmation
      // not mentioning a particular address
    },
    ...{ 'localhost:8080': ... }
  },
```
can we instead of `setNewHosts` do like this?

src/store/modules/permissions.js
```diff
@@ -37,18 +46,63 @@ export const checkPermissions = async (method, params = {}) => {
...
+    },
+    async setNewHosts({ state, rootState: { account }, commit }) {
+      if (!state)
+        [
+          'superhero.com',
+          ...(process.env.NODE_ENV === 'development' ? ['localhost'] : []),
+        ].forEach((host) => commit('addHost', { host, address: account.publicKey }));
+      Object.entries(await getNewHosts()).forEach(([host, [address]]) =>
+        commit('addHost', { host, address, limit: 1 }),
+      );
```
davidyuk, line 105: Don't you afraid to sync browser.storage and vuex manually?

src/lib/wallet.js
davidyuk: ok, I got how it works, but I the Base app I think we were handling this a way better

davidyuk (commented):
I think regular migrations is not the way we should proceed with this. I have some idea in mind, let me implement it above the develop in the next day. Proposing to switch @etharner to another task untill that.

src/popup/router/pages/PermissionsDetails.vue
```diff
@@ -0,0 +1,169 @@
...
+    ...mapMutations('permissions', ['togglePermission']),
+    setTransactionSignLimit(rawValue) {
+      const value = +rawValue;
+      this.transactionSignLimitError = Number.isNaN(value) || value < 0;
+      if (this.transactionSignLimitError) return;
+      this.$store.commit('permissions/setTransactionSignLimit', { host: this.host, value });
+    },
+    removePermissions() {
+      this.$store.commit('permissions/removeAeppPermissions', this.host);
+      this.$router.replace({ name: 'not-found' });
```
davidyuk, line 100: I got why it is needed, but won't be better to show 404 when navigating to an unexisting permission page like https://feature-applications-permissions.wallet.z52da5wt.xyz/settings/permissions/asdflkjhkljhk? It may be done by redirection to the 404 page in `created` hook if `host` does not exist in the permissions

src/store/modules/permissions.js
```diff
@@ -1,55 +1,94 @@
...
       const permissionsMethods = {
         'connection.open': 'address',
         'address.subscribe': 'address',
         'message.sign': 'messageSign',
       };
-      return state[permissionsMethods[method]];
+      return state?.[host][permissionsMethods[method]];
+    },
+    async requestAddressForHost({ state, commit, dispatch }, { host, address, connectionPopupCb }) {
+      if (state?.[host]?.addresses?.includes(address)) return true;
```
davidyuk, line 82:
```suggestion
      if (state.[host]?.addresses?.includes(address)) return true;
```

src/lib/wallet.js
```diff
@@ -133,68 +129,93 @@ export default {
...
+        } catch (error) {
+          action.deny();
+          if (error.message !== 'Rejected by user') throw error;
+        }
+      };
 
       const sdk = await RpcWallet.compose({
         methods: {
-          address: async () => store.getters.account.publicKey,
+          // eslint-disable-next-line no-shadow
```
davidyuk, line 162:
```suggestion
```
this is not shadowing anything right now

src/store/modules/permissions.js
```diff
@@ -1,55 +1,94 @@
...
-    address: true,
-    messageSign: true,
-    transactionSignLimit: 0,
-    transactionSignLimitLeft: 0,
-    transactionSignFirstAskedOn: null,
+    ...['superhero.com', 'localhost'].reduce(
+      (st, host) => ({ ...st, [host]: hostConfig([], true) }),
+      {},
+    ),
   },
```
davidyuk, line 27: probably it may be more redable using a new `Object.fromEntries` 🎉
```js
  state: Object.fromEntries(
    ['superhero.com', 'localhost'].map(domain => [domain, hostConfig([], true)])
  ),
```
so, no more `reduce` for collecting objects

davidyuk (approved):

mradkov (approved):

Liubov-crypto:
Tested, some options do not work correctly. 
I have set a limit of 2 ae tokens to spend without request per day, spent 12 ae, but they are not displayed in the Spent today field. I spent them on superhero UI as well as Youtube / Twitter:

perm set

Don't ask for message signing - works well.

davidyuk:
@Liubov-crypto ensure that there is a corresponding issue to don't forget about these bugs

Liubov-crypto:
> @Liubov-crypto ensure that there is a corresponding issue to don't forget about these bugs

moved to #733

PR #596, opened 2020-11-12
Update dependencies

davidyuk:


Liubov-crypto (commented):
Send tip - pass
Claim - pass
Activity - pass
Invite - pass
Tokens - pass
Top Up - pass
Withdraw - pass
Settings - pass
Names - extend - failed
Set name pointer - failed
Register a name - failed

Безіменний 24
Безіменний 25

Issue #600, opened 2020-11-16
Name extend error


mradkov:
Trying to extend a name throws this:
<img width="324" alt="Screenshot 2020-11-16 at 10 37 57" src="https://user-images.githubusercontent.com/13139371/99230262-cc203400-27f7-11eb-9eb4-94de2784debf.png">

Related to #592 &  #581

Issue #592, opened 2020-11-10
Set name pointer


Liubov-crypto:
The user can't set name pointer in SH wallet. 

Steps to reproduce the behaviour:
1. Go to sh wallet
2. Click on Names
3. Choose name
4. Click on Set name pointer
5. See an error:

![Знімок екрана 2020-11-10 о 12 27 00 пп](https://user-images.githubusercontent.com/69896204/98662152-0a2cdc00-2350-11eb-8423-24c74dd93ec4.png)

 - OS: windows 10
 - Browser: chrome,
 - Wallet Superhero 5d91965 / v.0.3.3

PR #603, opened 2020-11-16
Fix aens errors

### LINKED ISSUE #581: Can't register a new .chain name in the wallet
Steps to reproduce the behaviour:
1. Login into the wallet
2. Click on Burger Menu on Top
3. Click on Names
4. Click on the Register button
5. Enter a random name 
6. Click on +
7. See the error: 

Знімок екрана 2020-11-04 о 1 40 18 пп

This issue is reproducible in both web wallet and extensions (Chrome, FF).

Browser: chrome, Version 86.0.4240 (Mac OS)
OS: 10.15.7
Superhero ff63864 / v.0.4.0

etharner:
Fixes #581, #592, #600

### DIFF
--- a/src/lib/rpcWallet.js
+++ b/src/lib/rpcWallet.js
@@ -118,7 +118,10 @@ export default {
     return extractHostName(url);
   },
   async shouldOpenPopup(aepp, action) {
-    if (!NO_POPUP_AEPPS.includes(this.getAeppOrigin(aepp)) || (await checkPermissions(action)))
+    if (
+      !NO_POPUP_AEPPS.includes(this.getAeppOrigin(aepp)) ||
+      (await checkPermissions(action.method))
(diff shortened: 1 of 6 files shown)


mradkov (approved):
Changes are looking good.

I tried extending and it works as intended now.

Liubov-crypto (commented):
I tried to set the name pointer to a different name in my wallet but the button seems to be inactive:

Безіменний 12
Безіменний 13
Безіменний 17

The rest of the transactions with names work smoothly:
Register a name - pass
Name extend - pass

Liubov-crypto:
I understood why I cannot do extend, set name pointer and set as default with some names, apparently,  their blocks are already outdated: 
Безіменний 27

Checked, works with other names whose blocks are not out of date.

PR #607, opened 2020-11-17
permissions: Add panel to settings screen

### LINKED ISSUE #585: Permission settings are missed on the settings page
Screenshot 2020-11-06 at 14 59 27

etharner:
Closes #585.
Please correct wording if needed.

### DIFF
--- a/src/popup/router/pages/Settings.vue
+++ b/src/popup/router/pages/Settings.vue
@@ -17,6 +17,11 @@
         :info="$t('pages.settings.tabNetworksSmall')"
       />
       <PanelItem
+        to="/permissions-settings"
+        :title="$t('pages.titles.permissionsSettings')"
+        :info="$t('pages.settings.tabPermissionsSettingsSmall')"
+      />
+      <PanelItem
         @click="removeAccount"
         :title="$t('pages.settings.tabRemoveAccount')"
         :info="$t('pages.settings.tabRemoveAccountSmall')"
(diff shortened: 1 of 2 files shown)


src/popup/router/pages/Settings.vue
```diff
@@ -16,6 +16,11 @@
         :title="$t('pages.appVUE.networks')"
         :info="$t('pages.settings.tabNetworksSmall')"
       />
+      <PanelItem
+        to="/permissions-settings"
```
davidyuk, line 20: I would refer to the route by name, but it is ok to merge it as it is.

src/popup/router/components/PanelItem.vue
```diff
@@ -74,6 +74,7 @@ export default {
   .panel-item-info {
     color: #ccc;
     font-weight: normal;
+    line-height: normal;
```
davidyuk, line 77: Don't see a reason to add this.

davidyuk: | With | Without |
|---|---|
| Screenshot 2020-11-17 at 14 33 27 | Screenshot 2020-11-17 at 14 33 23 |

To fir to some phone screen?🤔

etharner: There are different line-height on button (Remove account) and a link (Permissions settings and upper panels). Browser set it by default for buttons and it rewrites our line height defined for html in https://github.com/aeternity/superhero-wallet/blob/develop/src/common/base.scss#L21, but for link it's not, so line-height became 1.5

mradkov: We anyway will have a different design for this page very soon, so I don't think we have to spend time refactoring it now.

etharner: @mradkov So, remove this line?

davidyuk (approved):

PR #608, opened 2020-11-17
List: prevent falling through checkbox in firefox

### LINKED ISSUE #605: When enabling/disabling name autoextend the name screen is shown
Browser: Firefox
Version:  Superhero ff63864 / v.0.4.0 

When clicking the auto-extend tickbox on a name which I own I'm sent to the screen with name details. When I click the tickbox I'm not expecting to be sent to another screen. Going back to my names it turns out the tickbox was indeed activated.

etharner:
Closes #605.

### DIFF
--- a/src/popup/router/pages/Names/List.vue
+++ b/src/popup/router/pages/Names/List.vue
@@ -22,7 +22,8 @@
         <span class="address">{{ owner }}</span>
         <CheckBox
           :value="autoExtend"
-          @input="value => $store.commit('names/setAutoExtend', { index, value })"
+          prevent
+          @click.native="$store.commit('names/setAutoExtend', { index, value: !autoExtend })"
         >
           {{ $t('pages.names.auto-extend') }}
         </CheckBox>


src/popup/router/pages/Names/List.vue
```diff
@@ -22,7 +22,8 @@
         <span class="address">{{ owner }}</span>
         <CheckBox
           :value="autoExtend"
-          @input="value => $store.commit('names/setAutoExtend', { index, value })"
+          prevent
+          @click.native="$store.commit('names/setAutoExtend', { index, value: !autoExtend })"
```
davidyuk, line 26:
```suggestion
          @input="value => $store.commit('names/setAutoExtend', { index, value })"
          @click.native.prevent
```

I think CheckBox's `prevent` shouldn't be used, can't it work this way? To use input event with CheckBoxes looks more semantic.

davidyuk: And let's don't put clickable inside clickable the next time...

etharner: I have this issue in firefox both in extension and web-wallet and I've tried this solution but it doesn't work 🤷

Liubov-crypto:
I can't reproduce this error. When I click on Auto Extend in my wallet it doesn't redirect me anywhere

mradkov:
> I can't reproduce this error. When I click on Auto Extend in my wallet it doesn't redirect me anywhere

Same here.

davidyuk (approved):
I can't reproduce this on macOS, but can in VM with Ubuntu😄 let's merge it as it is, and refactor CheckBox component and it's usages later.

PR #613, opened 2020-11-18
Remove unused options page

davidyuk:
To simplify migration to vue-cli.

How it was looking before:
Screenshot 2020-11-18 at 07 39 07
Screenshot 2020-11-18 at 07 39 05


src/options/options.html
```diff
@@ -1,12 +0,0 @@
-<!DOCTYPE html>
-<html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <title>Superhero - Options</title>
-  <link rel="stylesheet" href="options.css">
-</head>
-<body>
-  <div id="app"></div>
-  <!-- <script src="options.js"></script> -->
```
davidyuk, line 10: This options page is empty because js is disabled.

src/options/App.vue
```diff
@@ -1,18 +0,0 @@
-<template>
-  <div>
-    <!--eslint-disable-next-line vue-i18n/no-raw-text-->
-    <p>{{ $t('options.text') }}<a href="superhero.com">Superhero.com</a></p>
```
davidyuk, line 4: also, internationalization is not connected

mradkov (approved):

PR #621, opened 2020-11-19
Refactor CheckBox component

### LINKED ISSUE #616: Refactor CheckBox component and it's usages
See discussion in #608 

- drop `prevent` property (try to use `@click.native.prevent` outside)
- drop `val` property (use already defined `value` property instead)
- drop `checkedProxy` internal variable
- don't use v-model internally
- drop `checked` computed property
- ensure that it is compatible with v-model
- use `default` in property definition instead of `getType`

Useful links:
- https://vuejs.org/v2/guide/forms.html#v-model-with-Components
- https://vuejs.org/v2/guide/components.html#Using-v-model-on-Components
- https://vuejs.org/v2/guide/components-custom-events.html#Customizing-Component-v-model

CedrikNikita:
fixes #616

### DIFF
--- a/src/popup/router/components/CheckBox.vue
+++ b/src/popup/router/components/CheckBox.vue
@@ -1,14 +1,13 @@
 <template>
-  <label class="checkbox-container" @click.stop="handleClick">
+  <label class="checkbox-container">
     <input
       :disabled="disabled"
-      :value="val"
-      v-model="checked"
-      @change="onChange"
-      :type="getType"
+      :value="value"
+      :type="type"
       :name="name"
+      @change="$emit('input', $event.target.checked)"
     />
-    <span class="checkmark" :class="{ checked, disabled }" />
+    <span class="checkmark" :class="{ checked: value, disabled }" />
     <slot class="checkbox-holder" />
   </label>
 </template>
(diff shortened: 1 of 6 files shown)


src/store/modules/permissions.js
```diff
@@ -41,8 +41,8 @@ export default {
   namespaced: true,
 
   state: {
-    address: true,
-    messageSign: true,
+    address: false,
+    messageSign: false,
```
mradkov, line 45: That would mean turned off by default, right?

mradkov (approved):
Code changes look reasonable to me.

src/store/modules/permissions.js
CedrikNikita: yes, as they were before

davidyuk: Shouldn't the permission checking logic be inverted as well? Or it wasn't working properly before? cc: @etharner

davidyuk (approved):
The code looks ok, also I have tested it in Firefox

src/store/modules/permissions.js
etharner: @davidyuk Yes, it's inverted. I've asked for changes. I also propose to uninvert it in aepps permissions (this way false in store will mean "ask = _don't have_ permission) because it's messy to work with it

etharner (approved):

PR #642, opened 2020-11-22
Refactor SecuritySettings page

davidyuk:


src/common/base.scss
```diff
@@ -363,8 +363,6 @@ button {
   padding: 4px 20px;
   text-align: center;
   font-size: 16px;
-  word-break: break-all;
-  word-wrap: break-word;
```
davidyuk, line 367: It is a bad idea to have these styles global, but removing them may break something. I have checked all screens briefly, they look ok.

CedrikNikita (approved):
From code perspective changes looks good, from design side i would recommend more margins in verify section. @onvisions What do you think would be a better solution?

onvisions:
@CedrikNikita First we have to note that this version I'm seeing here is different in many ways from the new wallet design that can be seen in the Figma project. I know we discussed this but I repeat it just to avoid any confusion. 

Having in mind the above I'll suggest here just improvement design fix for what you have implemented because it really lacks space and "air". 

Font-size of the main text should  be changed from 16 to 15px. 
Font-weight of all button labels (text inside) and the font-weight of top bar screen title should be set to **font-weight:500;**.

As far as the margins are concerned I'll attach screenshots. 

Screenshot(1)
Screenshot(2)
Screenshot(3)

PR #643, opened 2020-11-22
Fix Avatar component

davidyuk:
| Before | After |
|---|---|
| Not rounded corners | Normal |
| Screenshot 2020-11-22 at 22 44 00| Screenshot 2020-11-22 at 22 43 55  |
| Shrinked | Normal |
| Screenshot 2020-11-22 at 22 41 27 | Screenshot 2020-11-22 at 22 41 32 |

Stan has changed Avatar's `object-fit` to `scale-down` for an [unknown reason](https://github.com/aeternity/superhero-wallet/commit/646f37321d1ef2e67b4fe45b04fd9cf348431bc8#diff-68c702f9f374859bd82baf6c6278fed4f219edbc1a17150ac08f7516f1d6c098R49). I'm guessing that Avatar was looking broken because of shrinking in a TokenListItem's flexbox. To fix this is more reasonable to disallow shrinking that I'm proposing to do in this PR.

related PRs: #572, #519

### DIFF
--- a/src/popup/router/components/Avatar.vue
+++ b/src/popup/router/components/Avatar.vue
@@ -46,7 +46,8 @@ $lg-size: 64px;
   border-radius: 50%;
   overflow: hidden;
   display: inline-block;
-  object-fit: scale-down;
+  object-fit: cover;
+  flex-shrink: 0;
 
   &.lg {
     height: $lg-size;


CedrikNikita (commented):
For my avatar change looks like this
| Before | After |
|---|----|
|image |  image|

My avatar is circle with a border in a white rect:
image

davidyuk:
[That avatar](https://raendom-backend.z52da5wt.xyz/profile/image/ak_2AVeRypSdS4ZosdKWW1C4avWU4eeC2Yq7oP7guBGy8jkxdYVUy) size is 130x128, it should be square to be displayed properly.

In general, `object-fit: cover` is a workaround, the avatar images should be cropped by the user to square on uploading. But on this side `object-fit: cover` fits the best because the avatar is not squashed and we don't have the empty space around it (as we have now).

stanislav-slavov (approved):

mradkov (approved):

PR #649, opened 2020-11-24
persistState: Explicitly mark that cloneDeep hack is needed for Firefox

davidyuk:
Hack explanation that was missed in https://github.com/aeternity/superhero-wallet/pull/108#discussion_r462054888:

> It's generally not possible to store other types, such as Function, Date, RegExp, Set, Map, ArrayBuffer, and so on. Some of these unsupported types will restore as an empty object, and some cause set() to throw an error.

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/set

I assume that `cloneDeep` converts reactive vue variables to plain objects that become compatible with Firefox's storage.

### DIFF
--- a/src/store/plugins/persistState.js
+++ b/src/store/plugins/persistState.js
@@ -1,10 +1,10 @@
 import { cloneDeep } from 'lodash-es';
+import { detect } from 'detect-browser';
 
 const KEY = 'state';
 
-const setState = async (state) => {
-  await browser.storage.local.set({ [KEY]: cloneDeep(state) });
-};
+const setState = (state) =>
+  browser.storage.local.set({ [KEY]: detect().name === 'firefox' ? cloneDeep(state) : state });
 
 export const getState = async () => (await browser.storage.local.get(KEY))[KEY];
 


src/store/plugins/persistState.js
```diff
@@ -1,10 +1,10 @@
 import { cloneDeep } from 'lodash-es';
+import { detect } from 'detect-browser';
 
 const KEY = 'state';
 
-const setState = async (state) => {
-  await browser.storage.local.set({ [KEY]: cloneDeep(state) });
-};
+const setState = (state) =>
+  browser.storage.local.set({ [KEY]: detect().name === 'firefox' ? cloneDeep(state) : state });
```
mradkov, line 7: Is it only Firefox that is problematic and requires `cloneDeep` to be used? And why not keep it consistent across browsers and stick to using `cloneDeep` for all of them?

davidyuk: > Is it only Firefox that is problematic and requires cloneDeep to be used?

I think so, also Firefox-based browsers can be problematic like Tor Browser, Pale Moon, and Waterfox

> And why not keep it consistent across browsers and stick to using cloneDeep for all of them?

Firstly, I want to have the explanation above be connected to git history. `cloneDeep` makes sense only for Firefox, also there needed not exactly the cloning, but proper evaluation of the Vuex state. Other browsers doing it by themselves, and without any mark (like firefox check) the current implementation looks like api misusage.

> Does cloneDeep introduce any performance drawbacks?

Nope.

mradkov: Sounds good. Thanks 👍

mradkov (approved):

Issue #560, opened 2020-10-20
Creating a new account after deleting an existing one


Liubov-crypto:
If you delete an existing account from the wallet and then want to create a new one, you get this error:

![Знімок екрана 2020-10-09 о 4 05 35 пп](https://user-images.githubusercontent.com/69896204/96580450-f5c16a80-12e0-11eb-9037-643788fe861b.png)

Steps to reproduce bug: 
1. Delete existing account
2. Close the wallet
3. Open wallet and create a new account
4. See the error.

Buttons in the popup are not responsive. User cannot create report and send via wallet.

Browser: Chrome, Brave
OS: Mac 10.15.7
superhero ui 0.2.1
superhero wallet 5d91965 / v.0.3.3

Liubov-crypto:
Also, the following error appears in firefox when performing the following steps: 
1. Delete existing account
2. Close the wallet
3. Open wallet and create a new account
4. See the error: 

![Знімок екрана 2020-10-21 о 5 21 22 пп](https://user-images.githubusercontent.com/69896204/96733755-7d32da80-13c2-11eb-942b-94b3ae5d2213.png)

Browser: FireFox
OS: Mac 10.15.7
superhero ui 0.2.1
superhero wallet 5d91965 / v.0.3.3

davidyuk:
well, the first one soould be fixed by 74e3da67ee79a74ac411b654bc5873aa0a019709
@mradkov can we make a new release?

CedrikNikita:
Fixed by this PR https://github.com/aeternity/superhero-wallet/pull/650

Issue #570, opened 2020-10-26
Creating account right after deleting previous doesn't work


CedrikNikita:
Steps to reproduce:
1. Create an account
2. Delete an account
3. Create an account (without closing the extension)
4. Close extension
5. Open extension; You don't have an account view, because you somehow doesn't create a new account
![Remove account](https://user-images.githubusercontent.com/9008074/97158351-054f2080-17c5-11eb-83f0-433cec6436ab.gif)

davidyuk:
probably it is the same bug #560

CedrikNikita:
Fixed by this PR https://github.com/aeternity/superhero-wallet/pull/650

PR #650, opened 2020-11-24
Simplify resetting by reloading page after that

### LINKED ISSUE #624: An error after deleting account
The error appears after I delete my account: 
2020-11-19 11 34 42 дп

An error when I submit a bug report to the wallet:
2020-11-19 11 34 48 дп

Superhero 366bffd/v.0.4.2 for Android

davidyuk:
fixes #624 

This approach we have been used in the Base aepp.

Pros:
- the most robust approach
- no need to test all new features against resetting (to make a correct resetting without reloading requires more accurate state handling)
- simplifies testing because not needed to use other tools (reinstall extension, incognito mode, android app settings) to completely reset the app state
- probably is more secure because all data is explicitly deleted and the app is restarted, so less probable to have something cached from the previous account

Cons:
- resettings happen slower because of page reload (shouldn't be critical because resetting assumed to be an edge case for the end user)


CedrikNikita (approved):
@davidyuk look at lgtb bot report. All the rest changes look good, actually this changes may close #560, #570 issues.

src/lib/rpcWallet.js
```diff
@@ -291,8 +291,6 @@ export default {
       browser.tabs.reload(aepp.connection.port.sender.tab.id);
       this.sdk.removeRpcClient(aepp.id);
     });
-    delete walletController.wallet;
-    this.initFields();
   },
   async [AEX2_METHODS.INIT_RPC_WALLET]({ address, network }) {
```
mradkov, line 295: Can we have a consistent function naming convention in this file?  Quoting this change: `async [AEX2_METHODS.LOGOUT]()` -> `async disconnect()`

davidyuk: I assuming `AEX2_METHODS` a deprecated object (mainly because it is not connected to AEX2, but message exchanging between background and popup) and planning to get rid of it in future PRs.

mradkov (approved):
Code changes look good.

Liubov-crypto (commented):
Logging into wallets after deleting an existing account and login in a new work smoothly. 

Found several bugs that this PR does not fix - synchronization of the wallet and superhero UI.
Does not upload user avatar and information to Superhero UI. Some data from my profile has disappeared. The avatar also disappeared from the wallet, although it was there before.

не загрузил аватар пользователя в брейв
пропала инфа и аватар пользователя
пропал аватар

I need to reload the page several times for an avatar and a photo on the cover to appear (there is no bio and location)
not sync info between sh UI and sh wallet

In the brave browser instead of the equivalent in currency, the wallet shows zeros:
нулевые суммы в брейв

Liubov-crypto:
Web wallet:
Send tip - pass
Claim tip-  was successful but the amount didn’t appear in my wallet
Activity - pass
Invite - pass
Withdraw - pass
Top up - pass
Names - pass

FF/Сhrome extension: 
Send tip- pass 
Claim tip- was successful but the amount didn’t appear in my wallet
Send tip via Twitter/Youtube button -passed, but the transaction is marked as pending in my wallet / from Twitter was successful.
Activity - pass
Top up -pass
Invite - pass
Withdraw - pass
Names - pass

Issue #641, opened 2020-11-20
Upon changing a FT balance the dropdown on the index page doesn't refresh


stanislav-slavov:

CedrikNikita:
fixed #652

Issue #629, opened 2020-11-19
Superhero walet performs an unbeliviable amount of height queries


gorbak25:
Go to wallet.superhero.com and open the dev console...
![image](https://user-images.githubusercontent.com/1352848/99664246-2de6d500-2a68-11eb-9485-dcc6e9ebfbed.png)
The height usually changes every 3 minutes -.-

davidyuk:
<img width="1387" alt="Screenshot 2020-11-19 at 20 30 38" src="https://user-images.githubusercontent.com/9007851/99701920-2d6d2f00-2aa6-11eb-8cd6-80eb91d2532e.png">

Well, I can't reproduce this. Can you specify the exact page that makes so lot of requests?

gorbak25:
@davidyuk I have a lot of infinitely pending transactions in my wallet: https://github.com/aeternity/superhero-wallet/issues/625

davidyuk:
That issue is closed, reopen it if the issue will persist in the next release

PR #652, opened 2020-11-24
Make topBlockHeight and balance observables

### LINKED ISSUE #646: Issue with claiming
While claiming tip was successful, the balance of my wallet has not changed (last transaction 11/17/20):

Безіменний 0004

I noticed that the wallet makes a transaction with a tip claim, increasing the amount. I sent 0.1 tokens, 0.2 is claimed (0.1 was already). Then I sent another 0.1 tokens, claimed already 0.3 tokens. It's strange. The balance in my wallet has not changed.

Superhero d311677 / v.0.4.2, 
Chrome 87.0.4280.67

CedrikNikita:
fixes #646, #641
potential fix for #629

### DIFF
--- a/src/popup/router/components/AmountSend.vue
+++ b/src/popup/router/components/AmountSend.vue
@@ -23,7 +23,7 @@
       <div class="balance-box">
         <p class="label">{{ $t('pages.tipPage.availableLabel') }}</p>
         <span class="secondary-text f-14 block l-1" data-cy="balance">
-          {{ selectedToken ? selectedToken.convertedBalance : tokenBalance }}
+          {{ selectedToken ? selectedToken.convertedBalance : tokenBalance.toFixed(2) }}
           {{ selectedToken ? selectedToken.symbol : $t('pages.appVUE.aeid') }}
         </span>
         <span class="f-14 block l-1 amount-currency" data-cy="balance-currency">
@@ -36,6 +36,7 @@
 </template>
 
 <script>
+import { pick } from 'lodash-es';
 import { mapGetters, mapState } from 'vuex';
 import Input from './Input';
 
@@ -44,8 +45,11 @@ export default {
     Input,
   },
   props: ['amountError', 'value', 'errorMsg', 'label'],
+  subscriptions() {
+    return pick(this.$store.state.observables, ['tokenBalance', 'balanceCurrency']);
+  },
   computed: {
-    ...mapGetters(['tokenBalance', 'balanceCurrency', 'formatCurrency']),
+    ...mapGetters(['formatCurrency']),
     ...mapState('fungibleTokens', ['selectedToken']),
     currencyAmount() {
       return ((this.value || 0) * this.$store.getters.currentCurrencyRate).toFixed(2);
--- a/src/popup/router/components/BalanceInfo.vue
+++ b/src/popup/router/components/BalanceInfo.vue
@@ -11,7 +11,7 @@
             is-custom
           />
           <span class="display-value text-ellipsis">{{
-            selectedToken ? selectedToken.convertedBalance : tokenBalance
+            selectedToken ? selectedToken.convertedBalance : tokenBalance.toFixed(2)
           }}</span>
           <span class="token-symbol">{{
             !selectedToken ? $t('pages.appVUE.aeid') : selectedToken.symbol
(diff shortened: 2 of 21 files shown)


src/popup/router/components/AmountSend.vue
```diff
@@ -23,7 +23,7 @@
       <div class="balance-box">
         <p class="label">{{ $t('pages.tipPage.availableLabel') }}</p>
         <span class="secondary-text f-14 block l-1" data-cy="balance">
-          {{ selectedToken ? selectedToken.convertedBalance : tokenBalance }}
+          {{ selectedToken ? selectedToken.convertedBalance : tokenBalance.toFixed(2) }}
```
mradkov, line 26: Why `.toFixed(2)`?

The fungible tokens can have as many decimal pieces as they want to. 🤔

src/popup/router/components/BalanceInfo.vue
```diff
@@ -11,7 +11,7 @@
             is-custom
           />
           <span class="display-value text-ellipsis">{{
-            selectedToken ? selectedToken.convertedBalance : tokenBalance
+            selectedToken ? selectedToken.convertedBalance : tokenBalance.toFixed(2)
```
mradkov, line 14: Same - why `toFixed(2)` 👆?

src/popup/router/components/AmountSend.vue
CedrikNikita: https://github.com/aeternity/superhero-wallet/pull/652/files#diff-42d961a783c297603d8cd3d38fb7471e62b81eb03a5ac50e7cb3124042e4b495L44-L48

mradkov: Ok. Then we need to discuss this functionality, but not in the scope of this PR.

src/store/plugins/observables.js
```diff
@@ -19,6 +20,31 @@ export default (store) => {
...
+            })
+            .then((result) => {
+              const balance = convertToAE(result);
+              if (balance !== store.state.balance) {
+                store.commit('updateBalance', balance);
+              }
+              return balance;
+            })
+        : Promise.resolve(store.state.balance),
+    ),
```
davidyuk, line 43: We can write it simpler filtering out the `null` values of sdk:
```suggestion
    switchMap((sdk) =>
      sdk.balance(store.state.account.publicKey).catch((error) => {
        if (!isNotFoundError(error)) {
          handleUnknownError(error);
        }
        return 0;
      }),
    ),
    map((balanceAettos) => {
      const balance = convertToAE(balanceAettos);
      if (balance !== store.state.balance) {
        store.commit('updateBalance', balance);
      }
      return balance;
    }),
```
So, proposing to rewrite `sdk$` from the above as:
```js
  const sdk$ = watchAsObservable(({ sdk }) => sdk, { immediate: true }).pipe(
    pluck('newValue'),
    filter((sdk) => sdk),
  );
```

src/store/plugins/observables.js
```diff
@@ -19,6 +20,31 @@ export default (store) => {
...
+    multicast(new BehaviorSubject(store.state.balance)),
```
davidyuk, line 44: The arguments to don't store the balance in store:
- developers may use the balance from the store not aware that it won't be updated without a subscription to the corresponding observable variable
- balance polling emits wasteful mutations that can complicate debugging
- the cached balance not needed to be handled the same careful (don't need to maintain in migrations) as other state fields persisted in local storage

Proposing to sync it with local storage explicitly, using a different key. For example, to get the balance by:
```js
JSON.parse(localStorage['rxjs']).balance
```
and to set it by:
```js
localStorage['rxjs'] = JSON.stringify({ ...JSON.parse(localStorage['rxjs']), balance })
```
even better would be to write a helper functions for that

src/store/plugins/observables.js
```diff
@@ -41,5 +67,28 @@ export default (store) => {
...
 
   // eslint-disable-next-line no-param-reassign
-  store.state.observables = { notifications: notifications$ };
+  store.state.observables = {
+    notifications: notifications$,
+    balance: balance$,
+    topBlockHeight: createSdkObservable(async (sdk) => (await sdk.topBlock()).height, 0),
+    tokenBalance: watchAsObservable(
+      (state) => ({
+        token: state.current.token,
```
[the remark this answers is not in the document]
CedrikNikita: Yeah, this PR is more than 1 month old, wallet got new changes

davidyuk (approved):

PR #671, opened 2020-11-27
Remove outdated tokens pages and store

davidyuk:
As I got, @stanislav-slavov is building fungible tokens support from scratch (for example, in #519 he created a new tokens vuex module instead of reusing the existing one), so not needed to keep the outdated implementation in the codebase. 

Also, the old implementation and a new one depends on different packages (aeternity-tokens@1.0.2 vs aeternity-fungible-token).


mradkov (approved):
In general, I like this cleanup.

Though let us think of how we can also allow users to add custom tokens (AEX9) of theirs that are currently not indexed by the word bazar.

stanislav-slavov (approved):

davidyuk:
> Though let us think of how we can also allow users to add custom tokens (AEX9) of theirs that are currently not indexed by the word bazar.

Isn't it a too specific thing for a wallet? Maybe better to make a separate aepp focused on the more experienced users?

mradkov:
> > Though let us think of how we can also allow users to add custom tokens (AEX9) of theirs that are currently not indexed by the word bazar.
> 
> Isn't it a too specific thing for a wallet? Maybe better to make a separate aepp focused on the more experienced users?

Nope, I would expect from a wallet to be able to work with any tokens according to that standard. As MetaMask can handle every ERC20 that you add - I would expect from superhero to allow me to add custom AEX9 tokens so I can use them with my wallet. 🤔

davidyuk:
Sure, I agree. I thought about deploying a token using superhero wallet.

mradkov:
> Sure, I agree. I thought about deploying a token using superhero wallet.

For deployment of a custom token - yeah let's leave that for the users(developers) to handle however they like without introducing additional explicit functionality for it.

PR #675, opened 2020-11-28
Extract TokenAmount component

davidyuk:
I'm proposing:
- to show all token amounts and fiat equivalents using the same component Screenshot 2020-11-28 at 17 45 16
- to hide insignificant zeroes, like `12.30 AE` => `12.3 AE`, and `0 AE (0 $)` => `0 AE`
- to refactor transaction list item, and token list item


CedrikNikita (approved):
Changes look reasonable, good clean up

Liubov-crypto (commented):
Checked - looks good. 

btw, when I tested this PR I got an error in the extension and web wallet: 

error on brave

Also, after connecting the extension and the superhero UI, the wallet on the superhero shows zero fiats. It is necessary to refresh the page for the equivalent in currencies to appear:

zeros on browsers ff:brave

davidyuk:
> btw, when I tested this PR I got an error in the extension and web wallet

I think this is a Middleware issue, I have created several, but there is no response yet: https://github.com/aeternity/ae_mdw/issues/104

> the superhero shows zero fiats

this should be fixed by https://github.com/aeternity/superhero-ui/pull/899

PR #676, opened 2020-11-28
Cleanup global styles

davidyuk:


CedrikNikita (approved):
The changes look reasonable, but some of them a little out of place.

Liubov-crypto (commented):
Tested - looks good. 

Note: not all fields are translated into Chinese (Tokens, Permissions settings). This bug is also on the developer version.

mradkov:
> Tested - looks good.
> 
> Note: not all fields are translated into Chinese (Tokens, Permissions settings). This bug is also on the developer version.

I wouldn't say it's a bug, it's just that we lack translations for all of the text on the wallet.

mradkov (approved):
Changes look good, though they seem to be touching lots of areas in the wallet. Let's maybe merge it after we finalize the other smaller PRs?

davidyuk:
Actually, the broadest change is to renaming `common` to `styles` that modifies a lot of components but it should be easy to rebase on top of that.

PR #686, opened 2020-12-01
calculateFee: Don't calculate max fee

davidyuk:


CedrikNikita (approved):
I don't understand what was the purpose to do it like it was before. The approach that this PR provides looks more reasonable.

PR #688, opened 2020-12-01
Send: Trim entered address

### LINKED ISSUE #612: Space after AE address invalidates withdraw transaction
When withdrawing through the wallet ext, having a space after the AE address makes it invalid.

CedrikNikita:
fixes #612


davidyuk (approved):
I like this fix because it uses built-in functionality. Also, I have tested it (and found another bug).

PR #682, opened 2020-11-30
Enable tip/retip using deeplinks with FT and v1/2/3 tips

stanislav-slavov:
This PR depends on https://github.com/aeternity/superhero-wallet/pull/640

PR #693, opened 2020-12-02
Refactor amount validation in Retip page

davidyuk:
inspired by #682


CedrikNikita (approved):
Also it is nice to have a proper error message.

PR #694, opened 2020-12-02
PermissionSettings: Simplify and fix validation

### LINKED ISSUE #690: Daily limit
if I  manually set the daily limit (Limit for spending without asking per day), then it is not saved, it disappears. 
If using a slider, it is saved. 

perm

This issue is reproduced on the development branch.
Mac OS 11.0.1
Superhero 3832d26 / v.0.4.2 
 Superhero ca07c27 / v.0.4.2

davidyuk:
closes #690 was broken in 770acec1f40e3874a480fa33d19ab338c679707e

### DIFF
--- a/src/popup/router/pages/PermissionsSettings.vue
+++ b/src/popup/router/pages/PermissionsSettings.vue
@@ -85,11 +85,11 @@ export default {
     async changePermission(name, value) {
       let newValue = value;
       if (name === 'transactionSignLimit') {
-        if (Number.isNaN(newValue) || !Number.isFinite(newValue) || newValue < 0) {
+        newValue = +value;
+        if (Number.isNaN(newValue) || newValue < 0) {
           this.error = true;
           return;
         }
-        newValue = +value;
         this.limitLeft = newValue;
         await setLimitLeft(newValue, undefined);
       }


src/popup/router/pages/PermissionsSettings.vue
```diff
@@ -85,11 +85,11 @@ export default {
     async changePermission(name, value) {
       let newValue = value;
       if (name === 'transactionSignLimit') {
-        if (Number.isNaN(newValue) || !Number.isFinite(newValue) || newValue < 0) {
```
davidyuk, line 88: @etharner did you used `Number.isFinite` to disallow typing `Infinity`? I think this is an edge case that we can ignore currently

etharner: It appears when user is typing very large number in input, but I agree we can omit this check

etharner (approved):

mradkov (approved):

Liubov-crypto (commented):
Tested, now limit settings work.

Btw, When I set a spending limit, sent tokens are not displayed in the "Spent Today" section.

perm2
sent

davidyuk:
@Liubov-crypto please create a separate issue for that bug

Liubov-crypto:
> @Liubov-crypto please create a separate issue for that bug

moved to #704

PR #707, opened 2020-12-04
Refactoring

CedrikNikita:


Liubov-crypto:
Tested, looks good. 
I tested it mainly on the testnet, because the connection to the mainnet does not work. 
Got small errors when generating an invite link:
invite link

when changing the network - from the testnet to mainnet:
error in networks FF

davidyuk (approved):
👍

PR #709, opened 2020-12-04
Refactor permissions, add Vuex to extension backend

davidyuk:


tests/e2e/integration/account.js
```diff
@@ -5,7 +5,7 @@ describe('Test cases for Account Page', () => {
 
   it('Checks dissapearing seed phrase backup button, opens/close sidebar, checks copy button, contains a claim name message', () => {
     cy.get('.seed-backup-notification')
-      .should('not.be.visible')
+      .should('be.visible')
```
davidyuk, line 8: I can't get why it shouldn't be visible. That tests using a usual [login command](https://github.com/aeternity/superhero-wallet/blob/47a0ad7262e2781c057d51651771a8ac8eda8dde/tests/e2e/integration/account.js#L3). This command calls [`getLoginState`](https://github.com/aeternity/superhero-wallet/blob/e215310d2a83eb87e99aff3efa86e7755c12333a/tests/e2e/support/commands.js#L98) with almost default arguments. That function without an agument switches [`backedUpSeed`](https://github.com/aeternity/superhero-wallet/blob/ddbb32b724833f3a69164e470709bb557bebb812/src/popup/utils/index.js#L23) to falsy. So, a notification should be displayed. I have tested on master branch, test passes with both `not.be.visible` and `be.visible`, probably the first one passes because of some delay in rendering.

Liubov-crypto (commented):
When I tested I set a limit of 1 ae, sent 2.2 ae and nothing changed. I do not see tokens in spent today section. The wallet does not ask me for some permission when I spend the amount over the daily limit: 

perms
am

etharner:
I propose this fix for transaction sign limit check problem which @Liubov-crypto mentioned

etharner (approved):

PR #713, opened 2020-12-07
ImportAccount: Show only Loading component on loading

CedrikNikita:
I propose to change behaviour of loading in the ImportAccount page
|Before|After|
|---|---|
|impot|import2|


davidyuk (approved):

davidyuk:
I've broke it in one of the previous refactorings

PR #715, opened 2020-12-07
Remove all !important modifiers

CedrikNikita:


davidyuk (approved):
the code changes look fantastic 🎉

src/styles/global.scss
```diff
@@ -80,12 +80,31 @@ p {
...
 .sendAmount .ae-dropdown .ae-dropdown-button {
-  width: 100% !important;
-  height: 16px !important;
+  width: 100%;
+  height: 16px;
   font-size: 1.3rem;
 }
 
-.loading {
+.v-tour--active .v-tour__target--highlighted {
```
davidyuk, line 88: As I understand it is only referenced in a single component. So, I'm proposing to keep it locally, but not `scoped` (or to use v-deep immediately).

davidyuk (approved):

src/styles/global.scss
CedrikNikita: v-deep in this case won't work, but i agree that locally is prefered variant

mradkov (approved):

PR #716, opened 2020-12-08
Add new designs for notifications

etharner:
@onvisions please check how it looks like because applying of new Header designs may be confusing on pages with old designs etc.
@davidyuk advices about what to do with `ThreeDotsMenu` and `SmallModal` are welcomed


src/store/mutations.js
```diff
@@ -74,6 +74,7 @@ export default {
       {
         ...payload,
         wallet: true,
+        type: 'wallet',
```
CedrikNikita, line 77:
```suggestion
        wallet: true,
        type: 'wallet',
```
Why do we need two properties to verify is notification from wallet or not?

src/popup/locales/en.json
```diff
@@ -599,6 +599,7 @@
     },
     "notifications": {
       "new": "New",
+      "all": "All",
       "wallet": "Wallet",
       "superhero": "Superhero.com",
       "read": "Read",
```
CedrikNikita, line 605:
```suggestion
```
Since there would be no `Read` tab we may delete it

CedrikNikita (changes requested):
`Claimed your retip` notifications is not shown in `Superhero.com` tab.

Three dots menu is moved around dependiong on the length of the name, plus it have different color
image

src/popup/locales/en.json
```diff
@@ -599,6 +599,7 @@
     },
     "notifications": {
       "new": "New",
```
CedrikNikita, line 601:
```suggestion
```
Since there would be no `New` tab we may delete it

src/store/mutations.js
onvisions: @CedrikNikita @maxkpower

Yes, not visible difference between **read** and **unread** notifications because it's not implemented according to design provided. I think on the mockup below the "unread" are clearly visible.
And yes, the **...** menu icon should be right aligned with the date and the padding should be always 16px to the right. Sync with me if you'd like to have the source file where styling and measures can be taken from.

Group 1591

davidyuk (commented):
> advices about what to do with ThreeDotsMenu and SmallModal are welcomed

I don't see any problem right now, we may refactor them later if it is needed.

src/filters/toRelativeTime.js
```diff
@@ -0,0 +1,16 @@
+import { i18n } from '../popup/utils/i18nHelper';
+
+export const toRelativeTime = (minutes) => {
```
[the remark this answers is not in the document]
davidyuk: the `toRelativeTime` fits better if it accepts a Date instance or a timestamp

src/popup/router/pages/Intro.vue
```diff
@@ -138,6 +138,10 @@ export default {
         privateKey: seed,
       };
       await this.$store.dispatch('setLogin', { keypair });
+      this.$store.commit('addNotification', {
+        text: this.$t('pages.account.seedNotification', [this.$t('pages.account.backup')]),
+        path: '/securitySettings',
+      });
```
davidyuk, line 144: Was it intended to show this only the first time, not on each opening as it was before?

src/store/actions.js
```diff
@@ -153,6 +153,21 @@ export default {
 
     return backendMethod(respondChallenge);
   },
+  async modifyNotifications({ state: { sdk }, getters: { activeNetwork } }, [ids, status, author]) {
```
davidyuk, line 156: isn't author always the same? Proposing to do something like this:
```suggestion
  async modifyNotifications({ state: { sdk, account: { publicKey } }, getters: { activeNetwork } }, [ids, status]) {
```

davidyuk (approved):

CedrikNikita (commented):
The wallet moving on changing the tab, witch is weird behaviour.
Moving

And this is actually a bad behaviour, but it's a small case.
Overlay

mradkov (approved):
Code changes look good.

PR #717, opened 2020-12-08
Fix firefox design related issues

### LINKED ISSUE #698: FT design bug in FF browser
I tested FT transactions and found some strange white lines in the FF extension:

design FF FT
ff design
ff design 2

 Superhero b500d03 / v.0.4.2

CedrikNikita:
fixes #698

### DIFF
--- a/src/popup/router/components/Modals/SignModal.scss
+++ b/src/popup/router/components/Modals/SignModal.scss
@@ -3,7 +3,6 @@
 .confirm-tx-sign-modal.modal .container {
   width: 100%;
   height: 100%;
-  overflow-y: scroll;
   padding: 0;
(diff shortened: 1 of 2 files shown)


src/popup/router/components/Modals/SignModal.scss
```diff
@@ -6,6 +6,7 @@
   overflow-y: scroll;
```
davidyuk, line 6: won't changing this to
```suggestion
  overflow-y: auto; // or even removing this
```
will work the same?

CedrikNikita: Yes, you are right

davidyuk (approved):

PR #721, opened 2020-12-09
Fix sdk of wallet in reverse frame

davidyuk:
Fixer error:
> [vuex] state field "sdk" was overridden by a module with the same name at "sdk"

The overlapping of `state.sdk` variable and `sdk` module is completely breaks sdk in iframe mode.


etharner (approved):

CedrikNikita (approved):
This is really fixing iframe wallet in superhero ui.

mradkov (approved):

PR #722, opened 2020-12-10
Make back button go up in the hierarchy

### LINKED ISSUE #465: To make the back button go up as it was before
Currently, the back button in Header navigates the user at the previous page instead of the parent one (account or intro) as it was before. I think the previous behavior was more useful because the current one already implemented by system controls in all environments:
- the traditional back button in desktop/mobile browsers
- a hardware/software back button at the bottom in Android
- a left swipe to go back in iOS

### LINKED ISSUE #644: Back button from Names should go straight to dashboard
If I go to the Names section and click between Your Names, Auctions, and Register, the back button remembers this and serves more like an Internet back button. So if I've clicked between the three tabs 20 times, I would have to click the back button 21 times to go back to the dashboard/home screen. 

Back button should just go back to Home.

CedrikNikita:
fixes #644,
fixes #465

### DIFF
--- a/src/popup/App.vue
+++ b/src/popup/App.vue
@@ -101,7 +101,7 @@ export default {
     if (!this.backedUpSeed) {
       this.$store.commit('addNotification', {
         text: this.$t('pages.account.seedNotification', [this.$t('pages.account.backup')]),
-        path: '/securitySettings',
+        path: '/settings/security',
       });
     }
     if (this.$store.state.notificationSettings.length === 0) {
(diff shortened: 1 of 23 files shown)


davidyuk (approved):

davidyuk:
@Liubov-crypto Here were massive changes in the router, briefly ensure that all pages displaying correctly.

Liubov-crypto (commented):
I do not see my names in the wallet (this may be due to a problem with connecting to the mainnet network)

no names found

Not all tokens are displayed in the FF/Chrome extension, only ae tokens. Also, more decimal numbers are displayed: 
only ae in ff extension


while I see other tokens in web wallet: 
webwallet mainnet

White stripes in FF extension: 
white stripes in FF ext

In general, the design of the pages has not changed.

PR #737, opened 2020-12-18
Fix migration and iframe permissions

etharner:


src/store/migrations/02-reset-permissions.js
```diff
@@ -1,5 +1,7 @@
+import { defaultDomains } from '../modules/permissions';
+
 export default async (state) => {
   const newState = { ...state };
-  newState.permissions = {};
+  newState.permissions = defaultDomains;
```
davidyuk, line 5: I think migrations immediately become a legacy code and they should be written once to avoid an unpredictable state if it was edited afterward. The extraction of common code doesn't work well here, because:
- you can't edit the extracted part because it requires an additional migration instead
- the export of `defaultDomains` looks unobvious because it wouldn't be necessary for a usual flow (without migrations)

So, I'm proposing to inline the value of `defaultDomains` here despite that it would lead to duplicate code

src/lib/wallet.js
```diff
@@ -186,9 +199,6 @@ export default {
             }
             return address;
           },
-          sign: (data) => store.dispatch('accounts/sign', data),
-          signTransaction: (txBase64, opt) =>
-            store.dispatch('accounts/signTransaction', { txBase64, opt }),
```
davidyuk, line 191: This removes the confirmation of generation the invite link, is it an intended change?

davidyuk (approved):

mradkov (approved):

PR #738, opened 2020-12-18
Update available token balances upon switching networks

stanislav-slavov:


src/store/modules/fungibleTokens.js
```diff
@@ -74,10 +77,17 @@ export default {
             updatedTokenInfo[contract] = { ...objectStructure };
             commit('setAvailableTokens', updatedTokenInfo);
           }
-
+          if (selectedToken && selectedToken.contract === objectStructure.contract) {
+            commit('setSelectedToken', null);
+            commit('setSelectedToken', objectStructure);
```
davidyuk, line 82: I can’t get why it’s called two times

mradkov: First, because it's a hotfix.

The problem is when you switch between networks the fungible tokens from the testnet were staying persisted in the dropdown with your balances, where your mainnet tokens should be.

This is checking if the current token is Ae and not an FT - if its an FT then we can reset the selection (this is when you have a token selected from Testnet and switch to Mainnet).

mradkov: It's not pretty but it fixes the problem. And if we have someone to take care of the CORS issue with mainnet middleware this is fine.

davidyuk: 👌

mradkov (approved):

Issue #723, opened 2020-12-10
Unable to connect to chosen node


Liubov-crypto:
I got an error while trying to connect to the mainnet network:

![connection with node](https://user-images.githubusercontent.com/69896204/101758078-96adf400-3ae0-11eb-86d7-757fd1c7f0bc.png)

FF browser
Superhero 1cafb53 / v.0.4.2 

This data is taken from a web wallet in the Safari browser:
![safari web wallet](https://user-images.githubusercontent.com/69896204/101759014-bbef3200-3ae1-11eb-9847-f2122b9c4c10.png)

davidyuk:
@CedrikNikita just ensure that this problem is not on the side of wallet

Liubov-crypto:
> @CedrikNikita just ensure that this problem is not on the side of wallet

should I move this issue to another section?

CedrikNikita:
That what is goind on
```
Access to fetch at 'https://mainnet.aeternity.io/mdw/swagger/swagger.json'
from origin 'https://develop.wallet.z52da5wt.xyz' has been blocked by CORS policy: No 'Access-Control-Allow-Origin'
header is present on the requested resource.
If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

popup.js:1 GET https://mainnet.aeternity.io/mdw/swagger/swagger.json net::ERR_FAILED
```

CedrikNikita:
Current version of wallet doesn't have such problems. Feel free to reopen this issue, if it will pop up again.

PR #741, opened 2020-12-19
Fungible Tokens: Middleware CORS issue workaround

mradkov:
This is a temporary solution to handle the problem with #723 which prevents some of the functionality with Fungible Tokens in the wallet to work properly.

### DIFF
--- a/src/popup/router/components/BalanceInfo.vue
+++ b/src/popup/router/components/BalanceInfo.vue
@@ -10,7 +10,7 @@
           is-custom
         />
         <span class="display-value text-ellipsis">
-          {{ selectedToken ? selectedToken.convertedBalance : tokenBalance.toFixed(2) }}
+          {{ selectedToken ? selectedToken.convertedBalance : tokenBalance }}
         </span>
         <span class="token-symbol">{{ !selectedToken ? $t('ae') : selectedToken.symbol }}</span>
         <ExpandedAngleArrow class="expand-arrow" />
--- a/src/popup/router/pages/Retip.vue
+++ b/src/popup/router/pages/Retip.vue
@@ -33,6 +33,7 @@
 <script>
 import { pick } from 'lodash-es';
 import { mapGetters, mapState } from 'vuex';
+import BigNumber from 'bignumber.js';
 import tipping from 'tipping-contract/util/tippingContractUtil';
 import { MAGNITUDE, calculateFee, TX_TYPES } from '../../utils/constants';
 import { convertToken } from '../../utils/helper';
@@ -47,7 +48,7 @@ export default {
   components: { AmountSend, UrlStatus, Button, BalanceInfo },
   data: () => ({
     tip: {},
-    amount: '',
+    amount: null,
     loading: false,
   }),
   subscriptions() {
@@ -62,6 +63,7 @@ export default {
       urlStatus(state, getters) {
         return getters['tipUrl/status'](this.tip.url);
       },
+
       validationStatus({ sdk }, { account, minTipAmount }) {
         if (!sdk || !this.tippingContract) {
           return { error: true };
@@ -69,7 +71,10 @@ export default {
         if (this.selectedToken && this.$route.query.id.includes('_v1')) {
           return { error: true, msg: this.$t('pages.tipPage.v1FungibleTokenTipError') };
         }
-        if (!this.selectedToken && this.amount < minTipAmount) {
+        if (!+this.amount) {
+          return { error: true, msg: this.$t('pages.tipPage.requiredAmountError') };
+        }
+        if (!this.selectedToken && +this.amount < minTipAmount) {
           return { error: true, msg: this.$t('pages.tipPage.minAmountError') };
         }
         const fee = calculateFee(TX_TYPES.contractCall, {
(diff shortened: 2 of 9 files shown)


Liubov-crypto (commented):
Checked, in general LGTM. 
Found this flying bug in FF extension when switching from mainnet to the testnet:

connecting to the testnet

src/popup/router/pages/Retip.vue
```diff
@@ -62,25 +62,31 @@ export default {
...
           callerId: account.publicKey,
         });
+        const calculatedMaxValue = this.balance > fee ? this.balance.minus(fee) : 0;
         if (
           this.selectedToken
-            ? this.selectedToken.convertedBalance < this.amount || this.balance < fee
-            : this.balance < fee + this.amount
+            ? convertToken(this.selectedToken.balance, -this.selectedToken.decimals) <
+                this.parsedAmount || this.balance < fee
+            : calculatedMaxValue - this.parsedAmount <= 0
```
davidyuk, line 89: I can't get the reasoning of this refactoring, the previous version was looking better. Did you try to do it more accurately?

src/popup/router/pages/Retip.vue
```diff
@@ -92,6 +98,14 @@ export default {
         ? this.tippingV2
         : this.tippingV1;
     },
+    parsedAmount() {
+      return this.amount
+        ? convertToken(
+            this.amount * 10 ** (this.selectedToken ? this.selectedToken.decimals : MAGNITUDE),
+            -(this.selectedToken ? this.selectedToken.decimals : MAGNITUDE),
+          )
+        : 0;
```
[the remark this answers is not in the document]
davidyuk: Actually, why is it needed at all?

src/store/plugins/observables.js
```diff
@@ -38,7 +38,7 @@ export default (store) => {
       }),
     ),
     map((balanceAettos) => {
-      const balance = convertToAE(balanceAettos);
+      const balance = convertToken(balanceAettos, -18);
```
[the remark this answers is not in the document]
davidyuk: I know it is not going to change, but I would use `MAGNITUDE` instead of `-18` for clarity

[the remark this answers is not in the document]
davidyuk: or even use `aettosToAe`

src/popup/router/pages/Tip.vue
```diff
@@ -115,6 +115,14 @@ export default {
     tippingContract() {
       return this.tippingV2 || this.tippingV1;
     },
+    parsedAmount() {
+      return this.amount
+        ? convertToken(
+            this.amount * 10 ** (this.selectedToken ? this.selectedToken.decimals : 18),
+            -(this.selectedToken ? this.selectedToken.decimals : 18),
+          )
+        : 0;
```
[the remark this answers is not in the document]
davidyuk: Also, let's rename it to `amountBigNumber`, because actually, here is no any parsing.

src/popup/router/pages/Tip.vue
```diff
@@ -140,10 +148,12 @@ export default {
           contractId: this.tippingContract.deployInfo.address,
           callerId: account.publicKey,
         });
+        const calculatedMaxValue = this.balance > fee ? this.balance.minus(fee) : 0;
```
davidyuk, line 151: I think this is an extra variable, the ternary operator also not needed.

src/popup/router/pages/Send.vue
```diff
@@ -253,15 +257,17 @@ export default {
           const { hash } = await this.$store.dispatch('fungibleTokens/transfer', [
             receiver,
             this.form.amount,
-            { waitMined: false, modal: false },
+            { waitMined: true, modal: false },
```
davidyuk, line 260: Are you sure want use to wait for the transaction? This is inconsistent with a usual sending

mradkov: Yes, I want to wait for it for FT, because currently we are using the list of available tokens per account from the backend and the cache is not fast enough. Which fires additional errors if not awaited.

Once MDW has the capability to track that for us - we will remove it.

src/popup/router/components/BalanceInfo.vue
```diff
@@ -10,7 +10,13 @@
           is-custom
         />
         <span class="display-value text-ellipsis">
-          {{ selectedToken ? selectedToken.convertedBalance : tokenBalance.toFixed(2) }}
+          {{
+            selectedToken
+              ? selectedToken.convertedBalance
+              : typeof tokenBalance === 'object'
+              ? tokenBalance.toFixed(2)
+              : tokenBalance
```
davidyuk, line 18: This is because `tokenBalance` is a sting before the actual balance arrives, I'm proposing to convert it to BigNumber on loading from `localStorage` or to keep it Number as it was before.
So:
```suggestion
              : tokenBalance.toFixed(2)
```
Or:
```suggestion
              : tokenBalance
```

src/popup/router/pages/Retip.vue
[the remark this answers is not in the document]
stanislav-slavov: For proper value checks.

src/popup/router/pages/Retip.vue
stanislav-slavov: I did it more accurate, yes.

stanislav-slavov (approved):

PR #747, opened 2020-12-22
wallet.js: Affect args to make it inserted into function

etharner:
This is an interesting bug(?) related to how babel transpile await functions https://github.com/babel/babel/issues/4219.

As I understand, if you spread `...args` in async function it will only be inserted if it is used somehow inside. That's strange, because we had getting last element of args array `args[args.length - 1]`, but transpiler hasn't inserted it, so when I've changed it to `args.pop()` it worked 🤷.

Because of this bug `address` permissions weren't working correctly, so I propose to include it to release if it pass @davidyuk review.

### DIFF
--- a/src/background/wallet.js
+++ b/src/background/wallet.js
@@ -74,7 +74,7 @@ export default {
           },
           async address(...args) {
             const address = store.state.account.publicKey;
-            const app = args[args.length - 1];
+            const app = args.pop();
             if (
               app instanceof App &&
               !(await store.dispatch('permissions/requestAddressForHost', {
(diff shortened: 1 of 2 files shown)


davidyuk (commented):
I'm fine to approve it if you find exactly the same issue or add one to the babel repo.

etharner:
I've reproduced this issue in babel standalone environment and created an issue https://github.com/babel/babel/issues/12552

davidyuk (approved):

mradkov (approved):

Issue #754, opened 2020-12-24
Fix insufficient balance check


davidyuk:
if balance - amount - fee is equal to 0 this means that we have enough funds, don't we?

This typo is seen in
https://github.com/aeternity/superhero-wallet/pull/741/files#diff-1a443786f57caf4444a2d47ba9eb8dcbdee7c2d2bfe4165d2889f65825771fe4R89
https://github.com/aeternity/superhero-wallet/pull/741/files#diff-dddf8f9c6093ce6ad4d6362f5c501e27942738cba5cd6353a3ec419dae66e57dR240
https://github.com/aeternity/superhero-wallet/pull/741/files#diff-7fe7768b1768e442dfe067deb4ef02b3074e8d9a20d5a3699e71dabd5f9746f9R151

cc: @stanislav-slavov

PR #756, opened 2020-12-24
Make amount a string instead of BigNumber

### LINKED ISSUE #752: balance observable should have a single type
https://github.com/aeternity/superhero-wallet/pull/741/files#diff-cb741da311f57d7eedca15dbc87add620536ad2bf015d2cf03a969754bbb6e85

Currently, it is a string after loading from localStorage and a BigNumber after fetching the current balance from the node.

cc: @stanislav-slavov

### LINKED ISSUE #753: Don't use BigNumber or use it with it's own methods instead of usual operators
https://github.com/aeternity/superhero-wallet/pull/741/files#diff-1a443786f57caf4444a2d47ba9eb8dcbdee7c2d2bfe4165d2889f65825771fe4

cc: @stanislav-slavov 

`new BigNumer(4) + new BigNumber(5)` will be converted to `4 + 5` firstly

so, the correct way to do it is:
```
new BigNumer(4).plus(new BigNumber(5)) // or new BigNumer(4).plus(5) (5 will be converted internally)
```
the same works for other numeric operators in JS, see [BigNumber docs](https://mikemcl.github.io/bignumber.js/) for corresponding methods

CedrikNikita:
part of #754, fixes #753, fixes #752

### DIFF
--- a/src/popup/router/components/AmountSend.vue
+++ b/src/popup/router/components/AmountSend.vue
@@ -23,7 +23,7 @@
       <div class="balance-box">
         <p class="label">{{ $t('pages.tipPage.availableLabel') }}</p>
         <span class="secondary-text f-14 block l-1" data-cy="balance">
-          {{ selectedToken ? selectedToken.convertedBalance : tokenBalance }}
+          {{ selectedToken ? selectedToken.convertedBalance : tokenBalance.toFixed(2) }}
           {{ selectedToken ? selectedToken.symbol : $t('ae') }}
         </span>
         <span class="f-14 block l-1 amount-currency" data-cy="balance-currency">
--- a/src/popup/router/components/BalanceInfo.vue
+++ b/src/popup/router/components/BalanceInfo.vue
@@ -10,7 +10,7 @@
           is-custom
         />
         <span class="display-value text-ellipsis">
-          {{ selectedToken ? selectedToken.convertedBalance : tokenBalance }}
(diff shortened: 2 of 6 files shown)


Liubov-crypto:
In the drop-down line, there is a big number instead of a short one both in the mainnet and in the testnet.

pr1
testnet error

I haven't used the wallet for a few minutes, then I came back and found this problem:

erro leave for a while
error in wallet

I also checked some transactions, but most of them failed due to compiler error:
Activity - failed
Send Tip - failed
Claim Tip - failed
Withdraw - pass
Invite link - pass
Settings: 
Language - pass
Security&Privacy - pass
Networks - failed
Remove account - pass

davidyuk:
> In the drop-down line, there is a big number instead of a short one both in the mainnet and in the testnet

It was introduced before, proposing to fix it in the future releases

Liubov-crypto:
Send Tip - pass
Claim tip - pass
Withdraw - pass
Invite link - pass
Settings: 
Language - pass
Security&Privacy - pass
Networks - pass
Withdraw - pass, via scan QR code - pass
Activity - pass
Top Up - pass
Names- pass
Invite link - pass
Settings: 
Language - pass
Security&Privacy - pass
Networks - pass
Remove account - pass

Got this flying bug in web wallet: 
web wallet issue

davidyuk (approved):

PR #761, opened 2020-12-29
Display big amount transactions properly

stanislav-slavov:


davidyuk (approved):
Ok, but the date still looks misaligned:

Screenshot 2020-12-30 at 23 05 36

we have to ask @onvisions about this. Also, proposing to solve it by a more general approach: #766

mradkov (approved):

PR #763, opened 2020-12-30
Removed scrollbar of texarea in  SuccessTip page on FF

stanislav-slavov:


davidyuk (commented):
Screenshot 2020-12-30 at 22 43 21

The scrollbar disappeared, but the text is still cropped, let's use div instead of textarea.

src/popup/router/components/Textarea.vue
```diff
@@ -49,6 +49,7 @@ export default {
   margin-left: auto;
   margin-right: auto;
   word-break: break-word;
+  scrollbar-width: none;
```
davidyuk, line 52: Also, we shouldn't disable scrollbars in all textareas.

Liubov-crypto (commented):
After logging in I got this issue: 

web wallet error

and this: 
FF error

Liubov-crypto:
Tested, now LGTM

test tip

src/popup/router/pages/SuccessTip.vue
```diff
@@ -116,4 +116,14 @@ export default {
   font-weight: normal;
   margin: 8px 0;
 }
+
+.note {
+  color: $text-color;
+  font-size: $font-size;
+  min-height: 100px;
```
davidyuk, line 123: This is not needed because divs are auto-sized.

davidyuk (approved):

mradkov (approved):

PR #776, opened 2021-01-25
Don't display scrollbar on blink and webkit engines

etharner:
I've decided to move this fix out of notifications PR because it affects other screens.
It fixes issue mentioned by @CedrikNikita in first gif here https://github.com/aeternity/superhero-wallet/pull/716#pullrequestreview-573009116
Because of this issue notifications, activity, names, terms of use, privacy policy (maybe more) screens were moving left a little bit because of 1px width transparent (😆) scrollbar appearance.

### DIFF
--- a/src/styles/global.scss
+++ b/src/styles/global.scss
@@ -13,8 +13,7 @@
 @import './variables';
 
 ::-webkit-scrollbar {
-  width: 1px;
-  background-color: transparent;
+  display: none;
 }
 
 ::-webkit-scrollbar-thumb {


src/styles/global.scss
```diff
@@ -13,8 +13,7 @@
 @import './variables';
 
 ::-webkit-scrollbar {
-  width: 1px;
-  background-color: transparent;
+  display: none;
```
davidyuk, line 16: Probably, it breaks accessibility in some cases (when not ios?), but scrollbars are already invisible.

davidyuk (approved):
The code changes look good.

src/styles/global.scss
etharner: We already have scrollbar hidden in firefox https://github.com/aeternity/superhero-wallet/blob/6f49b24c2599911f493b52c3d1b853b37ca9068f/src/styles/global.scss#L197
let's keep it consistent

mradkov (approved):

PR #777, opened 2021-01-25
Fix chain name outside viewport

### LINKED ISSUE #714: Copy button moved out of the viewport, if the name is to long
Copied

imollov:
fixes #714
depends on #776

recording

### DIFF
--- a/src/styles/global.scss
+++ b/src/styles/global.scss
@@ -13,8 +13,7 @@
 @import './variables';
 
 ::-webkit-scrollbar {
-  width: 1px;
-  background-color: transparent;
+  display: none;
 }
 
 ::-webkit-scrollbar-thumb {
--- a/src/popup/router/components/AccountInfo.vue
+++ b/src/popup/router/components/AccountInfo.vue
@@ -3,13 +3,15 @@
     <div class="title">
       <div class="account-name" data-cy="account-name">
         <Avatar :address="account.publicKey" :name="account.name" class="avatar" size="small" />
-        <template v-if="activeAccountName.includes('.chain')">{{ activeAccountName }}</template>
+        <span class="chainname" v-if="activeAccountName.includes('.chain')">{{
+          activeAccountName
+        }}</span>
         <router-link class="claim-chainname" to="/names" v-else
           >{{ $t('pages.account.claim-name') }}
         </router-link>
       </div>
       <div class="copied-alert" v-if="copied">{{ $t('pages.account.copied') }}</div>
-      <button data-cy="copy" @click="copy" v-clipboard:copy="account.publicKey">
+      <button v-else data-cy="copy" @click="copy" v-clipboard:copy="account.publicKey">
         {{ $t('pages.account.copy') }}
       </button>
     </div>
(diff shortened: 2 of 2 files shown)


src/styles/global.scss
```diff
@@ -13,8 +13,7 @@
 @import './variables';
 
 ::-webkit-scrollbar {
-  width: 1px;
-  background-color: transparent;
+  display: none;
```
mradkov, line 16: Isn't this conflicting with #776 ?

davidyuk (approved):
The code changes look good, but I'm not sure that the way long names showed is user-friendly. I would consider
- to hide the middle of the name, like `tttttttt...ttt.chain`
- or to implement an analog of marquee tag 😄 (maybe on hover)

cc: @onvisions

src/styles/global.scss
davidyuk: this is the same change, git should handle this well

imollov: I needed to do the same change as part of the fix. My intention is to rebase and remove the duplicated commit once #776 is merged. 

I've also added it as depending to that PR in the description. Please let me know if there's a more appropriate way to handle this case.

onvisions:
@davidyuk 
I discussed it already with @mradkov. What @imollov did was synced with me. It's a temporary design fix. I think it's the better option at the moment until we have special design for displaying very long names. I'll do this as part of UX/UI wallet rebranding and improvement I'm currently working on.

The problem with your suggestion above would be that:
1. on mobile we don't have hover interactions. This means that the user will have to click to show the full name "tag" but the .chain name element is **supposed to be link to the user account** (opened in blockchain explorer).
2. still there might be not enough space to display really long names within a tag or something...There is no max limit of names characters so virtually any length is possible. **Something to discuss about this** is whether we really need limitless .chain names as this is probably not very useful to anybody.

onvisions (commented):
@davidyuk
I discussed it already with @mradkov. What @imollov did was synced with me. It's a temporary design fix. I think it's the better option at the moment until we have special design for displaying very long names. I'll do this as part of UX/UI wallet rebranding and improvement I'm currently working on.

The problem with your suggestion above would be that:

   

1.  on mobile we don't have hover interactions. This means that the user will have to click to show the full name "tag" but the .chain name element is supposed to be link to the user account (opened in blockchain explorer).
2.   still there might be not enough space to display really long names within a tag or something...There is no max limit of names characters so virtually any length is possible. Something to discuss about this is whether we really need limitless .chain names as this is probably not very useful to anybody.

davidyuk:
> There is no max limit of names characters so virtually any length is possible.

Actually, AENS documentation says:

> Names SHOULD be at most 253 characters long, in order to enable interoperability with DNS.

https://aeternity.com/documentation-hub/protocol/AENS/

mradkov:
> > There is no max limit of names characters so virtually any length is possible.
> 
> Actually, AENS documentation says:
> 
> > Names SHOULD be at most 253 characters long, in order to enable interoperability with DNS.
> 
> https://aeternity.com/documentation-hub/protocol/AENS/

@imollov @onvisions Lets come up with a nice UX solution taking into account the limit of 253 chars. I doubt people would be having such long names anyways but still.

I'm will merge this PR now, and once we have a better UX suggestion we can refactor it.

PR #796, opened 2021-02-02
fix(global): revert scrollbar width until rebranding

etharner:
I propose to revert 7755be64 because it 
1) prevents scrolling in extension in Chrome
2) we have new scrollbar designs


mradkov (approved):
Ok, lets get this back.

PR #799, opened 2021-02-02
Hide chain name scrollbar

imollov:
Fix after reverting the global style hiding scrollbars #796


mradkov (approved):
This scopes the scrollbar hiding to this component only. Looks good now and should not break the rest of the scroll functionality.

PR #812, opened 2021-02-04
Rebrand Header component

imollov:


src/popup/router/components/Header.vue
```diff
@@ -88,91 +98,105 @@ export default {
...
+      cursor: pointer;
 
-    button {
-      padding: 0;
-      margin-left: 5px;
+      .count {
+        position: absolute;
+        left: -2px;
+        top: 20%;
+        z-index: 1;
```
etharner, line 182: I don't see any difference after removing it

etharner (changes requested):

src/popup/router/components/Header.vue
```diff
@@ -88,91 +98,105 @@ export default {
 
 <style lang="scss" scoped>
 @import '../../../styles/variables';
+@import '../../../styles/typography';
```
etharner, line 101: In this case we can remove `variables` import as it is already imported in `typography`

etharner (changes requested):

src/popup/router/components/Header.vue
davidyuk: [`@import`](https://sass-lang.com/documentation/at-rules/import) is deprecated, let's use [`@use`](https://sass-lang.com/documentation/at-rules/use) instead. `@use` working differently and probably we won't be needed to remove "duplicate" imports.

imollov: Using `@use` instead of `@import` will require making some global changes first.

mradkov: Let's log and address this separately and not stall this PR because of this change.

imollov: #832

mradkov (approved):
Changes looking good!

PR #813, opened 2021-02-05
Lint pre-commit hook

imollov:
https://www.npmjs.com/package/pre-commit


mradkov (approved):
Good, now we will mitigate problems with lint in the CI. Thank you!

PR #815, opened 2021-02-08
Tip shared URL on iOS/Android

### LINKED ISSUE #260: Mobile share functionality

imollov:
fixes #260

change log:
* forked a patched fork of `cordova-plugin-openwith` and adjusted hardcoded config to support only text sharing
* added URL prop to Tip page
* route to Tip page with URL prefilled on share intent

dev notes:
* run `cordova prepare` to init the new plugin for iOS/Android

screenshots:

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "clean-webpack-plugin": "^3.0.0",
     "copy-webpack-plugin": "^6.4.1",
     "cordova": "^10.0.0",
+    "cordova-plugin-openwith": "git+https://github.com/aeternity/cordova-plugin-openwith.git",
     "cordova-plugin-whitelist": "^1.3.4",
     "cordova-res": "^0.15.2",
     "core-js": "^3.8.2",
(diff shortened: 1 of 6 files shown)


imollov:
@aniputapet, could you please review the message below? 

It pops up if a user would share text which is not an URL (ex: article excerpt). Unfortunately these cannot be filtered out at the moment and hence the popup message. 

Thanks in advance!

package.json
```diff
@@ -131,7 +132,12 @@
         "URL_SCHEME": "superhero",
         "DEEPLINK_HOST": "wallet.superhero.com"
       },
-      "cordova-plugin-qrscanner": {}
+      "cordova-plugin-qrscanner": {},
+      "cordova-plugin-openwith": {
+        "IOS_URL_SCHEME": "hf938wofijfjhsdlsf23",
```
mradkov, line 137: Can you let us know what this `IOS_URL_SCHEME` is about?

mradkov (approved):
Looks good!

package.json
imollov: It's supposed to be random: https://github.com/aeternity/cordova-plugin-openwith#installation

aniputapet:
> @aniputapet, could you please review the message below?
> 
> It pops up if a user would share text which is not an URL (ex: article excerpt). Unfortunately these cannot be filtered out at the moment and hence the popup message.
> 
> Thanks in advance!
> 
> 

**Only URLs can be tipped at the moment!**
More options are currently being developed - coming to you soon!

package.json
mradkov: Cool. Thanks!

PR #795, opened 2021-02-02
Apply some general rebranding

etharner:
Partially closes #786,

PR #820, opened 2021-02-12
Update Index page designs

CedrikNikita:
depends on #795


src/popup/router/pages/Index.vue
```diff
@@ -38,68 +44,124 @@
...
+
+      &.agreed {
+        color: white;
+      }
 
-    ::v-deep .checkmark {
-      margin-right: 5px;
+      &:hover:not(.agreed),
+      &:active:not(.agreed) {
+        transition: all 0.12s ease-in-out;
```
etharner, line 88: Maybe we can use single 
```
    & > * {
      transition: all 0.12s ease-in-out;
    }
```
in parent `terms-agreement` and get rid of multiply transition? For me it will look equal

src/popup/router/pages/Index.vue
```diff
@@ -38,68 +44,124 @@
...
-
-    &:hover {
-      background: #4aabff;
-    }
+    width: 280px;
+    font-weight: 700;
 
     &:first-of-type {
-      margin-bottom: 20px;
+      margin-bottom: 24px;
```
etharner, line 113: Let's apply margin-bottom not only for first because in extension bottom button not fully shown
image

src/popup/router/components/Platforms.vue
```diff
@@ -84,27 +84,45 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-@import '../../../styles/variables';
+@import '../../../styles/typography';
 
 .platforms {
   font-size: 15px;
-  background-color: #21222c;
+  background-color: #000;
```
etharner, line 91: Let's use `$color-black` here as well, it's already imported with typography

src/popup/router/components/Platforms.vue
```diff
@@ -113,32 +131,34 @@ export default {
...
+          padding-bottom: 12px;
         }
 
         a {
-          padding: 5px;
-          border-radius: 10px;
-          height: 52px;
-          width: 52px;
+          &:first-of-type {
+            margin-right: 32px;
```
etharner, line 147: I think we can avoid this stuff by using 
```
display: flex;
justify-content: space-evenly;
```
in parent

etharner (changes requested):

src/popup/router/components/Platforms.vue
CedrikNikita: Yes, we can, but it wouldn't be pixel perfect.

src/popup/router/pages/Index.vue
CedrikNikita: @onvisions 
Aren't we planned to squeeze the new designs to a 600px height? I guess it should be done in other PR then, or provide a designs for that.

mradkov (approved):

src/popup/router/components/Platforms.vue
onvisions: I thought we already discussed it during the team meeting. 
Current designs have been done using standart screen size and ratio of an Android device (360x640px).
Furthermore:
_"High-density screens can be a bit confusing, a pixel isn't just a pixel anymore: the actual pixel count is a multiple of the declared pixel count.
That particular resolution is common because it's the simplest math for HD formats:
360x640 pixels, at 200% DPI, is 720x1280 -- which is the viewport size of the HDTV standard (aka 720p).
At 300% DPI it is 1080x1920 -- which is the 1080p standard."_

The need to decrease the viewport height from 640 to 600 px appears because of the Google Chrome (and chromium based browsers) limitation of viewport max-height to 600px. As it turned out none of us was aware of this limitation (I have asked before starting the designs). Currently it will be time consuming to adjust all the screens and the prototype interactions for just 40px decrease of viewport height. At the same time it won't be so good idea as current size is better for the native apps. 

My suggestion will be to apply "display: flex; justify-content: space-evenly;" suggested above but only for chromium-based browsers. The difference of 40px distributed evenly in this case won't be a problem. For web-based wallet, apps and non chromium browsers we may still have 640px height (as mentioned above it's very standard aspect ratio of the viewport).

On the welcome screens it will be better to diminish the size of the margins between the spinning animation and the rest of the content by 20px on the top and another 20px on the bottom. It also can be done quickly with a browser-specific extra line of CSS. This is the most obvious and quick solution here. 

For the screens where we already have scrollable area this won't be a problem.

maxkpower: Fully agree with Tsvetan here!

PR #830, opened 2021-02-18
Rebrand RecentTransactions component

imollov:


src/popup/router/components/PendingTxs.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div v-if="transactions.length" data-cy="pending-txs">
+  <div v-if="transactions.length" class="list" data-cy="pending-txs">
```
davidyuk, line 2:
```suggestion
  <div v-if="transactions.length" class="pending-txs" data-cy="pending-txs">
```
In multiple projects we agreed to use the component name as a class of the root node, it simplifies styles binging on the side of parent component.

mradkov (approved):

PR #834, opened 2021-02-19
Fix blank screen after splash on iOS

### LINKED ISSUE #833: Blank screen after splash on iOS below 14
Introduced in PR #525

imollov:
fixes #833

QA:
* QR code read related functionality must be tested well on all platforms

### DIFF
--- a/src/popup/router/components/Modals/QrCodeReader.vue
+++ b/src/popup/router/components/Modals/QrCodeReader.vue
@@ -14,7 +14,6 @@
 </template>
 
 <script>
-import { BrowserQRCodeReader } from '@zxing/library/esm/browser/BrowserQRCodeReader';
 import Modal from '../Modal';
 import openUrl from '../../../utils/openUrl';
 
@@ -30,7 +29,7 @@ export default {
   data: () => ({
     // allow camera while QRScanner is loading to not show cameraNotAllowed before actual check
     cameraAllowed: process.env.PLATFORM === 'cordova',
-    browserReader: !(process.env.PLATFORM === 'cordova') && new BrowserQRCodeReader(),
+    browserReader: null,
     style: null,
     headerText: '',
   }),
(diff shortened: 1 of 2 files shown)


davidyuk (commented):
The black screen was because of es features of `@zxing/library` or what?

imollov:
> The black screen was because of es features of `@zxing/library` or what?

Probably but can't tell for sure because no errors are logged. 

In any case - not importing a browser-only package on mobile must be beneficial, right?

mradkov:
Can you adapt the e2e test to the changes as well?

davidyuk:
> Probably but can't tell for sure because no errors are logged.

Have you tried to connect Safari/Chrome debugger to the mobile app built in develop mode? You should see the exception in the console then.

> In any case - not importing a browser-only package on mobile must be beneficial, right?

Agree. In general, I think would be best to split this component into two for web and Cordova because scanning approaches are too different.

imollov:
> > Probably but can't tell for sure because no errors are logged.
> 
> Have you tried to connect Safari/Chrome debugger to the mobile app built in develop mode? You should see the exception in the console then.
> 
> > In any case - not importing a browser-only package on mobile must be beneficial, right?
> 
> Agree. In general, I think would be best to split this component into two for web and Cordova because scanning approaches are too different.

Here's the exception causing the black screen:
Screenshot 2021-02-22 at 14 59 50

And an issue to refactor as suggested: #837

davidyuk (approved):

davidyuk:
Actually, this is a workaround. We won't be able to use QR code scanner in a web version on iOS below 14:



this is a simulator, iOS 12.4, iPhone 7

mradkov (approved):
Let us fix the actual app problems, and we can resolve the web safari issue separately. Feel free to merge.

PR #856, opened 2021-02-26
Update RecentTransactions spinner

imollov:


src/icons/animated-spinner-white.svg
```diff
@@ -0,0 +1,43 @@
...
+        fill="remove"
+        from="360 0 0"
+        repeatCount="indefinite"
+        restart="always"
+        to="0 0 0"
+        type="rotate"
+      />
+    </path>
+  </g>
+</svg>
```
CedrikNikita, line 43:
```suggestion
</svg>

```

davidyuk (commented):
Isn't breaking of animation fixed in the latest version of SVG optimizer? Do you have an issue number to track fixing progress on their side?

Why don't inline the spinner, change color using `currentColor`, and disable optimized specifically for this import or even for the whole project? This is better than duplicating spinner image 🤷‍♀️

imollov:
> Isn't breaking of animation fixed in the latest version of SVG optimizer? Do you have an issue number to track fixing progress on their side?
> 
> Why don't inline the spinner, change color using `currentColor`, and disable optimized specifically for this import or even for the whole project? This is better than duplicating spinner image 🤷‍♀️

We use the latest version of `vue-svg-loader` and the issue is still present. Unfortunately, I couldn't find any relatable issue that directly addresses the problem.

I decided to not introduce new loader for inline svg only for this case. Instead I added an option to load without `svgo`.

Disabling optimization entirely results in increasing the bundle size by ~31KB.

src/popup/router/components/RecentTransactions.vue
```diff
@@ -139,6 +138,10 @@ export default {
     width: 56px;
     height: 56px;
     margin: 0 auto;
+
+    > path {
+      fill: $color-white;
+    }
```
davidyuk, line 144:
```suggestion
    color: $color-white;
```
won't it work the same?

davidyuk (commented):
In general looks good, would be nice to report to SVGO about our problem.

imollov:
> In general looks good, would be nice to report to SVGO about our problem.

Actually, the unexpected behavior can be avoided by disabling one of `svgo` plugins: `moveGroupAttrsToElems`.

However, configuring it globally bumped assets size by ~14KB. And the number could increase as we add more files.

imollov:
> > In general looks good, would be nice to report to SVGO about our problem.
> 
> Actually, the unexpected behavior can be avoided by disabling one of `svgo` plugins: `moveGroupAttrsToElems`.
> 
> However, configuring it globally bumped assets size by ~14KB. And the number could increase as we add more files.

Just to clarify - I don't consider this an issue. 

@davidyuk do you still think it should be reported?

mradkov:
@imollov @davidyuk whats the state of this, shall we merge it?

davidyuk (approved):

davidyuk:
> @davidyuk do you still think it should be reported?

Yep, the default settings break animated svgs. Actually, it wasn't a blocker to merge this.

PR #857, opened 2021-03-01
fix: use TruncateMid only for custom titles

### LINKED ISSUE #848: Truncated header title looks confusing
Screenshot 2021-02-23 at 16 49 38
Screenshot 2021-02-23 at 16 49 54

seen at #812

cc: @CedrikNikita

CedrikNikita:
fixes #848

### DIFF
--- a/tests/e2e/integration/names.js
+++ b/tests/e2e/integration/names.js
@@ -1,10 +1,11 @@
-const name = 'test.chain';
+const name = 'test1.chain';
+const resultName = 'test1\u200E.chain';
 describe('Test cases for names', () => {
   it('Contains .chain name in sidebar menu and account info', () => {
     cy.login({ name })
       .openMenu()
       .get('[data-cy=chain-name]')
-      .truncateStringShouldContain('[data-cy=chain-name] .truncate-mid', name)
-      .truncateStringShouldContain('[data-cy=account-name] .truncate-mid', name);
+      .truncateStringShouldContain('[data-cy=chain-name] .truncate-mid', resultName)
+      .truncateStringShouldContain('[data-cy=account-name] .truncate-mid', resultName);
   });
 });
--- a/src/popup/router/components/Header.vue
+++ b/src/popup/router/components/Header.vue
@@ -15,10 +15,10 @@
     </div>
 
     <div class="title">
-      <TruncateMid
-        :str="pageTitle || (title && $t(`pages.titles.${title}`)) || $t('pages.titles.home')"
-        class="text"
-      />
+      <TruncateMid v-if="pageTitle" :str="pageTitle" class="text" />
+      <span v-else class="text">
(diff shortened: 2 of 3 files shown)


imollov (commented):
Shouldn't we see the "..." in the middle here?
 
Screenshot 2021-03-01 at 11 43 17

imollov:
Looks much better now!

Not sure if this is important but the space before "..." makes the title look like 2 different elements visually. Maybe @onvisions can weigh in here.

Screenshot 2021-03-04 at 12 45 54

onvisions:
As far as I know the last decision discussed with the team and @Max was to have token abbreviation limited to 6 characters. Token short name (abbreviation) is needed exactly to make token names user friendly across the UI.

onvisions:
@imollov  As far as I know the last decision discussed with the team and @Max was to have token abbreviation limited to 6 characters. Token short name (abbreviation) is needed exactly to make token names user friendly across the UI. Also in hte new designs we have not only token name but also token identicon is displayed there ( to distinguish from other tokens it's auto-generated and kept on the blockchain).

imollov:
> @imollov As far as I know the last decision discussed with the team and @max was to have token abbreviation limited to 6 characters. Token short name (abbreviation) is needed exactly to make token names user friendly across the UI. Also in hte new designs we have not only token name but also token identicon is displayed there ( to distinguish from other tokens it's auto-generated and kept on the blockchain).
> Generally in my opinion to have unlimited length token names (without short abbreviation name to be used across the UI) is nonsense.

Makes sense. Yet this solution is used and will be used on other places - token header title was just an example.

mradkov:
> Looks much better now!
> 
> Not sure if this is important but the space before "..." makes the title look like 2 different elements visually. Maybe @onvisions can weigh in here.
> 
> Screenshot 2021-03-04 at 12 45 54

@CedrikNikita can you address this, so we can merge the PR?

CedrikNikita:
> > Looks much better now!
> > Not sure if this is important but the space before "..." makes the title look like 2 different elements visually. Maybe @onvisions can weigh in here.
> > Screenshot 2021-03-04 at 12 45 54
> 
> @CedrikNikita can you address this, so we can merge the PR?

I don't see an easy way to do it. In the current Header designs it looks like this.
image

mradkov:
> Since we are going to use token abbreviation limited to 6 characters, `space` won't be a problem.

Actually the final decision was 14 chars. But now looks fine.

tests/e2e/integration/names.js
```diff
@@ -1,10 +1,11 @@
-const name = 'test.chain';
+const name = 'test1.chain';
+const resultName = 'test1\u200E.chain';
```
davidyuk, line 2: I was going to say that it breaks text if I copy it from there, but actually, the text in TruncateMid is not selectable because of before/after.

PR #859, opened 2021-03-01
Refactor popup classes into App component

etharner:
The idea of this PR came from requested changes for desktop web-wallet. scrolling of Recent Transactions and general rebranding (getting rid of `global.scss` file).
I particularly don't like having some global classes and spread it to every page. Instead of that we can use inheritance of vue components and add classes on top of parent classes, if needed. 

Pros:
1. it helps clear `global.scss` file
2. every page will have root class corresponding to it's name (like we have in components)
2. we can mix old and new designs before we will complete rebranding (so in the end we should get rid of `no-rebrand` prop)
3. we can clear `App,vue` from unnecessary complications (like, moving `Header`, `Loader`, `Node connection status` etc into `Page`) and make css more clear
4. it will probably help to mix modals and popups designs in the future

Cons:
1. for now `no-rebrand` prop looks kinda hacky
2. we should test this PR in every environment to see if everything is ok in terms of layout

### DIFF
--- a/src/popup/App.vue
+++ b/src/popup/App.vue
@@ -1,17 +1,11 @@
 <template>
-  <div
-    :class="[
-      'ae-main',
-      aeppPopup ? 'ae-main-popup ae-main-wave' : waveBg ? 'ae-main-wave' : '',
-      iframe && ($route.path === '/intro' || $route.path === '/') ? 'iframe' : '',
-    ]"
-  >
-    <Header
-      v-if="!(iframe && $route.path === '/intro')"
-      @toggle-sidebar="showSidebar = !showSidebar"
-    />
+  <div id="app">
+    <Header v-if="showHeader" @toggle-sidebar="showSidebar = !showSidebar" />
 
-    <RouterView />
+    <RouterView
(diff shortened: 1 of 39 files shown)


davidyuk (commented):
In general, the changes look fine. I'm not sure about the purpose of App.vue. From my point of view, firstly we need to move all initialization logic into the store plugin and/or main.js. You think that we need to move all layout, what will stay in App.vue then? I would propose to leave commonly displayed components there, like Header, SIdebar, ConnectionStatus.

We have two approaches:
- the root component of page is Page or something else or plain div
- the root component of page is plain div, rendered inside App.vue that contains base layout (can be adjusted by route's meta)

In the first approach, we can use or not use Page component, allowing us to make a completely different layout (I'm not sure that we need it). Also, it allows us to pass properties to "parent layout". Currently, it is only `no-rebrand` property, that is going to be removed at the end, instead of this, we can define a key in the route's meta. So, I don't see a strong reason to introduce Page component, but I like the code clean-up that you did here.

davidyuk (approved):

mradkov (approved):

PR #865, opened 2021-03-02
fix(commitlint): revert run commitlint from first valid commit

davidyuk:
This reverts commit 404972ea13d4d55f33d2fc402b88ff373012cac9.

I can't get the reasoning for changes made in #791. From my perspective reference to develop is better, because:
- running `lint:commit` I will get a short list of commit messages made by me, without commits already merged
- reference to develop makes it possible to move the setup to another repository without extra adjustments

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "scripts": {
     "lint:vue": "eslint --ext .js,.vue src tests",
     "lint:css": "stylelint src/**/*.{css,scss,vue}",
-    "lint:commit": "commitlint --from 151392b0 --verbose",
+    "lint:commit": "commitlint --from develop --verbose",
     "lint": "npm run lint:commit && npm run lint:vue && npm run lint:css",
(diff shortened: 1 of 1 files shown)


etharner:
> This reverts commit [404972e](https://github.com/aeternity/superhero-wallet/commit/404972ea13d4d55f33d2fc402b88ff373012cac9).
> 
> I can't get the reasoning for changes made in #791. From my perspective reference to develop is better, because:
> 
> * running `lint:commit` I will get a short list of commit messages made by me, without commits already merged
> * reference to develop makes it possible to move the setup to another repository without extra adjustments

I've made that hotfix because for some reason commitlint had been checking more commits than diff from develop (and there were some old styled commits) and it failed, so I've decided to explicitly point first valid commit to avoid this problem.
And as I don't like my solution, let's merge this and if this problem will appear one more time let's do proper investigation.

etharner (approved):

mradkov (approved):
As @etharner said. Lets see if this brings problems, otherwise looks good.

PR #866, opened 2021-03-02
refactor: prefer 'mounted' hook instead of 'created'

davidyuk:
I think by default we should use `mounted` because:
- in `mounted` hook the component's DOM is available making a wider range of operations available
- `mounted` is not executed in SSR context, it is not related to this repository, but it seems to be a good habit to don't use `created` by default because of that

### DIFF
--- a/src/phishing/App.vue
+++ b/src/phishing/App.vue
@@ -35,7 +35,7 @@ export default {
       hostname: null,
     };
   },
-  created() {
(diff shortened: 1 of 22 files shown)


CedrikNikita (approved):
Code changes look good.

mradkov (approved):
LGTM

PR #868, opened 2021-03-02
Switch to a testing backend

davidyuk:
related to https://github.com/aeternity/superhero-ui/pull/1026/


Liubov-crypto:
Got this error when tested backend with iframe wallet: 
er wal1

PR #869, opened 2021-03-03
Adjust Index page styles

CedrikNikita:


src/popup/router/components/Platforms.vue
```diff
@@ -90,26 +90,20 @@ export default {
   font-size: 15px;
   background-color: $color-black;
   word-break: break-word;
+  cursor: default;
```
davidyuk, line 93: I don't think that we should change `text` cursor to `default`. Do we want to disallow text selection in wallet UI?

CedrikNikita: it's mot my call
image

davidyuk: @onvisions yep, and for highlighting the text there is a separate "text" cursor, why do you want to change it? Let's disallow text selection in the whole app to look like a native app.

onvisions: I meant something else here.
The cursor should be pointer over interactive elements such as buttons (when they arre active, not disabled).
The cursor should be arrow over elements that are not interactive (although these elements can have hover effect applied). For example in the design here when the mouse cursor enters the whole card with browser buttons and info text the background changes its color (it has hover effect) but the cursor should remain arrow. 
The default behavior over texts is a separate matter. I don't see a problem to leave the default text cursor over larger portions of text. However I think @davidyuk is right text selection should be disallowed for most of the places with a **few exceptions.**
I think text selection will be useful where we have **"Terms of use"**, other meaningful texts that the user may want to copy and for the **addresses** (although we have "copy" button...).

davidyuk (approved):

PR #878, opened 2021-03-10
refactor(event-bus): use store plugin instead

davidyuk:


CedrikNikita (approved):
Code changes look good

PR #886, opened 2021-03-11
Scroll to top on route change

imollov:
Prevent from new pages being opened in the same scroll position.

Demo of the issue:
demo

### DIFF
--- a/src/popup/router/index.js
+++ b/src/popup/router/index.js
@@ -31,6 +31,7 @@ Vue.component('Loader', LoaderComponent);
 const router = new VueRouter({
   routes,
   mode: process.env.PLATFORM === 'web' ? 'history' : 'hash',
+  scrollBehavior: (to, from, savedPosition) => savedPosition || { x: 0, y: 0 },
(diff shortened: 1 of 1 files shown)


src/popup/router/index.js
```diff
@@ -31,6 +31,7 @@ Vue.component('Loader', LoaderComponent);
 const router = new VueRouter({
   routes,
   mode: process.env.PLATFORM === 'web' ? 'history' : 'hash',
+  scrollBehavior: () => ({ x: 0, y: 0 }),
```
davidyuk, line 34: this is fine, but let's also preserve position stored in history
```suggestion
  scrollBehavior: (to, from, savedPosition) => savedPosition || { x: 0, y: 0 },
```
https://router.vuejs.org/guide/advanced/scroll-behavior.html

davidyuk (approved):

mradkov (approved):

Liubov-crypto:
Checked, LGTM.

PR #895, opened 2021-03-16
Feat: Add Stale bot

mradkov:
Add stale bot to close inactive PRs.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,39 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Limit only for Pull Requests
+only: pulls
+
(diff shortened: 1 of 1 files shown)


gorbak25:
I've added the bot the the organization ;)

davidyuk (commented):
what is the motivation behind it?

mradkov:
> what is the motivation behind it? Can we add a label for such PRs instead? Somebody who doesn't want to see such PRs may filter PRs by that. The decision to close a PR should be done by humans I think.

The motivation behind is to have people make PRs with smaller changes which can be easily merged in couple of days, and prevent the current situation of huge PRs being introduced which are hanging for months.

PR #911, opened 2021-03-22
Rebrand input search bar

etharner:
This refactoring came for 3 reasons:
1) Rebrand text-number input corresponding to new designs (except for error messages, it will came later in Send and Tip rebranding)
2) Extract search-bar code into separate component as it will be used in two screens (Transaction List and Balances)
3) Futher clean-up of `global.scss` file


src/popup/router/pages/Names/Claim.vue
```diff
@@ -2,13 +2,13 @@
...
-      <Input
+      <InputField
         v-model="name"
         :placeholder="$t('pages.names.claim.name-placeholder')"
-        label=".chain"
-        labelPosition="right"
         :error="!validName"
-      />
+      >
+        <span slot="right">{{ `.chain` }}</span>
```
davidyuk, line 10:
```suggestion
        <!-- eslint-disable vue-i18n/no-raw-text -->
        <span slot="right">.chain</span>
```
proposing to drop this rule here: https://github.com/aeternity/superhero-wallet/pull/918

src/popup/router/components/InputField.vue
```diff
@@ -0,0 +1,109 @@
...
+    <div :class="{ focus, error }" class="container">
+      <slot v-if="!error" name="left" />
+      <Error v-else />
+      <input
+        :type="type"
+        class="input"
+        :placeholder="placeholder"
+        :value="value"
+        @input="$emit('input', $event.target.value)"
+        @focus="(f) => (focus = f)"
```
davidyuk, line 15:
```suggestion
        @focus="focus = true"
        @blur="focus = false"
```
don't you forget to reset the focus state on blur? Also, proposing to rename `focus` to `focused` or `isFocused`

src/popup/router/components/SearchBar.vue
```diff
@@ -0,0 +1,39 @@
...
+  <InputField
+    class="search-bar"
+    :v-model="value"
+    :value="internalValue"
+    v-bind="$attrs"
+    @input="input"
+    type="text"
+  >
+    <Search slot="right" v-if="!internalValue.length" />
+    <Eraser slot="right" v-else @click="input('')" />
```
davidyuk, line 11:
```suggestion
    :value="value"
    v-bind="$attrs"
    @input="$emit('input', $event)"
  >
    <Search slot="right" v-if="!value" />
    <Eraser slot="right" v-else @click="$emit('input', '')" />
```
In general, it looks like that this component can be stateless.

src/popup/router/components/SearchBar.vue
```diff
@@ -0,0 +1,39 @@
...
+  },
+  data: () => ({
+    internalValue: '',
+  }),
+  methods: {
+    input(value) {
+      this.internalValue = value;
+      this.$emit('input', this.internalValue);
+    },
+  },
```
davidyuk, line 37:
```suggestion
```

src/popup/router/components/InputField.vue
```diff
@@ -0,0 +1,110 @@
...
+  .container {
+    display: flex;
+    align-items: center;
+    padding: 8px 16px;
+    height: 40px;
+    background-color: $color-bg-2;
+    border: 1px solid transparent;
+    border-radius: 6px;
+
+    &.isFocused {
```
davidyuk, line 70:
```suggestion
    &:focus-within {
```
🚀

davidyuk: @etharner didn't it work for some reason?
https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within

davidyuk (approved):

PR #918, opened 2021-03-23
refactor: disable vue-i18n/no-raw-text rule

davidyuk:
I think this rule is too strict, it shouldn't make developers
- write `<!--eslint-disable-next-line vue-i18n/no-raw-text-->` to add a tilda
- write `{{ '~' }}` to add a tilda
- write `{{ $t('AE') }}` and `"AE": "AE"` in locale file to add "AE"


davidyuk:
Tests are failing, but it is not related to changes made in this PR.

CedrikNikita (approved):

Issue #896, opened 2021-03-16
What happened to my name?


gorbak25:
![image](https://user-images.githubusercontent.com/1352848/111301268-4cde2800-8652-11eb-8632-93c36700eb62.png)
![image](https://user-images.githubusercontent.com/1352848/111301298-55cef980-8652-11eb-953f-aba6c4573c27.png)

gorbak25:
gorbak25.chain -> gorbak2chain.5 <3

davidyuk:
I believe it is because of the TruncateMid component. cc: @CedrikNikita

CedrikNikita:
Was fixed here https://github.com/aeternity/superhero-wallet/pull/857

CedrikNikita:
Current version of wallet doesn't have such problems. Feel free to reopen this issue, if it will pop up again.

PR #921, opened 2021-03-24
refactor(truncate-mid): simplify TruncateMid into OverflowableText

etharner:
This PR simplifies TruncateMid component and make it use native browser text-overflow. It's also an another solution to #896.


src/popup/router/components/OverflowableText.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <span class="overflowable-text">
+    <span class="text">{{ isName ? isName : text }}</span>
+    <span v-if="isName">{{ `.chain` }}</span>
```
davidyuk, line 4:
```suggestion
    <span class="text">{{ text.substr(0, text.length / 2) }}</span>
    <span>{{ text.substr(text.length / 2) }}</span>
```
Will it work the same as TruncateMid then?

src/popup/router/components/OverflowableText.vue
```diff
@@ -0,0 +1,30 @@
...
+</template>
+
+<script>
+export default {
+  props: {
+    text: { type: String, required: true },
+  },
+  computed: {
+    isName() {
+      return this.text.match(/([a-zA-Z0-9]+)\.chain/)?.[1];
```
davidyuk, line 15:
```suggestion
      return this.text.match(/^([a-zA-Z0-9]+)\.chain$/)?.[1];
```
I wouldn't like to specify name structure here, I would check for `.endsWith('.chain')` instead

davidyuk (approved):
I see two changes:

- rewriting TruncateMid with a different approach
- truncating not the middle, but the end of the title

what is the motivation for the second one?

In general, code changes looks good.

src/popup/router/components/OverflowableText.vue
davidyuk, line 4:
```suggestion
    <!--eslint-disable-next-line vue-i18n/no-raw-text-->
    <span v-if="isName">.chain</span>
```
and please check #918

davidyuk (commented):
I have registered "victorfrieeeeeeeeeeeeeeeeees.chain" name, the previous version was showing it better in a side menu:


Screenshot 2021-04-09 at 03 03 33
Screenshot 2021-04-09 at 03 04 17

In general, the problem of the new approach is that the only first part getting truncated making it looking asymmetric, also it won't work at all if the second part is too long.

Screenshot 2021-04-09 at 03 04 49
Screenshot 2021-04-09 at 03 04 55

I see that the modern web is not friendly to truncating the middle of text, so for simplicity, I would propose
- to truncate only the end of the title (not necessary to see the end of it)
- revert the previous implementation of this component (showing `abc....chain`) and use it only for aens names

or to keep status quo

PS I have rebased just to restart tests (didn't help)

PR #926, opened 2021-03-24
refact(settings): redirect back to settings after backup seed

mradkov:
> When you select this from
	the Settings menu, and go “show seed phrase” + “I’ve already
	done this” it takes you back home instead of showing the Settings
	menu again. Adds an unnecessary step in case you wish to continue
	browsing the Settings menu.

### DIFF
--- a/src/popup/router/pages/SecuritySettings.vue
+++ b/src/popup/router/pages/SecuritySettings.vue
@@ -97,7 +97,7 @@ export default {
   methods: {
     setBackedUpSeed() {
       this.$store.commit('setBackedUpSeed');
-      this.$router.push({ name: 'account' });
+      this.$router.push({ name: 'settings' });
     },
     verifyLastStep() {
       const mnemonicSelected = this.selectedWordIds
--- a/src/popup/router/routes/index.js
+++ b/src/popup/router/routes/index.js
@@ -1,46 +1,44 @@
-import Index from '../pages/Index';
-import Account from '../pages/Account';
 import ConfirmTransactionSign from '../components/Modals/ConfirmTransactionSign';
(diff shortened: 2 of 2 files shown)


src/popup/router/pages/SecuritySettings.vue
```diff
@@ -97,7 +97,7 @@ export default {
   methods: {
     setBackedUpSeed() {
       this.$store.commit('setBackedUpSeed');
-      this.$router.push({ name: 'account' });
+      this.$router.push({ name: 'settings' });
```
imollov, line 100: The settings route doesn't have `name`. 

So either add it or:
```suggestion
      this.$router.push({ path: '/settings' });
```

Or even go back (seems less coupled to me):

```suggestion
      this.$store.dispatch('router/back');
```

mradkov: The `router/back` seems like the nicer approach to me as well. Thanks!

src/popup/router/pages/SecuritySettings.vue
```diff
@@ -97,7 +97,7 @@ export default {
   methods: {
     setBackedUpSeed() {
       this.$store.commit('setBackedUpSeed');
-      this.$router.push({ name: 'account' });
+      this.$store.dispatch('router/back');
```
CedrikNikita, line 100:
```suggestion
      this.$store.dispatch('router/back');
```
This is not a good idea. If you will open extension on this page, than this button will lead you to an account page.

mradkov: Then 9ada9dd

davidyuk (approved):

PR #928, opened 2021-03-24
refactor(link-button): prevent default in cordova

imollov:
depends on #929
closes #914

PR #929, opened 2021-03-24
fix(cordova): open urls without leaving app

imollov:
related to #928

### DIFF
--- a/src/popup/utils/openUrl.js
+++ b/src/popup/utils/openUrl.js
@@ -13,7 +13,7 @@ export default (url, newTab) => {
     }
     case 'cordova':
       document.addEventListener('deviceready', () => {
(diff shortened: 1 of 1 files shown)


src/popup/utils/openUrl.js
```diff
@@ -13,7 +13,7 @@ export default (url, newTab) => {
     }
     case 'cordova':
       document.addEventListener('deviceready', () => {
-        cordova.InAppBrowser.open(url, '_system');
+        cordova.InAppBrowser.open(url, '_blank');
```
davidyuk, line 16: We were using inappbrowser plugin just to open system browser. Right now it works without the plugin, so I'm proposing to remove it in #943

PR #872, opened 2021-03-04
General rebranding of web wallet layout

etharner:
This PR makes wallet centered on desktop as it is in Figma prototype and fixes emerged old designs layout issues.
Depends on #859.

PR #931, opened 2021-03-25
Refactor Plate component

### LINKED ISSUE #920: Refactor Plate component
Notes after checking #903:
- use hard shadow instead of the border
- remove negative margins (won't be needed to hide it after finishing #872)
- I think `position: fixed` should be defined by the parent component, also consider using `sticky` instead
- at the end `overflow-x: hidden` shouldn't be needed
- remove extra div

cc: @imollov

imollov:
closes #920, depends on #872

### DIFF
--- a/src/popup/router/components/Plate.vue
+++ b/src/popup/router/components/Plate.vue
@@ -1,8 +1,6 @@
 <template>
   <div class="plate">
-    <div class="content">
-      <slot />
-    </div>
+    <slot />
   </div>
 </template>
 
@@ -11,16 +9,8 @@
 
 .plate {
   width: $container-width;
-  overflow-x: hidden;
-  position: fixed;
-  z-index: 1;
-
-  .content {
-    width: #{2px + $container-width};
-    margin: -1px 0 0 -1px;
-    border: 1px solid rgba($color-border, 0.5);
-    border-radius: 0 0 10px 10px;
-    background: linear-gradient(180deg, $color-bg-3 0%, $color-black 100%);
-  }
+  box-shadow: rgba($color-border, 0.5) 0 1px;
+  border-radius: 0 0 10px 10px;
+  background: linear-gradient(180deg, $color-bg-3 0%, $color-black 100%);
 }
 </style>
--- a/src/popup/router/pages/Account.vue
+++ b/src/popup/router/pages/Account.vue
@@ -48,17 +48,18 @@ export default {
 @import '../../../styles/variables';
 
 .account {
+  .plate {
+    position: sticky;
+    top: 48px;
+    z-index: 1;
+  }
+
   .menu-carousel {
     padding: 24px;
   }
 
   .recent-transactions {
     margin-top: -10px;
-    padding-top: 279px;
-  }
-
-  &.has-reminder .recent-transactions {
-    padding-top: 313px;
   }
 
   .seed-backup-notification {
--- a/src/popup/router/pages/TransactionDetails.vue
+++ b/src/popup/router/pages/TransactionDetails.vue
@@ -1,87 +1,85 @@
 <template>
   <div class="transaction-details">
-    <div class="header">
+    <Plate>
       <TokenAmount :amount="amount" :symbol="symbol" :direction="direction" large />
+    </Plate>
+    <div class="visual-overview"><!-- TODO --></div>
+    <div class="data-grid">
+      <DetailsItem
+        v-if="tipUrl"
+        :label="$t('pages.transactionDetails.tipUrl')"
+        class="span-2-columns"
+        data-cy="tip-url"
+      >
+        <CopyButton slot="label" :value="tipUrl" message="URL copied" />
+        <LinkButton slot="value" :to="tipUrl">{{ tipUrl }}</LinkButton>
+      </DetailsItem>
+      <DetailsItem
+        :value="hash"
+        :label="$t('pages.transactionDetails.hash')"
+        class="span-2-columns"
+        data-cy="hash"
+        small
+      >
+        <CopyButton slot="label" :value="hash" message="Hash copied" />
+      </DetailsItem>
+      <DetailsItem
+        v-if="transaction.microTime"
+        :value="transaction.microTime | formatDate"
+        :secondary="transaction.microTime | formatTime"
+        :label="$t('pages.transactionDetails.timestamp')"
+        data-cy="timestamp"
+      />
+      <DetailsItem
+        v-if="transaction.blockHeight && transaction.blockHeight > 0"
+        :value="transaction.blockHeight"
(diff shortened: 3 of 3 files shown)


src/popup/router/components/Plate.vue
```diff
@@ -11,16 +9,8 @@
...
-  z-index: 1;
-
-  .content {
-    width: #{2px + $container-width};
-    margin: -1px 0 0 -1px;
-    border: 1px solid rgba($color-border, 0.5);
-    border-radius: 0 0 10px 10px;
-    background: linear-gradient(180deg, $color-bg-3 0%, $color-black 100%);
-  }
+  box-shadow: rgba($color-border, 0.5) 0 0 0 1px;
```
davidyuk, line 12:
```suggestion
  box-shadow: rgba($color-border, 0.5) 0 1px;
```
this should show the border only at the bottom of node (as it is in the designs) instead of showing it around
Screenshot 2021-03-29 at 23 27 02

config.xml
```diff
@@ -48,6 +48,7 @@
         <preference name="UseSwiftLanguageVersion" value="4.2" />
         <allow-intent href="itms:*" />
         <allow-intent href="itms-apps:*" />
+        <preference name="DisallowOverscroll" value="true" />
```
davidyuk, line 51: What is the motivation behind this? In general, I'm against this because it needs
- to behave consistently with other iOS apps (to don't break iOS users experience)
- to behave consistently with the web version of wallet

The next time would be better to write "why something is done" instead of "what is done" in the commit message. For example, "fix(iOS): disable overscroll to fix XYZ" or just "fix(iOS): XYZ" 🤔

davidyuk (commented):
> fix iOS 12.x border radius issue

I wouldn't care so much about iOS 12

config.xml
imollov: > What is the motivation behind this? In general, I'm against this because it needs
> 
>     * to behave consistently with other iOS apps (to don't break iOS users experience)
> 
>     * to behave consistently with the web version of wallet
> 

Totally agreed on not breaking the UX of the native platform.

Disallowing overscroll was in attempt to achieve the scrolling behaviour of the interactive prototype. In combination with the proposal to use `sticky` instead of `fixed`. 

This is how it behaves without overscroll disabled:

https://user-images.githubusercontent.com/482351/112992432-d6244d00-9170-11eb-83eb-783706952edd.mp4

If we want to achieve the native UX + consistency with the web wallet + resemblance with the interactive prototype - then I guess the element should be positioned `fixed`, so only the area beneath is scrolled/overscrolled. 

Any better solutions?

> 
> The next time would be better to write "why something is done" instead of "what is done" in the commit message. For example, "fix(iOS): disable overscroll to fix XYZ" or just "fix(iOS): XYZ" 🤔

It is consistent with the repo commit message history, in my opinion.

Indeed an explanation maybe in the PR description or commit body would've been helpful and it is my bad for not including it.

onvisions: I synced with @imollov  and I agree with his suggestion to keep the native functionality but make fixed both the header and the following component. The behavior visible in the video just looks strange. 
Also I noticed a few other visual bugs which are described in this screenshot:
Screenshot_1

davidyuk: Is this still actual in regards to the latest changes in the develop branch?

imollov: The side remarks by @onvisions are fixed in develop and #951. 

What remains unsolved is the scrolling behavior on iOS. Simply put - the plate should not scroll by design and it currently does.

Although with the latest changes to support web layout I don't see an easy way to achieve it.

src/popup/router/components/Plate.vue
davidyuk: > It is consistent with the repo commit message history, in my opinion.

It may be, but it doesn't make it a good practice 😄 

Technically we can make Plate fixed on mobile (see #982), but I don't think that this code complication is reasonable, let's just fix the color of the gap between the page header and plate component on overscrolling

onvisions:
@davidyuk 
Fixes made by @imollov are satisfactory to me. The previous version was looking weird and broken on scroll. However from the demo video he sent me I can undesratnd that the component that was supposed to fixed to top by design is implemented in a different way (See the attached gif below where apparently the -1AE transaction info is hidden outside the visible viewport while user scrolling). Is that so for all instances of the same component like User account cards switcher, Balances, Payments, Tips... and many more to come probably as by design the intended behaviour is "fixed component" and the plan was to reuse it across the screens and functionalities. The ratio behind this is to have the important features and information accessible and visible on the screen (in front of user's eyes) all the time even when page is scrolled. 
So... what's the reasoning behind such significant change in the behavior of the component? And has this been synced with @maxkpower?

hiding fixed elements

The GIF above shows the wrong behavior I'm talking about. And if this is porbably not a big issue for the specific screen it will completely change things for Home screen and user accounts switcher. 
Also having this as a scrollable (not fixed element) makes my work on expandable/collapsable account section (on payments and tips screens) completely pointless. What is the reason to have compact mode of this panel if the whole thing is scrollable anyway?
I don't agree with the suggested UX approach and I'm curious to know what's the ratio behind it.

imollov:
> @davidyuk
> Fixes made by @imollov are satisfactory to me. The previous version was looking weird and broken on scroll. However from the demo video he sent me I can undesratnd that the component that was supposed to fixed to top by design is implemented in a different way (See the attached gif below where apparently the -1AE transaction info is hidden outside the visible viewport while user scrolling). Is that so for all instances of the same component like User account cards switcher, Balances, Payments, Tips... and many more to come probably as by design the intended behaviour is "fixed component" and the plan was to reuse it across the screens and functionalities. The ratio behind this is to have the important features and information accessible and visible on the screen (in front of user's eyes) all the time even when page is scrolled.
> So... what's the reasoning behind such significant change in the behavior of the component? And has this been synced with @maxkpower?
> 
> ...
> 
> The GIF above shows the wrong behavior I'm talking about. And if this is porbably not a big issue for the specific screen it will completely change things for Home screen and user accounts switcher.
> Also having this as a scrollable (not fixed element) makes my work on expandable/collapsible account section (on payments and tips screens) **completely pointless**. What is the reason to have compact mode of this panel if the whole thing is scrollable anyway?
> I don't agree with the suggested UX approach and I'm curious to know what's the ratio behind it.

regarding header areas which should stay on top while scrolling:
* I created a fix for Transaction details page https://github.com/aeternity/superhero-wallet/pull/1010. 
* Account/Home page is OK
* I will refactor Accounts page to use Plate and it should also achieve the desired behavior

here's a demo of https://github.com/aeternity/superhero-wallet/pull/1010 combined with #1006 which I think meets the design requirements mentioned above by @onvisions 

https://user-images.githubusercontent.com/482351/114916762-d9d0f880-9e2d-11eb-86e9-1a1006e10a44.mov

PR #935, opened 2021-03-26
fix(names): fix condition for a revoked names

### LINKED ISSUE #902: Names shown as revoked
Wire 2021-03-16 at 17-06-15

CedrikNikita:
https://github.com/aeternity/ae_mdw#naming-system
fixes #902

### DIFF
--- a/src/store/plugins/names.js
+++ b/src/store/plugins/names.js
@@ -58,36 +58,29 @@ export default (store) => {
         const names = await Promise.all([
           getPendingNameClaimTransactions(),
           middleware.getOwnedBy(account.address).then(({ active }) =>
-            active
-              .map(({ info, name }) => ({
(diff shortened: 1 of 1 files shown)


src/store/plugins/names.js
```diff
@@ -85,7 +85,7 @@ export default (store) => {
                         { root: true },
                       );
                     return {
-                      ...(revoked || oldName.revoked ? { revoked: true } : {}),
+                      ...(name.revoke ? { revoked: true } : {}),
```
davidyuk, line 88: So, `revoke` will be renamed to `revoked` only if `oldName` exist? I can't get why an `oldName` check is needed at all. Also, does middleware return **revoked** names in the list of **active** names?? Is there an issue about this?

PR #941, opened 2021-03-29
refactor(webpack): simplify path

davidyuk:
@Liubov-crypto ensure that it doesn't break the injected tip button styles


Liubov-crypto (commented):
Checked, LGTM.

PR #951, opened 2021-04-01
TransactionOverview component

imollov:
- used in Tx details page
- to be used in Sign and Connect popups

screenshot of claim tip tx

Screenshot 2021-04-01 at 20 53 48


Liubov-crypto (commented):
Tested on both test and mainnet, LGTM. 

send/receive tokens - pass
tip/retip  - pass
name claiming - pass
create/claim back invite link- pass
claim tip - pass
contract call -pass (send FT)
contract create - pass (create a voting poll)
name update - pass
name pre-claim - pass

src/popup/router/components/TransactionOverview.vue
```diff
@@ -0,0 +1,282 @@
...
+        <CopyAddress v-if="transaction.recipient.address" :value="transaction.recipient.address" />
+      </div>
+    </div>
+  </div>
+</template>
+
+<script>
+import { mapGetters } from 'vuex';
+import { OBJECT_ID_TX_TYPE } from '@aeternity/aepp-sdk/es/tx/builder/schema';
+import { TX_TYPES, TX_TYPES_MDW } from '../../utils/constants';
```
davidyuk, line 59: let's use `TX_TYPE` from SDK instead of `TX_TYPES`, #955

src/popup/router/components/TransactionOverview.vue
```diff
@@ -0,0 +1,282 @@
...
+    transaction() {
+      if (this.connect)
+        return {
+          sender: {
+            address: this.app.host,
+            name: this.app.name,
+          },
+          recipient: this.currentAccount,
+          title: this.$t('pages.connectConfirm.connecting'),
+          desc: '...',
```
davidyuk, line 96: please, use `description` instead, firstly I thought that it is about ordering (like asc/desc in SQL)

src/popup/router/components/TransactionOverview.vue
```diff
@@ -0,0 +1,282 @@
...
+    AensIcon,
+    ContractIcon,
+  },
+  props: {
+    tx: { type: Object },
+    claim: { type: Boolean },
+    event: { type: String },
+    contract: { type: String },
+    connect: { type: Boolean },
+    app: { type: Object },
```
davidyuk, line 83: please specify for each property the default value or mark it as required, otherwise, linter settings from #877 will warn about this

src/popup/router/components/TransactionOverview.vue
```diff
@@ -0,0 +1,282 @@
...
+              name: this.getDisplayName(this.tx.senderId),
+              label: this.$t('transaction.overview.accountAddress'),
+            },
+            recipient: {
+              address: this.tx.recipientId,
+              name: this.getDisplayName(this.tx.recipientId),
+              label: this.$t('transaction.overview.accountAddress'),
+            },
+            title: this.$t('transaction.type.spendTx'),
+            desc: this.txType,
```
davidyuk, line 122: please, use `description` instead, firstly I thought that it is about ordering (like asc/desc in SQL)

src/popup/router/components/TransactionOverview.vue
```diff
@@ -0,0 +1,282 @@
...
+      if (this.claim)
+        return {
+          sender: {
+            address: this.contract,
+            label: this.$t('transaction.overview.contract'),
+          },
+          recipient: this.currentAccount,
+          title: this.$t('transaction.overview.receiveTips'),
+          desc: this.event,
+        };
```
davidyuk, line 107: I would handle this as a general contract call transaction, this approach doesn't care about third-party developers, show the contract's method name instead.

src/popup/router/components/TransactionOverview.vue
```diff
@@ -0,0 +1,282 @@
...
+    txType() {
+      return TX_TYPES_MDW[this.tx.type] || OBJECT_ID_TX_TYPE[this.tx.tag];
+    },
+    currentAccount() {
+      return {
+        address: this.account.address,
+        name: this.account.name,
+        label: this.$t('transaction.overview.accountAddress'),
+      };
+    },
```
davidyuk, line 170: proposing to modify the `account` getter instead:
```js
...mapState({
  account(state, { account }) {
    return {
      ...account,
      label: this.$t('transaction.overview.accountAddress'),
    };
  },
}),
```
then we won't have `account` and `currentAccount` that looks different but means almost the same

src/popup/router/components/TransactionOverview.vue
```diff
@@ -0,0 +1,282 @@
...
+  },
+  methods: {
+    getDisplayName(address) {
+      return this.account.address === address ? this.account.name : '';
```
davidyuk, line 174: in the Base aepp we had a getter that was returning a name optionally making a request to MDW, would be nice to use something similar here to show names of other users (may be solved in a separate PR)

src/popup/router/components/TransactionOverview.vue
```diff
@@ -0,0 +1,282 @@
...
+    .description {
+      color: $color-dark-grey;
+
+      &::before {
+        content: '(';
+      }
+
+      &::after {
+        content: ')';
+      }
```
davidyuk, line 201: it is because of the i18n linter? we have disabled it in #918, I'm proposing to move brackets to the markup to keep it shorter

src/popup/router/components/TransactionOverview.vue
```diff
@@ -0,0 +1,282 @@
...
+        <TriangleRight />
+        <div class="line" />
+      </div>
+      <div v-if="transaction.aens" class="avatar">
+        <AensIcon class="icon" />
+      </div>
+      <div v-else-if="transaction.contractCreate" class="avatar">
+        <ContractIcon class="icon" />
+      </div>
+      <Avatar v-else v-bind="transaction.recipient" />
```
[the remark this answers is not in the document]
imollov: Currently, all supported transactions have a recipient defined.

src/popup/router/components/TransactionOverview.vue
imollov: Tip claim transactions are fetched from https://raendom-backend.z52da5wt.xyz/cache/events/?address=...&event=TipWithdrawn and have different fields from a general contract call transaction:

```
{
    "event": "TipWithdrawn",
    "caller": "ak_HryHrgaVUSxezLQBfNENhptKeXk6tpg4kjWBk88nsU5RJvhNk",
    "nonce": 11408,
    "height": 380886,
    "hash": "th_KCGPBwhc7mKo2rnuQfSpAKaFsh6zoyMmnLW93BiUP5QSmmHFf",
    "time": 1612287712669,
    "contract": "ct_2AfnEfCSZCTEkxL5Yoi4Yfq6fF7YapHRaFKDJK3THMXMBspp5z",
    "address": "ak_DUFY68J5QvRTDUEoAKSVXgYGovV86dyboNk8RVq3gvwQ1fQ5y",
    "amount": "120000000000000000",
    "url": "https://github.com/imollov",
    "tokenContract": null
  }
```

src/popup/router/components/TransactionOverview.vue
imollov: #958

src/popup/router/components/CopyAddress.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+  <div @click="copy" v-clipboard:copy="value" class="copy-address">
```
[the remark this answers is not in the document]
imollov: #960

src/popup/router/components/TransactionOverview.vue
davidyuk: Later middleware will provide a single list of transactions that have changed user's balance, for we can mimic it by something like:
```js
    let txs = await Promise.all([
      state.middleware.getTxByAccount(address, limit, page).then(({ data }) => data),
      dispatch('fetchPendingTransactions'),
      fetchJson(
        `${getters.activeNetwork.backendUrl}/cache/events/?address=${address}&event=TipWithdrawn${
          recent ? `&limit=${limit}` : ``
        }`,
      )
        .then((response) =>
          response.map(({ hash }) => state.middleware.getTxByHash(hash)),
        )
        .catch(() => []),
    ]);
```
[here](https://github.com/aeternity/superhero-wallet/blob/28cdf10f00df58ecc7efe7bd8593b6e9939e4ffe/src/store/actions.js#L40-L57)

src/popup/router/components/TransactionOverview.vue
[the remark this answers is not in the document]
davidyuk: but then you will be able to support all transactions to some extent, contract icon doesn't look like something specific to contracts

src/popup/router/components/TransactionOverview.vue
```diff
@@ -0,0 +1,272 @@
+<template>
+  <div class="transaction-overview">
+    <div class="title">
+      {{ transaction.title }}
+      <span class="description">({{ transaction.description }})</span>
```
davidyuk, line 5:
```suggestion
```
with @onvisions we [agreed](https://www.figma.com/file/ewTV7C01eFlKVODGHokUI0/Superhero-Wallet-UI?node-id=31%3A1533) to remove this

src/popup/router/components/TransactionOverview.vue
```diff
@@ -0,0 +1,272 @@
...
+    TruncateMid,
+    CopyAddress,
+    TriangleRight,
+    AensIcon,
+    ActionIcon,
+  },
+  props: {
+    tx: { type: Object, default: null },
+    connect: { type: Boolean, default: false },
+    app: { type: Object, default: null },
```
davidyuk, line 82: Splitting this component into Overview, TransactionOverview, ConnectOverview (?), would allow to make it more readable and to check props stricter

src/popup/router/components/TransactionOverview.vue
```diff
@@ -0,0 +1,272 @@
...
+      <div class="mid">
+        <TriangleRight />
+        <div class="line" />
+      </div>
+      <div v-if="transaction.aens" class="avatar">
+        <AensIcon class="icon" />
+      </div>
+      <div v-else-if="transaction.contractCreate" class="avatar">
+        <ActionIcon class="icon" />
+      </div>
```
davidyuk, line 19:
```suggestion
```
this doesn't seem to be needed because it would be handled the same by the last `v-else`

src/popup/router/components/TransactionOverview.vue
imollov: I extracted Overview from TransactionOverview. Connect modal can use Overview with little adaptation.

davidyuk (approved):

PR #952, opened 2021-04-05
Rebranded design adjustments

imollov:


davidyuk (approved):
changes looks fine, but `style` as a convertional commit type is misused, it is intented to be used for changes in code style (like a lite version of `refactor`), not the changes in CSS:

> style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)

https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

I'm proposing to use one of: `fix(styles)` or `fix(side-menu)` instead.

davidyuk (approved):

PR #959, opened 2021-04-07
Fix commitlint

davidyuk:


davidyuk:
I have tested it and added it to develop branch

PR #969, opened 2021-04-09
Rebrand ConfirmTransactionSign modal

### LINKED ISSUE #881: Don't stretch confirm modals to the whole page
In a web wallet, the pop-up should be on the whole page? is this correct?
full screen

_Originally posted by @Liubov-crypto in https://github.com/aeternity/superhero-wallet/issues/864#issuecomment-792632730_

cc: @onvisions

### LINKED ISSUE #890: Name claim fee
Name register fee is not added in the total fee for the transaction, rather displayed below.

imollov:
fixes #890, fixes #881

Screenshot 2021-04-22 at 20 18 39

Screenshot 2021-04-22 at 20 19 05


src/popup/router/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -1,24 +1,53 @@
...
+      >
+        <Minus v-if="showAdvanced" />
+        <Plus v-else />
+        <span class="text">{{ $t('pages.signTransaction.advanced') }}</span>
+      </button>
+      <CopyButton :value="code" :message="$t('copied')" />
+    </div>
+
+    <div class="code" v-if="showAdvanced">
+      <DetailsItem :value="code" :label="$t('pages.signTransaction.code')" />
```
davidyuk, line 37: I think we need to redesign and use `SignDetailsList` here. Different transaction types have fields that would be useful to show here, not only the `code`.

src/popup/router/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -29,21 +58,118 @@
...
   },
   mixins: [mixin],
   props: {
     transaction: { type: Object, required: true },
   },
+  data: () => ({ showAdvanced: false }),
+  computed: {
+    ...mapGetters(['getTxSymbol', 'getTxAmountTotal']),
+    code() {
+      return this.transaction.payload || this.transaction.callData;
```
davidyuk, line 92: I wouldn't name `payload` a `code`, also would be better to decode it than to show something like `ba_Xfbg4g==` (actually, it should be an empty string)

src/popup/router/components/Button.vue
```diff
@@ -46,6 +47,7 @@ export default {
     inactive: Boolean,
     to: [String, Object],
     bold: Boolean,
+    secondaryModal: Boolean,
```
davidyuk, line 50: can it be combined with `dark`? I don't think that we need so lot of types of buttons

src/popup/router/components/Modals/ConfirmTransactionSign.vue
imollov: Which other fields do you think are useful for different transaction types?

src/popup/router/components/Button.vue
imollov: `dark` is used in modals and a few pages that are not rebranded yet. 

In the design system modals have special secondary button, not sure if the same will be used on the pages using `dark` at the moment (designs not available).

davidyuk: @onvisions can we make all old `dark` buttons to have the same color as in new designs?

src/popup/router/components/Modals/ConfirmTransactionSign.vue
davidyuk: in Base aepp we were rendering these fields:
https://github.com/aeternity/aepp-base/blob/e25875692ea3f0be766e0911507daed59b8fe05a/src/components/mobile/ConfirmTransactionSignModal.vue#L101-L111
We can add nonce as far as it is displayed in transaction details. In the future, we may add fields of oracle-related transactions, state channels and so.

src/popup/router/components/Button.vue
onvisions: @davidyuk , yes we can. :) I synced the issue with @imollov.
The reason for different secondary button color in the modals was the contrast between modal background and button but I think we can use #232323 for all secondary buttons.

onvisions (approved):
We synced with @imollov and I think it looks perfect. Approved on my side.

Liubov-crypto (commented):
Missing reject / confirm button in popup in FF extension:



Oversized modals: #1036

Liubov-crypto:
Android app LGTM, except error with QR scanner: https://github.com/aeternity/superhero-wallet/issues/1001#issuecomment-826737724

Liubov-crypto:
Rechecked on superhero feed (voting and retip), (create an invite link, registering a name) now looks good to me.

davidyuk (approved):
Code changes look good 👍

etharner:
@imollov Please rebase and LGTM

etharner (approved):

PR #971, opened 2021-04-12
fix(web-iframe-popups): make iframe popups notRebranded

etharner:
This PR fixes this behaviour in web-iframe-popups:
image


davidyuk (approved):

davidyuk:
Screenshot 2021-04-14 at 03 39 33

"generate" e2e test fails because the menu is not visible

PR #972, opened 2021-04-12
Limit recent transactions count

### LINKED ISSUE #917: Recent transactions contain all transactions after opening the list of transactions
this is a side effect of this workaround: https://github.com/aeternity/superhero-wallet/pull/883/files#diff-42c781502dd0badc5c59da2f8253cdbee661a4cb9ec87b7ad4703bad7fb0f4b3R125

imollov:
fixes #917

### DIFF
--- a/src/popup/router/components/RecentTransactions.vue
+++ b/src/popup/router/components/RecentTransactions.vue
@@ -50,7 +50,7 @@ export default {
   methods: {
     async updateTransactions() {
       this.$store.commit(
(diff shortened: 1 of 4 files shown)


Liubov-crypto (commented):
Web wallet LGTM.

Chrome extension is missing an item - View Full Transaction History:
chrome

the FF extension is hidden under the panel and I cannot log in:
hidden in FF

davidyuk (approved):

PR #981, opened 2021-04-14
fix(tx-details): expand content and fill bg

imollov:
temporary layout fix until #951 is merged


src/popup/router/pages/TransactionDetails.vue
```diff
@@ -186,7 +186,8 @@ export default {
 @import '../../../styles/typography';
 
 .transaction-details {
-  overflow-x: hidden;
+  height: 100%;
```
davidyuk, line 189: I can't get the reasoning, also I don't think that usual pages should set theit height explicitly. Would be nice to avoid extra scrolling of this page and others.

PR #1006, opened 2021-04-15
Fix app background color

imollov:
fixes:

* pages with different background color from header 

| before | after |
| ------ | ----- |
|Screenshot 2021-04-15 at 15 21 23 | Screenshot 2021-04-15 at 15 21 35|

* overscroll (solution suggested here https://github.com/aeternity/superhero-wallet/pull/931#discussion_r613115772)

| before | after |
| ------ | ----- |
|Screenshot 2021-04-15 at 13 12 08| Screenshot 2021-04-15 at 13 14 20 |

### DIFF
--- a/src/popup/App.vue
+++ b/src/popup/App.vue
@@ -118,6 +118,10 @@ body {
   background-color: variables.$color-black;
 }
 
+body.color-bg-3 {
+  background-color: variables.$color-bg-3;
+}
+
 html,
 body {
   height: var(--height);
--- a/src/popup/App.vue
+++ b/src/popup/App.vue
@@ -184,14 +184,18 @@ body {
     }
   }
 
-  &.show-header .main {
-    padding-top: 48px;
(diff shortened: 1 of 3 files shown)


src/popup/App.vue
```diff
@@ -117,6 +117,10 @@ body {
   background-color: $color-black;
 }
 
+body.bg {
```
davidyuk, line 120: the class name probably is too unspecific

davidyuk (approved):
lgtm

src/popup/App.vue
imollov: Agreed. Suggestions?

imollov: how about `body.header-bg`?

imollov: or `body.main-bg`?

imollov: or `body.color-bg-3`?

PR #1011, opened 2021-04-16
refactor(accounts): use Plate component

imollov:
| before | after |
| ------ |  ---- |
| recording (13)|recording (12)|


davidyuk (approved):
Corresponding designs: https://www.figma.com/proto/ewTV7C01eFlKVODGHokUI0/Superhero-Wallet-UI?node-id=575%3A12040&scaling=min-zoom

PR #1015, opened 2021-04-16
fix(actions): get available tokens when switching network

### LINKED ISSUE #907: Error when switching networks
> Screenshot 2021-03-17 at 16 09 53
...

Now I got this error when switched between networks.

_Originally posted by @Liubov-crypto in https://github.com/aeternity/superhero-wallet/issues/883#issuecomment-801130363_

tested on develop 226adef

quick findings:
* to reproduce you must have a ft transfer transaction
* seems like the available ft tokens in the state are not in sync with transactions for a quick moment after switching networks

related to #905

imollov:
fixes #907

### DIFF
--- a/src/store/actions.js
+++ b/src/store/actions.js
@@ -13,9 +13,10 @@ import {
 import { i18n } from './plugins/languages';
 
 export default {
-  switchNetwork({ commit }, payload) {
+  async switchNetwork({ commit, dispatch }, payload) {
     commit('switchNetwork', payload);
     commit('setTransactions', []);
+    await dispatch('fungibleTokens/getAvailableTokens');
   },
   async fetchPendingTransactions({
     state: { sdk },


src/store/actions.js
```diff
@@ -13,9 +13,10 @@ import {
 import { i18n } from './plugins/languages';
 
 export default {
-  switchNetwork({ commit }, payload) {
+  async switchNetwork({ commit, dispatch }, payload) {
     commit('switchNetwork', payload);
     commit('setTransactions', []);
+    await dispatch('fungibleTokens/getAvailableTokens');
```
davidyuk, line 19: I'm proposing to make `fungibleTokens` a plugin and to subscribe for network updates from there instead of making network switcher aware of multiple parts of the app.

davidyuk (changes requested):

src/store/actions.js
imollov: Actually that's done in #1016 so I'm closing this one.

PR #1017, opened 2021-04-19
fix(tokens): show not null token info

### LINKED ISSUE #968: All fungible tokens disappeared from the Balance tab
After this error, all fungible tokens disappeared from the Balance tab:
зображення_viber_2021-04-08_11-02-35

_Originally posted by @Liubov-crypto in https://github.com/aeternity/superhero-wallet/issues/957#issuecomment-815559362_

CedrikNikita:
fixes #968

### DIFF
--- a/src/popup/router/components/FungibleTokens/TokensList.vue
+++ b/src/popup/router/components/FungibleTokens/TokensList.vue
@@ -52,13 +52,15 @@ export default {
      * Converts the token information object into a searchable list
      */
     convertedTokenInfo() {
-      return Object.entries(this.availableTokens).map(([contract, tokenData]) => ({
-        name: tokenData.name,
-        symbol: tokenData.symbol,
-        contract,
-        decimals: tokenData.decimals,
-        convertedBalance: tokenData.convertedBalance,
-      }));
+      return Object.entries(this.availableTokens)
+        .filter(([, tokenData]) => tokenData)
+        .map(([contract, tokenData]) => ({
+          name: tokenData.name,
+          symbol: tokenData.symbol,
+          contract,
+          decimals: tokenData.decimals,
+          convertedBalance: tokenData.convertedBalance,
+        }));
     },
     filteredResults() {
       const tokensInfo = [


Liubov-crypto:
Tested, LGTM.

src/popup/router/components/FungibleTokens/TokensList.vue
```diff
@@ -52,13 +52,15 @@ export default {
...
-      }));
+      return Object.entries(this.availableTokens)
+        .filter(([, tokenData]) => tokenData)
+        .map(([contract, tokenData]) => ({
+          name: tokenData.name,
+          symbol: tokenData.symbol,
+          contract,
+          decimals: tokenData.decimals,
+          convertedBalance: tokenData.convertedBalance,
+        }));
```
davidyuk, line 63: in which case the value in [token info](https://raendom-backend.z52da5wt.xyz/tokenCache/tokenInfo) response can be `null`? Is it a backend issue? cc: @kenodressel 

if it is a case for some reason then proposing to filter it at [the top level](https://github.com/aeternity/superhero-wallet/blob/b03bde3e4ee32196de5a95c7bdcc20ba1d7c75f4/src/store/modules/fungibleTokens.js#L33-L35), right after data received to avoid breaking in other places for the same reason

davidyuk (changes requested):

src/popup/router/components/FungibleTokens/TokensList.vue
kenodressel: It can be null because we are fetching the token meta info from the contract via the node for every token. If a call fails it can be `null´.

CedrikNikita: > in which case the value in [token info](https://raendom-backend.z52da5wt.xyz/tokenCache/tokenInfo) response can be `null`? Is it a backend issue? cc: @kenodressel
> 
> if it is a case for some reason then proposing to filter it at [the top level](https://github.com/aeternity/superhero-wallet/blob/b03bde3e4ee32196de5a95c7bdcc20ba1d7c75f4/src/store/modules/fungibleTokens.js#L33-L35), right after data received to avoid breaking in other places for the same reason

I agree that this should be filtered out on the top level, and yes the endpoint returns null for this contract `ct_cxp2TozAG6nzkQ84A1PyTWcKtgtPA52xTFkJPe2yxAeSdFnbY`

`ct_cxp2TozAG6nzkQ84A1PyTWcKtgtPA52xTFkJPe2yxAeSdFnbY":null,`

https://test-tipping.aeternity.art/tokenCache/tokenInfo

CedrikNikita:
https://github.com/aeternity/superhero-wallet/pull/1017#discussion_r617148831

PR #1027, opened 2021-04-21
Fix failing tests

### LINKED ISSUE #1032: cypress: Find a better way to assert node content
https://github.com/aeternity/superhero-wallet/blob/d02410462c1d3d9e31d50f92f1214791c97803b1/tests/e2e/integration/amount-input.js#L8

https://github.com/aeternity/superhero-wallet/blob/d02410462c1d3d9e31d50f92f1214791c97803b1/tests/e2e/integration/amount-input.js#L12

Consider using should instead as far as it is "automatically retried until they pass or time out" (https://docs.cypress.io/api/commands/should).

CedrikNikita:
fixed #1032

### DIFF
--- a/tests/e2e/integration/networks.js
+++ b/tests/e2e/integration/networks.js
@@ -84,7 +84,7 @@ describe('Test cases for networks page', () => {
       .should('contain', defaultNetwork.middlewareUrl)
       .goBack()
       .goBack()
-      .get('.transaction-list')
+      .get('.transaction-list', { timeout: 8000 })
       .should('exist')
 
       .openNetworks()
--- a/tests/e2e/integration/transactions.js
+++ b/tests/e2e/integration/transactions.js
@@ -1,42 +1,17 @@
 import { TXS_PER_PAGE } from '../../../src/popup/utils/constants';
-import { testAccount } from '../../../src/popup/utils/config';
-import { formatDate, formatTime } from '../../../src/popup/utils';
 
 describe('Tests cases for transactions page', () => {
-  beforeEach(() => {
-    cy.login();
-  });
-
   it('Load transactions, load additional transactions on scroll', () => {
-    cy.openTransactions().get('[data-cy=list]').children().should('have.length', TXS_PER_PAGE);
-
+    cy.login()
+      .openTransactions()
+      .get('[data-cy=list]')
+      .children()
+      .should((el) => expect(el).to.have.length.greaterThan(1));
     cy.scrollTo('bottom')
       .get('[data-cy=loader]')
       .should('be.visible')
       .get('[data-cy=list]')
       .children()
-      .should('have.length', TXS_PER_PAGE * 2);
-  });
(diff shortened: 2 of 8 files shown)


etharner (approved):

tests/e2e/integration/transactions.js
```diff
@@ -1,42 +1,20 @@
...
-  });
-
   it('Load transactions, load additional transactions on scroll', () => {
-    cy.openTransactions().get('[data-cy=list]').children().should('have.length', TXS_PER_PAGE);
+    cy.login()
+      .openTransactions()
+      .get('[data-cy=list]')
+      .children()
+      .its('length')
+      .should('be.gte', 0);
```
davidyuk, line 10: isn't the length of anything is always greater than or equal to 0?
```suggestion
      .should('be.gt', 0);
```

tests/e2e/integration/amount-send.js
```diff
@@ -1,7 +1,9 @@
 describe('Test cases AmountSend component', () => {
   it('Calculates currency on enter amount, validates entered amount, shows correct balance', () => {
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
     cy.login()
       .openWithdraw()
+      .wait(1000)
       .get('[data-cy=balance-currency]')
       .invoke('text')
       .then((text) => expect(text).not.to.eq('$0.00'))
```
davidyuk, line 9:
```suggestion
    cy.login()
      .openWithdraw()
      .get('[data-cy=balance-currency]')
      .invoke('text')
      .should('not.equal', '$0.00')
```

tests/e2e/integration/networks.js
```diff
@@ -84,6 +85,7 @@ describe('Test cases for networks page', () => {
       .should('contain', defaultNetwork.middlewareUrl)
       .goBack()
       .goBack()
+      .wait(2500)
       .get('.transaction-list')
       .should('exist')
```
davidyuk, line 90: I can't get why `wait` is needed here.
> Create an assertion. Assertions are automatically retried until they pass or time out.
https://docs.cypress.io/api/commands/should#Syntax

tests/e2e/integration/transactions.js
CedrikNikita: That is obviously a mistake, yes

tests/e2e/integration/networks.js
CedrikNikita: The goal of this PR was to find the weak spots, and prove them by restarting test with empty commits. That's why there were status/hodl label, this was not the final stage of development (intentionally). You are right it's is better to increase timeout for this get.

tests/e2e/integration/amount-send.js
CedrikNikita: It won't work since text are not trimmed.
https://github.com/cypress-io/cypress/issues/3887

PR #1030, opened 2021-04-22
feat(tabs): add Tabs component and split payments and tips screens

etharner:
This PR implements and uses Tabs component which is needed for rebranding of Send/Receive, Tip/Claim and AENS


davidyuk (changes requested):
The suggested solution looks interesting, but I'm proposing to use a straightforward router-based solution instead.

davidyuk:
Screenshot 2021-04-22 at 17 41 47

probably we need to remove extra account info on payments-receive page

davidyuk (approved):

src/popup/router/pages/Tips.vue
```diff
@@ -0,0 +1,36 @@
...
+    <BalanceInfo />
+    <Tabs>
+      <RouterLink
+        :to="{ name: 'tips-send' }"
+        exact-path
+      >
+        <TipIcon /> {{ $t('pages.titles.send-tip') }}
+      </RouterLink>
+      <RouterLink :to="{ name: 'tips-claim' }">
+        <ClaimTipsIcon /> {{ $t('pages.titles.claim-tips') }}
```
imollov, line 13: Icon seems to be different from designs in Figma, as well as letter casing:

Screenshot 2021-04-26 at 11 29 36

src/popup/router/components/Header.vue
```diff
@@ -121,8 +121,10 @@ export default {
     ...mapMutations(['setNotificationsStatus']),
     back() {
       const fallBackRoute = this.isLoggedIn ? '/account' : '/';
+      let { fullPath } = this.$route;
+      fullPath = fullPath.endsWith('/') ? fullPath.slice(0, -1) : fullPath;
```
imollov, line 125: Isn't it a bad practice to mutate router object props?

etharner: I'm making destructuring assignment of string, it'll just copy the value

davidyuk (approved):

PR #1035, opened 2021-04-22
fix(tab-bar): color home button area on iOS

davidyuk:
Before

Screenshot 2021-04-22 at 20 41 52

After

Screenshot 2021-04-22 at 20 39 18

### DIFF
--- a/src/popup/router/components/TabBar.vue
+++ b/src/popup/router/components/TabBar.vue
@@ -38,7 +38,7 @@ export default {
 .tab-bar {
   display: flex;
   background-color: variables.$color-bg-3;
-  padding-bottom: env(safe-area-inset-bottom);
+  border-bottom: env(safe-area-inset-bottom) solid variables.$color-bg-2;
 
   > a {
     height: 48px;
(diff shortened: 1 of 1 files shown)


onvisions:
> 
> 
> Before
> 
> Screenshot 2021-04-22 at 20 41 52
> 
> After
> 
> Screenshot 2021-04-22 at 20 39 18
Suggested changes definitely make the UI look better than before. However I would suggest getting rid of the left, right and bottom menu borders for iOS devices. Because of the specifics of the device viewport it will look better (from visual perspective). Isn't it possible to have this approach applied for iOS device screens only? I can sync specific styling approach in sync with @imollov to solve this issue (briefly: to have first and last buttons left and right borders otside the viewport and the bottom border color changed to the color of background).
@davidyuk @maxkpower 
Screenshot_7

davidyuk:
The latest version looks this way

Screenshot 2021-04-23 at 23 25 53

etharner (approved):

Liubov-crypto:
Checked in Safari and FF mobile browser, LGTM.

davidyuk:
I think it is not fine, but it can't be caused by this PR.

Issue #999, opened 2021-04-15
Network Error


Liubov-crypto:
Enter to the new wallet:
![enter to the new wallet ](https://user-images.githubusercontent.com/69896204/114860885-37951e80-9df5-11eb-9ba1-16925b8861a1.png)

Liubov-crypto:
Maybe this issue related to #907 ?

Liubov-crypto:
How I tested it: I logged into the wallet, then switch off the internet connection, waited a few minutes, then restored the internet and reopened the wallet.
FF browser: 
TypeError: NetworkError when attempting to fetch resource: 
![ff netw err](https://user-images.githubusercontent.com/69896204/116384053-28b06200-a820-11eb-9c76-592f695aa6fd.png)
![fff err](https://user-images.githubusercontent.com/69896204/116384061-29e18f00-a820-11eb-8bf6-8ee0e042747b.png)

![compiler](https://user-images.githubusercontent.com/69896204/116384925-fe12d900-a820-11eb-9a20-c1fc51d04ec9.png)
![comp](https://user-images.githubusercontent.com/69896204/116384935-00753300-a821-11eb-9c53-fc03f1aa966b.png)


Brave browser: 
![brave](https://user-images.githubusercontent.com/69896204/116384345-6f05c100-a820-11eb-9ac9-4df95bf2d9fb.png)
![brave1](https://user-images.githubusercontent.com/69896204/116384360-70cf8480-a820-11eb-86d1-806ac13f1e7e.png)

Safari browser:
![netw err](https://user-images.githubusercontent.com/69896204/116384718-cb68e080-a820-11eb-9016-9a52e0d59901.png)
![saf1](https://user-images.githubusercontent.com/69896204/116384735-cf94fe00-a820-11eb-8f87-6391c4f5b44a.png)

development branch 1decc7f / v.0.6.0

imollov:
findings: 
- triggered when internet connection is interrupted
- network errors are caused by `fetch` in the wallet and `axios` from using the SDK
- `fetch` is implemented differently across browser and generally doesn't provide reliable info for the connection

suggested next steps:
- fetch wrapper that handles interrupted connection cross-browser and cross-platform
- basic offline support

Liubov-crypto:
Can't reproduce this error on the current branch. Can be closed.

PR #1042, opened 2021-04-27
Fix network errors

imollov:
notes:
* currently only fixes issues in extension

todo:
* fix web version

### DIFF
--- a/src/manifest.js
+++ b/src/manifest.js
@@ -26,6 +26,10 @@ module.exports = (isProd) => ({
     'https://twitter.com/*',
     'https://*.youtube.com/watch?v=*',
     'https://wallet.superhero.com/*',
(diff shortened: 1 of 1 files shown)


src/manifest.js
```diff
@@ -20,12 +20,17 @@ module.exports = (isProd) => ({
     'webRequest',
     'webRequestBlocking',
     'webNavigation',
+    '*://localhost/*',
```
imollov, line 23: https://github.com/github/fetch/issues/310

Liubov-crypto:
Found these errors. They are also reproducible on develop branch https://github.com/aeternity/superhero-wallet/issues/999#issuecomment-828324187 (after the internet is falling):
ff ext
ff ext2

Issue #1028, opened 2021-04-21
Error when sending FT to .chain name


Liubov-crypto:
![chainname](https://user-images.githubusercontent.com/69896204/115547801-61e05380-a2af-11eb-96bc-88d51f88354c.png)
![ttt](https://user-images.githubusercontent.com/69896204/115547810-6442ad80-a2af-11eb-9ef2-7342df8210bd.png)
![11111](https://user-images.githubusercontent.com/69896204/115548061-b1268400-a2af-11eb-94a0-7f5fab9a5822.png)

develop 68a2ffe / v.0.6.0

onvisions:
Error messages should be displayed in the modal template for the wallet and not in this wide version (as in the screenshot with error above).

davidyuk:
@onvisions can you send a link to proper designs?

onvisions:
@davidyuk , I am discussing here only the issue with full screen popups in web wallet suggesting to use the regular modal template. Its design is in "Superhero Wallet UI" (FIgma) and there are several examples of different modals included. I won't publish the direct link as I was adviced before not to include links to internal resources in our public discussions.

thepiwo:
I think sending fungible tokens should forbid send to name functionality a its not native to AEX-9

PR #1045, opened 2021-04-28
chore: update SDK to develop version

davidyuk:


Liubov-crypto:
During testing, I found the following errors:
This error appeared constantly:
ok errr
ok error

No connection with sh UI. I clicked on create a new post in the sh feed and was redirected to the extension. When I tried to go back I got the following error:
errr in ff ext
t error
mainteinance error on th eui

All my .chain names have disappeared from the wallet. No names found:
no names found

No connection with Jelly Swap. There are no URLs in the permission settings for Jelly Swap, UI testing, etc.

Broken manifest for chrome extension, can't install it.

These errors also remain #1028 https://github.com/aeternity/superhero-wallet/issues/907#issuecomment-809553301

Liubov-crypto:
Got this error after clicking on the Notifications bell icon and after sending ae from the main account to sub-account in the wallet (although the transaction was made):
sub acc wal
send ae from 1 acc to sub acc in the wallet
bell icon
bell


Tx history appears with delay in sub-accounts:

https://user-images.githubusercontent.com/69896204/116556845-96cf5480-a906-11eb-829e-88ae75da62c1.mov

After the registration, the name disappears, does not hang as a pending in the wallet (herosuperhero.chain):
th_2iDFdZ8zF5W7hPfvVQeuWz9Zzp1vgJ42bwVZd3FBjbw9JH7kHX

davidyuk:
Should be fixed:
- OK exception
- ability to install into Chrome

Liubov-crypto:
if I click on Receive in Balances, then there is no QR code:
receive error

when sending tokens, the wallet turns black, then everything works again:
https://user-images.githubusercontent.com/69896204/116678168-b3749680-a9b1-11eb-9bc4-d8fc08f5329b.mov

Design bug: 
design bug

Should be fixed:
* OK exception - fixed
* Ability to install into Chrome - fixed
* connection with Jelly Swap - fixed  May be fixed (can't reproduce):
* chain names have disappeared - fixed

etharner (approved):

PR #1064, opened 2021-05-05
Fix warnings

### LINKED ISSUE #871: Avoid console.log statements to fix the most of linter warnings
https://eslint.org/docs/rules/no-console

CedrikNikita:
fixes #871

### DIFF
--- a/src/popup/router/components/FungibleTokens/TokensListItem.vue
+++ b/src/popup/router/components/FungibleTokens/TokensListItem.vue
@@ -50,8 +50,8 @@ import TokenAmount from '../TokenAmount';
 export default {
   components: { Avatar, TruncateMid, TokenAmount },
   props: {
-    tokenData: Object,
-    name: String,
+    tokenData: { type: Object, default: null },
+    name: { type: String, default: '' },
   },
   computed: mapGetters(['formatCurrency']),
 };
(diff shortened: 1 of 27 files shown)


src/popup/router/components/FungibleTokens/TokensListItem.vue
```diff
@@ -50,8 +50,8 @@ import TokenAmount from '../TokenAmount';
 export default {
   components: { Avatar, TruncateMid, TokenAmount },
   props: {
-    tokenData: Object,
-    name: String,
+    tokenData: { type: Object, default: null },
+    name: { type: String, default: '' },
```
etharner, line 54:
```suggestion
```
Looks like it isn't used in this component

src/popup/router/components/TokenAmount.vue
```diff
@@ -19,11 +19,12 @@ export default {
   props: {
     amount: { type: Number, required: true },
     symbol: { type: String, default: 'AE' },
-    altText: { type: String },
+    altText: { type: String, default: '' },
     hideFiat: { type: Boolean },
     direction: {
       type: String,
       validator: (value) => ['sent', 'received'].includes(value),
+      default: 'sent',
```
etharner, line 27: Using this approach we will have `-amount` with red color where it's not needed, for example in invite links:
image
Let's maybe introduce empty direction which will looks neutral
image

```suggestion
      validator: (value) => ['sent', 'received', ''].includes(value),
      default: '',
```

CedrikNikita: prefer `default: undefined` in this case instead of changing validator

etharner (approved):

PR #1068, opened 2021-05-05
perf(icons): optimize iframe icons

### LINKED ISSUE #987: Optimise src/icons/iframe/sendAndReceive.svg
The current size of sendAndReceive.svg after optimization is 233kb, this is approximately equal to the size of JS SDK.

I see two ways to improve it:
- each circle at the background is exported as a path instead of a circle that is much expensive than to store x, y, and radius
- chars are exported as paths as well, would be more efficient to keep them as text, and to reuse the font that is used in UI

cc: @onvisions

CedrikNikita:
fixes #987 
used this https://jakearchibald.github.io/svgomg/ to optimize svgs


etharner:
Maybe this can be further improved by manual techniques but anyway it's good result already.

etharner (approved):

PR #1077, opened 2021-05-13
fix(pending): fix duplicate pending transactions

etharner:
We catched this because currently txs aren't mined on testnet and we have time to see this problem.

### DIFF
--- a/src/popup/router/components/PendingTxs.vue
+++ b/src/popup/router/components/PendingTxs.vue
@@ -21,15 +21,7 @@ export default {
   computed: mapState({
     transactions: ({ transactions: { pending } }) => pending
       .filter(({ amount, hash }) => !Number.isNaN(+amount) && hash)
-      .map(({
-        microTime, hash, tipUrl, ...tx
-      }) => ({
-        microTime,
-        hash,
-        tipUrl,
-        tx,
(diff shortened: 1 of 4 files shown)


src/popup/router/components/PendingTxs.vue
```diff
@@ -21,15 +21,7 @@ export default {
...
-      .map(({
-        microTime, hash, tipUrl, ...tx
-      }) => ({
-        microTime,
-        hash,
-        tipUrl,
-        tx,
-        pending: true,
-      })),
+      .map((tx) => ({ ...tx, pending: true })),
```
CedrikNikita, line 24: The original structure shouldn't be broken
https://github.com/aeternity/superhero-wallet/blob/develop/src/store/getters.js#L106 this won't work properly for example

CedrikNikita (changes requested):

src/popup/router/components/PendingTxs.vue
CedrikNikita: Didn't know that the sructure was broken (it was transaction.tx.tx instead of transaction.tx) and that you are fixing it here.

CedrikNikita (approved):

PR #1086, opened 2021-05-19
fix(tx-pointers): fix name pointers field error in confirm tx modal

etharner:
`DetailsItem` and `CopyButton` couldn't work with `pointers` tx field because it's an array


etharner:
> @etharner which tx type has `pointers` field?

Update pointer transactions

imollov (approved):
LGTM

PR #1092, opened 2021-05-19
fix(details): block buttons and don't watch before sdk init

etharner:
This fixes reloading of name details page leading to watch before sdk init and corresponding error.


CedrikNikita (approved):
Code changes look good.

PR #1093, opened 2021-05-20
Fix <ul> <ol> usage, remove timeout in Transactions

### LINKED ISSUE #944: Remove weird setTimeout
https://github.com/aeternity/superhero-wallet/blob/0a5d109ce898f6fd6e45173a611edad521e047c0/src/popup/router/pages/Transactions.vue#L87
I can't recall why it was introduced in https://github.com/aeternity/superhero-wallet/commit/94afe3cac6508fe6f1f2730505194b1f4fb2c4a7 🤷‍♀️

### LINKED ISSUE #945: Use <ul>, <ol> only when you need a usual list of text items
For example, the transaction list shouldn't use `<ul>`, but the list of places from where the wallet can be downloaded should use`<ol>`.
https://github.com/aeternity/superhero-wallet/blob/0a5d109ce898f6fd6e45173a611edad521e047c0/src/popup/router/pages/Transactions.vue#L6
https://github.com/aeternity/superhero-wallet/blob/c4f8573f3f0300cf3f955a755dee39553d5d5f6d/src/popup/router/pages/PrivacyPolicy.vue#L33-L34

AtanasKrondev:
Fixes #944 
Fixes #945

### DIFF
--- a/src/popup/router/pages/Names/AuctionList.vue
+++ b/src/popup/router/pages/Names/AuctionList.vue
@@ -27,7 +27,7 @@
       {{ $t('pages.names.auctions.bid') }}
     </Button>
 
-    <ul
+    <div
       v-if="activeAuctions.length"
       class="list"
     >
@@ -46,7 +46,7 @@
           {{ (expiration - topBlockHeight) | blocksToRelativeTime }}
         </div>
       </NameRow>
-    </ul>
+    </div>
 
     <p v-else>
       {{ $t('pages.names.auctions.no-auctions') }}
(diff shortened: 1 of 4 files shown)


CedrikNikita (commented):
I don't like the way #944 is solved, because this will end up in making hundreds of request within seconds, when the user will reach the end of transaction history.
This issue should be fixed with a better structure to avoid such behaviour.

Liubov-crypto (commented):
Checked, Tx list LGTM.
The image of loading transactions is shifted to the bottom panel when quickly scrolling down (reproducible on develop branch commit 120c024 / v.0.7.0): 

https://user-images.githubusercontent.com/69896204/120315978-98b17c80-c2e5-11eb-9785-d623c09e7f96.mov

the same transaction (withdraw, tip) is displayed 2 times:
2 trans view
2 times transaction

PR #1096, opened 2021-05-20
Enable unfinished features

etharner:


CedrikNikita (changes requested):
I think that this PR should contain all the changes, that was made additional with disabling unfinished features.
Example on this branch:
width

CedrikNikita (approved):
Changes look reasonable

PR #1100, opened 2021-05-21
Rebrand modals

### LINKED ISSUE #1036: Oversized modals in the Web wallet
send tip conf
rrr

imollov:
fixes #1036

Screenshot 2021-05-21 at 16 15 44

Screenshot 2021-05-21 at 16 12 35

Screenshot 2021-05-21 at 16 12 14

Screenshot 2021-05-21 at 17 00 23

Screenshot 2021-05-21 at 17 01 03

Screenshot 2021-05-21 at 17 04 45

Screenshot 2021-05-21 at 17 09 01

Screenshot 2021-06-03 at 16 15 50

Simulator Screen Shot - iPhone 12 mini - 2021-05-21 at 17 52 46

Screenshot_1621608870


Liubov-crypto (approved):
tested, LGTM.

src/popup/router/components/Modals/Default.vue
```diff
@@ -4,13 +4,27 @@
...
-    >
-      {{ title }}
+    <template slot="header">
+      <Component
+        :is="icon"
+        v-if="icon"
+        class="icon"
+        :class="type"
+      />
+      {{ title || $t(`modals.${type}`).title || '' }}
```
CedrikNikita, line 14: I think that there shoul be `no-dynamic-keys`, since this PR is called `Rebrand modals` we may fix it here

src/icons/help-circle.svg
```diff
@@ -0,0 +1,4 @@
+<svg viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg">
```
CedrikNikita, line 1: I would choose name `question-circle` or `question-mark-circle`

src/popup/router/components/Modals/Default.vue
```diff
@@ -4,13 +4,27 @@
...
     </template>
-    {{ msg || $t(`modals.${type}.msg`) }}
+
+    <span v-if="msg">
+      {{ msg }}
+    </span>
+    <i18n
+      v-else
+      :path="`modals.${type}.msg`"
```
CedrikNikita, line 22: same here

CedrikNikita (commented):
`type: 'remove',`
Should be placed here aswell
https://github.com/aeternity/superhero-wallet/blob/f95e720203ac4556a4529e63929e7d0857e8679f/src/store/actions.js#L227-L230

src/popup/router/components/Modals/Default.vue
imollov: That approach wasn't introduced in this PR and is used at a few places. Any idea why `vue-i18n/no-dynamic-keys` does not detect an error here?

To avoid using dynamic keys, I must create separate components for each modal type. Please advise.

src/popup/locales/en.json
```diff
@@ -707,7 +707,7 @@
       "from": "From",
       "hash": "Transaction Hash",
       "home": "Home",
-      "scanAddress": "Scan AE Address",
+      "scanAddress": "Scan aeternity address",
```
davidyuk, line 718: Here are all about aeternity, maybe better "account address"? Or meant that tokens can be sent to smart contract or oracle?

imollov: > Here are all about aeternity, maybe better "account address"? Or meant that tokens can be sent to smart contract or oracle?

Yes, perhaps. 

To get the rationale behind you need to talk to the product & design people. Starting a discussion on figma might be the quickest way.

src/popup/router/components/Modals/TipUrlStatus.vue
```diff
@@ -1,11 +1,13 @@
 <template>
```
CedrikNikita, line 1: The whole file should be simplified using Default component. Also we will not doube the future changes made to the Default component here.

src/popup/locales/en.json
onvisions: The reason to change the text is that "Scan AE address" to "Scan aeternity address" was that "AE address" is no longer relevant with the introduction of different tokens. What @davidyuk suggested (Scan account address) sounds good to me but we may even replace it with just "Scan QR" as far as the modal opens in the context of account address input field on the "Send" screen.

paolomolo: The popup is only shown when the wallet doesn't have permission to access the camera.

So it should be the following:

image

```suggestion
      "scanAddress": "Allow camera access to scan QR code",
```

Thanks!

src/popup/router/components/Modals/Default.vue
imollov: Main reason was that some modal messages need new lines and it requires component interpolation. I separated the only use case so far in a component: https://github.com/aeternity/superhero-wallet/pull/1100/files#diff-bf1f126eebbafd355e30fb983bcc776954252488c787cb5448ee0b543a87f296R7-R13

The rest can use the default as others.

imollov: There are only prior usages of modal with `type` now and I'm going to do a refactor in this PR or another.

davidyuk (approved):

src/popup/locales/en.json
imollov: What about when access is allowed? Which title should we use?

Screenshot 2021-06-25 at 10 04 15

Currently this is a single popup and changing title conditionally or splitting into two modals will require additional changes.

src/popup/router/components/Modals/Default.vue
imollov: #1150

etharner (approved):

PR #1107, opened 2021-05-26
Unify arrow svg

CedrikNikita:


Liubov-crypto:
bug with drop-down menu (the same on develop branch commit cce633c / v.0.7.2):
bug with drop down menu

long token amount truncated (the same on develop branch commit cce633c / v.0.7.2):
long token amount trucated

leave a wallet for a while:
leave a wallet for a while
cons

Initially, the registered name is not visible in the subaccounts. The user needs to go to the names tab and then it will appear  (the same on develop branch commit cce633c / v.0.7.2)

if the user sets the default name, then it will change itself to the previous one (the same on develop branch commit cce633c / v.0.7.2).

Wallet:
Balances - pass 
Send ae - pass 
Send ae via QR - pass 
Receive ae- pass 
Send FT - pass 
Receive FT - pass 
TX history - pass 
Info about a transaction - pass 
Send Tip - pass 
Claim tip - pass
Create an invite link - pass 
Top -up - pass 
Claim back - pass 
Register a name - pass 
Set as default - pass, but then changed itself 
Set name pointer - pass 
Extend - pass 
Settings - pass

etharner (approved):

PR #1108, opened 2021-05-27
fix(comment-new): redirect to /account on successful comment send

### LINKED ISSUE #681: Deeplinking problem with Android mobile app when commenting on a post
**Description of the bug:**

When I try to make a comment on Superhero.com for the second time (at the first time everything works fine) I get a deeplinking connection to my Superhero mobile wallet (Android) to confirm the comment. For some reason when making a comment the second time, the approval window inside the Superhero mobile app shows the text for the previous comment I approved before and when I try to continue it crashes. I need to manually close down the Superhero mobile app, re-create the comment I wanted to post on a tip on Superhero.com, and then when it comes to the deeplinking, my new comment correctly appears in the approval window in the Superhero mobile app and I can confirm it successfully. This is obviously a non-optimal condition for a good user experience (most users would be very unlikely to manually close their Superhero app, write a new comment and make a second intent to confirm the comment, just to make a simple comment on a post).

**To Reproduce**
1. Write a comment on any tip post on superhero.com
2. Confirm it through Superhero mobile app (Android)
3. Make another comment with a different text (e.g. same tip post)
4. Impossible to confirm on Superhero mobile app + shows previous comment text, not the new comment I wanted to post
5. Close Superhero mobile app
6. Write another comment and go to confirmation
7. Only now the newer comment text will appear in the confirmation window on Superhero mobile app (Android)

**Smartphone:**
 - Device: Samsung Galaxy A40
 - OS: Android 10
 - Browser: Chrome
 - Version 86.0.4240.198

CedrikNikita:
fixes #681,
or we may setInterval and check if the text changed (not a very good solution)
or we may implement more complex solution with cordova plugin that will check `deviceready` and just reload from the query text,
something like https://github.com/katzer/cordova-plugin-background-mode

### DIFF
--- a/src/popup/router/pages/CommentNew.vue
+++ b/src/popup/router/pages/CommentNew.vue
@@ -57,6 +57,7 @@ export default {
           this.parentId,
         ]);
         this.openCallbackOrGoHome(true);
+        this.$router.push('/account');
       } catch (e) {
         this.$store.dispatch('modals/open', { name: 'default', type: 'transaction-failed' });
         e.payload = pick(this, ['id', 'parentId', 'text']);


etharner (approved):
Futher improvment is welcomed as separate issue/suggestion, let's merge this.

PR #1119, opened 2021-06-03
Fix address text overflow

CedrikNikita:
fix addres overflowing issue.
For example:
|Before|After|
|---|---|
|image|image|

### DIFF
--- a/src/popup/router/components/AccountInfo.vue
+++ b/src/popup/router/components/AccountInfo.vue
@@ -330,6 +330,7 @@ export default {
   .ae-address {
     display: block;
     text-decoration: none;
+    text-align: center;
     color: variables.$color-light-grey;
     letter-spacing: -0.4px;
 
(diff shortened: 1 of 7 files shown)


src/popup/router/components/AccountInfo.vue
```diff
@@ -330,10 +330,13 @@ export default {
...
     display: block;
     text-decoration: none;
+    text-align: center;
     color: variables.$color-light-grey;
     letter-spacing: -0.4px;
 
-    @extend %face-sans-10-medium;
+    @extend %face-mono-10-medium;
+
+    font-size: 9.62px;
```
etharner, line 339: I don't see any strong reason to use such fractional value, can we just use 10 or 9? @onvisions

onvisions: The reason is obviously that font-size 10 is too large for the specific UI element to fit in the real estate we have therea and to have some marrgins to the card edge. Also it's tested with longest account addresses. 
On the other hand we are trying to keep the address readable so that we are trying to use the highest font size value that is both readable and firts the card. Which happens to be 9.62px. If you have problem with that go with 9 although I think it is very difficult to read. What's the specific issue or problem that prevents us from using 9.5 or 9.64... or whatever value. I think that fractional values are not something strange especially if defined in rem or em units in the CSS.

etharner (approved):

Issue #1049, opened 2021-04-29
Utilize line-height properly


etharner:
https://github.com/aeternity/superhero-wallet/pull/1038#discussion_r621135379

subhod-i:
@CedrikNikita closing this as we are using this in the new design

PR #1125, opened 2021-06-09
AccountItem component

imollov:
* with chain name
Screenshot 2021-06-09 at 17 33 22

* without
Screenshot 2021-06-09 at 17 27 01

### DIFF
--- a/src/popup/router/components/AccountItem.vue
+++ b/src/popup/router/components/AccountItem.vue
@@ -10,12 +10,14 @@
       <Avatar
         :address="address"
         :name="name"
+        size="mid"
       />
       <span class="name">{{ name }}</span>
     </div>
     <Avatar
       v-else
       :address="address"
+      size="mid"
     />
     <span class="address">{{ address }}</span>
   </div>
@@ -42,8 +44,6 @@ export default {
   align-items: center;
 
   .avatar {
-    width: 32px;
-    height: 32px;
     margin-right: 8px;
   }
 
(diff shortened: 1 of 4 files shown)


src/popup/router/components/AccountItem.vue
```diff
@@ -0,0 +1,73 @@
...
+@use '../../../styles/variables';
+@use '../../../styles/typography';
+
+.account-item {
+  display: flex;
+  align-items: center;
+
+  .avatar {
+    width: 32px;
+    height: 32px;
```
CedrikNikita, line 46:
```suggestion
  .avatar {
    width: 32px;
    height: 32px;
```
Maybe we shall introduce new Avatar size, due to usage in several places.
here,
SidebarMenu,
New Names page designs.

src/styles/typography.scss
```diff
@@ -76,6 +76,13 @@
   line-height: 24px;
 }
 
+%face-sans-14-bold {
+  font-family: $font-sans;
+  font-weight: bold;
+  font-size: 14px;
+  line-height: 24px;
+}
+
```
CedrikNikita, line 85: Will this be used somewhere else, if not we may just define `font-weight` here.

imollov: Yes, in auctions, though with different `line-height`. 

There's still no clearance on #1049 but I guess it would be done like here: https://github.com/aeternity/superhero-wallet/pull/1124/files#diff-72b5bff734d57e7b2cf89d4e386f24a599c0a029d5a53c30422bf427c04f385dR79-R81

I also spotted `%face-sans-15-bold` and `%face-sans-20-bold` are not used anywhere and few others have only 1 occurrence. Should it be a concern at the moment, given the new design implementation is underway?

CedrikNikita (approved):

PR #1137, opened 2021-06-18
fix(qr-code): inline transparent style to show preview in cordova

etharner:
Fixes lack of camera preview in Cordova 
image


Liubov-crypto:
Tested in Chrome, FF, Brave, Safari, and FF browser for Android - LGTM.

kenodressel (approved):

kenodressel:
looks a bit weird but I assume there is no better way than to modify the body background :see_no_evil:

etharner:
> looks a bit weird but I assume there is no better way than to modify the body background 🙈

This is how this plugin works

PR #1145, opened 2021-06-23
Rebrand auction page

imollov:
Screenshot 2021-06-29 at 16 22 31

Screenshot 2021-06-29 at 16 22 53

Screenshot 2021-07-01 at 16 12 28

Screenshot 2021-06-29 at 16 23 21

Screenshot 2021-06-29 at 16 23 38

todo:
* implement #1163 in bid history ✅


Liubov-crypto (approved):
Tested, LGTM.

src/icons/auction.svg
```diff
@@ -0,0 +1,8 @@
+<svg viewBox="0 0 25 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<g opacity="0.7">
```
CedrikNikita, line 2:
```suggestion
```
I think we should also remove opacity inside svg here

src/popup/router/components/AuctionCard.vue
```diff
@@ -0,0 +1,84 @@
...
+    line-height: 16px;
+    margin-top: 16px;
+  }
+
+  .auction-overview {
+    margin-top: 16px;
+  }
+
+  .help {
+    position: absolute;
```
CedrikNikita, line 72: We may just use negative margin to allign `.help`, instead of adding z-index. This won't introduce this level of z-index complexity to an upper elements.

CedrikNikita (commented):
General remarks:
- loading on the direct link of auction will lead to the infinite loading, if user on wrong network
- there should be no TabBar on Bidding page
- the last element in the Bid History should not have blue bottom border
- transaction fee should not have equivalent in fiat

CedrikNikita (approved):
Code changes look good.

src/popup/router/pages/Names/Auction.vue
```diff
@@ -0,0 +1,92 @@
...
+@use '../../../../styles/variables';
+
+.auction {
+  .plate {
+    height: 108px;
+  }
+
+  .tabs {
+    position: sticky;
+    top: calc(146px + env(safe-area-inset-top));
```
CedrikNikita, line 82: Should be rebased on top of the develop, there would be new challenges because now you can not do this due to expanding and collapsing ability.

src/popup/router/components/Modals/AuctionsHelp.vue
```diff
@@ -0,0 +1,23 @@
+<template>
```
CedrikNikita, line 1: There would be nice to have one modal covering all the help cases to not add each time a new one.

CedrikNikita (changes requested):

src/popup/router/pages/Names/Auction.vue
imollov: Do you mean after #1169 is merged to develop?

CedrikNikita: https://github.com/aeternity/superhero-wallet/pull/1148 is merged now, and we should change the AccountInfo on the top to the AccountSwitcher

src/popup/router/components/Modals/Help.vue
```diff
@@ -0,0 +1,47 @@
+<template>
```
CedrikNikita, line 1: I don't think that this version is good solution either, we have 2 or even 3 more places with hints.
Will you propose to v-else the same code 3 additional times?

imollov: Well, the reason for this, as well as the solution before is to avoid using dynamic localization keys.

It is caused by the need to have new lines in the modal message which is achieved by component interpolation: https://kazupon.github.io/vue-i18n/guide/interpolation.html.

I'm open for better solutions, of course.

CedrikNikita: Something similar to this https://github.com/aeternity/aepp-base/blob/develop/src/components/Guide.vue,
but simplier

Liubov-crypto (commented):
Tested, LGTM.

src/popup/router/components/TemplateRenderer.vue
```diff
@@ -0,0 +1,37 @@
...
+        case 'em':
+          return createElement('em', renderNodeContent(createElement, n));
+        case 'mark':
+          return createElement('mark', renderNodeContent(createElement, n));
+        case 'strong':
+          return createElement('strong', renderNodeContent(createElement, n));
+        case 'br':
+          return createElement('br');
+        case 'p':
+          return createElement('p', renderNodeContent(createElement, n));
```
CedrikNikita, line 17: Let's add everything that is needed for now, not for every __potentinal__ use case.

CedrikNikita (approved):

PR #1152, opened 2021-06-28
Update tab bar component

CedrikNikita:


src/popup/locales/en.json
```diff
@@ -1,6 +1,7 @@
 {
   "ok": "OK",
   "ae": "AE",
+  "home": "Home",
```
etharner, line 4: Let's maybe move in into `pages.titles` as other tab bar text

CedrikNikita: Since in the Header component very strange logic around the home title i will improve it and include this into titles.

etharner (approved):

PR #1156, opened 2021-06-28
Update Loader to new designs

### LINKED ISSUE #1171: Unify z-index
Also avoid using negative z-index
https://github.com/aeternity/superhero-wallet/blob/e33aeb6eb06ef93dda24fe642c944f17b2ea8b2b/src/popup/router/components/Header.vue#L198

imollov:
fixes #1171

Screenshot 2021-06-28 at 13 38 44


src/popup/router/components/Loader.vue
```diff
@@ -3,91 +3,51 @@
...
-  }
-
-  .main-loader {
-    position: fixed;
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  z-index: 9;
```
etharner, line 30: Maybe let's get rid of z-indexes bigger than 1 (in Header for example) and set this to 2. We also have corresponding issue #1171

CedrikNikita:
I think the fact that you decided not to change z-index different than one in the `Tour` component in this PR is actually OK.

CedrikNikita (approved):

PR #1016, opened 2021-04-19
Fix names and fungible tokens usage with subaccounts

CedrikNikita:

PR #1161, opened 2021-07-01
fix(actions): add missing transaction type to transcation history

CedrikNikita:
depends on #1016


src/store/plugins/fungibleTokens.js
```diff
@@ -117,6 +121,21 @@ export default (store) => {
...
         );
       },
+      async getTokensHistory({ rootState, rootGetters: { activeNetwork, account }, commit }) {
+        const { address } = account;
+        const rawTransactions = ([
+          ...(await fetchJson(`${activeNetwork.middlewareUrl}/aex9/transfers/to/${address}`)),
+          ...(await fetchJson(`${activeNetwork.middlewareUrl}/aex9/transfers/from/${address}`)),
+        ]);
+
+        const transactions = await Promise.all(
```
kenodressel, line 131: why do you still need to get the full tx? I see that this endpoint: https://testnet.aeternity.io/mdw/aex9/transfers/to/ak_2AVeRypSdS4ZosdKWW1C4avWU4eeC2Yq7oP7guBGy8jkxdYVUy is updated as https://github.com/aeternity/ae_mdw/issues/198 is fixed

CedrikNikita: On the Details screen you would be able to see the details

etharner (approved):

PR #1163, opened 2021-07-01
feat: show chain name of other users

### LINKED ISSUE #958: Resolve chain name from account address
in the Base aepp we had a getter that was returning a name optionally making a request to MDW, would be nice to use something similar here to show names of other users (may be solved in a separate PR)

_Originally posted by @davidyuk in https://github.com/aeternity/superhero-wallet/pull/951#discussion_r607960997_

imollov:
closes #958 

Screenshot 2021-07-01 at 14 31 57

### DIFF
--- a/src/popup/router/components/TransactionOverview.vue
+++ b/src/popup/router/components/TransactionOverview.vue
@@ -19,6 +19,7 @@ export default {
   },
   computed: {
     ...mapGetters(['getTxType', 'getTxDirection', 'getExplorerPath']),
+    ...mapGetters('names', ['getPreferred']),
     ...mapState({
       account(_, { account }) {
         return {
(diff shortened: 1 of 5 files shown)


Liubov-crypto (commented):
Tested, in general LGTM. 

Got this error when clicked on pending transaction: 
bad request
bad request1

CedrikNikita (changes requested):
You are proposing to show the name that may not have a pointer to this account, instead it may have pointer to another account. ( in case user will decide to send ae to this name, he will end up sending to the different account that he wanted).
`preferredChainName`  that is returned from backend already done the job of actually checking that for you.

CedrikNikita:
I think maybe we should drop (or start to do so) adding `name` part in the name of the mutations and getters and actions within `names` plugin. For me it would be much clear to see `names/get` instead of `names/getName` etc.

Liubov-crypto (commented):
I sent tokens to a name in my account but pointed to another wallet address. This is what I see on the screenshots:
tst1
tsts2


Hash: https://explorer.aeternity.io/transactions/th_21QZw3ozUCwpFJUeaatXn8QCfSDGKJVjqM1zQDAjgpqJW9vsai

imollov:
> I sent tokens to a name in my account but pointed to another wallet address. This is what I see on the screenshots:

@Liubov-crypto The name must be set as default for the account. 

It behaves like that because of the issue Nikita described here https://github.com/aeternity/superhero-wallet/pull/1163#pullrequestreview-711336434.

CedrikNikita (approved):
Code changes look good, did not test new version properly.

kenodressel:
@Liubov-crypto maybe you can test again :)

Liubov-crypto (commented):
Tested, LGTM.

PR #1182, opened 2021-07-06
chore: rename Payments page to Transfer

### LINKED ISSUE #1174: Rename payments tab to tranfers

imollov:
closes #1174


etharner:
I understand it's a safe fast solution but we should rename routes and components as well

etharner (approved):

PR #1185, opened 2021-07-06
feat(payments-send): drop trailing zeros from amount

### LINKED ISSUE #1177: TX Confirmation: Skip useless decimal places
**Describe the bug**

The wallet shows X.000 tokens / ae on the transaction confirmation screen. It should omit useless (trailing zeros) decimal places. So if I send a single token, it should be 1 AE, if i send 0.1 AE, it should be like that. Currently its 1.000 and 0.100 respectively. 

**Screenshots**

image

imollov:
closes #1177

### DIFF
--- a/src/popup/router/pages/PaymentsSend.vue
+++ b/src/popup/router/pages/PaymentsSend.vue
@@ -89,7 +89,7 @@
               <span
                 data-cy="review-amount"
                 class="amount"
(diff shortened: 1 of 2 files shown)


Liubov-crypto (approved):
Tested, LGTM.

etharner (approved):

PR #1190, opened 2021-07-07
feat(txlist): adds realtime capability via websocket

kenodressel:
Disclaimer: 
This only works for a sample token: `Bitcoin` aka `ct_jzUHLhMK112thY2YkZb4tzgUHtYbtzqHyLZnEjxmKzsp88Hkt`

We will likely be able to scale it up but it would probably better to add an intermediary (external service) that filters the tx beforehand.

### DIFF
--- a/src/store/plugins/chainListener.js
+++ b/src/store/plugins/chainListener.js
@@ -1,3 +1,5 @@
+import { TOKEN_TEST_CONTRACT } from '../../popup/utils/constants';
+
 export default (store) => {
   let wsconnection = null;
   const subscribeToContract = (contract) => {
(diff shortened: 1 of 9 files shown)


src/store/plugins/chainListener.js
```diff
@@ -0,0 +1,57 @@
...
+  store.watch(
+    ({ middleware }) => middleware,
+    (middleware) => {
+      if (!middleware) return;
+
+      const { middlewareUrl } = store.getters.activeNetwork;
+      const websocketUrl = new URL(middlewareUrl);
+      websocketUrl.protocol = websocketUrl.protocol === 'https:' ? 'wss' : 'ws';
+      websocketUrl.pathname += '/websocket';
+      connectToWebsocket(websocketUrl);
```
kenodressel, line 52:
```suggestion
      wsconnection.onClose = () => null;
      connectToWebsocket(websocketUrl);
```

etharner (approved):

PR #1201, opened 2021-07-08
fix(tx): move view more link to tx list component

kenodressel:


src/popup/router/components/TransactionList.vue
```diff
@@ -99,7 +109,7 @@ export default {
             if (this.displayMode.latestFirst) arr.reverse();
             return arr[0] - arr[1];
           })
-          .slice(0, this.maxLength);
+          .slice(0, this.maxLength || 999999999);
```
etharner, line 112: We can use `Infinity` here instead of random big number

etharner (approved):

PR #1202, opened 2021-07-08
fix(transaction-list): fix spinner centering

etharner:
This will fix this behavior on iPhone
image


imollov:
Should we try to also center it vertically, as it is by design?

imollov:
@etharner pls check #1204

PR #1229, opened 2021-07-20
style(checkbox): update checkbox color

etharner:
Before:

image

After:

image

### DIFF
--- a/src/popup/router/components/CheckBox.vue
+++ b/src/popup/router/components/CheckBox.vue
@@ -40,14 +40,10 @@ export default {
   -ms-user-select: none;
   user-select: none;
 
-  &:hover .checkmark {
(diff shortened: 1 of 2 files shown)


onvisions:
The color change is ok. A few comments:
1. By design checkbox **unchecked** state needs to have box-shadow: inset 0 0 4px rgba(0,0,0,.25);
2. The **checked** state has no box-shadow and no visible border so the border color and the background color should be the same (#00FF9D).
3. Not sure if this is part of the scope of current commit but there is some UI bug to be fixed related to checkboxes here:

screen

onvisions (approved):
Approved after the last changes applied (https://github.com/aeternity/superhero-wallet/pull/1229#issuecomment-883228138).

imollov (approved):

PR #1230, opened 2021-07-20
Fixes for TransactionList

### LINKED ISSUE #1195: Shifted text in the Tx history
text displaying
text
dev

Reproducible on develop branch.

### LINKED ISSUE #1211: No explanatory text if the token is not found
Wire 2021-07-09 at 14_31

### LINKED ISSUE #1228: Transaction lists with none or few items reload infinitely
recording (19)

imollov:
fixes #1228
fixes #1211
fixes #1195

### DIFF
--- a/src/popup/router/components/TransactionList.vue
+++ b/src/popup/router/components/TransactionList.vue
@@ -21,7 +21,7 @@
       data-cy="loader"
     />
     <div
-      v-else-if="!transactions.length"
+      v-else-if="!filteredTransactions.length"
       class="message"
     >
       <p>{{ $t('pages.recentTransactions.noTransactionsFound') }}</p>
--- a/src/popup/router/components/TransactionList.vue
+++ b/src/popup/router/components/TransactionList.vue
@@ -193,7 +193,6 @@ export default {
 .transaction-list {
   display: flex;
   flex-direction: column;
-  padding-bottom: 48px;
 
   .list {
     background: variables.$color-black;
@@ -206,6 +205,7 @@ export default {
     flex-grow: 1;
     display: flex;
     align-items: center;
+    padding-bottom: 48px;
   }
 
   .message > p {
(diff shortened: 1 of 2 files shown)


CedrikNikita:
loading

src/popup/router/components/TransactionList.vue
```diff
@@ -21,7 +21,7 @@
       data-cy="loader"
     />
     <div
-      v-else-if="!transactions.length"
+      v-else-if="!filteredTransactions.length"
```
etharner, line 24: How can I reproduce it? I have this message on newly created account on develop version:
image

etharner: Ok, got it, it's after filtering

imollov:
> loading

This is reproducible on develop. 

I propose to not show the loader when fetching next page for now as it's not in the designs either.

CedrikNikita:
> 
> I propose to not show the loader when fetching next page for now as it's not in the designs either.

At this case user won't know that he scrolled to the bottom. I propose to drop last commit and merge the rest, since the issue i've mentioned was not supposed to be fixed within this PR.

src/popup/router/components/TransactionList.vue
```diff
@@ -157,10 +158,7 @@ export default {
         this.updateTransactions(transactions);
       } finally {
         this.loading = false;
-      }
-      if (transactions.length) {
         this.page += 1;
-        this.checkLoadMore();
```
CedrikNikita, line 163: Without this check we will get in the situation like this, when the chunk contain not enought info to fill the whole height, and the user is not able to scroll to the bottom cause of that
Ivelin
or even like this:

this PR
image

develop
image

CedrikNikita (changes requested):

CedrikNikita:
I still have an infinite loading
Ivelin new

imollov:
> I still have an infinite loading
> Ivelin new

I couldn't fix it with the current implementation of load more behavior.

https://github.com/aeternity/superhero-wallet/issues/1234

> At this case user won't know that he scrolled to the bottom. I propose to drop last commit and merge the rest, since the issue i've mentioned was not supposed to be fixed within this PR.

Agreed.

CedrikNikita (approved):

PR #1231, opened 2021-07-21
Custom network default values

### LINKED ISSUE #1151: Provide default value when adding own node
We should leave a default value for the compiler url field pointing to the current hosted compiler, so people can more easily try out their own node without having to mess with setting up a compiler, too. E.g. :

image

AtanasKrondev:
fixes #1151


etharner (approved):

davidyuk (commented):
lgtm

PR #1232, opened 2021-07-22
Use explorerUrl from constants

CedrikNikita:


src/popup/router/components/AccountInfo.vue
```diff
@@ -135,7 +136,7 @@ export default {
     },
     explorerUrl() {
       const { address } = this.accounts[this.idx];
-      return `https://explorer.aeternity.io/account/transactions/${address}`;
+      return `${defaultNetwork.explorerUrl}/account/transactions/${address}`;
```
etharner, line 139: Will it point to wrong explorer if user will change network?

etharner (approved):

PR #1238, opened 2021-07-22
Align node status bar in collapsed iframe

### LINKED ISSUE #1120: Misaligned node status bar in iframe when collapsed
Screenshot 2021-06-03 at 10 59 58

AtanasKrondev:
fixes #1120 

Screenshot 2021-07-22 at 19 07 10

+ TabBar hidden in collapsed view as in my opinion collapsed view should be just for info reference, not complicated navigating

### DIFF
--- a/src/popup/router/components/NodeConnectionStatus.vue
+++ b/src/popup/router/components/NodeConnectionStatus.vue
@@ -62,4 +62,12 @@ export default {
 .connect-node {
   background: variables.$color-bg-3;
 }
+
+@include mixins.collapsed {
+  .node-connection-status,
+  .connect-error,
+  .connect-node {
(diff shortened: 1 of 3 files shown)


src/popup/router/components/NodeConnectionStatus.vue
```diff
@@ -63,4 +63,12 @@ export default {
 .connect-node {
   background: variables.$color-bg-3;
 }
+
+@media screen and (max-height: 250px) {
```
etharner, line 67: Maybe let's add this to `mixins.scss` in case we will need more design changes for collapsed wallet

CedrikNikita (approved):

PR #1246, opened 2021-08-02
docs: correct grammar and spelling in README

jamesko:
Corrects/clarifies the text and makes tense consistent in sub-section headers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Superhero Wallet
 
-**Superhero is a browser extension that allows you to send and receive value to URLs and content providers across the internet.**
+**Superhero is a browser extension that allows you to send and receive value to URLs and content producers across the internet.**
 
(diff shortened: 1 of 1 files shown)


mradkov:
Can you rephrase the commit message so it follows the format we use, please?

Check conventional commits for reference.

jamesko:
Hope that is more consistent.

In the description, should it include that the extension can also send value to other people with a wallet? Or is the emphasis going to remain on URLs and content producers?

etharner:
@jamesko Please squash commits and name the resulting commit the same way as you made with PR name (in your case it's `docs: correct grammar and spelling in README`). Otherwise it won't pass checks (as @mradkov mentioned before)

CedrikNikita:
The changes in this PR was already merged with #1249, to avoid additional work from the author.

PR #1265, opened 2021-08-30
fix(status): set bottom to 0 if tab bar is hidden

etharner:
Fixes this issue on screens which have tab bar hidden (details, bidding):
image


imollov (approved):

Liubov-crypto:
Tested, LGTM.

PR #1121, opened 2021-06-04
Rebrand send screen

etharner:
fixes #1101

PR #1280, opened 2021-09-09
Add validation support for inputs

etharner:
This PR is implementing base for refactoring of #1121 with including of vee-validate and example of using in send screen.

### DIFF
--- a/src/popup/locales/en.json
+++ b/src/popup/locales/en.json
@@ -1018,10 +1018,8 @@
       "backToTokenMigration": "Back to token migration info",
       "backendServiceUrl": "Backend Service Url"
     },
-    "aeAddressInput": {
-      "label": "Aeternity Address",
-      "identicon": "Your identicon",
-      "error": "Invalid AE Address"
+    "addressInput": {
+      "placeholder": "Enter recepient’s address or .chain name"
     },
     "welcome": {
       "send": "Send",
--- a/src/popup/locales/en.json
+++ b/src/popup/locales/en.json
@@ -1095,7 +1095,11 @@
     }
   },
   "validation": {
-    "required": "This field is required"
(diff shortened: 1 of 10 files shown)


Liubov-crypto:
Tested in Chrome, FF, Safari, Brave browsers, LGTM.

src/popup/locales/en.json
```diff
@@ -1094,6 +1092,13 @@
       "empty": "You aren't connected to any aepps"
     }
   },
+  "validation": {
+    "required": "This field is required",
+    "address": "Enter valid blockchain address or .chain name",
+    "name_registered_address": "This name isn't registered or pointed to address",
```
CedrikNikita, line 1098: Let's not introduce new case in the wordings.

src/popup/utils/helper.js
```diff
@@ -196,3 +196,5 @@ export const readValueFromClipboard = async () => {
   }
   return value;
 };
+
+export const isAensName = (value) => /^[a-z]+\.chain$/i.test(value);
```
CedrikNikita, line 200: Since this new name check is better, let's use it istead of old one.

src/popup/router/pages/TransferSend.vue
```diff
@@ -55,8 +49,8 @@
             <Button
               data-cy="review-withdraw"
               half
-              :disabled="!validAddress || !+form.amount || form.amount <= 0"
-              @click="step = 2"
+              :disabled="!+form.amount || form.amount <= 0"
```
CedrikNikita, line 52: Disabled logic now is not the same, it won't be disabled if there are errors in the address input.

CedrikNikita (approved):
Code changes look reasonable.

PR #1281, opened 2021-09-10
feat(send): rebrand address and permission input

etharner:
image
image


Liubov-crypto:
Tested in Chrome, FF, Safari, Brave browsers, LGTM.

Liubov-crypto:
Tested in Chrome, FF, Safari, Brave browsers, LGTM.

src/popup/router/components/AddressInput.vue
```diff
@@ -0,0 +1,117 @@
...
+
+export default {
+  components: {
+    InputField,
+    Valid,
+    QrScan,
+    Paste,
+  },
+  props: {
+    tokenSymbol: { type: String, default: 'AE' },
```
etharner, line 51: Remove this

src/popup/router/components/AddressInput.vue
```diff
@@ -0,0 +1,117 @@
...
+    </InputField>
+    <button
+      v-if="!address"
+      class="right-button"
+      data-cy="scan-button"
+      @click="scan"
+    >
+      <QrScan />
+    </button>
+    <button
```
CedrikNikita, line 27: It's not user friendly to not have paste button visible until you type somehitng. @onvisions can you elaborate on this?

Liubov-crypto:
Tested in Chrome, Safari, FF,  Brave browsers and FF ext and Chrome ext, LGTM.

Liubov-crypto:
Tested in Chrome, Safari, FF, Brave browsers LGTM.

CedrikNikita (commented):
Behavior of permission input is not obvious.
- if you will enter a letter, it will be removed
- If you will enter a letter, it would mark field as an error, but will not show you a letter and will not turn back into not error state after unfocus. You will be able to turn it back only by typing number or typing letter and then deleting it fast enough.
- if you want to write down the amount in exponentially notation, you need to make it fast, due to above problems. 

Since this PR is not about changing behaviour, i will create and additional issue for that. @etharner

CedrikNikita (approved):
Code changes look reasonable

PR #1301, opened 2021-09-27
Update Receive screen

etharner:


Liubov-crypto:
Tested web wallet and extensions as well, LGTM.

kenodressel:
When I click copy address the size of the address field gets smaller

config.xml
```diff
@@ -8,6 +8,7 @@
     <content src="index.html" />
     <plugin name="cordova-plugin-whitelist" spec="1" />
     <plugin name="cordova-clipboard" spec="^1.3.0" />
+    <plugin name="cordova-plugin-x-socialsharing" spec="^6.0.3" />
```
CedrikNikita, line 11: Version `^6.0.0` includes support for Android X, build on Android is failing because of that.
`This project uses AndroidX dependencies, but the 'android.useAndroidX' property is not enabled. Set this property to true in the gradle.properties file and retry`

CedrikNikita (commented):
In the section cordova plugins in the package.json, new plugin should be named `cordova-plugin-x-socialsharing` instead of `cordova-cordova-plugin-x-socialsharing`

src/popup/router/components/Modals/ShareQr.vue
```diff
@@ -0,0 +1,92 @@
...
+        @click="resolve"
+      >
+        {{ $t('modals.shareQr.close') }}
+      </Button>
+      <Button
+        v-if="!IS_MOBILE_DEVICE"
+        v-clipboard:copy="qrStr"
+        icon-text
+      >
+        <Copy />
```
CedrikNikita, line 31: When you click on copy the qr code should be also highlighted
image

CedrikNikita (changes requested):

CedrikNikita (approved):
Let's merge this PR, i've created an additional issue https://github.com/aeternity/superhero-wallet/issues/1324

PR #1307, opened 2021-09-30
fix(helper): fix aens check

etharner:
It now allows any language\digits


kenodressel (approved):

Liubov-crypto:
Tested, LGTM, except for a small bug: 
If the user accidentally adds a space before .chain, then the wallet will show that everything is correct, but the Review button will not let go further:
Безіменний 2
ghj

kenodressel (approved):

Liubov-crypto:
Retested, LGTM.

PR #1311, opened 2021-10-04
Fix redirection page

CedrikNikita:
Better would be to update sdk to a `9.0.0`, this issues is fixed there.


CedrikNikita:
implemented better on #1314

PR #1322, opened 2021-10-07
ci(docker): initial version of docker build and push tags

mmpetarpeshev:
We have to decide - Do we want PR builds against  main branch or only against develop.To me looks like unnecessary duplication in most of the cases , but maybe could be useful if there is something like hot fixes(if they goes through PR's).


.github/workflows/docker.yml
```diff
@@ -0,0 +1,47 @@
+name: Publish Docker Superhero Wallet image
+
+on:
+  push:
+    branches: [ master, develop ]
```
kenodressel, line 5: here i guess we could only check for master

mmpetarpeshev: Done

.github/workflows/docker.yml
```diff
@@ -0,0 +1,47 @@
...
+    tags:
+      - 'v*'
+  pull_request:
+    branches: [ master, develop ]
+
+jobs:
+  main:
+    name: Wallet Build/Test/Docker
```
kenodressel, line 13: I guess its only docker right?

.github/workflows/main.yml
```diff
@@ -1,77 +0,0 @@
-name: Test, build, and deploy
```
kenodressel, line 1: I dont think we can remove this quite yet, there is a lot more in here

mmpetarpeshev: We are not going to merge , before everything is done - like staging deployments.

.github/workflows/docker.yml
mmpetarpeshev: Depends on how we look on it , inside the docker we have buid, test and we have docker build and push

kenodressel:
Why did we delete the issue templates?

PR #1325, opened 2021-10-12
Rewrite several e2e tests to unit

CedrikNikita:


jest.config.js
```diff
@@ -1,3 +1,28 @@
+const packagesToTranspile = [
+  'lodash-es',
+  '@aeternity/aepp-sdk',
+  '@aeternity/hd-wallet',
+];
+
 module.exports = {
-  preset: '@vue/cli-plugin-unit-jest',
```
etharner, line 2: Maybe we can stick with some preset which is more suitable for us instead of manually writing all rules

PR #1327, opened 2021-10-14
feat(claim): show name price before submit

CedrikNikita:


src/popup/locales/en.json
```diff
@@ -565,6 +565,7 @@
         "register-name": "Register .chain name",
         "name-placeholder": "Enter .chain name",
         "button": "Register name",
+        "button-price": "Register name for {price} AE",
```
etharner, line 568: This isn't an unobvious to get the meaning of parameter here, for other places we've used default `{0}`, maybe let's stick with that as well

src/popup/router/pages/Names/Claim.vue
```diff
@@ -36,18 +36,23 @@
       :disabled="!sdk || !name || errors.any()"
       @click="claim"
     >
-      {{ $t('pages.names.claim.button') }}
+      {{
+        validName ?
+          $t('pages.names.claim.button-price', { price: nameFee })
```
etharner, line 41: Meaning first comment, in this case you can just pass `[nameFee]`

PR #1328, opened 2021-10-15
Rebrand transfer and tips pages

CedrikNikita:
fixed issue missed in #1121

### DIFF
--- a/src/popup/router/routes/index.js
+++ b/src/popup/router/routes/index.js
@@ -277,6 +277,7 @@ export default [
       props: true,
       meta: {
         title: 'transfer',
+        backButton: false,
       },
     }, {
       path: 'receive',


CedrikNikita:
> QR scan does not work in the web wallet.
> 
> If I enter the amount of tokens in the receiver screen, then the QR scan will also not work (scans, but there is no automatic filling of the address and the amount, which are encrypted in QR Code), both in the web wallet and in the extension, perhaps because such a link is being formed: https://wallet.superhero.com/transfer?account=ak_cEjh5on5RvoBw8zqCGS4VWeY2bmNYNPjn3SCZutYyVLWKf7X1&token=AE&amount=0.1

To test that you have to just insert `https://develop.wallet.z52da5wt.xyz/` instead `: https://wallet.superhero.com/`

PR #1332, opened 2021-10-18
chore(release): 0.9.3

kenodressel:

PR #1335, opened 2021-10-19
Fix popups

CedrikNikita:
fixes https://github.com/aeternity/superhero-wallet/pull/1332#issuecomment-946113215


Liubov-crypto:
FF browser LGTM.

PR #1366, opened 2021-10-29
Fix truncate scrolling on string change

### LINKED ISSUE #1247: Several Truncate components on the same screen not scrolling properly
if the name is fetching for the first time in the details, then it won't scroll.
Truncate

CedrikNikita:
fixes #1247,
also fixing scroll issue when preffered/default name changed
scroll

### DIFF
--- a/src/popup/router/components/Truncate.vue
+++ b/src/popup/router/components/Truncate.vue
@@ -26,26 +26,31 @@ export default {
       };
     },
   },
-  mounted() {
-    if (this.fixed) return;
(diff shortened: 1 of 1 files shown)


CedrikNikita:
> yeah pausing would be nice.

I would rather make a separate issue for this.

etharner:
> yeah pausing would be nice.

I think we should do this in separate PR as this PR fixes concrete problem

etharner (approved):

PR #1367, opened 2021-10-29
Drop workaround for the duplicate transactions from mdw

### LINKED ISSUE #1053: Constantly fails because of duplicate transactions in MDW output (edge case)
On the `/account` screen:

Screenshot 2021-04-30 at 09 31 04

It happened when sending coins to itself. Should be fixed on the MDW side, also we can make a workaround.

https://github.com/aeternity/ae_mdw/issues/162

CedrikNikita:
was fixed in https://github.com/aeternity/ae_mdw/pull/292
closes #1053

### DIFF
--- a/src/store/actions.js
+++ b/src/store/actions.js
@@ -1,5 +1,5 @@
 import {
-  flatten, orderBy, uniq, uniqBy,
+  flatten, orderBy, uniq,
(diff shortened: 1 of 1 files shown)


Liubov-crypto:
Tested, LGTM.

PR #1369, opened 2021-10-29
Utilize line-height in rebranded pages

CedrikNikita:
part of #1049 (or even fixes the whole issue what do you think @etharner?)

### DIFF
--- a/src/popup/router/components/AccountInfo.vue
+++ b/src/popup/router/components/AccountInfo.vue
@@ -255,7 +255,6 @@ export default {
     justify-content: center;
     margin-top: 8px;
     margin-bottom: 6px;
-    line-height: 21px;
 
     @extend %face-sans-14-medium;
 
--- a/src/popup/router/components/AuctionCard.vue
+++ b/src/popup/router/components/AuctionCard.vue
@@ -33,6 +33,7 @@ export default {
 
 <style lang="scss" scoped>
 @use '../../../styles/variables';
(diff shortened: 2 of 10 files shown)


src/popup/router/components/AccountInfo.vue
```diff
@@ -255,11 +255,8 @@ export default {
     justify-content: center;
     margin-top: 8px;
     margin-bottom: 6px;
-    line-height: 21px;
 
-    @extend %face-sans-14-medium;
```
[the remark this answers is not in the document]
etharner: Yes, that's what Paolo requested until designs will cover long addresses issues

src/popup/router/components/AuctionCard.vue
```diff
@@ -49,8 +50,8 @@ export default {
   }
 
   .name {
-    font-weight: bold;
```
etharner, line 52: Looks like in designs we still need bold here 
image

etharner (commented):
In general I support this changes, but there are still many places where we have something like that:
image
And this maybe should be resolved on design side first: it needs to be standardized and consistent to prevent further misusing of fonts.

src/popup/router/components/AuctionCard.vue
CedrikNikita: Was going to use %face-sans-14-bold here.

etharner (approved):

PR #1373, opened 2021-11-03
refactor(claim): disable register button if not enough ae

CedrikNikita:


src/popup/router/pages/Names/Claim.vue
```diff
@@ -5,7 +5,12 @@
     </span>
     <InputField
       v-model="name"
-      v-validate="'required|name|name_unregistered'"
+      v-validate="{
+        required: true,
+        name: true,
+        name_unregistered: true,
+        enough_ae: (+getFee + +nameFee).toString(),
```
CedrikNikita, line 12: Maybe we need to multiply `getFee` by two, cause we actually send two transactions, but it still wouldn't be accurate enough.

Liubov-crypto:
Tested, LGTM.

src/popup/router/pages/Names/Claim.vue
etharner: Maybe this two fees is not the same in this case we can't multiply or we should agree to add some extra space to cover this case.

CedrikNikita:
depends on https://github.com/aeternity/aepp-sdk-js/issues/1328

CedrikNikita:
fixed in https://github.com/aeternity/superhero-wallet/pull/2186

PR #1375, opened 2021-11-03
Build for higher target sdk version

etharner:
This changes have been made because from 1st November, 2021 Google rejects any releases built with target sdk version lower than 30.

image

And because some of our Cordova plugins isn't updated for that additional plugin for converting code to android >10 was needed.

v0.9.3 android release already uploaded with this changes.


kenodressel (approved):

Liubov-crypto:
Tested App for android, LGTM.

CedrikNikita (approved):
Code changes look reasonable.

PR #1396, opened 2021-12-27
refactor(invite claim): remove workarounds for sdk

davidyuk:
can't test it locally due to #1398 
tested on deployed branch, looks working

### DIFF
--- a/src/store/modules/invites.js
+++ b/src/store/modules/invites.js
@@ -1,4 +1,4 @@
-import { Universal, MemoryAccount, Crypto } from '@aeternity/aepp-sdk';
+import { Crypto } from '@aeternity/aepp-sdk';
 import { i18n } from '../plugins/languages';
 
 export default {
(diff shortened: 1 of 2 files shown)


src/store/modules/invites.js
```diff
@@ -13,16 +13,13 @@ export default {
...
-      // TODO: Remove this after fixing https://github.com/aeternity/aepp-sdk-js/issues/1261
-      const { name, instance } = sdk.pool.get(current.network);
-      // TODO: Remove this after merging https://github.com/aeternity/aepp-sdk-js/pull/1060
-      const s = await Universal({
-        nodes: [{ name, instance }],
-        accounts: [MemoryAccount({ keypair: { publicKey, secretKey } })],
-      });
-      await s.transferFunds(1, account.address, { payload: 'referral', verify: false });
+      await sdk.transferFunds(
+        1,
```
marsic3, line 19: It would be good to add constant for this value, people might get confused what does number 1 means

src/store/modules/invites.js
```diff
@@ -13,16 +13,13 @@ export default {
...
+        account.address,
+        { payload: 'referral', onAccount: { publicKey, secretKey }, modal: false },
```
marsic3, line 21: also for the referral would be good to be enum value

src/store/modules/invites.js
davidyuk: Ideally, they would check sdk documentation. This code shouldn't document sdk methods.

src/store/modules/invites.js
davidyuk: We don't have any other hardcoded payloads. And we are not checking for this payload somewhere in the wallet. I assume this payload was set to track usage of this feature or so from outside. I think it is fine to keep it as it is, also the intention of PR was to remove the specific workaround.

CedrikNikita:
When you click claim back your own invite link there is an error.
```popup.0987a077.js:1 TypeError: t.onAccount.sign is not a function
    at Object.signTransaction
```

PR #1414, opened 2022-03-22
Release/0.9.7

kenodressel:


kenodressel:
@Liubov-crypto please check this :)

Liubov-crypto:
> release-0-9-7.wallet.z52da5wt.xyz

Can't open the URL for testing. Got this issue: 
ffff

Liubov-crypto:
If I enter the wallet, then the name is not set by default, I need to go to the names tab and set it manually. After a while, the name disappears and I have to set it manually again:
chain nam

No automatic connection (pop-ups) between web wallet and superhero.com. 

When click on Notifications: 

notifications

When switching between networks:
switch

Fungible tokens have disappeared from the wallet, so I can't do transactions with them. 

Web wallet:
Create a new wallet - pass
Change currency - pass
Send ae - pass
Receive ae - pass
Balances view- pass
Send via scan QR - pass
AE is displayed in Tx History (All/Sent/Received) - pass
Send Tip - pass
Claim tip - pass
Tx history - pass
Tx details - pass
Invite link: create - pass
top up -pass
claim back - pass
Register a name - pass
Set name pointer - pass
Set as default name - pass
Extend - pass
Settings - pass
Networks - pass

kenodressel:
@Liubov-crypto could you check which of these issues already exist in 0.9.6?

Liubov-crypto:
> @Liubov-crypto could you check which of these issues already exist in 0.9.6?

1. If I enter the wallet, then the name is not set by default, I need to go to the names tab and set it manually. After a while, the name disappears and I have to set it manually again.

2. Fungible tokens have disappeared from the wallet, so I can't do transactions with them.

3. There is no possibility make the name by default after registration - this issue is also reproducible from time to time (when you register 1 name and after a while 2 name, second name will be without make default option)

PR #1421, opened 2022-04-13
Feature/don't depend on aepp-raendom

pstachel:


CedrikNikita (changes requested):
Don't merge develop into your branch, rebase it over develop. Only reasonable commits should be in the PR, do not add something, then remove it in another commit.

CedrikNikita:
Commit type should not be `feat` for this changes
https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

CedrikNikita (approved):

PR #1422, opened 2022-04-13
fix: welcome screen message for mobile

ifaouibadi:


src/popup/router/pages/Index.vue
```diff
@@ -74,6 +77,7 @@ export default {
   data: () => ({
     termsAgreed: false,
     IS_WEB: process.env.PLATFORM === 'web',
+    IS_MOBILE: process.env.PLATFORM === 'cordova',
```
davidyuk, line 80:
```suggestion
    IS_MOBILE: window.IS_MOBILE_DEVICE,
```
Do cordova have a special user agent not containing `Mobi`? Or if it is only for cordova then I would rename the variable to `IS_CORDOVA`.

src/popup/locales/en.json
```diff
@@ -581,6 +581,7 @@
         "mobile-app": "Mobile app"
       },
       "webVersion": "Or continue with our web version",
+      "mobileVersion": "More than just a way to receive, store & send tokens on the aeternity blockchain",
```
davidyuk, line 584:
```suggestion
      "mobileVersion": "More than just a way to receive, store & send coins on the aeternity blockchain",
```
They rather meant AE then actual tokens 🤔

PR #1423, opened 2022-04-13
fix: open cancel callback urls in the same tab

pstachel:


CedrikNikita (commented):
why did you make it opened in the same tab only on not success?

CedrikNikita (changes requested):
This approach doesn't work in cordova version.

CedrikNikita (commented):
This PR should be also rebased on the develop.

CedrikNikita (commented):
image

If the url would be without protocol.

davidyuk (approved):
code changes looks fine, but squash commits in the branch or use squash merging

kenodressel:
Apparently not working on android version @kenodressel will test :)

PR #1425, opened 2022-04-14
fix: load currencies after a time interval

pstachel:


CedrikNikita:
@Liubov-crypto can you check it?

.gitignore
```diff
@@ -10,6 +10,7 @@ tests/e2e/screenshots
 .DS_Store
 /tests/e2e/videos/
 /tests/e2e/screenshots/
+.idea
```
davidyuk, line 13: consider set up a global .gitignore to don't be distracted by this in future https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

src/store/plugins/observables.js
```diff
@@ -107,6 +107,18 @@ export default (store) => {
...
       refCountDelay(1000),
     ),
+    currencies: watchAsObservable(
+      ({ currencies }) => currencies,
+      { immediate: true },
+    ).pipe(
+      pluck('newValue'),
+      switchMap(() => {
+        const delay = 3600000;
+        return timer(0, delay).pipe(switchMap(() => store.dispatch('getCurrencies', delay)));
```
davidyuk, line 117: I'm proposing to define a global interval (in a sore plugin or App.vue) instead, and to use `currencies` as before. https://github.com/aeternity/superhero-wallet/issues/1432

src/store/actions.js
```diff
@@ -80,19 +80,22 @@ export default {
     txs = orderBy(flatten(txs), ['microTime'], ['desc']);
     commit('addTransactions', recent ? txs.slice(0, limit) : txs);
   },
-
-  async getCurrencies({ state: { nextCurrenciesFetch }, commit }) {
-    if (!nextCurrenciesFetch || nextCurrenciesFetch <= new Date().getTime()) {
+  async getCurrencies({ state: { nextCurrenciesFetch, currencies }, commit }, delay) {
+    if (!nextCurrenciesFetch
+    || Number(((nextCurrenciesFetch - new Date().getTime()) / 1000).toFixed()) < 1
```
davidyuk, line 85: `new Date().getTime()` can be replaced with `Date.now()`
Did you mean to use `Math.floor/round/ceil` instead of `Number(x.toFixed())`?
But I don't understand change in general, the previous version looks fine.

pstachel: I've changed this condition because when I was testing it on smaller intervals it gave me some small differences in time but I think `Math.floor` will fix that

Liubov-crypto:
I got this issue: 
iss

The rest LGTM.

src/store/actions.js
davidyuk: `nextCurrenciesFetch` is a time at some moment + 1 hour, before we had a check `nextCurrenciesFetch < Date.now()` that is clear, you proposing to change it to `nextCurrenciesFetch < Date.now() + 1 second`,  why do we need this second?

Or you mean to compare `nextCurrenciesFetch` and `Date.now()` accurate to the second? Then you missed `Math.abs`. But still I don't see the reason 🤔

pstachel: I want the difference to be less than 1 second because when I run `setInterval` I sometimes get few millisecond difference between `Date.now()` and `nextCurrenciesFetch` and because of that the currencies will not fetch. `Math.abs()` is not needed here because I want to accept all differences that are smaller than 1s.

CedrikNikita (changes requested):
You should not introduce changes and then remove them in the next commit. There should be no random changes such as `z-index` manupulations.

src/popup/App.vue
```diff
@@ -66,8 +67,9 @@ export default {
 
     await this.$watchUntilTruly(() => this.isRestored);
 
-    this.$store.dispatch('getCurrencies');
     this.$store.dispatch('fungibleTokens/getAeternityData');
+    this.getCurrenciesInterval();
```
davidyuk, line 71:
```suggestion
    this.$store.dispatch('pollCurrencies');
```

src/store/actions.js
```diff
@@ -81,13 +81,14 @@ export default {
     txs = orderBy(flatten(txs), ['microTime'], ['desc']);
     commit('addTransactions', recent ? txs.slice(0, limit) : txs);
   },
-
-  async getCurrencies({ state: { nextCurrenciesFetch }, commit }) {
-    if (!nextCurrenciesFetch || nextCurrenciesFetch <= new Date().getTime()) {
+  async getCurrencies({ state: { nextCurrenciesFetch }, commit }, delay) {
```
davidyuk, line 84:
```suggestion
  async pollCurrencies({ commit }) {
    executeAndSetInterval(() => {
      try {
        const { aeternity } = await fetchJson(CURRENCIES_URL);
        commit('setCurrencies', aeternity);
      } catch (e) {
        handleUnknownError(e);
      }
    }, 3600000);
  }
```

davidyuk (commented):
If you want to know my opinion on how to write this code better, so here it is:
- as we are implementing continuous polling then we don't need `nextCurrenciesFetch` at all
- `getCurrencies` is not called anywhere else, so we can combine the cone in App.vue with it
- leaving `App.vue` is not a case, so I think we can don't clear that interval

CedrikNikita (approved):

Issue #1400, opened 2022-02-04
Show more than 10 auctions, sort among all auctions


davidyuk:
https://github.com/aeternity/superhero-wallet/blob/68bed5ca3367c0ec58539dcbe0d09abd5e9a9483/src/store/plugins/names.js#L98-L107

This returns only 10 auctions.

https://github.com/aeternity/superhero-wallet/blob/a362af60afee0a981b3c02c65c9870e94ed9fe30/src/popup/router/pages/Names/AuctionList.vue#L68-L79

This would sort only fetched auctions, need to do this among all auctions (on the middleware side).

https://github.com/aeternity/superhero-wallet/blob/a362af60afee0a981b3c02c65c9870e94ed9fe30/src/popup/router/pages/Names/AuctionList.vue#L78

Use [`reverse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse) instead?

subhod-i:
@pstachel

Liubov-crypto:
This issue has been fixed. On the current development branch https://github.com/aeternity/superhero-wallet/commit/79750b2a873a8d24259b16487d7a54654f72987d, I see more than 10 auctions. Can be closed.

PR #1429, opened 2022-04-25
WIP: Fix/sort among all auctions

pstachel:


src/store/plugins/names.js
```diff
@@ -95,15 +97,23 @@ export default (store) => {
...
         if (!middleware) return [];
-        return (
-          await middleware.getAllAuctions({ by: 'expiration', direction: 'forward' })
-        ).data.map(({ name, info }) => ({
-          name,
-          expiration: info.auctionEnd,
-          lastBid: info.lastBid.tx,
-        }));
+        const response = await (next
+          ? fetchJson(`${activeNetwork.middlewareUrl}${next}`)
```
davidyuk, line 106: I'm proposing to add a fetch-by-path method to middleware: https://github.com/aeternity/superhero-wallet/blob/f1d03c7d1104cc80b38eac9dc379f4c6f4a2c170/src/lib/wallet.js#L73
to extract `fetchJson`, `middlewareUrl`, `camelcaseKeysDeep `

src/store/plugins/names.js
```diff
@@ -95,15 +97,23 @@ export default (store) => {
 
         commit('set', names);
       },
-      async fetchAuctions({ rootState: { middleware } }) {
+      async fetchAuctions({
+        rootState: { middleware },
+        rootGetters: { activeNetwork },
+      }, { next, filterBy = 'expiration', filterDirection = 'forward' }) {
```
davidyuk, line 103: shouldn't these defaults be defined in component?

src/store/plugins/fungibleTokens.js
```diff
@@ -65,6 +65,9 @@ export default (store) => {
           `${activeNetwork.middlewareUrl}/aex9/by_name`,
         ).catch(handleUnknownError);
 
+        // TODO - remove
+        if (!response?.length || typeof response === 'string') return null;
```
davidyuk, line 69: I would remove `.catch(handleUnknownError)` from above and use `try/catch` instead

src/popup/router/pages/Names/AuctionList.vue
```diff
@@ -75,15 +90,61 @@ export default {
...
     this.loading = false;
   },
+  methods: {
+    ...mapActions('names', ['fetchAuctions']),
+    setScrollWatcher() {
+      const app = document.querySelector('#app');
+      const throttledLoadMore = () => throttle(this.checkLoadMore, 700);
+      app.addEventListener('scroll', throttledLoadMore());
+      this.$on('hook:destroyed', () => {
+        app.removeEventListener('scroll', throttledLoadMore());
```
davidyuk, line 109: as I know, you need to pass here exact the same function as in `addEventListener`, but `throttledLoadMore()` returns a new one each time

src/popup/router/pages/Names/AuctionList.vue
```diff
@@ -75,15 +90,61 @@ export default {
...
-    this.activeAuctions = await this.$store.dispatch('names/fetchAuctions');
+    await this.fetchAuctionsChunk();
+    this.setScrollWatcher();
     this.loading = false;
   },
+  methods: {
+    ...mapActions('names', ['fetchAuctions']),
+    setScrollWatcher() {
+      const app = document.querySelector('#app');
+      const throttledLoadMore = () => throttle(this.checkLoadMore, 700);
```
davidyuk, line 106: `throttle` shouldn't be needed, because `checkLoadMore` should do an inexpensive check in sync and to trigger async action if needed, if the action is already triggered then is should do nothing
similar to https://github.com/aeternity/superhero-ui/blob/3ad2490571b7aa2c65a4e2fc56eab3464b5f40eb/src/components/FeedPagination.vue#L71-L87

src/popup/router/pages/Names/AuctionList.vue
```diff
@@ -75,15 +90,61 @@ export default {
...
+      app.addEventListener('scroll', throttledLoadMore());
+      this.$on('hook:destroyed', () => {
+        app.removeEventListener('scroll', throttledLoadMore());
+        this.isDestroyed = true;
+      });
+    },
+    async checkLoadMore() {
+      if (this.isDestroyed || !this.nextPageUrl) return;
```
davidyuk, line 114:
```suggestion
      if (this.isDestroyed || !this.nextPageUrl || this.loading) return;
```
according to the above suggestion

src/popup/router/pages/Names/AuctionList.vue
```diff
@@ -75,15 +90,61 @@ export default {
...
+    },
+    async fetchAuctionsChunk() {
+      const filterBy = filterMap[this.displayMode.sort];
+      const filterDirection = this.displayMode.rotated ? 'backward' : 'forward';
+      const { data, next } = await this.fetchAuctions({
+        next: this.nextPageUrl,
+        filterBy,
+        filterDirection,
+      });
+      this.activeAuctions = [...this.activeAuctions, ...data];
```
davidyuk, line 132:
```suggestion
      this.activeAuctions.push(...data);
```
🤷‍♀️

src/popup/router/pages/Names/AuctionList.vue
```diff
@@ -1,25 +1,29 @@
 <template>
   <div class="auction-list">
     <Filters
-      v-if="activeAuctions.length || auctions.length || loading"
-      v-model="displayMode"
+      v-if="hasAuctions || loading"
+      :value="displayMode"
       :filters="filters"
+      @input="handleFilter"
```
davidyuk, line 7: alternatively, you can keep `v-model` and to use `watch: { displayMode: { handler, immediate: true } }`

src/popup/router/pages/Names/AuctionList.vue
```diff
@@ -57,15 +67,20 @@ export default {
...
     return pick(this.$store.state.observables, ['topBlockHeight']);
   },
   computed: {
-    ...mapGetters(['getNameFee']),
+    ...mapGetters(['getNameFee', 'activeNetwork']),
+    hasAuctions() {
+      return this.activeAuctions.length || this.auctions.length;
+    },
     auctions() {
+      // TODO - change to middleware sorting
```
davidyuk, line 83: so, this PR is not fixing #1400 completely

src/store/plugins/names.js
```diff
@@ -15,6 +16,7 @@ export default (store) => {
       preferred: {},
       auctions: {},
       pendingAutoExtendNames: [],
+      nextPageUrl: null,
```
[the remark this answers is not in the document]
pstachel: Yes, in `fetchAuctionsChunk` method

src/store/plugins/fungibleTokens.js
CedrikNikita: - Firstly, i think that was a temprorary fix since the PR is named `WIP...`.
- Secondary, that won't work since 400 and 500 is not an error for the fetch. We need to start throwing and actuall errors on side of fetchJson, but also we need to handle them throug out the whole project.

That's why i don't think this is part of this PR.

src/popup/router/pages/Names/AuctionList.vue
pstachel: I thought it will be clearer and more efficient to have a method here since it is changed only on input event in `Filters.vue`

src/store/plugins/names.js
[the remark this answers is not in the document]
davidyuk: `fetchAuctionsChunk` uses `this.nextPageUrl` that is defined in component state, not binded to store

src/store/plugins/fungibleTokens.js
davidyuk: 👌 https://github.com/aeternity/superhero-wallet/issues/1433

I didn't notice `WIP`, used to look at draft/not draft mark

PR #1437, opened 2022-05-06
fix(transactions): show tipped url only in transaction with function equal to 'tip'

pstachel:


CedrikNikita (changes requested):
This will result in excluding dex swap transactions from the TransactionList.

pstachel:
@CedrikNikita dex swap transactions are not excluded from transaction list (at least in my case) - it removes them only when we select "Tips" filter

CedrikNikita:
> @CedrikNikita dex swap transactions are not excluded from transaction list (at least in my case) - it removes them only when we select "Tips" filter

It is my mistake, i was looking at something else.

CedrikNikita (approved):

PR #1440, opened 2022-05-12
fix: prevent null as valid parameter at formatDate and formatTime

pstachel:


CedrikNikita (commented):
We will end up gettring a error, but as far as i know we have an issue showing from date based on null. But what would we have now instead of wrong date? a error? Is it handled in a suitable way?

pstachel:
@CedrikNikita when `null` will be provided as a parameter it will not be parsed but treated same as other invalid parameters, so we will receive "Invalid date" info - maybe we should have translations for that case?

pstachel:
The default behaviour should be empty space instead of "Invalid date"

PR #1441, opened 2022-05-12
feat: enable simplex

pstachel:


CedrikNikita (changes requested):
To actually enable simplex you need to also enable the buttn that will lead to the simplex page.

src/popup/router/components/FungibleTokens/TokensList.vue
```diff
@@ -21,7 +21,6 @@
           {{ $t('pages.fungible-tokens.receiveTokens') }}
         </Button>
         <Button
-          v-if="UNFINISHED_FEATURES"
```
CedrikNikita, line 24: The whole `data` function can be removed from the state.

src/popup/router/pages/FungibleTokens/TokenDetails.vue
```diff
@@ -20,7 +20,7 @@
           <ReceiveIcon />{{ $t('pages.token-details.receive') }}
         </BoxButton>
         <BoxButton
-          v-if="UNFINISHED_FEATURES && id === 'aeternity'"
+          v-if="id === 'aeternity'"
```
CedrikNikita, line 23: `UNFINISHED_FEATURES` can be removed from the `data`

CedrikNikita (changes requested):
You need also to enable the `/buy` page in the router.

CedrikNikita (approved):

PR #1444, opened 2022-05-20
feat: adjust to mdw v2

pstachel:


src/store/plugins/names.js
```diff
@@ -69,27 +69,31 @@ export default (store) => {
...
+              owner: tx.accountId,
+            })));
+
+        const getNames = (address) => middleware.getNamesOwnedBy(address)
+          .then(({ active }) => active.map(({ info, name, hash }) => ({
+            createdAtHeight: info.activeFrom,
+            expiresAt: info.expireHeight,
+            owner: info.ownership.current,
+            pointers: info.pointers,
+            nameHash: hash,
```
CedrikNikita, line 87: Why did you decide to not just add a `nameHash: hash`, but extract the whole `getNames` function?

pstachel: Only for better code readability and consistency with the second function in `names` assignment

src/store/plugins/fungibleTokens.js
```diff
@@ -60,17 +62,18 @@ export default (store) => {
...
+      async getAvailableTokens({ rootState: { middleware }, commit }) {
+        if (!middleware) return;
+        const result = await getAllPages(() => middleware.fetchByPath('/v2/aex9?by=name&limit=100'), middleware.fetchByPath);
+        if (!result?.length) {
+          commit('setAvailableTokens', {});
+          return;
+        }
 
-        const availableTokens = response.reduce((obj, { contract_id: contract, ...other }) => ({
+        const availableTokens = result.reduce((obj, { contract_id: contract, ...other }) => ({
```
CedrikNikita, line 73: This code is not adapted to work with `camelcaseKeysDeep`ed data that is now returning, which results in having wrong data in availableTokens.

CedrikNikita (changes requested):
- If you think that fetchByPath is needed function, why don't you use it everywhere?
- why account balances wasn't moved to a v2 version as well?

PR #1446, opened 2022-05-23
fix: load all auctions at once

pstachel:


src/store/plugins/names.js
```diff
@@ -97,9 +97,11 @@ export default (store) => {
       },
       async fetchAuctions({ rootState: { middleware } }) {
         if (!middleware) return [];
+
+        // TODO: Switch to onscroll loading after/while resolving https://github.com/aeternity/superhero-wallet/issues/1400
         return (
-          await middleware.getAllAuctions({ by: 'expiration', direction: 'forward' })
-        ).data.map(({ name, info }) => ({
+          await getAllPages(() => middleware.getAllAuctions({ by: 'expiration', direction: 'forward' }), middleware.fetchByPath)
```
CedrikNikita, line 103: Suggest to add maximum field `limit` value.

CedrikNikita (approved):
@Liubov-crypto can you check that auctions are working properly in the `names/auctions` page. And that amount of actuions is the same as in the `https://explorer.aeternity.io/auctions`?

Liubov-crypto:
Tested, LGTM

PR #1451, opened 2022-06-01
test: improve unit testing coverage

pstachel:
part of #1445

### DIFF
--- a/tests/unit/button-plain.spec.js
+++ b/tests/unit/button-plain.spec.js
@@ -1,0 +1,14 @@
+import { shallowMount } from '@vue/test-utils';
+import ButtonPlain from '../../src/popup/router/components/ButtonPlain.vue';
+
+describe('ButtonPlain', () => {
+  it('should have default slot', async () => {
+    const testPhrase = 'Slot Content';
+    const wrapper = shallowMount(ButtonPlain, {
+      slots: {
+        default: testPhrase,
+      },
+    });
+    expect(wrapper.find('.button-plain').text()).toContain(testPhrase);
+  });
+});
--- a/tests/unit/helper-functions/convert-token.spec.js
+++ b/tests/unit/helper-functions/convert-token.spec.js
@@ -1,0 +1,17 @@
+import BigNumber from 'bignumber.js';
+import { convertToken } from '../../../src/popup/utils/helper';
+
+const invalidInput = [null, 'test', undefined, {}, [], NaN, '1', '-1'];
+
+const powers = Array.from(Array(61), (_, index) => index - 30);
+
+describe('convertToken', () => {
+  powers.forEach((power) => it('should move decimal point n places to desired direction', () => {
+    const value = '1234.9876123456789123456789123456789';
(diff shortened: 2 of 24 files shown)


tests/unit/button-plain.spec.js
```diff
@@ -0,0 +1,21 @@
+import { shallowMount } from '@vue/test-utils';
+import ButtonPlain from '../../src/popup/router/components/ButtonPlain.vue';
+
+describe('ButtonPlain', () => {
+  it('should emit click event on click', async () => {
```
CedrikNikita, line 5: IMO, I don't think tha we should test that button is emiting click event, every other test is using button, and if the click would fail the test would fail as well.

CedrikNikita (commented):
I also believe that there is mainly `component test` type presented here, but the issue was about unit tests.

https://vuejs.org/guide/scaling-up/testing.html

tests/unit/helper-functions/convert-token.spec.js
```diff
@@ -0,0 +1,14 @@
+import { convertToken } from '../../../src/popup/utils/helper';
+
+const invalidInput = [null, 'test', undefined, {}, [], NaN, '1', '-1'];
+
+describe('convertToken', () => {
+  [2, -2, 0].forEach((precision) => it('should move decimal point n places to desired direction', () => {
+    const value = 1234.9876;
+    const expectedResult = value * (10 ** precision);
```
CedrikNikita, line 8: You've missed the point during creation of this test. We need to test that it converts with the maximum precision each digit should still be there. 
We need to test here edge cases something like 20 30 digits powering. Need to use BigInt or BigNumber here to test properly.

tests/unit/tab-bar.spec.js
```diff
@@ -0,0 +1,28 @@
...
+describe('TabBar', () => {
+  routes.forEach((route, index) => it('should have assigned route', () => {
+    const wrapper = shallowMount(TabBar, {
+      stubs: {
+        RouterLink: RouterLinkStub,
+      },
+    });
+    expect(wrapper.findAllComponents(RouterLinkStub).at(index).props().to).toStrictEqual(route);
+  }));
+  it('should have same amount of routes and RouterLinks', () => {
```
CedrikNikita, line 20: In this case we would need to change this test everytime we would decide to change the routes in the component.

src/popup/router/components/AccountCard.vue
```diff
@@ -28,8 +28,8 @@ export default {
   },
   props: {
     idx: { type: Number, required: true },
-    color: { type: String, required: true },
-    shift: { type: Number, required: true },
+    color: { type: String },
```
CedrikNikita, line 31: The `color` and `shift` is actually required, why do you decide to change that.

pstachel: Because when `AccountCard` is used somewhere it always receive all of the props through `v-bind` and it resulted in errors in the tests

CedrikNikita (approved):

Issue #1458, opened 2022-06-10
Simplex Button not working on IOS


kenodressel:

CedrikNikita:
fixed in #1459

PR #1459, opened 2022-06-10
fix: simplex button on mobile #1458

ifaouibadi:


kenodressel (approved):

CedrikNikita (commented):
If you think that we need to open simplex in the new tab everywhere, then lets complelty remove the route an the component.

PR #1462, opened 2022-06-10
style(iOS): Set min height for main router-view

pstachel:


CedrikNikita:
THe issue that this PR was about to fix, already is not presented in the develop.

PR #1468, opened 2022-06-16
[SW-61] More – Main screen - OLD

ifaouibadi:
Fixes: https://aeternity-blockchain.atlassian.net/browse/SW-61

UI Ref: https://www.figma.com/proto/3oGLWzSH0oJljo4RETZtur/Superhero-Wallet-UI-(%E2%9C%94%EF%B8%8FUpdated)?node-id=13969%3A315682&scaling=min-zoom&page-id=10439%3A144915&starting-point-node-id=10641%3A135489


Liubov-crypto:
Tested, LGTM. 

Maybe this should be fixed (view of bell button): 
fix it

Idk if this related to this pr:

I was not able to do a retip on superhero.com: 
retip on sh feed

When I switched to the testnet and then to the mainnetwork, the wallet automatically switched me to testnet by default , which is incorrect.

CedrikNikita (commented):
Looks good from code perspective

Is this difference acceptable @paolomolo?
Actually everything is not in place, including Header
diff
.

CedrikNikita (approved):
The `color-bg-3` change should be agreed, else is fine from code perspective.

CedrikNikita (approved):
Everything is fine, except for the some screens are now will be without the Home button, but it is again up to @paolomolo.

PR #1478, opened 2022-06-20
Improve transaction fetching

CedrikNikita:


CedrikNikita:
@Liubov-crypto ensure that every transactions is loading and everything transaction related is working properly.

Liubov-crypto:
In general LGTM, except that the wallet is very slow and doesn't load all transactions.

Liubov-crypto:
if I  switch the wallet from the test network to the main network, then the wrong balance of tokens is showed up:
mainnet

Should be: 
mainnet 1

To see the maximum number of transactions, I need to scroll the list of transactions to the maximum down, otherwise I will not see transactions for the previous year.

The wallet is still slow and when the user scrolls fast, the list of transactions may be visible with breaks.

src/store/actions.js
```diff
@@ -38,6 +38,30 @@ export default {
       .filter((transaction) => !transactions.pending.find((tx) => tx?.hash === transaction?.hash))
       .map((transaction) => ({ ...transaction, pending: true }));
   },
+  // TODO: remove it after https://github.com/aeternity/tipping-community-backend/issues/405, 406 will be resolved
```
kenodressel, line 41: why remove it, seems like we will need it regardless?

CedrikNikita: I've rephrased it.

kenodressel (approved):

PR #1479, opened 2022-06-22
[SW-46] Network – Choose network connection

pstachel:


src/popup/router/components/NodeConnectionStatus.vue
```diff
@@ -49,7 +52,6 @@ export default {
     position: sticky;
   }
 
-  bottom: 48px;
```
CedrikNikita, line 52: We have that backup property to support old browser versions.
Let's consider dropping of support for the old browsers in other PR. Or at least investigate it. If you want you can create an issue to think about it.

src/popup/utils/constants.js
```diff
@@ -54,6 +54,16 @@ export const calculateNameClaimFee = (name) => calculateFee(SCHEMA.TX_TYPE.nameC
...
+  {
+    url: 'https://mainnet.aeternity.io',
+    networkId: 'ae_mainnet',
+    middlewareUrl: 'https://mainnet.aeternity.io/mdw',
+    explorerUrl: 'https://explorer.aeternity.io',
+    compilerUrl: 'https://compiler.aepps.com',
+    backendUrl: 'https://raendom-backend.z52da5wt.xyz',
+    tipContractV1: 'ct_2AfnEfCSZCTEkxL5Yoi4Yfq6fF7YapHRaFKDJK3THMXMBspp5z',
+    name: 'Mainnet',
+  },
```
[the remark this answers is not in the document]
pstachel: So should I remove it from defaultNetworks or just prevent it from being selected as default?

src/popup/router/components/SettingsLabel.vue
```diff
@@ -0,0 +1,17 @@
+<template>
```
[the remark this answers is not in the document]
CedrikNikita: Since component doesn't now anything about the `settings`, being just a `p` with some css, i was proposing to not name it `settings` related.

src/popup/router/components/ScreenLabel.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <p class="settings-top-label">
```
CedrikNikita, line 2: Class name remain previous.

src/popup/router/components/NodeConnectionStatus.vue
```diff
@@ -79,4 +93,28 @@ export default {
     bottom: 0;
   }
 }
+
+.node-connection-status {
+  &.relative {
```
CedrikNikita, line 98:
```suggestion
.node-connection-status.relative {
```
In such case we don't need to split it into two rows.

src/popup/utils/constants.js
```diff
@@ -127,6 +127,15 @@ export const NOTIFICATION_SETTINGS = [
...
 
+export const NETWORK_PROPS = {
+  url: null,
+  name: null,
+  error: false,
+  middlewareUrl: defaultNetwork.middlewareUrl,
+  compilerUrl: defaultNetwork.compilerUrl,
+  backendUrl: defaultNetwork.backendUrl,
+};
+
```
CedrikNikita, line 138: Is not used anywhere.

src/popup/router/components/NetworkRow.vue
```diff
@@ -0,0 +1,103 @@
...
+    isActive() {
+      return this.network === this.activeNetwork;
+    },
+  },
+  methods: mapActions(['selectNetwork']),
+};
+</script>
+
+<style lang="scss" scoped>
+@use '../../../styles/variables';
```
CedrikNikita, line 69:
```suggestion
```
Is not used in that file.

src/popup/router/components/NetworkRow.vue
```diff
@@ -0,0 +1,103 @@
...
+@use '../../../styles/typography';
+
+.network-row {
```
CedrikNikita, line 72: We are using a nesting approach with selectors.
You shall consider using it. https://www.w3.org/TR/css-nesting-1/
Or to reconsider this approach with your own arguments and create an issue about it.

src/popup/router/components/NetworkRow.vue
```diff
@@ -0,0 +1,103 @@
...
+}
+
+.network-row-wrapper {
+  text-align: left;
+
+  &.row-disabled {
+    opacity: 0.5;
+  }
+
+  p {
```
konradodo, line 83: use class

src/popup/App.vue
```diff
@@ -5,6 +5,7 @@
       'not-rebrand': $route.meta.notRebrand,
       'show-header': showStatusAndHeader,
       'hide-tab-bar': $route.meta.hideTabBar,
+      'new-ui': $route.meta.newUI,
```
peronczyk, line 8: This is not needed anymore.

src/popup/App.vue
```diff
@@ -5,6 +5,7 @@
       'not-rebrand': $route.meta.notRebrand,
```
peronczyk, line 5: This was removed, that's weird its still here.

src/popup/App.vue
```diff
@@ -5,6 +5,7 @@
...
       'show-header': showStatusAndHeader,
       'hide-tab-bar': $route.meta.hideTabBar,
```
peronczyk, line 7: This was also removed.

src/popup/router/components/Button.vue
```diff
@@ -67,6 +65,34 @@ export default {
 @use '../../../styles/variables';
 @use '../../../styles/typography';
 
+.new-ui {
```
peronczyk, line 68: I strongly recommend to not do this change. Components should be an encapsulated things and it's states should be only changed with the props. Only exception for this are the css variables.

src/popup/router/components/NetworkRow.vue
```diff
@@ -0,0 +1,103 @@
...
+          data-cy="network-name"
+        >
+          {{ network.name }}
+        </p>
+      </RadioButton>
+      <p
+        class="url shifted"
+        data-cy="network-url"
+      >
+        <span class="url-label">{{ $t('pages.network.url') }}</span> {{ network.url }}
```
peronczyk, line 24: Please move the `{{ network.url }}` to next line.

src/popup/router/components/NetworkRow.vue
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  .name {
+    @extend %face-sans-14-medium;
+  }
+
+  .url {
+    @extend %face-sans-14-regular;
+
+    &.shifted {
```
peronczyk, line 94: As far as I see there is no element with only `url` class and without `shifted`. I think those styles should be merged.

src/popup/router/components/NodeConnectionStatus.vue
```diff
@@ -1,21 +1,32 @@
...
-  </div>
+  <transition :name="isRelative ? 'fade' : ''">
+    <div
+      v-if="isVisible"
+      :data-cy="nodeStatus !== 'error' ? 'connect-node' : ''"
+      class="node-connection-status"
+      :class="[
+        `connect-${nodeStatus === 'error' ? 'error' : 'node'}`,
+        {
+          'hide-tab-bar': $route.meta.hideTabBar,
```
peronczyk, line 10: This was removed.

src/popup/router/components/NodeConnectionStatus.vue
peronczyk: I don't agree. The `.relative` is a state so it's easier to follow the CSS when this is done like Paweł did it.
But to be honest I think this component should not control its position relative to its parent component. The parent should place it in correct place.

src/popup/router/components/NodeConnectionStatus.vue
```diff
@@ -79,4 +93,28 @@ export default {
...
+.fade-enter-active,
+.fade-leave-active {
+  transition: opacity 0.15s ease;
+  opacity: 0.75;
+}
+
+.fade-enter,
+.fade-leave-to {
+  opacity: 0;
+}
```
peronczyk, line 119: We already heave the global fade transition placed in `trnasitions.scss` file.

src/popup/router/components/RadioButton.vue
```diff
@@ -0,0 +1,91 @@
...
+    />
+    <slot class="radio-holder" />
+  </label>
+</template>
+
+<script>
+export default {
+  props: {
+    value: { type: [String, Number, Boolean], default: '' },
+    type: { type: String, default: 'checkbox' },
```
peronczyk, line 25: Input type radio has only one type: `radio`. I think this prop makes no sense.

src/popup/router/components/RadioButton.vue
```diff
@@ -0,0 +1,91 @@
...
+    name: { type: String, default: '' },
+    disabled: { type: Boolean },
```
peronczyk, line 27: This can be shortened.

src/popup/router/components/RadioButton.vue
```diff
@@ -0,0 +1,91 @@
...
+  },
+  methods: {
+    handleChange(event) {
+      if (this.disabled) return;
+      this.$emit('input', event.target.checked);
```
peronczyk, line 32: I suggest following:
```
if (!this.disabled) {
  this.$emit(...
}
```

Please bare in mind that it is a good practice to decide if function should return something or not.

src/popup/router/components/RadioButton.vue
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  &.disabled {
+    opacity: 50%;
+    cursor: not-allowed;
+  }
+
+  .radio-dot {
+    border-radius: 50%;
+    border: 1px solid rgba(255, 255, 255, 0.5);
```
peronczyk, line 61: Please use colors taken from variables.

src/popup/router/components/RadioButton.vue
```diff
@@ -0,0 +1,91 @@
...
+    width: 20px;
+    height: 20px;
+    background: variables.$color-bg-3;
+    margin-right: 6px;
+    position: relative;
```
peronczyk, line 66: I suggest to organise CSS properties according to commonly used methodology: https://9elements.com/css-rule-order/

src/popup/router/pages/Networks.vue
```diff
@@ -1,337 +1,55 @@
...
-          @click="deleteNetwork(network.index)"
-        >
-          <ae-icon name="delete" />
-          {{ $t('pages.network.delete') }}
-        </ButtonPlain>
-      </ActionsMenu>
-    </div>
+      :network="network"
+    />
     <Button
```
peronczyk, line 9: Please use the `new-ui` prop.

src/popup/router/pages/Networks.vue
```diff
@@ -1,337 +1,55 @@
...
-    </div>
-  </div>
 </template>
 
 <script>
-import { mapGetters } from 'vuex';
+import { mapActions, mapGetters } from 'vuex';
+import NetworkRow from '../components/NetworkRow.vue';
+import SettingsTopLabel from '../components/ScreenLabel.vue';
+import PlusCircle from '../../../icons/plus-circle.svg?vue-component';
```
peronczyk, line 25: I suggest to add `Icon` suffix to icon components.

src/popup/router/routes/index.js
```diff
@@ -142,9 +142,11 @@ export default [
   {
     path: '/settings/networks',
     component: Networks,
+    name: 'networks',
     props: true,
     meta: {
       title: 'networks',
+      newUI: true,
```
peronczyk, line 149: This is not used anymore.

src/store/actions.js
```diff
@@ -23,6 +23,14 @@ export default {
     commit('setMiddleware', null);
     commit('initTransactions');
   },
+  async selectNetwork({ dispatch, getters }, network) {
+    await dispatch('switchNetwork', network);
+    if (getters.tippingSupported) return;
+    await dispatch('modals/open', {
+      name: 'default',
```
peronczyk, line 30: Please use modal names taken from a constants.

src/styles/variables.scss
```diff
@@ -6,6 +6,8 @@ $font-sans: 'IBM Plex Sans', sans-serif;
 $font-mono: 'IBM Plex Mono';
 
 $color-white: #fff;
+$color-white-alpha-15: rgba(255, 255, 255, 0.15);
+$color-white-alpha-85: rgba(255, 255, 255, 0.85);
```
peronczyk, line 10: Please avoid using SCSS variables for the opacity.

peronczyk (changes requested):

src/popup/router/components/RadioButton.vue
pstachel: I see that it's used in few places as a checkbox or as a radio (same with checkbox that has different styles) so I think id should stay like that

src/popup/components/NodeConnectionStatus.vue
```diff
@@ -1,18 +1,26 @@
 <template>
-  <div
-    v-if="nodeStatus && account.address && isLoggedIn"
-    :data-cy="nodeStatus !== 'error' ? 'connect-node' : ''"
-    class="node-connection-status"
-    :class="`connect-${nodeStatus === 'error' || nodeStatus === 'offline' ? 'error' : 'node'}`"
-  >
-    {{ statuses[nodeStatus] }}
-  </div>
+  <transition :name="isSettings ? 'from-settings-fade' : ''">
```
konradodo, line 2: <transition :name="isSettings && 'from-settings-fade'">

konradodo (approved):

src/store/actions.js
```diff
@@ -14,14 +14,24 @@ import {
...
     commit('setMiddleware', null);
     commit('initTransactions');
   },
+
+  async selectNetwork({ dispatch, getters }, network) {
+    await dispatch('switchNetwork', network);
+    if (getters.tippingSupported) return;
+    await dispatch('modals/open', {
+      name: MODAL_DEFAULT,
+      ...i18n.t('modals.tip-mainnet-warning'),
```
peronczyk, line 32: I'm not sure about this. I think we should pass data to the modal as the object prop values.

src/popup/components/Header.vue
```diff
@@ -120,7 +120,7 @@ export default {
       return this.$route.meta.hideNotificationsIcon;
     },
     notificationsCount() {
-      return [...this.notifications, ...this.superheroNotifications].filter(
+      return [...(this.notifications || []), ...(this.superheroNotifications || [])].filter(
```
peronczyk, line 123: I suggest to split this in to 2 lines. I mean the elements added to the array in my opinion should be in separate lines.

peronczyk (changes requested):

src/popup/components/NetworkRow.vue
```diff
@@ -0,0 +1,98 @@
...
+
+<style lang="scss" scoped>
+@use '../../styles/typography';
+
+.network-row {
+  margin-bottom: 12px;
+}
+
+.network-row-wrapper {
+  &.row-disabled {
```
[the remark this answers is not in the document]
pstachel: I'll change the name from disabled to inactive to no mislead about it's behaviour

src/popup/components/NetworkRow.vue
```diff
@@ -0,0 +1,98 @@
...
+  },
+  props: {
+    network: {
+      type: Object,
+      required: true,
+    },
+  },
+  computed: {
+    ...mapGetters(['activeNetwork']),
+    isActive() {
```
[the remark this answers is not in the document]
pstachel: This component is used only in this place and I don't think there's a need for moving that condition to the parent

peronczyk (approved):

PR #1481, opened 2022-06-23
Improve fungible tokens perfomance

CedrikNikita:


CedrikNikita:
@Liubov-crypto You have to check that every fungible token balance is still there, and fungible token related operatios work as in develop. Also check will the initialization of the wallet be fater for you?

kenodressel (approved):

PR #1482, opened 2022-06-24
feat: show all tokens user had activity with

CedrikNikita:


CedrikNikita:
@Liubov-crypto can you check that fungible token list consist only with tokens, that have transaction history and that they are also can be selected as the current selected token for the wallet.

Liubov-crypto:
The tokens with 0 balance are not displayed on the development version on dev version: 
tok

on this pr: 
pr

Also, this issue on the dev version too: #1483 

The rest LGTM.

Liubov-crypto:
Tested, LGTM, except this small issue https://github.com/aeternity/superhero-wallet/issues/1483

Issue #1457, opened 2022-06-10
Create Custom View TX Screen for DEX Swaps


kenodressel:
Augment the TX view screen and probably TX list with information about the swap.

PR #1495, opened 2022-07-12
[SW-78] DEX transaction – Update transaction list

### LINKED ISSUE #1297: Show the state of transactions
The middleware currently provides also aborted / reverted transactions in the list of transactions. We should filter these as they are not relevant or at least flag them visually as aborted / reverted. 
Examples in https://github.com/aeternity/mdw-frontend/issues/89

CedrikNikita:
fixes #1297,
part of #1457


src/popup/router/components/TransactionItem.vue
```diff
@@ -3,45 +3,82 @@
...
-        class="type"
-      >
-        ({{ ($t('transaction.type')[txType]) || txType }})
-      </span>
+    <div class="header">
+      <span />
+      <div class="status">
+        <template v-if="transaction.pending">
+          <Pending class="icon" />
+          <span class="pending">Pending...</span>
```
pstachel, line 11: shouldn't we have `Pending` in `en.json`?

src/popup/router/components/TransactionItem.vue
```diff
@@ -3,45 +3,82 @@
...
+        v-for="token in tokens.filter(({ amount }) => amount)"
+        :key="token.symbol"
+        :class="['token-row', token.isReceived ? 'received': 'sent', { error: isErrorTransaction }]"
+      >
+        <span class="amount">
+          {{
+            `${token.isReceived ? '+' : '-'}
+            ${amountRounded(token.decimals
+            ? convertToken(token.amount || 0, -token.decimals) : token.amount)}`
+          }}
```
pstachel, line 45: wouldn't it be possible to use TokenAmount.vue here?

src/popup/router/components/TransactionItem.vue
```diff
@@ -50,19 +87,82 @@ export default {
...
+        ...this.transaction.tx,
+        amount: this.isAllowanceTransaction
+          ? convertToken(this.transaction.tx.fee, -MAGNITUDE)
+          : this.getTxAmountTotal(this.transaction),
+        symbol: this.isAllowanceTransaction ? 'AE' : this.getTxSymbol(this.transaction),
+        isReceived: this.getTxDirection(this.transaction) === 'received',
+        isAe: this.isAllowanceTransaction || this.getTxSymbol(this.transaction) === 'AE',
+      }];
+    },
+    labels() {
```
pstachel, line 118: Maybe we could polish this function a bit? I'd suggest putting more complicated ifs in computed. For example 
```suggestion
  isDex() {
  return transactionResolver[camelCase(this.transaction.tx.function)]
        && (this.getDexContracts.wae.includes(this.transaction.tx.contractId)
        || this.getDexContracts.router.includes(this.transaction.tx.contractId))
}, 
    labels() {
    ...
     if (this.isDex) {
     ['DEX', FUNCTION_TYPE_DEX.pool.includes(this.transaction.tx.function) ? 'POOL' : 'SWAP']
     }
    ...
```

pstachel (changes requested):

src/popup/router/components/TransactionItem.vue
```diff
@@ -3,45 +3,82 @@
...
+          <span data-cy="time">{{ transaction.microTime | formatTime }}</span>
+          <div
+            v-if="isErrorTransaction"
+            class="error"
+          >
+            <Warning
+              v-if="transaction.tx.returnType === 'abort'"
+              class="icon"
+            />
+            <Reverted
```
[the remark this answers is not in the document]
CedrikNikita: no, it has `revert

src/popup/router/components/TransactionItem.vue
CedrikNikita: No, because the fiat value of the tokens should be shown in the `footer`, despite it appearing in the rows. Plus having TokenAmount.vue shoud be rewritten an include way more logic that it have right now. I would rather do it when we are agreed on changing logic in it in several or all places. 
We can mark that with `TODO: extract logic into TokenAmount` for now.

PR #1447, opened 2022-05-24
Feature/adjust to mdw v2 1.9.2

pstachel:

PR #1496, opened 2022-07-12
Improve names list performance

CedrikNikita:
based on #1447


Liubov-crypto:
tested, works the same as development version

pstachel (approved):

PR #1499, opened 2022-07-13
refactor: show pending transaction until middleware returns it

### LINKED ISSUE #1353: remember broadcasted tx and display "pending" state
see https://github.com/aeternity/superhero-wallet/issues/1352#issuecomment-950797422

I updated a name (setting a pointer) and didn't know about the state of the tx. it isn't displayed at all which can be confusing to users. they might broadcast the same tx multiple times again (which then works because the nonce is adjusted accordingly).

we should prevent this and the user should know that the tx is still not included in a block.

as we are relying on the mdw backend for displaying the info that might be tricky. the mdw needs to sync and index the info first but the tx might actually be included for a long time already. cc @kenodressel

CedrikNikita:
fixes #1353


pstachel (approved):

Liubov-crypto:
Tested, LGTM.

PR #1509, opened 2022-07-22
[SW-118, SW-107] Account card Swipe and arrow switching

ifaouibadi:


CedrikNikita (commented):
- Each of icons you added presented in the `icons/account-card`, i propose to move them from `icons/account-card`, to `icons`.
- Whole account swithing logic is already presented try to reuse it, and remove the old one afterwards

CedrikNikita:
`/icons/arrow.svg` is the same component just need to be rotated, BUT maybe roateted svg would be seen through other objects

package.json
```diff
@@ -52,12 +52,14 @@
     "qrcode.vue": "^1.7.0",
     "rxjs": "^6.6.7",
     "rxjs-etc": "^10.6.2",
+    "swiper": "^5.4.5",
```
CedrikNikita, line 55: Do we need both packages: `swiper` and `vue-awesome-swiper`?

CedrikNikita (commented):
- arrows have different style than in the designs
- arrow in the designs are part of the card, @paolomolo is it ok if they won't be?
- shall we maybe connect the add account card with an actual functionality of adding account

- visual bug
Swipe

Liubov-crypto:
I confirm the visual bug and the difference with the arrows in the design.

In the wallet, the arrow is located near the edge of the card, in the Figma, there is an indent from the edge of the card. 

Also, the card in the wallet looks narrow:
arrow

package.json
ifaouibadi: Yes, we had to use those specific version to make it work our current Vue version

subhod-i:
My suggestion is to merge this with https://github.com/aeternity/superhero-wallet/pull/1510 as they are part of the same story

onvisions:
@ifaouibadi  @subhod-i @Liubov-crypto @CedrikNikita @paolomolo
About the account navigation arrows (previous and next): the best will be to follow design (and prototype) and keep the arrows within the cards as buttons linked to the switcher component and to pay attention to the positioning and the icons as they seem to be different from Figma design. Their positioning fits to the new layout and design of the card so within the old card design it will look weird. IMO this is the best thing to do here.

**Workaround suggestion:** I can think of a quick workaround solution to be experimented with to improve the current issue with arrows being detached visually from the card component while swiping animation is in progress (which I think looks weird). My suugestion is to make the arrow invisible as soon as it's clicked (or cards swiping starts with mouse drag) and keep it this way until the swiping animation continues. Then when swiping transition is over to make it visible again with some nice opacity transition for 0,3 sec. However I'm not sure if this will save you time and efforts and even if it will look good enough to be considered a fix. It's more an alternative idea in case there are serious issues preventing us from exact implementation of the design.

pstachel:
@ifaouibadi I think not all elements should be selectable by user (or event none of them)
image

pstachel (changes requested):
I think not all elements should be selectable by the user (or even none of them)
image

pstachel:
There are some weird black lines next to longer account names 
image

paolomolo:
It would be good if we could change the gradient color for the truncation from black to the color of that account card

src/popup/router/components/AccountSwitcher.vue
```diff
@@ -50,87 +101,79 @@ export default {
       return this.accounts.map((a, index) => ({ ...a, i: index })).filter((a) => a.showed);
     },
     selectedCardNumber() {
-      return this.filteredAccounts.findIndex((a) => a.i === this.activeIdx);
+      return this.filteredAccounts.findIndex((a) => a.i === this.activeAccount);
```
CedrikNikita, line 104: This and this can be remove completly, since it is no longer used in that component.
https://github.com/aeternity/superhero-wallet/blob/fa38bb548bc23bc2760489cd9761e732bf673174/src/popup/router/components/AccountSwitcher.vue#L94-L99
https://github.com/aeternity/superhero-wallet/blob/fa38bb548bc23bc2760489cd9761e732bf673174/src/popup/router/components/AccountSwitcher.vue#L103-L105

CedrikNikita (commented):
Is removing of arrows on cordova a new decision @paolomolo?

CedrikNikita:
I think we can omit the changes in e2e tests, since they were not fixing problems with them.

pstachel (approved):

pstachel (approved):

CedrikNikita:
More screen margin
There isa margin in the more screen, we need to remove it for the for the Plate, but leave it for the rest of the page to avoid this jumping. Or just not to leave it like that for a while, because we will change more page in future.

src/popup/locales/en.json
```diff
@@ -324,7 +324,9 @@
       "claim-name": "Claim your .chain name"
     },
     "accounts": {
-      "showWallet": "Show on Wallet Home"
+      "showWallet": "Show on Wallet Home",
+      "add-account": "Add accounts",
```
CedrikNikita, line 328: `Add account` in Figma 
originally mentioned by @Liubov-crypto

PR #1519, opened 2022-08-02
[SW-119] Bullet switcher

pstachel:


ifaouibadi (approved):

src/popup/router/components/BulletSwitcher.vue
```diff
@@ -0,0 +1,145 @@
...
+          v-for="(_, idx) in optionsSize"
+          :key="idx"
+          class="bullet"
+          :class="{
+            medium: getMediumCondition(idx),
+            big: getBigCondition(idx),
+            active: idx === currentIdx,
+          }"
+          :style="getBulletTranslation(idx)"
+          @click="() => $emit('change', idx)"
```
CedrikNikita, line 18:
```suggestion
          @click="$emit('change', idx)"
```
Why shouldn't we do it like that?

src/popup/router/components/BulletSwitcher.vue
```diff
@@ -0,0 +1,145 @@
...
+        />
+      </div>
+    </div>
+    <PlusCircle
+      :class="{ active: currentIdx === optionsSize }"
+      @click="() => $emit('change', optionsSize)"
```
CedrikNikita, line 24:
```suggestion
      @click="$emit('change', optionsSize)"
```

Liubov-crypto:
As I see there are different fonts, and the account address in the wallet is longer than in Figma. Also in the wallet, the send/receive buttons are white, and in Figma they are transparent.
@pstachel  @ifaouibadi  please check it
figma : wall

src/styles/variables.scss
```diff
@@ -7,13 +7,15 @@ $font-mono: 'IBM Plex Mono';
 
 $color-white: #fff;
 $color-black: #000;
+$color-purple: #3270ff;
```
peronczyk, line 10: This variable is not used anymore.

ifaouibadi (approved):

PR #1522, opened 2022-08-09
Fix transactions

CedrikNikita:
Since all the dex contract calls contain user address, `txs/backward` will return them and will have more information than `aex9/transfers/to/`. I propose to exclude them from the list which `getTokenHistory` is returning.
image


pstachel (approved):

Liubov-crypto:
Tested, in general LGTM.

PR #1527, opened 2022-08-10
fix(transaction-details): remove fiat amount for tokens other than ae

pstachel:
...


csiarab:
Bug was fixed. Now no fiat amount can be seen for the tokens both on the confirmation screen and transaction details.

CedrikNikita (approved):

PR #1532, opened 2022-08-15
[SW-19] Receive modal

peronczyk:


src/popup/router/components/AssetSelector.vue
```diff
@@ -0,0 +1,96 @@
+<template>
```
peronczyk, line 1: Please skip this file while reviewing as it will be totally changed by @pstachel .

src/popup/router/components/Button.vue
```diff
@@ -43,9 +53,9 @@ export default {
     inline: Boolean,
     inactive: Boolean,
     to: { type: [String, Object], default: null },
-    bold: Boolean,
     backgroundless: Boolean,
     iconText: Boolean,
+    newUi: Boolean,
```
pstachel, line 58: I think the flag we used to use was `newUI`

src/popup/router/components/InputField.vue
```diff
@@ -1,35 +1,96 @@
...
   props: {
     value: { type: [String, Number], default: null },
+    label: { type: String, default: '' },
+    error: Boolean,
+    errorMessage: { type: String, default: '' },
+    warning: Boolean,
+    warningMessage: { type: String, default: '' },
     readonly: Boolean,
+    plain: Boolean,
+    newUi: Boolean,
```
pstachel, line 85: I'd suggest to group the props by type

src/popup/router/components/Modals/TransferReceive.vue
```diff
@@ -0,0 +1,294 @@
...
+
+      <div class="request-specific-amount">
+        <RequestAmount
+          v-model="amount"
+          own-validation
+          :validation="{ min_value_exclusive: 0 }"
+          :label="$t('modals.receive.requestAmount')"
+          @handleAssetSelection="handleAssetSelection"
+        />
+      </div>
```
pstachel, line 59: Do we need to wrap it into another div?

src/popup/router/components/Modals/TransferReceive.vue
```diff
@@ -0,0 +1,294 @@
...
+
+.transfer-receive-modal {
+  font-weight: 500;
+  color: variables.$color-white;
+
+  .transfer-receive {
+    display: flex;
+    align-content: center;
+    justify-content: center;
+    flex-direction: column;
```
pstachel, line 197: You can use mixins.flex here

src/popup/router/components/Modals/TransferReceive.vue
```diff
@@ -0,0 +1,294 @@
...
+      }
+
+      .truncate {
+        display: block;
+      }
+
+      .account-name {
+        @extend %face-sans-16-medium;
+
+        padding-left: 7.69px;
```
pstachel, line 229: It won't work

src/popup/router/components/RequestAmount.vue
```diff
@@ -0,0 +1,137 @@
...
+        >
+          <span v-if="$attrs.value">&thickapprox;</span>
+          {{ formatCurrency(totalAmount) }}
+        </span>
+
+        <span class="request-amount-desc-at">
+          @{{
+            (selectedAsset && selectedAsset.current_price)
+              ? formatCurrency(currentTokenFiatPrice)
+              : 'price: n/a'
```
pstachel, line 47: should be in translation file

pstachel (changes requested):

src/popup/router/components/Modals/TransferReceive.vue
peronczyk: After some tests it turns out that the flex definitions serves no purpose here so I totally removed this block of code.

src/popup/router/components/Modals/TransferReceive.vue
peronczyk: Not that much but it gives the separation of concerns so I would prefer to leave it as it is.

src/popup/router/components/InputField.vue
peronczyk: I kind of agree, but now the props are grouped by the purpose. error + errorMessage.

src/popup/router/components/Button.vue
peronczyk: If I would set the prop as `newUI` then the prop in template should be `new-u-i` which in my opinion is rather bad. What do you think?

pstachel: You're right

src/popup/router/components/Modals/TransferReceive.vue
```diff
@@ -0,0 +1,282 @@
...
+        </a>
+      </div>
+
+      <div class="request-specific-amount">
+        <RequestAmount
+          v-model="amount"
+          own-validation
+          :validation="{ min_value_exclusive: 0 }"
+          :label="$t('modals.receive.requestAmount')"
+          @asset-selected="selectedAsset = $event"
```
pstachel, line 57: I think we don't assign an event to the `selectedAsset`. It may be better to move it to methods

pstachel (changes requested):

src/popup/router/components/Modals/TransferReceive.vue
peronczyk: Done

src/popup/router/components/RequestAmount.vue
```diff
@@ -0,0 +1,137 @@
...
+  },
+};
+</script>
+
+<style lang="scss" scoped>
+@use '../../../styles/variables';
+@use '../../../styles/typography';
+@use '../../../styles/mixins';
+
+.request-amount {
```
subhod-i, line 118: Missing one small feature: 
    
Screenshot 2022-08-22 at 1 18 45 PM

  One of the solution: 
  Within `.request-amount`
  ```scss
  :focus-within .asset-field .asset-selector {
    background-color: #171717;
  }
  ```

peronczyk: Asset selector was refactored by Paweł within the scope of a different task: https://aeternity-blockchain.atlassian.net/browse/SW-130

subhod-i: Okay. @pstachel knows about this.

pstachel (approved):

PR #1533, opened 2022-08-16
[SW-152] Move Names into Account popup

ifaouibadi:


src/popup/router/components/Modals/AccountDetails.vue
```diff
@@ -355,6 +385,45 @@ export default {
...
+        }
+      }
+
+      .tabs {
+        padding: 6px;
+        gap: 16px;
+
+        @include mixins.flex(flex-start, center, row);
+
+        .button-plain {
```
peronczyk, line 411: I suggest to separate the concerns. Parent component should not use the child component's classes as those might change.

src/popup/router/components/Modals/AccountDetails.vue
```diff
@@ -115,6 +143,8 @@ export default {
     return {
       activeTab: 'tokens',
       tabs: ['tokens', 'transactions', 'names'],
```
peronczyk, line 145: I suggest to not use magic strings. Instead maybe some constants would be better.

src/popup/router/components/Modals/AccountDetails.vue
```diff
@@ -73,6 +73,30 @@
             :display-filter="showSearchBox"
           />
         </div>
+        <div
+          v-if="activeTab === 'names'"
+          class="names-list-wrapper"
+        >
+          <div class="tabs">
```
peronczyk, line 80: This styling is something more common so I suggest to create separate component for the tabs.

peronczyk (changes requested):

PR #1537, opened 2022-08-23
feat: enable add account feature

pstachel:


src/popup/router/components/AccountCard.vue
```diff
@@ -1,42 +1,75 @@
 <template>
   <div
-    :class="[
-      'account-card',
-      { subaccount: idx !== 0, 'first-subaccount': idx === 1, minified: cardMinified }]"
-    :style="cssVar"
+    class="account-card"
+    :style="cardCssProps"
+    @click.prevent="$store.dispatch('modals/open', {
```
konradodo, line 5: better use action from mapAction

PR #1539, opened 2022-08-25
[SW-192, SW-179, SW-177, SW-176] 

ifaouibadi:


src/popup/router/components/AccountSwitcher.vue
```diff
@@ -150,14 +150,17 @@ export default {
     top: 81px;
     color: variables.$color-white;
     z-index: 1;
+    height: 60px;
+    width: 30px;
+    margin-top: -20px;
```
pstachel, line 155: Maybe we should change the `top` property instead of using minus margins?

pstachel (changes requested):

src/popup/router/components/AccountSwitcher.vue
```diff
@@ -147,17 +147,19 @@ export default {
   .swiper-button {
     position: absolute;
     opacity: 0.5;
-    top: 81px;
+    top: 61px;
```
peronczyk, line 150: I think placing thing in the middle is better to accomplish with `top: 50%` and then `transform: translateY(-50%)`.

src/popup/router/components/AddAccountCard.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <div class="add-account-card">
+  <div
```
peronczyk, line 2: In my opinion using plain `div` with the ability to click on it is an antipattern: https://stackoverflow.com/a/32659785/2399521. I think it should be a link or button or at least have the tabindex.

src/popup/router/components/AccountInfo.vue
```diff
@@ -60,15 +70,13 @@ export default {
   },
   mixins: [CopyMixin],
   props: {
-    accountIdx: { type: Number, default: -1 },
     color: { type: String, default: 'black' },
-    showCopyIcon: { type: Boolean, default: false },
+    canCopyAddress: { type: Boolean, default: false },
```
peronczyk, line 74: As the default state of the Boolean props is `false` this can be shortened to `canCopyAddress: Boolean`.

peronczyk (changes requested):

src/popup/router/components/AccountSwitcher.vue
ifaouibadi: we can't use `top: 50%` here, because the swiper button has a specific height.

src/popup/router/components/AddAccountCard.vue
```diff
@@ -1,20 +1,26 @@
 <template>
-  <div class="add-account-card">
+  <ButtonPlain
+    class="add-account-card"
+    @click="createAccount"
```
peronczyk: Clicking on the add account card should not create new account immediately. It should open a separate modal. There's another task for this ready to be merged.

peronczyk (approved):

src/popup/router/components/AccountInfo.vue
```diff
@@ -61,22 +61,18 @@ export default {
   },
   mixins: [CopyMixin],
   props: {
-    accountIdx: { type: Number, default: -1 },
     color: { type: String, default: 'black' },
-    showCopyIcon: { type: Boolean, default: false },
+    canCopyAddress: Boolean,
+    i: { type: Number, default: 0 },
```
peronczyk, line 66: I strongly suggest to not use one letter variables/prop names. Hard to find, refactor and know the context.

src/popup/router/components/AccountSwitcher.vue
peronczyk: I'll try to do it on my side.

src/popup/router/components/Modals/AccountDetails.vue
```diff
@@ -11,7 +11,7 @@
         <AccountInfo
           v-bind="$attrs"
```
peronczyk, line 12: I suggest to not pass all attrs to child components as this creates undiscoverable layer of abstraction. We should be aware what data we are passing.

peronczyk (changes requested):

src/popup/router/components/AccountInfo.vue
```diff
@@ -3,25 +3,25 @@
     <div class="title">
```
peronczyk, line 3: I suggest to create one computed prop (eg.: `currentAccount`) to replace all `accounts[i]`.

peronczyk (changes requested):

peronczyk (changes requested):

peronczyk (approved):

PR #1545, opened 2022-08-31
[SW-61, SW-41, SW-63, SW-39, SW-48, SW-41] More - Main Screen

ifaouibadi:


src/popup/router/components/BoxButton.vue
```diff
@@ -68,6 +68,21 @@ export default {
     }
   }
 
+  &.light {
+    &:hover {
+      background: rgba(variables.$color-white, 0.1);
+
+      &,
+      ::v-deep svg {
```
peronczyk, line 76: Please avoid styling child component's inner code (v-deep).

src/popup/router/components/Header.vue
```diff
@@ -1,15 +1,14 @@
 <template>
   <div
     v-if="showNavigation && !aeppPopup"
-    class="header"
-    :class="{ 'not-logged-in': !isLoggedIn }"
+    :class="['header', { 'not-logged-in': !isLoggedIn, 'new-ui': $route.meta.newUI }]"
```
peronczyk, line 4: I don't see any reason to combine the static CSS class names and the dynamic ones.

src/popup/router/pages/More.vue
```diff
@@ -1,60 +1,94 @@
...
 import About from '../../../icons/circle-i.svg?vue-component';
 import Buy from '../../../icons/buy.svg?vue-component';
 import Dex from '../../../icons/dex.svg?vue-component';
-import { DEX_URL, SIMPLEX_URL } from '../../utils/constants';
+import { DEX_URL, SIMPLEX_URL, BUG_REPORT_URL } from '../../utils/constants';
 
 export default {
   components: {
-    Plate, AccountSwitcher, BoxButton, Invites, Settings, About, Buy, Dex,
+    BoxButton, Invites, Settings, About, Buy, Dex, ArrowRight, ExternalLink, BugReport,
```
peronczyk, line 86: I suggest to split this line into separate ones.

src/popup/router/pages/More.vue
```diff
@@ -67,12 +101,84 @@ export default {
...
+      margin: 8px;
+      padding: 8px 12px;
+      width: auto;
+      height: 48px;
+      border-radius: 10px;
+      justify-content: space-between;
+      flex-direction: row;
+      color: rgba(variables.$color-white, 0.85);
+      font-weight: 400;
+      -webkit-transition: 0.4s;
```
peronczyk, line 116: This will be handled by autoprefixer so it's not necessary here.

src/popup/router/pages/More.vue
```diff
@@ -67,12 +101,84 @@ export default {
   margin: 0 auto;
 
   .navigation {
-    padding: 12px;
-    display: flex;
-    flex-wrap: wrap;
+    width: 100%;
 
     .box-button {
```
peronczyk, line 106: I suggest not to use class names that does not exist in the current component as we can't be 100% sure that someone else will remove/replace it in the future. And in general makes the code much harder to refactor.

src/popup/router/pages/More.vue
```diff
@@ -67,12 +101,84 @@ export default {
...
+
+      &:hover {
+        color: rgba(variables.$color-white, 1);
+      }
+
+      .box {
+        display: inline-flex;
+        align-items: center;
+
+        svg {
```
peronczyk, line 128: I suggest to use class names instead of html element names.

peronczyk (changes requested):

src/popup/router/components/Header.vue
ifaouibadi: it's better than we have duplicated property

src/popup/router/components/BoxButton.vue
peronczyk: If something is placed inside of the slot you can just use it's class. In this case `.icon` as all the icons has it.

src/popup/router/components/Header.vue
peronczyk: Let's agree to disagree as this is a subjective thing. Static class is not processed by the vue internals. Dynamic class needs to be processed. Avoiding unnecessary work in my opinion is a good thing.

Another thing is the situation where we have more static classes. It would be pain in the ass to write those as an array.

After all I'm leaving the decision to you.

src/popup/router/pages/More.vue
```diff
@@ -67,12 +121,92 @@ export default {
...
+          color: variables.$color-white;
+        }
+
+        .setting-icon {
+          margin-right: 8px;
```
peronczyk, line 154: I think it would be better to avoid setting right padding based on the icon size to align the labels. Maybe the button labels could have a big padding on the left to accommodate the space for icons and the icons would be position there with position absolute?

Or wrap all icons with a wrapper that would be little bigger and center the icons inside?

src/popup/router/pages/ResetWallet.vue
```diff
@@ -0,0 +1,22 @@
...
+  padding: 0;
+  font-size: 0.95rem;
+
+  .header {
+    padding: 0.75rem;
+    color: variables.$color-light-grey;
+    font-size: 1.05rem;
+  }
+}
+</style>
```
peronczyk, line 22: As far as I can see this component is never used.

src/popup/router/pages/TipsClaim.vue
```diff
@@ -1,5 +1,9 @@
 <template>
   <div class="tips-claim">
+    <AccountInfo
+      v-bind="$attrs"
```
peronczyk, line 4: In my opinion this case of use `v-bind="$attrs"` makes the code less readable and can lead to further problems. The `AccountInfo` components requires only 3 props and here the binding will pass all unnecessary stuff to it.

src/popup/router/pages/TipsClaim.vue
```diff
@@ -132,6 +142,10 @@ export default {
     }
   }
 
+  ::v-deep .account-info .title {
```
peronczyk, line 145: Please avoid using ::v-deep.

peronczyk (changes requested):

PR #1554, opened 2022-09-07
[SW-62] about screen ui updates

ifaouibadi:


src/popup/router/pages/About.vue
```diff
@@ -1,76 +1,190 @@
...
-    </p>
-    <hr>
-    <div class="waellet-links">
+    <div class="table">
+      <div class="table-item">
+        <div class="name">
+          {{ $t('pages.about.name') }}
+        </div>
+        <div class="value">
+          Superhero Wallet
```
CedrikNikita, line 9: `font-weight` should be 400
image
image

ifaouibadi: it's due to figma font difference with 300 it looks exact as 400

PR #1559, opened 2022-09-08
Feature/fix qr scanner

ifaouibadi:


src/popup/router/components/Modals/QrCodeReader.vue
```diff
@@ -162,23 +162,25 @@ export default {
           ['body', '#app'].forEach((s) => {
             document.querySelector(s).style = 'background: transparent';
           });
-          document.querySelector('.main').style.display = 'none';
-          document.querySelector('.transfer-send-modal').style.display = 'none'; // TODO: refactor cordova qr code reading
-          this.$store.commit('setPageTitle', this.$t('modals.qrCodeReader.scanQr'));
+          document.querySelector('.app-inner').style.display = 'none';
+
+          const closeButton = document.querySelector('#camera-close');
+          closeButton.style.display = 'block';
```
peronczyk, line 168: In my opinion here we should use the Vuex state and the close button should read the state to be displayed or not.

src/popup/router/components/Modals/QrCodeReader.vue
```diff
@@ -162,23 +162,25 @@ export default {
...
-        ['body', '#app', '.main'].forEach((s) => {
+        ['body', '#app'].forEach((s) => {
           document.querySelector(s).style = '';
         });
-        document.querySelector('.transfer-send-modal').style.display = 'flex';
-        // https://github.com/bitpay/cordova-plugin-qrscanner/issues/234
-        window.plugins.webviewcolor.change('#090909');
-        this.$store.commit('setPageTitle', '');
+        document.querySelector('.app-inner').style = '';
+        document.querySelector('#camera-close').style.display = 'none';
```
peronczyk, line 181: Also this one in my opinion should be replaced with Vuex state.

src/popup/App.vue
```diff
@@ -9,6 +9,9 @@
       'as-modal': $route.meta.asModal,
     }"
   >
+    <button id="camera-close">
```
peronczyk, line 12: I suggest to avoid using IDs.

src/popup/App.vue
```diff
@@ -167,6 +172,16 @@ body {
     box-shadow: variables.$color-border 0 0 0 1px;
   }
 
+  #camera-close {
```
peronczyk, line 175: Same here - the styling should use css classes.

peronczyk (changes requested):

peronczyk (changes requested):

PR #1565, opened 2022-09-08
[SW-152] Account details names tab

peronczyk:


src/popup/router/pages/Names/List.vue
```diff
@@ -38,8 +42,8 @@ export default {
     }),
   },
   mounted() {
-    const id = setInterval(() => this.$store.dispatch('names/fetchOwned'), 10000);
-    this.$once('hook:destroyed', () => clearInterval(id));
+    // const id = setInterval(() => this.$store.dispatch('names/fetchOwned'), 10000);
```
konradodo, line 45: if that code is not used, then delete

peronczyk: No, my fault. Need to uncomment this.

Issue #1575, opened 2022-09-08
The hidden details on confirm transaction screen


Liubov-crypto:
The transaction details (DEX related) are hidden under the buttons, I can't scroll through the details: 

![popup1](https://user-images.githubusercontent.com/69896204/189166522-3648a2a3-1c76-493f-932b-62daf4658192.png)

Tested on https://github.com/aeternity/superhero-wallet/pull/1536

PR #1534, opened 2022-08-19
[SW-18] Send modal

peronczyk:

PR #1555, opened 2022-09-07
[SW-225] Page transitions tweaks

peronczyk:

PR #1576, opened 2022-09-09
Refactoring

CedrikNikita:
- **fix**: aligning of node connection status **[SW-235]**
- **fix**: ability to set pointer, pointer fitting the screen **from [SW-152]**
- **fix**: back button behaviour on auctions screen
- **fix**: footer buttons on confimation screen
- **fix**: ability to scroll in confirm modal #1575
- **fix**: display error in `PermissionsDetails` correctly
`Invalid prop: type check failed for prop "message". Expected String, Object, got Boolean with value false....`
image
- **fix**: showing correct wording for `SeedPhraseVerifySettings`
image
- **fix(dashboard)**: set correct link for a name claim
- **fix(deeplinks)**: handle transfer deeplink correctly (missed from #1534 during rebasing)
- **fix(index)**: make terms an internal link (that would've result in not working link on release)
- **fix(transfer-receive)**: correct share link, make it backwards compatible


src/popup/router/components/Modal.vue
```diff
@@ -56,8 +56,8 @@
           class="footer"
         >
           <slot name="footer" />
-          <NodeConnectionStatus v-if="fullScreen" />
         </div>
+        <NodeConnectionStatus v-if="fullScreen" />
```
peronczyk, line 60: Maybe we could have this component only in App scope positioned above the modal?

src/popup/router/components/NameItem.vue
```diff
@@ -317,11 +317,17 @@ export default {
     .input-field {
       margin: 8px 0;
 
-      ::v-deep .wrapper {
+      ::v-deep .input-wrapper {
```
peronczyk, line 320: In general styling child component's elements is a bad thing and we should do it.

src/popup/router/components/NameItem.vue
```diff
@@ -317,11 +317,17 @@ export default {
...
         svg + .input {
           margin-left: 4px;
         }
 
+        input {
```
peronczyk, line 325: I suggest to avoid styling elements by tag names. Class name should be used.

src/popup/router/components/NodeConnectionStatus.vue
```diff
@@ -49,34 +46,19 @@ export default {
...
-  bottom: 48px;
-  bottom: calc(48px + env(safe-area-inset-bottom));
   left: 0;
   right: 0;
-  background: variables.$color-blue;
+  background: variables.$color-bg-3-new;
+  padding: 4px 0;
   color: variables.$color-white;
   text-align: center;
+  bottom: env(safe-area-inset-bottom);
```
peronczyk, line 55: In my opinion this component should not position itself. Parent components should take care on this.

src/popup/router/pages/Dashboard.vue
```diff
@@ -160,10 +165,11 @@ export default {
 @use '../../../styles/typography';
 @use '../../../styles/mixins';
 
-.account {
+.dashboard {
   display: flex;
   flex-direction: column;
   min-height: 100%;
+  background-color: variables.$color-bg-3-new;
```
peronczyk, line 172: This is not necessary. This background is set when the route has the `newUI` meta set to true. But in general I fixed the BG behaviour here: https://github.com/aeternity/superhero-wallet/pull/1555

peronczyk (changes requested):

src/popup/router/pages/PermissionsDetails.vue
```diff
@@ -20,7 +20,7 @@
     <div class="transaction-sign-limit">
       <InputField
```
peronczyk, line 21: You can also add `new-ui` prop to this input to align it with the interface design changes,

peronczyk (changes requested):

src/popup/router/pages/Dashboard.vue
CedrikNikita: image

pstachel (approved):

src/popup/router/components/Modal.vue
CedrikNikita: I did not introduce it in that PR, and i don't think we have time for now to change the approach.

src/popup/router/components/NameItem.vue
CedrikNikita: I did not introduce it in that PR, and i don't think we have time for now to change the approach.

src/popup/router/components/NodeConnectionStatus.vue
CedrikNikita: I did not introduce it in that PR, and i don't think we have time for now to change the approach.

src/popup/router/pages/PermissionsDetails.vue
CedrikNikita: Can be a separate PR

PR #1578, opened 2022-09-09
fix: account card address copy

ifaouibadi:


src/popup/router/components/AccountInfo.vue
```diff
@@ -20,7 +20,7 @@
           {{ $t('pages.account.heading') }} {{ accountIdx + 1 }}
         </div>
         <ButtonPlain
-          v-if="truncatedAddress && truncatedAddress.length"
+          v-if="canCopyAddress && truncatedAddress.length"
```
peronczyk, line 23: I already fixed this in a little bit different way. Maybe more extensive, but here event if the `canCopyAddress` is false there is still hover effect. Also on my side I aligned the visual styling of the copied layer with figma.

peronczyk (changes requested):

ifaouibadi:
fixed in #1581

PR #1579, opened 2022-09-09
Improve asset selector performance & submit textarea on enter

pstachel:


konradodo (approved):

src/popup/router/components/FormTextarea.vue
```diff
@@ -40,12 +41,20 @@ export default {
...
       type: String,
       default: null,
       validator: (val) => SIZES.includes(val),
     },
   },
+  methods: {
+    handleEnterClick(event) {
+      if (!this.enterSubmit) return;
+      this.$emit('submit');
+      event.preventDefault();
```
peronczyk, line 55: I think it's better to:
```
if (this.enterSubmit) {
    this.$emit('submit');
    event.preventDefault();
}
```

src/popup/router/components/FormTextarea.vue
```diff
@@ -70,6 +79,7 @@ export default {
     word-break: break-word;
     font-size: 15px;
     line-height: var(--base-line-height);
+    resize: none;
```
peronczyk, line 82: In my opinion this should be configurable with a bool prop.

peronczyk (changes requested):

src/popup/router/components/FormTextarea.vue
pstachel: I think otherwise, you end the function execution when it's not matching the conditions

PR #1580, opened 2022-09-09
[SW-234, SW-237] Fix account switcher

CedrikNikita:
Will not remove the whole logic behind it for two reasons:
- this might break in weird places,
- @paolomolo said that we want to return it soon


csiarab:
LGTM. New accounts link are correct and we are not limited to 9 accounts.

PR #1585, opened 2022-09-09
chore(refactor): clean new-ui temporary variable

ifaouibadi:


src/popup/router/components/NodeConnectionStatus.vue
```diff
@@ -49,7 +49,7 @@ export default {
   bottom: env(safe-area-inset-bottom);
   left: 0;
   right: 0;
-  background: variables.$color-bg-3-new;
+  background: variables.$color-bg-3;
```
peronczyk, line 52: If we want to be sure that the message has the same bg color as the page I suggest to use CSS variable: `--screen-bg-color`.

src/popup/router/components/Plate.vue
```diff
@@ -26,7 +26,7 @@
 
   &.only-top,
   .top {
-    background-color: variables.$color-bg-3-new;
+    background-color: variables.$color-bg-3;
```
peronczyk, line 29: Same here, I guess the Plate should inherit the color from current screen with the use of `--screen-bg-color`.

peronczyk (changes requested):

src/popup/router/pages/AccountDetails.vue
```diff
@@ -222,7 +222,7 @@ export default {
     gap: 16px;
     padding: 4px;
     border-radius: 14px;
-    background-color: variables.$color-bg-3;
+    background-color: var(--screen-bg-color);
```
peronczyk, line 225: This is actually wrong, because the tabs background should be darker than the screen background.

peronczyk (changes requested):

src/popup/router/pages/AccountDetails.vue
ifaouibadi: i only replaced the backgrounds that have the same variables bg-3

pstachel (approved):

paolomolo (approved):

PR #1593, opened 2022-09-09
fix: status bar background color

ifaouibadi:


config.xml
```diff
@@ -106,6 +106,7 @@
     </platform>
     <preference name="SplashScreen" value="screen" />
     <preference name="ShowSplashScreenSpinner" value="false" />
+    <preference name="StatusBarOverlaysWebView" value="falseU" />
```
subhod-i, line 109: `falseU`? Is it correct or is it supposed to be `false`?

src/popup/router/pages/AccountDetails.vue
```diff
@@ -147,6 +147,19 @@ export default {
           return null;
       }
     },
+    mobile() {
+      return process.env.PLATFORM === 'cordova';
+    },
+  },
+  mounted() {
+    if (this.mobile) {
+      window.StatusBar.backgroundColorByHexString('#191919');
```
subhod-i, line 156: A todo could be better for the color code

PR #1600, opened 2022-09-12
[SW-13] fix: modal footer fix

konradodo:


src/popup/router/components/Button.vue
```diff
@@ -247,12 +247,8 @@ export default {
     &.secondary {
```
peronczyk, line 247: Please avoid changing the buttons as I introduced major changes to them.

src/popup/router/components/Modal.vue
```diff
@@ -201,9 +201,10 @@ export default {
       padding: 8px var(--screen-padding-x);
       background:
         linear-gradient(
-          0deg,
-          rgba(var(--screen-bg-color), 0.9) 50%,
-          rgba(var(--screen-bg-color), 0) 100%
+          180deg,
+          rgba(25, 25, 25, 0) 0%,
+          rgba(25, 25, 25, 0.8) 43.08%,
+          rgba(25, 25, 25, 0.9) 90.79%
```
peronczyk, line 207: First problem I see here is the direct use of colors. We should take them from SCSS variables.
The second and bigger problem is that I tweaked this part of the Modal in my PR here: https://github.com/aeternity/superhero-wallet/pull/1601/files#diff-81f7e7dfe274e46405414664589d76fde15fe58185d5e85eb32372ad60a67bb9R206

peronczyk (changes requested):

pstachel (changes requested):

src/popup/router/components/Modals/BtnClose.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+  <button
```
peronczyk, line 2: I think this button should have the same hover effect as the BtnIcon so maybe using BtnIcon here would be better to avoid unnecessary styling?

src/styles/variables.scss
```diff
@@ -21,6 +21,9 @@ $color-silver: #cbcbcb;
 $color-light-grey: #babac0;
 $color-medium-grey: #373737;
 $color-dark-grey: #787878;
+$color-darker-grey: #161616;
+$color-darker-grey-hover: #272727;
```
peronczyk, line 25: I know that we already have the `$color-dark-grey` but I think we should stick to "from general to specific" rule sam as you did with `$color-red-dark`. So in this case:
`$color-grey-darker` and `$color-grey-darker-hover`.
And the `$color-dark-grey` could be refactored then.
What do you think?

src/styles/variables.scss
```diff
@@ -21,6 +21,9 @@ $color-silver: #cbcbcb;
...
+$color-red-dark: #ff4746;
```
peronczyk, line 26: This color is already defined in this file as `$color-danger`.

peronczyk (changes requested):

src/popup/router/components/Modals/BtnClose.vue
konradodo: close button have hover like is in the figma

pstachel (approved):

peronczyk (approved):

PR #1607, opened 2022-09-15
[SW-257] Add TypeScript and VUE Composition API support

peronczyk:


package.json
```diff
@@ -105,6 +111,7 @@
     "fs-extra": "^10.0.1",
     "sass": "^1.50.0",
     "sass-loader": "^10.2.1",
+    "typescript": "~4.3.5",
```
subhod-i, line 114: @peronczyk SDK uses typescript version 4.7.4
https://github.com/aeternity/aepp-sdk-js/blob/609555aae2d27454e7f6b90f6135d138785cd05e/package.json#L108

subhod-i (approved):

package.json
peronczyk: Not an issue. Misalignment between app TS version and the imported package dev dependency TS should not affect us. At least if it's not difference on the minor version. I tested the new version of the SDK on this branch and everything was fine.

We'll raise the TS version when moving to Vue 3.

PR #1608, opened 2022-09-15
[SW-189] Apply sticky rule for filters

pstachel:


src/popup/router/components/Filters.vue
```diff
@@ -1,8 +1,8 @@
 <template>
   <div
-    class="filters"
     data-cy="filters"
-    :class="{ fixed }"
+    :style="cssVars"
+    :class="['filters', { sticky, 'with-search': withSearch }]"
```
peronczyk, line 5: I suggest to not merge the static CSS classes with dynamic ones. I know this is tempting to have only one line if we have only one static class. But it would look really bad if there would be more static classes.

src/popup/router/components/Filters.vue
```diff
@@ -64,17 +77,18 @@ export default {
...
-  width: 100%;
+  height: 36px;
   display: flex;
   gap: 8px;
-  padding: 13px 0;
+  padding-bottom: 12px;
+  background: var(--screen-bg-color);
+  padding-inline: var(--screen-padding-x);
+  margin-left: calc(-1 * var(--screen-padding-x));
+  width: calc(100% + (2 * var(--screen-padding-x)));
```
peronczyk, line 87: Instead of this I suggest to use margin-right with the same value as margin-left.

src/popup/router/components/Filters.vue
```diff
@@ -35,15 +35,28 @@ export default {
...
   data() {
     return {
       rotatableFilters: this.filters,
     };
   },
+  computed: {
+    cssVars() {
+      return this.withSearch ? {
+        '--header-height': '116px',
+        '--account-info-height': '62px',
```
peronczyk, line 50: This makes this component dependant on the fact that the parent component is the AccountInfo. In my opinion components should be parent-agnostic.

src/popup/router/components/Filters.vue
```diff
@@ -35,15 +35,28 @@ export default {
...
+      } : {
+        '--header-height': '104px',
+        '--account-info-height': '62px',
+      };
+    },
+  },
   methods: {
     handleClick([filter, { rotated = null }]) {
+      document.querySelector('.app-inner').scrollTo({ top: 140, behavior: 'smooth' });
```
peronczyk, line 59: I strongly recommend to not make the components dependant on the parent structure. Also please avoid magic numbers.

src/popup/router/components/TransactionList.vue
```diff
@@ -1,10 +1,11 @@
 <template>
-  <div class="transaction-list">
+  <div :class="['transaction-list', { 'sticky-filter': stickyFilter }]">
```
peronczyk, line 2: IDEM

src/popup/router/pages/AccountDetailsNames.vue
```diff
@@ -43,11 +41,12 @@ export default {
 
 .account-details-names {
   .tabs {
+    position: sticky;
+    top: calc(126px + env(safe-area-inset-top));
```
peronczyk, line 45: I suggest to avoid using magic numbers.

peronczyk (changes requested):

src/popup/pages/Names/AuctionList.vue
```diff
@@ -4,7 +4,8 @@
       v-if="activeAuctions.length || auctions.length || loading"
       v-model="displayMode"
       :filters="filters"
-      :fixed="filtersFixed"
+      sticky
+      :sticky-top-height="166"
```
konradodo, line 8: magic number :D pls create variable for this

konradodo (approved):

src/popup/pages/AccountDetailsNames.vue
```diff
@@ -42,12 +40,16 @@ export default {
...
   .tabs {
+    position: sticky;
+    top: calc(var(--header-height) + env(safe-area-inset-top));
     gap: var(--gap);
-    border-radius: 14px;
     padding-top: 2px;
     background-color: var(--screen-bg-color);
-    margin-bottom: 10px;
+    padding-bottom: 10px;
+    z-index: 2;
```
peronczyk, line 52: I suggest to order the properties according to this rules: https://9elements.com/css-rule-order/

peronczyk (changes requested):

src/popup/components/Filters.vue
```diff
@@ -35,15 +36,26 @@ export default {
   props: {
     value: { type: Object, required: true },
     filters: { type: Object, required: true },
-    fixed: Boolean,
+    stickyTopHeight: { type: Number, default: 0 },
+    sticky: Boolean,
```
peronczyk, line 40: There's a rule that I red once and liked a lot, that says the component should not position itself, the parent scope should control the positioning of child elements.

So I have this crazy idea: maybe the filters should not have the `sticky` prop? Then the parent would organise the elements and position the filters (make it sticky). The benefit would be encapsulation of the CSS code that controls the sizes and the dependancies in offsets.

What do you think?

peronczyk (changes requested):

peronczyk (approved):

PR #1609, opened 2022-09-15
[SW-200] Ability to enter letters in the number field

### LINKED ISSUE #1356: In the Send / Receive/Invite input field, I can enter alphabetic values
In the Send / Receive input field, I can enter alphabetic values (in FF browser):
fff input
ff input
ff inp

In Chrome browser in the Amount field, I can enter e, E, comma, point, +, -. The rest of the characters will be blocked for input:
chrome

chr

chr inp

The same behavior on development branches.

### LINKED ISSUE #1538: The ability to enter letters in the amount field
There is a possibility to enter letters in the amount field.
Reproducible on commit f25fad6 / v.0.9.11 
let

pstachel:
fixes #1538,
fixes #1356

### DIFF
--- a/src/popup/components/InputField.vue
+++ b/src/popup/components/InputField.vue
@@ -51,6 +51,7 @@
           <input
             :id="inputId"
             v-bind="$attrs"
+            :type="type"
             class="input"
             autocomplete="off"
             step="any"
@@ -58,6 +59,7 @@
             :data-cy="$attrs.type ? `input-${$attrs.type}` : 'input'"
             :disabled="readonly"
             @input="$emit('input', $event.target.value)"
+            @keydown="checkIfNumber"
             @focusin="focused = true"
             @focusout="focused = false"
           >
@@ -100,6 +102,7 @@ export default {
   props: {
     value: { type: [String, Number], default: null },
     label: { type: String, default: '' },
+    type: { type: String, default: 'text' },
     message: {
       type: [String, Object],
       validator(value) {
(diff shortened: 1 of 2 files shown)


src/popup/router/components/InputField.vue
```diff
@@ -142,6 +143,12 @@ export default {
       return !!this.message;
     },
   },
+  methods: {
+    checkIfNumber(event) {
+      if (this.$attrs.type !== 'number' || ['Enter', 'Backspace'].includes(event.key)) return;
```
peronczyk, line 148: Please avoid using `$attrs` in the component logic as it makes the component less predictable. If we want this component to be able to receive some data it has to have the prop for this.

Also please use consistent returns within the scope of the functions. I know that early returns are tempting, but it won't work in typescript, which will be used in some time.

I think merging these two if statements can would be the best solution.

peronczyk (changes requested):

src/popup/router/components/InputField.vue
```diff
@@ -142,6 +145,13 @@ export default {
       return !!this.message;
     },
   },
+  methods: {
+    checkIfNumber(event) {
+      if (!(this.type !== 'number' || ['Enter', 'Backspace'].includes(event.key)) && !/^[0-9/-]$/.test(event.key)) {
```
peronczyk, line 150: I suggest to avoid double negations, especially as the code should target the type number fields.
Maybe this will be OK:
```
if (this.type === 'number' && !['Enter', 'Backspace'].includes(event.key) && !/^[0-9/-]$/.test(event.key)) {
```
(not tested, only suggestion)

peronczyk (changes requested):

konradodo (approved):

peronczyk (approved):

subhod-i (approved):

PR #1611, opened 2022-09-19
[SW-262] Fix permissions layout

CedrikNikita:
- `PermissionsDetails`: on permission deletion, pressing back button was resulting in 404 
- margin for `BtnMain` was missed in #1601

### DIFF
--- a/src/popup/components/RegisterName.vue
+++ b/src/popup/components/RegisterName.vue
@@ -1,9 +1,12 @@
 <template>
   <div class="register-name">
-    <p>
+    <p class="message">
       {{ msg }}
     </p>
-    <BtnMain :to="{ name: 'account-details-names-claim' }">
+    <BtnMain
+      class="btn"
+      :to="{ name: 'account-details-names-claim' }"
+    >
       {{ $t('pages.names.list.register-name') }}
     </BtnMain>
   </div>
@@ -25,12 +28,16 @@ export default {
 @use '../../styles/typography';
 
 .register-name {
-  p {
+  .message {
     @extend %face-sans-15-medium;
 
     margin: 32px;
     text-align: center;
     color: variables.$color-light-grey;
+  }
+
+  .btn {
+    margin-inline: auto;
   }
 }
 </style>
--- a/src/popup/components/buttons/BtnMain.vue
+++ b/src/popup/components/buttons/BtnMain.vue
@@ -79,6 +79,7 @@ export default defineComponent({
 
   @extend %face-sans-16-regular;
 
+  margin: 0 auto;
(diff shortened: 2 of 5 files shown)


src/popup/router/components/buttons/BtnMain.vue
```diff
@@ -73,6 +73,7 @@ export default {
 
   @extend %face-sans-16-regular;
 
+  margin: 0 auto;
```
CedrikNikita, line 76: It was `margin: 8px auto;` not sure shall we have it 8 or 0.

src/popup/router/pages/PermissionsDetails.vue
```diff
@@ -124,6 +124,7 @@ export default {
   font-size: 15px;
   text-align: left;
   color: variables.$color-white;
+  padding: 6px;
```
ifaouibadi, line 127: `padding: 16px;`
the common padding in settings screens are 16px

ifaouibadi (changes requested):

src/popup/router/pages/PermissionsDetails.vue
CedrikNikita: https://github.com/aeternity/superhero-wallet/blob/c6ebb4fcfe282e6ccccee15de9bf01860d21fa47/src/popup/pages/PermissionsSettings.vue#L50

Liubov-crypto:
LGTM.

src/popup/router/components/buttons/BtnMain.vue
peronczyk: In my opinion buttons should not have any margins on their own.
And also I really like the rule, that I red once, which says that the components should not position themselves (on their own). The parent should take care about placing the child elements in correct places. This is why I'm rather against adding x-axis margins as auto. The default behaviour of buttons in HTML is that they are inline-blocks so they align with the text.

src/popup/router/pages/PermissionsDetails.vue
peronczyk: I suggest to use `--screen-padding-x` for x-axis paddings to maintain consistency within the pages.

peronczyk (changes requested):

peronczyk (approved):

src/popup/router/pages/PermissionsDetails.vue
peronczyk: Fixed

ifaouibadi (approved):

peronczyk:
Closing the pull request as this was already done by @ifaouibadi .

Issue #1470, opened 2022-06-17
Large number of notifications goes beyond blue circle


CedrikNikita:
Maybe this should be fixed (view of bell button): 
![fix it ](https://user-images.githubusercontent.com/69896204/174297133-9052c4c8-4b33-4c9e-ba13-8b832a857432.png)

_Originally posted by @Liubov-crypto in https://github.com/aeternity/superhero-wallet/issues/1468#issuecomment-1158825069_

PR #1612, opened 2022-09-21
fix: bell icon and notification badge [#1470]

konradodo:


src/popup/components/Header.vue
```diff
@@ -312,17 +314,19 @@ export default {
...
       height: 14px;
-      background: variables.$color-blue;
-      border-radius: 50%;
+      background: variables.$color-pink;
+      border-radius: 7px;
       text-align: center;
-
-      @extend %face-sans-12-regular;
+      line-height: 14px;
+      padding: 0 3.5px;
```
pstachel, line 329: I think there is no thing as half of a pixel

pstachel (changes requested):
The icon shouldn't have width & height as well as the color should be set to `currentColor`

src/popup/components/Header.vue
```diff
@@ -312,17 +314,19 @@ export default {
     position: relative;
 
     .badge {
```
peronczyk, line 316: I think we should have separate component for the badges.

peronczyk (changes requested):

src/popup/components/Header.vue
```diff
@@ -120,9 +120,11 @@ export default {
       return this.$route.meta.hideNotificationsIcon;
     },
     notificationsCount() {
-      return [...this.notifications, ...this.superheroNotifications].filter(
+      const count = [...this.notifications, ...this.superheroNotifications].filter(
```
pstachel, line 123: `this.notifications` and `this.superheroNotifications` may be undefined

pstachel (changes requested):

pstachel (approved):

peronczyk (approved):

PR #1614, opened 2022-09-24
[SW-67, SW-263, SW-290, SW-289, SW-301] feat: add/edit permission ui updates

ifaouibadi:


src/popup/components/InputField.vue
```diff
@@ -28,10 +28,14 @@
...
       <div
-        v-if="$slots['label-after']"
+        v-if="$slots['label-after'] || textLimit"
         class="label-after"
       >
-        <slot name="label-after" />
+        <slot name="label-after">
+          {{
+            (value && value.length) ? textLimit - value.length : textLimit
+          }}
```
pstachel, line 37: I think it should be in computed

src/popup/components/InputField.vue
```diff
@@ -58,6 +62,7 @@
             :value="value"
             :data-cy="$attrs.type ? `input-${$attrs.type}` : 'input'"
             :disabled="readonly"
+            :maxlength="textLimit && textLimit"
```
pstachel, line 65: why?

src/popup/components/PermissionManager.vue
```diff
@@ -0,0 +1,224 @@
...
+
+    <div
+      v-if="permission.dailySpendLimit"
+      class="transaction-sign-limit"
+    >
+      <InputField
+        v-model="permission.transactionSignLimit"
+        :message="
+          (Number.isNaN(permission.transactionSignLimit) || permission.transactionSignLimit < 0)
+            ? { status: 'error' } : null
```
pstachel, line 49: I'd suggest putting it into computed

src/popup/components/PermissionManager.vue
```diff
@@ -0,0 +1,224 @@
...
+        "
+        :label="$t('pages.permissions.transaction-sign')"
+        placeholder="always ask"
```
pstachel, line 52: That should be in en.json file

pstachel (changes requested):

peronczyk (changes requested):

src/popup/components/InputField.vue
```diff
@@ -143,6 +150,9 @@ export default {
       }
       return !!this.message;
     },
+    availableTextLimit() {
+      return this.value?.length ? this.textLimit - this.value.length : this.textLimit;
```
peronczyk, line 154: There is a possibility of type mismatching if the `textLimit` prop is not provided. In this case JS will try to subtract the text length from `null`. This is causing any JS error, but can be considered as a bad practice.

src/popup/components/InputField.vue
```diff
@@ -203,6 +213,9 @@ export default {
 
     &-after {
       margin-left: auto;
+      color: variables.$color-dark-grey;
+
+      @extend %face-sans-15-regular;
```
peronczyk, line 218: Please put the @extend / @include at the beginning.

peronczyk (changes requested):

src/popup/pages/PermissionManager.vue
```diff
@@ -0,0 +1,360 @@
...
+        v-model="permission.dailySpendLimit"
+        :label="$t('pages.permissions.daily-spending-limit')"
+      />
+    </div>
+
+    <div
+      v-if="permission.dailySpendLimit"
+      class="transaction-sign-limit"
+    >
+      <InputField
```
peronczyk: This can be replaced with the InputAmountV2 component.

src/popup/pages/PermissionManager.vue
```diff
@@ -0,0 +1,323 @@
...
+    this.$watch('permission', (permission) => {
+      this.permissionChanged = (!!permission.name && !!permission.host);
+    }, { deep: true });
+  },
+  methods: {
+    removePermissions() {
+      this.$store.commit('permissions/removePermission', this.host);
+      this.$router.push({ name: 'permissions-settings' });
+    },
+    onSavePermission() {
```
peronczyk, line 213: The `removePermissions` is plural. This `onSavePermission` is singular. I suggest to stick with one form.

peronczyk (changes requested):

src/popup/components/SelectAsset.vue
```diff
@@ -72,11 +75,11 @@ export default {
...
 
-  &:hover {
+  &.enabled:hover {
     border-color: rgba(variables.$color-white, 0.15);
   }
 
-  &.focused {
+  &.enabled.focused {
     background-color: rgba(variables.$color-white, 0.05);
   }
```
peronczyk, line 84: This lines can be merged:
```
&.enabled {
  &:hover {}
  &:focused {}
}
```

src/popup/components/SelectAsset.vue
```diff
@@ -21,6 +21,7 @@ export default {
   },
   props: {
     value: { type: Object, default: null },
+    canSelectAsset: { type: Boolean, default: true },
```
peronczyk, line 24: In case of the boolean props I suggest to not use the `true` as the default as it looks bad in the templates when we want to disable something. I suggest to inverse the logic. So maybe `disabled` prop would be better?

src/popup/components/InputField.vue
```diff
@@ -143,6 +150,10 @@ export default {
       }
       return !!this.message;
     },
+    availableTextLimit() {
+      return (this.textLimit && this.value?.length)
+        ? this.textLimit - this.value.length : this.textLimit;
```
peronczyk, line 155: Maybe we can agree on one approach of writing one line conditionals?
For me personally it's way better to follow with eyes the conditions written in following form:
```
return (someTest)
  ? firstResult
  : secondResult;
```
It's like a poll: question on the first line and the possible answers on separate lines.

peronczyk (changes requested):

src/popup/components/SelectAsset.vue
```diff
@@ -1,11 +1,11 @@
 <template>
   <BtnPlain
     class="select-asset"
-    :class="{ focused }"
+    :class="{ focused, enabled: !disabled }"
```
peronczyk, line 4: Maybe inverting the logic would be better to do on the CSS level. So here we'll have only `{ focused, disabled }` and in the css:
```
:not(.disabled) {
```

src/popup/components/SelectAsset.vue
```diff
@@ -35,6 +36,8 @@ export default {
       this.$emit('input', token);
     },
     openAssetSelector() {
+      if (!this.canSelectAsset) return;
```
peronczyk, line 39: I guess `canSelectAsset` is now `disabled`.

peronczyk (changes requested):

src/popup/pages/PermissionsSettings.vue
```diff
@@ -61,6 +69,23 @@ export default {
...
+      margin-top: 26px;
+      width: 100%;
+
+      @extend %face-sans-16-regular;
+
+      .icon {
+        width: 20px;
+        height: 20px;
+        margin-right: 4px;
+      }
```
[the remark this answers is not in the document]
ifaouibadi: Updated, but i had to addthe  ::v-deep to BtnMain cause has-icon style didn’t apply style to a link button, cause it will be RouterLink

src/popup/pages/PermissionManager.vue
```diff
@@ -0,0 +1,296 @@
...
+        this.permission = {
+          ...this.$store.state.permissions[this.host],
+          host: this.host,
+        };
+      }
+    }
+
+    this.$watch('permission', (permission) => {
+      this.permissionChanged = (!!permission.name && !!permission.host);
+    }, { deep: true });
```
pstachel, line 209:
```suggestion
watch: {
  permission: {
    (permission) => {
       this.permissionChanged = (!!permission.name && !!permission.host);
     }
   }, {
      deep: true,
      immediate: true,
   });
},
```

pstachel (changes requested):

src/popup/pages/Popups/Connect.vue
```diff
@@ -80,6 +88,19 @@ export default {
...
   },
+  methods: {
+    confirm() {
+      this.$store.commit('permissions/addPermission', {
+        address: false,
+        messageSign: false,
+        transactionSignLimit: 0,
+        transactionSignLimitLeft: 0,
+        transactionSignFirstAskedOn: null,
+        ...{ ...this.app },
```
peronczyk, line 99: I guess there is no need to destructure `this.app` twice.

src/store/modules/permissions.js
```diff
@@ -19,9 +21,7 @@ export class App {
 export default {
   namespaced: true,
 
-  state: Object.fromEntries(
-    ['superhero.com', 'localhost'].map((domain) => [domain, hostConfig([], true)]),
-  ),
+  state: {},
 
   mutations: {
     togglePermission(state, { host, name }) {
```
CedrikNikita, line 27: I think we should also go with `Vue.set` in `togglePermission`.

peronczyk (changes requested):

src/popup/pages/PermissionManager.vue
```diff
@@ -0,0 +1,302 @@
...
+    </div>
+
+    <div class="permission-row switch">
+      <SwitchButton
+        v-model="permission.dailySpendLimit"
+        :label="$t('pages.permissions.daily-spending-limit')"
+      />
+    </div>
+
+    <div
```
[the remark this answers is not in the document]
ifaouibadi: We have an upcoming task for transitions [SW-69]

peronczyk (approved):

src/icons/delete.svg
```diff
@@ -0,0 +1,3 @@
+<svg viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
pstachel, line 1: We have that icon already - check `trash.svg`

pstachel (changes requested):

pstachel (approved):

PR #1616, opened 2022-09-26
fix: buy link fix [SW-268]

konradodo:


src/popup/pages/FungibleTokens/TokenDetails.vue
```diff
@@ -287,10 +286,16 @@ export default {
...
-    ...mapGetters(['formatCurrency', 'formatNumber', 'accounts']),
+    ...mapGetters(['formatCurrency', 'formatNumber', 'accounts', 'account']),
     ...mapState('accounts', ['activeIdx']),
     ...mapState('fungibleTokens', ['aePublicData', 'availableTokens']),
     ...mapGetters('fungibleTokens', ['tokenBalances']),
+    getSimplexLink() {
+      const link = new URL(SIMPLEX_URL);
+      link.searchParams.set('wallet_address', this.account.address);
+      return link.toString();
+    },
```
pstachel, line 297: I think that method should be extracted somewhere else to not repeat it every time we want to use it

src/popup/pages/Dashboard.vue
```diff
@@ -121,12 +121,16 @@ export default {
...
     buyBackground,
     chainNameBackground,
-    SIMPLEX_URL,
   }),
   computed: {
     ...mapState(['backedUpSeed', 'transactions']),
-    ...mapGetters(['getAccountPendingTransactions']),
+    ...mapGetters(['getAccountPendingTransactions', 'account']),
     ...mapState('accounts', ['activeIdx']),
+    getSimplexLink() {
```
pstachel, line 129: Computed is not a function but a property, so it shouldn't be named `getSimplexLink` but `simplexLink`

pstachel (changes requested):

pstachel (approved):

PR #1620, opened 2022-09-27
[SW-269]: Fix viewport on android

peronczyk:


src/popup/App.vue
```diff
@@ -37,10 +38,14 @@
 
 <script>
 import { mapGetters, mapState } from 'vuex';
-import { detect } from 'detect-browser';
```
CedrikNikita, line 40: If you think that `IS_FIREFOX` variable can detect firefox browser, i suggest to replace `detect` function everywhere (also mock of it from snapshot test) and remove `detect-browser` from wallet dependencies.

peronczyk: In most of the places the detect-browser module was replaced with the constants. But there is this logger.js file where we couldn't replace the `detect` method so easilly. I created a spike for this: https://aeternity-blockchain.atlassian.net/browse/SW-356

pstachel (approved):

PR #1623, opened 2022-09-29
[SW-286, SW-297, SW-323] Limit amount input to numbers and "."

pstachel:


peronczyk (changes requested):

konradodo (approved):

src/popup/components/InputField.vue
```diff
@@ -158,11 +158,20 @@ export default {
     },
   },
   methods: {
+    checkModifierKey(event) {
```
peronczyk, line 161: In my opinion there is no need to create separate method for this logic as it's used only once.

src/popup/components/InputField.vue
```diff
@@ -158,11 +158,20 @@ export default {
...
+      return ['ctrlKey', 'metaKey'].some((key) => event[key]);
+    },
     checkIfNumber(event) {
-      if (this.type === 'number' && !['Enter', 'Backspace'].includes(event.key) && !/^[0-9/-]$/.test(event.key)) {
+      if (this.type !== 'number' || event.key.length > 1 || this.checkModifierKey(event)) return;
+      if ((this.value?.includes('.') && [',', '.'].includes(event.key)) || !/^([0-9]+|,|\.)$/.test(event.key)) {
```
peronczyk, line 166: This should be `else if`.

src/popup/components/InputAmountV2.vue
```diff
@@ -4,6 +4,7 @@
     class="input-amount"
     type="number"
     placeholder="0.00"
+    inputmode="decimal"
```
peronczyk, line 7: In my opinion this should be moved into InputField component and be based on input type.

src/popup/components/InputAmount.vue
```diff
@@ -11,6 +11,7 @@
     :message="$attrs['message'] || errors.first('amount')"
     class="input-amount"
     type="number"
+    inputmode="decimal"
```
peronczyk, line 14: In my opinion this should be moved into InputField component and be based on input type.

peronczyk (changes requested):

src/popup/components/InputField.vue
```diff
@@ -159,10 +163,16 @@ export default {
   },
   methods: {
     checkIfNumber(event) {
-      if (this.type === 'number' && !['Enter', 'Backspace'].includes(event.key) && !/^[0-9/-]$/.test(event.key)) {
+      if (!(this.type !== 'number' || event.key.length > 1 || ['ctrlKey', 'metaKey'].some((key) => event[key]))
```
peronczyk, line 166: With the multiline if statements I suggest to do the same approach as with the object properties. So we are able to do this:
```
{ a: 1, b: 2, c: 3 }
```
and this:
```
{
  a: 1,
  b: 2,
  c: 3
}
```
Nothing in between, so this is wrong:
```
{ a: 1, b: 2,
c: 3 }
```
I think the same rule should be applied to if statements. So I suggest to do it as follows:
```
if (
  (test1 || test2 || test3)
  && (test4 || test5)
  && test6
) {
  ...
}
```

peronczyk (changes requested):

peronczyk (approved):

PR #1629, opened 2022-10-03
Added testcases for Tokens & TotalAmount Component

NitinRamnani:
I've added testcases for Tokens & TotalAmount Components which have increased code coverage of those components.

Before PR:

Before PR

After PR:

After PR

After PR -2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ tests/e2e/screenshots
 .DS_Store
 /tests/e2e/videos/
 /tests/e2e/screenshots/
+/coverage
 
 # local env files
 .env.local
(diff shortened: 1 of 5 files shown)


ifaouibadi (approved):

NitinRamnani:
Fixed the snapshot testcases to compare HTML instead of vm of the component.

CedrikNikita:
In general, I would skip matching the snapshot and would rather test the logic of the component, because:
1) if we will change the class name the snapshot matching would fail;
2) we would need to update it none structure based tests less often

subhod-i:
@NitinRamnani There are few suggestions, would you work on fixing it? We can merge this once the PR is approved.

PR #1633, opened 2022-10-09
[SW-288] Fix top navigation bar

konradodo:


src/popup/components/Header.vue
```diff
@@ -3,66 +3,68 @@
...
+        >
+          <Logo />
+        </Component>
+        <BtnPlain
+          v-if="showBack"
+          class="icon-btn"
+          @click="back"
+        >
+          <Back data-cy="back-arrow" />
+        </BtnPlain>
```
pstachel, line 25: You should use BtnIcon for it

src/popup/components/Header.vue
```diff
@@ -3,66 +3,68 @@
...
-
-      <RouterLink
-        v-if="$route.path !== '/more' && !$route.meta.closeButton"
-        class="icon-btn"
-        to="/more"
-        data-cy="page-more"
+        <Truncate
+          :str="pageTitle ?
+            pageTitle :
+            title ? $t(`pages.titles.${title}`) : ''"
```
pstachel, line 34: I'd suggest moving it into computed

pstachel (changes requested):

src/icons/logo-header.svg
```diff
@@ -0,0 +1,10 @@
+<svg width="33" height="32" viewBox="0 0 33 32" fill="none" xmlns="http://www.w3.org/2000/svg">
```
peronczyk, line 1: After some testing I think adding new logo file makes new sensy as the old one was OK. The size difference could be fixed by changing CSS.

peronczyk (changes requested):

src/popup/components/Header.vue
konradodo: @peronczyk so i leave it now as agreed.

src/popup/components/Header.vue
```diff
@@ -222,8 +226,6 @@ export default {
     }
 
     svg {
```
peronczyk, line 228: This styles can be moved to `.home-icon`.

peronczyk (changes requested):

pstachel (approved):

peronczyk (approved):

PR #1634, opened 2022-10-11
[SW-266] Buttons to close to bottom edge on popups

pstachel:


src/popup/components/Modal.vue
```diff
@@ -111,6 +111,8 @@ export default {
 .modal {
   --screen-padding-x: 24px; // Default spacing
   --screen-bg-color: #{variables.$color-bg-modal};
+  --from-bottom-button-bottom-padding: 25px;
+  --normal-button-padding: 48px;
```
peronczyk, line 115: I have few problems with these 2 lines:
1. naming convention: these paddings are not a button paddings. They are applied to footer. And in general it's better to name things starting from the more generic word and end with most specific. Eg.: `--footer-padding-bottom`.
2. this approach on having 2 separate CSS variables is a rather bad pattern. It is better to have only one variable set here with its default value, then used in line 202, and changed in the modal variants.
3. 48px is a value that should not be used at any situation regarding the bottom space of the .footer. 48px is a space between the buttons and bottom edge in case don't put the buttons into the footer, so the buttons won't be sticky. So maybe having this CSS variable is totally obsolete?

Example:
```
--footer-padding-bottom: 24px;

.footer {
  padding: 8px var(--screen-padding-x) var(--footer-padding-bottom);
}

&.some-variant {
  --footer-padding-bottom: 666px;
}
```

src/popup/components/Modals/ErrorLog.vue
```diff
@@ -14,7 +14,6 @@
       <span>{{ $t('modals.error-log.sub-title') }}</span>
       {{ $t('modals.error-log.content') }}
     </div>
-
```
peronczyk, line 17: Why removing this space? It's useful to make the code easier to view and analyse.

src/popup/components/Modals/Default.vue
```diff
@@ -23,7 +23,6 @@
     </slot>
 
     <slot />
-
```
peronczyk, line 26: Why removing this space? It's useful to make the code easier to view and analyse.

src/popup/components/Modals/AccountImport.vue
```diff
@@ -22,16 +22,18 @@
...
-      :disabled="!mnemonic || (error != null)"
-      data-cy="import"
-      class="import-button"
-      extend
-      center
-      @click="importAccount"
-    >
-      {{ $t('pages.index.importAccount') }}
-    </BtnMain>
+    <template #footer>
```
peronczyk, line 25: Maybe adding empty line before the template would make the HTML more easy to view.

src/popup/components/Modals/QrCodeReader.vue
```diff
@@ -221,11 +220,11 @@ export default {
...
-  .button-wrapper {
-    padding-top: 20px;
-    padding-bottom: 0;
-    width: 100%;
-  }
+  //.button-wrapper {
+  //  padding-top: 20px;
+  //  padding-bottom: 0;
+  //  width: 100%;
+  //}
```
peronczyk, line 227: To be removed?

peronczyk (changes requested):

src/popup/components/Modals/AccountImport.vue
```diff
@@ -22,16 +22,19 @@
...
-      extend
-      center
-      @click="importAccount"
-    >
-      {{ $t('pages.index.importAccount') }}
-    </BtnMain>
+
+    <template #footer>
+      <BtnMain
+        :disabled="!mnemonic || (error != null)"
```
konradodo, line 28: :disabled="!mnemonic && error"

konradodo (approved):

peronczyk (changes requested):

peronczyk (approved):

PR #1636, opened 2022-10-13
chore: upgrade cordova android sdk 32

ifaouibadi:


package.json
```diff
@@ -186,4 +202,4 @@
       }
     ]
   }
-}
+}
```
CedrikNikita, line 205: Obviously, should have a newline in the end of the file.

CedrikNikita:
The `cordova-plugin-qrscanner` is breaking for me and Keno, when trying to build this on Android.

PR #1637, opened 2022-10-13
[SW-275] Fix app and modal rounded corners

peronczyk:
Mostly related to the chrome extension.

### DIFF
--- a/src/popup/App.vue
+++ b/src/popup/App.vue
@@ -3,7 +3,8 @@
     id="app"
     :class="{
       'show-header': showStatusAndHeader,
-      'display-as-mobile-app': displayAsMobileApp,
+      'is-web': IS_WEB,
+      'is-extension': IS_EXTENSION,
     }"
   >
     <button
@@ -41,6 +42,7 @@ import { mapGetters, mapState } from 'vuex';
 import { watchUntilTruthy } from './utils/helper';
 import { NOTIFICATION_DEFAULT_SETTINGS } from './utils/constants';
 import {
+  IS_WEB,
   IS_IOS,
   IS_ANDROID,
   IS_CORDOVA,
@@ -56,6 +58,10 @@ export default {
     NodeConnectionStatus,
     Close,
   },
+  data: () => ({
+    IS_WEB,
+    IS_EXTENSION,
+  }),
   computed: {
     ...mapGetters(['isLoggedIn']),
     ...mapState(['isRestored', 'backedUpSeed', 'qrScannerOpen']),
@@ -69,9 +75,6 @@ export default {
     },
     modals() {
       return this.$store.getters['modals/opened'];
-    },
-    displayAsMobileApp() {
-      return !IS_ANDROID;
     },
   },
   watch: {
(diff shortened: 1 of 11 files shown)


src/lib/environment.ts
```diff
@@ -20,6 +22,10 @@ export const IS_ANDROID = !!(userAgentLowerCase.includes('android')
 
 export const IS_MOBILE_DEVICE = userAgentLowerCase.includes('mobi');
 
+export const IS_CHROME = userAgentLowerCase.includes('chrome');
```
pstachel, line 25: what about Chrome based browsers?

pstachel (changes requested):

src/lib/environment.ts
peronczyk: I renamed the constant to be more precise. Is this OK in your opinion?

pstachel (approved):

src/popup/App.vue
```diff
@@ -3,7 +3,8 @@
     id="app"
     :class="{
       'show-header': showStatusAndHeader,
-      'display-as-mobile-app': displayAsMobileApp,
```
CedrikNikita, line 6: Computed property `displayAsMobileApp` can be removed, because it is no longer used in that file.

CedrikNikita (changes requested):

src/popup/App.vue
peronczyk: Done

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -3,13 +3,17 @@
     full-screen
     class="confirm-transaction-sign"
     data-cy="popup-aex2"
-    absolute-footer
   >
-    <TransactionOverview :tx="completeTransaction" />
+    <TransactionOverview
+      class="transaction-overview"
```
CedrikNikita, line 8: But transaction overview component already have a `transaction-overview` class.

src/popup/App.vue
CedrikNikita: It is not, yet.

src/popup/components/Modals/ConfirmTransactionSign.vue
peronczyk: It doesn't matter. If we are styling some class name within the <style> block the class should exist in the scope of the <template> block. So we should avoid styling stuff out of the current context. This is related to the "encapsulation" rules.

When I joined the team I've seen a lot of places where the parent component was trying to apply some styling on class names that should exists in the child scope, but someone changed the name or the structure of the child component so at the end it was messed up.

In general parent components should be blind for the child component's internals.

src/popup/App.vue
peronczyk: Ups, yes - you're right. Now I removed it.

src/popup/components/Modals/ConfirmTransactionSign.vue
CedrikNikita: - Now all code is consistent
- With that approach we would still need to rename classes names in case we would rename the component to at least match the logic
- In that case we would need to rewrite the whole project in that to be consistent

Would this approach will bring us more value than it would took effort?

peronczyk: I removed my change from the PR on rounded corners. There is an open conversation on how we should behave. But I think it should not block the task.

CedrikNikita (approved):
Code changes look reasonable.

PR #1646, opened 2022-10-18
[SW-342] terms privacy scrollable

ifaouibadi:


src/popup/components/Header.vue
```diff
@@ -248,6 +248,11 @@ export default {
     margin-right: 8px;
   }
 
+  .btn-close {
```
peronczyk, line 251: This class does not exist in the scope of the component. If you want to style something within the scope of the component you need to add additional class to this element.

peronczyk (changes requested):

src/popup/components/Scrollable.vue
```diff
@@ -1,11 +1,22 @@
 <template>
   <div class="scrollable">
-    <div class="scrollable-inner">
+    <div
+      class="scrollable-inner"
+      :style="{'scrollbar-width': scrollbarWidth}"
```
peronczyk, line 5: CSS scrollbar-width is not supported in most of the browsers: https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width#browser_compatibility

src/popup/pages/PrivacyPolicy.vue
```diff
@@ -1,351 +1,440 @@
 <template>
   <div class="privacy-policy">
-    <p><em>The present Privacy Policy is effective since on 9th April 2020.</em></p>
-    <ol class="policy-list">
-      <li class="first">
-        <h3><strong>Introduction</strong></h3>
+    <Scrollable scrollbar-width="auto">
```
peronczyk, line 3: I strongly recommend to not use the `Scrollable` component for the purpose of making the page scroll visible. The default scrollable area of the app is the `app-inner` located in App.vue.

In my opinion the page scroll should be controlled by the route meta settings as discussed on one of the meetings.

Using custom scrollable area will allow to scroll part of the screen, but the content would not be scrolled under the header, which soon will become semi-transparent.

src/popup/pages/TermsOfService.vue
```diff
@@ -1,943 +1,1149 @@
...
-        <p>
-          1.3. &ldquo;<strong>Blockchain</strong>&ldquo; means the list of records, also known as
-          <em>blocks</em>, which are linked with the method of cryptography. The technology allows
-          large groups of people to connect and reach agreement on data, which is afterwards
-          permanently recorded without the presence of central authority. <strong>USERS</strong> can
-          access, create and save information on the <strong>Aeternity blockchain</strong>. The
-          decentralized, distributed and public digital ledger is used for the purpose of recording
-          transactions in blocks of information with the help of <strong>Validators</strong>, which
-          cannot be altered retroactively.
+    <Scrollable scrollbar-width="auto">
```
peronczyk, line 3: IDEM

peronczyk (changes requested):

src/popup/router/routes.js
```diff
@@ -327,6 +327,7 @@ export default [
       closeButton: true,
       hideNotificationsIcon: true,
       ifNotAuth: true,
+      scrollable: true,
```
peronczyk, line 330: All pages are scrollable. This is about making the scrollbar visible. So maybe `displayScrollbar` or `scrollbarVisible` would be better here?

src/popup/App.vue
```diff
@@ -18,12 +18,16 @@
...
 
-      <transition :name="$route.meta.asModal ? 'pop-transition' : 'page-transition'">
-        <RouterView
-          :class="{ 'show-header': showStatusAndHeader }"
-          class="main"
-        />
-      </transition>
+      <div :class="['app-container', {scrollable: $route.meta.scrollable}]">
+        <transition :name="$route.meta.asModal ? 'pop-transition' : 'page-transition'">
+          <Scrollable :hide-scrollbar="!$route.meta.scrollable">
```
peronczyk, line 23: The "Scrollable" component should not be used to wrap the whole page in order to allow to scroll it. The whole `app-inner` should scroll as in the future the header will be semi-transparent and user should see the content under the header.

src/popup/components/Modals/TransferReceive.vue
```diff
@@ -31,7 +31,10 @@
           :copied="copied"
           @click="copyAddress()"
         >
-          <Scrollable class="address-scrollable-area">
+          <Scrollable
+            class="address-scrollable-area"
+            hide-scrollbar
```
peronczyk, line 36: On figma designs this place should have the scrollbar visible.

peronczyk (changes requested):

src/popup/components/Modals/TransferReceive.vue
ifaouibadi: https://www.figma.com/file/3oGLWzSH0oJljo4RETZtur/Superhero-Wallet-UI-(%E2%9C%94%EF%B8%8FUpdated)?node-id=11533%3A143300

src/popup/App.vue
ifaouibadi: if we keep `app-inner` scrollable the scrollbar will be cut off by the header see [Liubov Comments](https://aeternity-blockchain.atlassian.net/browse/SW-342?focusedCommentId=11514)

src/popup/components/Scrollable.vue
```diff
@@ -1,11 +1,22 @@
 <template>
   <div class="scrollable">
-    <div class="scrollable-inner">
+    <div
+      class="scrollable-inner"
+      :class="{'hide-scrollbar': hideScrollbar}"
```
pstachel, line 5: please add spaces 
```suggestion
      :class="{ 'hide-scrollbar': hideScrollbar }"
```

pstachel (changes requested):

src/popup/App.vue
```diff
@@ -19,12 +19,16 @@
...
 
-      <transition :name="$route.meta.asModal ? 'pop-transition' : 'page-transition'">
-        <RouterView
-          :class="{ 'show-header': showHeader }"
-          class="main"
-        />
-      </transition>
+      <div class="app-container">
+        <transition :name="$route.meta.asModal ? 'pop-transition' : 'page-transition'">
+          <Scrollable :hide-scrollbar="!$route.meta.scrollbarVisible">
```
peronczyk, line 24: In my opinion the `<Scrollable>` component should not be used to scroll whole page. It was created to scroll parts of the screen.

src/popup/App.vue
```diff
@@ -19,12 +19,16 @@
...
       <Header v-if="showHeader" />
 
-      <transition :name="$route.meta.asModal ? 'pop-transition' : 'page-transition'">
-        <RouterView
-          :class="{ 'show-header': showHeader }"
-          class="main"
-        />
-      </transition>
+      <div class="app-container">
+        <transition :name="$route.meta.asModal ? 'pop-transition' : 'page-transition'">
```
peronczyk, line 23: Transition component should be put immediately above the element that changes in order to apply the transition. The structure has changed and there is no page transition at all.

src/popup/App.vue
```diff
@@ -19,12 +19,16 @@
     >
       <Header v-if="showHeader" />
 
-      <transition :name="$route.meta.asModal ? 'pop-transition' : 'page-transition'">
-        <RouterView
-          :class="{ 'show-header': showHeader }"
-          class="main"
-        />
-      </transition>
+      <div class="app-container">
```
peronczyk, line 22: I don't quite get the reason of another element added to wrap the content. The CSS for this element is only width and height set to 100%. Maybe we could remove it?

src/popup/components/Filters.vue
```diff
@@ -43,7 +43,7 @@ export default {
   },
   methods: {
     handleClick([filter, { rotated = null }]) {
-      const appContainer = document.querySelector('.app-inner');
+      const appContainer = document.querySelector('.scrollable-inner');
```
peronczyk, line 46: I think this is a terrible idea to match a class name of a reusable element. This can lead to major problems. I strongly suggest to use different solution for this.

src/popup/components/TransactionList.vue
```diff
@@ -157,7 +157,7 @@ export default {
   mounted() {
     this.loadMore();
     const polling = setInterval(() => this.getLatest(), 10000);
-    const appInner = document.querySelector('.app-inner');
+    const appInner = document.querySelector('.scrollable-inner');
```
peronczyk, line 160: Same here - I really discourage you from using reusable element's classes.

src/popup/components/TransactionList.vue
```diff
@@ -174,7 +174,7 @@ export default {
       if (this.isDestroyed || !this.transactions.nextPageUrl) return;
       const isDesktop = document.documentElement.clientWidth > 480 || process.env.IS_EXTENSION;
       const { scrollHeight, scrollTop, clientHeight } = isDesktop
-        ? document.querySelector('.app-inner') : document.documentElement;
+        ? document.querySelector('.scrollable-inner') : document.documentElement;
```
peronczyk, line 177: IDEM

src/popup/router/index.js
```diff
@@ -62,16 +62,16 @@ router.beforeEach(async (to, from, next) => {
...
 
   if (to.fullPath === '/transactions' && savedScrollPosition) {
     setTimeout(() => {
-      (process.env.IS_CORDOVA ? document.scrollingElement : document.querySelector('.app-inner')).scroll(0, savedScrollPosition);
+      (process.env.IS_CORDOVA ? document.scrollingElement : document.querySelector('.scrollable-inner')).scroll(0, savedScrollPosition);
       savedScrollPosition = 0;
     }, 50);
   } else {
-    (process.env.IS_CORDOVA ? document.scrollingElement : document.querySelector('.app-inner')).scroll(0, 0);
+    (process.env.IS_CORDOVA ? document.scrollingElement : document.querySelector('.scrollable-inner')).scroll(0, 0);
```
peronczyk, line 74: IDEM

src/popup/components/Modals/TransferReceive.vue
peronczyk: I asked Tsvetan about this and he confirmed what I stated.

src/popup/App.vue
peronczyk, line 22: Previously when user was scrolling the content it was going under the header. There is an idea of making the header semi-transparent to allow user to see the content that goes under the header. Following changes changed this behaviour.

src/popup/App.vue
```diff
@@ -222,6 +239,10 @@ export default defineComponent({
     box-shadow: variables.$color-border 0 0 0 1px;
     transform: translate(0, 0); // Create custom viewport for fixed elements
 
+    @include mixins.desktop {
+      border-radius: variables.$border-radius-app;
```
peronczyk, line 243: The `#app` already has the border-radius. Is this necessary to add another one? And if yes I suggest to use the CSS variable `--screen-border-radius` or just inherit it.

peronczyk (approved):

pstachel (approved):

PR #1648, opened 2022-10-18
[SW-271] added superhero logo to qr code generator

ifaouibadi:


src/popup/components/QrCode.vue
```diff
@@ -0,0 +1,81 @@
...
+      cornersSquareOptions: { type: 'extra-rounded', color: '#000000' },
+      cornersSquareOptionsHelper: {
+        colorType: { single: true, gradient: false },
+        gradient: {
+          linear: true,
+          radial: false,
+          color1: '#000000',
+          color2: '#000000',
+          rotation: '0',
+        },
```
peronczyk, line 52: This can be removed

src/popup/components/QrCode.vue
```diff
@@ -0,0 +1,81 @@
...
+      cornersDotOptions: { type: 'dot', color: '#000000' },
+      cornersDotOptionsHelper: {
+        colorType: { single: true, gradient: false },
+        gradient: {
+          linear: true,
+          radial: false,
+          color1: '#000000',
+          color2: '#000000',
+          rotation: '0',
+        },
```
peronczyk, line 63: This can be removed

src/popup/components/QrCode.vue
```diff
@@ -0,0 +1,81 @@
...
+      },
+      backgroundOptionsHelper: {
+        colorType: { single: true, gradient: false },
+        gradient: {
+          linear: true,
+          radial: false,
+          color1: '#ffffff',
+          color2: '#ffffff',
+          rotation: '0',
+        },
```
peronczyk, line 73: This can be removed

src/popup/components/QrCode.vue
```diff
@@ -0,0 +1,81 @@
...
+
+import SHLogo from '../../icons/logo-smal-blue.png';
+
+export default {
+  props: {
+    value: { type: String, required: true },
+    size: { type: Number, required: true },
+  },
+  mounted() {
+    this.qrCode = new QRCodeStyling({
```
peronczyk, line 15: There's not `qrCode` set in the `data` section.

src/popup/components/QrCode.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+  <div id="qr-canvas" />
+</template>
+<script>
```
peronczyk, line 4: Please add empty line between the template and the script.

src/popup/components/QrCode.vue
```diff
@@ -0,0 +1,81 @@
...
+import QRCodeStyling from '@solana/qr-code-styling';
+
```
peronczyk, line 6: This line can be removed.

src/popup/components/QrCode.vue
```diff
@@ -0,0 +1,81 @@
...
+        gradient: {
+          linear: true,
+          radial: false,
+          color1: '#ffffff',
+          color2: '#ffffff',
+          rotation: '0',
+        },
+      },
+    });
+    this.qrCode.append(document.getElementById('qr-canvas'));
```
peronczyk, line 76: Please avoid matching elements with the use of vanilla JS. In the scope of Vue components it's better to use references.

src/popup/components/QrCode.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+  <div id="qr-canvas" />
```
peronczyk, line 2: In the scope of Vue apps it's considered as a bad pattern to use `id` because this component can be placed more than once in one view.

src/popup/components/QrCode.vue
```diff
@@ -0,0 +1,81 @@
...
+</template>
+<script>
+import QRCodeStyling from '@solana/qr-code-styling';
+
+import SHLogo from '../../icons/logo-smal-blue.png';
+
+export default {
```
peronczyk, line 9: To be honest this component begs to be written with the composition API and TS :)
Using TypeScript here would give us the benefit of type checking of the settings passed to the `QRCodeStyling`.

And this would result of finding that the `typeNumber: '0',` is wrong, because `typeNumber` accepts only numbers and undefined.

If you want help with that please ping me.

src/popup/components/QrCode.vue
```diff
@@ -0,0 +1,81 @@
...
+  },
+  mounted() {
+    this.qrCode = new QRCodeStyling({
+      data: this.value,
+      width: this.size,
+      height: this.size,
+      margin: 0,
+      padding: 0,
+      qrOptions: {
+        typeNumber: '0',
```
peronczyk, line 22: This should be a number.

src/popup/components/QrCode.vue
```diff
@@ -0,0 +1,81 @@
...
+      },
+      imageOptions: {
+        hideBackgroundDots: true,
+        imageSize: 0.4,
+        margin: 0,
+      },
+      dotsOptions: { type: 'extra-rounded', color: '#000000' },
+      backgroundOptions: {
+        color: '#ffffff',
+        gradient: null,
```
peronczyk, line 35: null is not assignable here.

src/popup/components/QrCode.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+  <div id="qr-canvas" />
+</template>
+<script>
+import QRCodeStyling from '@solana/qr-code-styling';
+
+import SHLogo from '../../icons/logo-smal-blue.png';
```
peronczyk, line 7: smal => small

peronczyk (changes requested):

src/popup/components/QrCode.vue
```diff
@@ -0,0 +1,53 @@
...
+        imageSize: 0.4,
+        margin: 0,
+      },
+      dotsOptions: { type: 'square', color: '#1161fe' },
+      image: SHLogo,
+      cornersSquareOptions: { type: 'square', color: '#6da0fe' },
+      cornersDotOptions: { type: 'square', color: '#1161fe' },
+    });
+
+    watch(() => props.value, (data: any) => qrCode.update({ data }));
```
peronczyk, line 45: The `data` here has the same type as the `value` prop, so it's a string.

peronczyk (changes requested):

peronczyk (approved):

PR #1650, opened 2022-10-19
[SW-368] Refactor: Update css color variable names

peronczyk:


src/styles/variables.scss
```diff
@@ -39,7 +36,7 @@ $color-hover: #1c1c1c;
 $color-disabled: #212121;
 
 $color-warning: #ffaa29;
-$color-danger: #ff4746;
+$color-danger: #ff2b5e;
```
pstachel, line 39: I don't think the "pink" colour should be labeled as "danger" - we should separate them.

pstachel (changes requested):

konradodo (approved):

src/styles/variables.scss
peronczyk: First of all it is a bad pattern to name the variable that holds a color by how it looks. Mostly because the color schemas changes.
I discussed this with Tsvetan and he told me that he is using this color as a "danger" thing - when user need to make an action because something bad happened or could happen. So the naming here makes sense.

pstachel (approved):

PR #1655, opened 2022-10-24
[SW-314, SW-320] The inconsistencies on the transaction details screen

pstachel:


src/popup/components/Overview.vue
```diff
@@ -39,70 +39,29 @@
     </div>
 
     <div class="details">
```
peronczyk, line 41: By the way of doing changes in this component I suggest to rename it. The name should not use only one word as this is not allowed in Vue 3. Also the name should be more precise and describe the purpose of this component.

src/popup/components/AvatarWithChainName.vue
```diff
@@ -9,12 +9,13 @@
     />
     <span
       class="text"
-    >{{ showAddress ? truncateAddress(address) : name }}</span>
+    >{{ showAddress ? splitAddress(address) : name }}</span>
```
peronczyk, line 12: Maybe instead of this the AddressFormatted component can be used?

src/popup/components/OverviewDetailsParty.vue
```diff
@@ -0,0 +1,100 @@
+<template>
+  <div
+    class="overview-details-party"
+    :data-cy="isRecipient ? 'recipient' : 'sender'"
```
peronczyk, line 4: In most cases element IDs used for testing cases should not be dynamic. I suggest to apply the data-cy properties on parent level to avoid being dynamic.

src/popup/components/OverviewDetailsParty.vue
```diff
@@ -0,0 +1,100 @@
...
+
+    display: block;
+    margin-bottom: 8px;
+    color: variables.$color-white;
+    text-decoration: none;
+    white-space: nowrap;
+    line-height: 16px;
+
+    &:hover {
+      .truncate::v-deep span {
```
peronczyk, line 84: This line can be removed. The hover styling will be applied to the `name` class which is an anchor `<a>` so the underline will be applied.

src/popup/components/OverviewDetailsParty.vue
```diff
@@ -0,0 +1,100 @@
...
+
+<style lang="scss" scoped>
+@use '../../styles/variables';
+@use '../../styles/typography';
+
+.overview-details-party {
+  width: 100%;
+
+  &.recipient {
+    .truncate {
```
peronczyk, line 68: This class does not exist in current scope.

peronczyk (changes requested):

peronczyk (approved):

konradodo (approved):

src/popup/components/Plate.vue
```diff
@@ -1,7 +1,7 @@
 <template>
```
peronczyk, line 1: In my opinion we need to remove the Plate component instead of tweaking it. There are only 2 places that uses it and the Figma designs does not use it anymore.

src/popup/components/Tokens.vue
```diff
@@ -46,6 +47,8 @@
 import AeIcon from '../../icons/tokens/ae.svg';
 import { AETERNITY_CONTRACT_ID, AETERNITY_SYMBOL } from '../utils/constants';
 
+const SIZES = ['rg', 'md', 'xl']; // TBD - add more sizes to icons
```
peronczyk, line 50: In my opinion there is no need to add the "TBD" comment here. And If you feel that we really need it I suggest to use "TODO:" pattern as described here: https://www.jetbrains.com/help/webstorm/using-todo.html#default_patterns
It allows to scrape the todos and some IDEs/addons color the todo comments in red if written this way.

src/popup/components/Tokens.vue
```diff
@@ -123,9 +134,21 @@ export default {
     vertical-align: middle;
   }
 
+  &.rg {
+    --icon-size: 16px;
```
peronczyk, line 138: This is the default size set in line 113. I think there is no need to set it twice.

src/popup/pages/AccountDetails.vue
```diff
@@ -188,9 +188,7 @@ export default {
   font-weight: 500;
   color: variables.$color-white;
   background-color: var(--screen-bg-color);
-  box-shadow:
-    0 0 0 1px variables.$color-border,
-    0 0 50px rgba(variables.$color-black, 0.6);
+  box-shadow: 0 0 0 1px variables.$color-border, 0 0 50px rgba(variables.$color-black, 0.6);
```
peronczyk, line 191: In my opinion the previous (splitted) version was better :)

peronczyk (changes requested):

peronczyk (approved):

PR #1658, opened 2022-10-26
fix(transaction-tokens): show none AE related transactions correctly

CedrikNikita:


konradodo (approved):

peronczyk (approved):

src/popup/components/TransferReview.vue
```diff
@@ -129,6 +129,10 @@ export default {
       'tippingV1',
       'tippingV2',
     ]),
+    isAex9() {
```
peronczyk, line 132: I suggest to use more informative naming here. Like for example `isSelectedAssetAex9`.

src/popup/components/TransferReview.vue
```diff
@@ -269,6 +273,7 @@ export default {
             contractId: this.tippingContract.deployInfo.address,
             type: SCHEMA.TX_TYPE.contractCall,
             function: 'tip',
+            selectedTokenId: selectedAsset.contractId,
```
peronczyk, line 276: I see some possible naming collision here. One one side we have `tokenId` which suggests that this is the `id` of the token. On the right side it is a `contractId` which is a different thing.
The `IToken` type has the `id` and `contractId` and I guess those 2 things should not be mixed.

CedrikNikita: There is no such thing as `tokenId` in the whole wallet. In this case we have an id of a contract that was used to sent tip and id of token (which is also a contract) that was sent.

peronczyk: Discussed in private.

CedrikNikita: 346de93245cff36e58f5df570362a1f2e136866c

peronczyk (approved):

PR #1664, opened 2022-10-28
[SW-304] Replace Balance List mixin with composable

peronczyk:


src/popup/components/Modals/AssetSelector.vue
```diff
@@ -25,74 +25,88 @@
...
+    }
+
     /**
      * Delay displaying tokens list until the modal transition is finished to prevent
      * performance issues when both animating the modal and rendering large amount of data.
      */
-    onModalOpen() {
-      this.$nextTick(() => {
-        this.isFullyOpen = true;
+    function onModalOpen() {
```
konradodo, line 91: you can add async, and add await on nextTick -> 0 callback

peronczyk: Makes sense. Done.

konradodo (approved):

PR #1666, opened 2022-11-02
[SW-324] - Transaction successfully submitted - css fix

konradodo:


peronczyk (approved):

src/popup/components/buttons/BtnMain.vue
```diff
@@ -130,6 +130,8 @@ export default defineComponent({
 
   &.secondary {
     --color: #{variables.$color-grey-medium};
+
+    padding: 0 32px;
```
peronczyk, line 134: I suggest to use CSS variable fox X axis paddings to avoid changing the Y axis paddings to 0.

src/popup/components/buttons/BtnMain.vue
```diff
@@ -85,7 +85,7 @@ export default defineComponent({
   flex: 1 1 0;
   width: fit-content;
   min-height: 40px;
-  padding: 5px 16px;
+  padding: 0 16px;
```
peronczyk, line 88: The top and bottom padding should not be set to 0. Actually I see that a change was made in figma regarding the button settings. The padding top and bottom should be equal to `8px`, and the `line-height` to `24px` instead of `20`.

peronczyk (changes requested):

peronczyk:
Please remove the `transactionTokens.ts` file from this PR.

pstachel (approved):

peronczyk (approved):

PR #1669, opened 2022-11-04
[SW-400] Fix connected and sign modals margins

peronczyk:


src/popup/components/Modal.vue
```diff
@@ -296,6 +306,11 @@ export default defineComponent({
     --screen-padding-x: 0;
   }
 
+  // This is not working correctly in Firefox extension
+  &.blur-bg {
+    backdrop-filter: blur(5px);
+  }
+
```
pstachel, line 313: Maybe there could be some kind of workaround for FF

pstachel (changes requested):

pstachel (approved):

konradodo (approved):

PR #1671, opened 2022-11-04
[SW-309] Allow register of names with all characters allowed in names spec

pstachel:


src/popup/components/TransactionInfo.vue
```diff
@@ -14,14 +14,18 @@
     </div>
 
     <div class="parties">
-      <Avatar v-bind="sender" />
+      <Avatar
+        v-bind="sender"
+        :name="punycodeToName(sender.name)"
```
konradodo, line 19: punycodeToName(sender.name) -> move to computed, you run this function twice

src/popup/components/TransferReview.vue
```diff
@@ -130,7 +131,8 @@ export default {
       'tippingV2',
     ]),
     isRecipientName() {
-      return this?.recipientAddress && checkAensName(this.recipientAddress);
+      return this.transferData?.address
```
peronczyk, line 134: I suggest to make the multiline conditions in a way that every item is in separate line.
This way:
```
return (
  condition
  && condition
)
```

src/popup/utils/names.ts
```diff
@@ -0,0 +1,15 @@
+import { toASCII, toUnicode } from 'tr46';
+
+const options = {
```
peronczyk, line 3: I suggest to rename this to `PUNYCODE_OPTIONS`.

src/popup/components/TransactionInfo.vue
```diff
@@ -14,14 +14,18 @@
...
+      />
       <div class="mid">
         <TriangleRight class="triangle" />
         <div class="line" />
       </div>
       <Avatar
         v-if="recipient.address"
         v-bind="recipient"
+        :name="punycodeToName(recipient.name)"
```
davidyuk, line 28: This is something that should be on the sdk side, it is the right way to encode names in not english?

pstachel: Yes, it's encoded in accordance with the [documentation](https://docs.aeternity.com/protocol/AENS.html#name)

PR #1674, opened 2022-11-08
Fix: Prevent errors in case of connection errors

peronczyk:


src/store/modules/accounts/ledger.js
```diff
@@ -58,7 +58,7 @@ export default {
       const address = await dispatch('request', { name: 'getAddress', args: [account.idx] });
       if (account.address !== address) {
         if (!process.env.IS_EXTENSION) {
-          dispatch('modals/open', { name: 'dafault', icon: 'alert', title: 'account not found' }, { root: true });
+          dispatch('modals/open', { name: 'default', icon: 'alert', title: 'account not found' }, { root: true });
```
pstachel, line 61: use `MODAL_DEFAULT` constant

pstachel (changes requested):

subhod-i:
I tested it and the bug doesn't exist anymore. Works as intended.

pstachel (approved):

PR #1684, opened 2022-11-13
Fix QR code reading on aeternity token

subhod-i:


pstachel (approved):

CedrikNikita:
The issue was fixed in #1685

PR #1698, opened 2022-11-15
Use SDK v12.1.3 (Bart's v2)

peronczyk:


package.json
```diff
@@ -36,6 +36,7 @@
     "@vue/composition-api": "^1.0.3",
     "@zxing/library": "^0.19.1",
     "aeternity-fungible-token": "github:aeternity/aeternity-fungible-token#v2.0.0",
+    "aeternity-sdk-12": "github:aeternity/aepp-sdk-js#v12.1.3",
```
subhod-i, line 39: @peronczyk Is this needed?

peronczyk: Yes, we'll use the dist directory script from this package.
But for now please skip reviewing the PR.

davidyuk: `dist` should be in npm package as well 🤔 installing from github works slower because it needs to be built

peronczyk: @davidyuk 
As mentioned already: this is a POC PR. So please skip reviewing it as it won't be merged.
I used the github direct link only because I didn't wanted to remove the previous SDK version.

PR #1702, opened 2022-11-17
[SW-411] AE Information Dissapears When The User Closes And Reopens The Extension

pstachel:


konradodo (approved):

konradodo (approved):

src/popup/pages/FungibleTokens/TokenDetails.vue
```diff
@@ -187,7 +187,7 @@ export default {
     },
   },
   async mounted() {
-    if (this.id.includes('ct_')) {
+    if (this.id?.includes('ct_')) {
```
peronczyk, line 190: To add some meaning to this check I suggest to create utility function.

src/popup/pages/FungibleTokens/TokenDetails.vue
```diff
@@ -154,7 +154,7 @@ export default {
       return this.availableTokens[this.id];
     },
     isAe() {
-      return this.id === AETERNITY_CONTRACT_ID;
+      return this.$route.name === 'coin-details' || this.id === AETERNITY_CONTRACT_ID;
```
peronczyk, line 157: In my opinion the components should not be aware about the scope around them. The encapsulation should make them blind and the only function for them is to receive some data - be controlled by the scope. So I suggest to add some prop to route meta and check it here instead of comparing the route name.

src/popup/pages/FungibleTokens/TokenDetails.vue
```diff
@@ -128,7 +128,7 @@ export default {
     Tab,
   },
   props: {
-    id: { type: String, required: true },
+    id: { type: String, default: null },
```
peronczyk, line 131: As the null is default in almost every case I suggest to do it like this:
```
id: String,
```

peronczyk (changes requested):

src/popup/pages/FungibleTokens/TokenDetails.vue
pstachel: It's mandatory to set default value for props with type other than `Boolean`

src/popup/utils/utils.ts
```diff
@@ -33,28 +41,56 @@ export function formatTime(time: number) {
...
 
-  if (fullHash.endsWith('.chain')) {
-    valid = true;
-    endpoint = addressTypes.nm;
-  } else if (addressTypes[prefix] && HASH_REGEX.test(hash)) {
-    valid = true;
-    endpoint = addressTypes[prefix];
+export function validateHash(fullHash?: string) {
+  if (!fullHash) {
+    return {};
```
peronczyk, line 64: This function should always return the same structure. Returning empty object breaks this because when this function is used you can't be sure what will be returned.

src/popup/utils/utils.ts
```diff
@@ -33,28 +41,56 @@ export function formatTime(time: number) {
...
   }
+  const { addressTypes, prefix, hash } = getHashPrefix(fullHash);
+
+  const isNameHash = fullHash.endsWith('.chain');
+  const valid: boolean = (addressTypes[prefix] && HASH_REGEX.test(hash)) || isNameHash;
+
+  return {
+    valid, isNameHash, addressTypes, prefix,
```
peronczyk, line 72: As this function name is `validate` in my opinion it should only return bool and optionally the prefix. No more data.

src/popup/utils/utils.ts
```diff
@@ -33,28 +41,56 @@ export function formatTime(time: number) {
   });
 }
 
-export function checkHashType(fullHash: string) {
+export function getHashPrefix(fullHash: string) {
```
peronczyk, line 44: I think this function should not be used within the `validateHash` but the `validateHash` should be used here. So the opposite.

src/popup/utils/utils.ts
```diff
@@ -33,28 +41,56 @@ export function formatTime(time: number) {
...
   const addressTypes: Record<string, string> = {
```
peronczyk, line 45: This logic should be put into `getHashEndpoint`.

src/popup/utils/utils.ts
```diff
@@ -33,28 +41,56 @@ export function formatTime(time: number) {
...
-  } else if (addressTypes[prefix] && HASH_REGEX.test(hash)) {
-    valid = true;
-    endpoint = addressTypes[prefix];
+export function validateHash(fullHash?: string) {
+  if (!fullHash) {
+    return {};
   }
+  const { addressTypes, prefix, hash } = getHashPrefix(fullHash);
+
+  const isNameHash = fullHash.endsWith('.chain');
```
peronczyk, line 68: Get `.chain` from existing constant.

peronczyk (changes requested):

src/popup/utils/utils.ts
```diff
@@ -33,28 +38,41 @@ export function formatTime(time: number) {
...
 
-export function checkHashType(fullHash: string) {
-  const addressTypes: Record<string, string> = {
-    ak: 'account/transactions',
-    ct: 'contracts/transactions',
-    nm: 'names',
-    ok: 'oracles/queries',
-    th: 'transactions',
-  };
+export const validateHash = (fullHash?: string) => {
```
peronczyk, line 41: There is no need to return twice. What about this approach:
```
export const validateHash = (fullHash?: string) => {
  const isName = !!fullHash?.endsWith(AENS_DOMAIN);
  let valid = false;
  let prefix = null;
  let hash = null;

  if (fullHash) {
    [prefix, hash] = fullHash.split('_');
    valid = (ADDRESS_TYPES[prefix] && HASH_REGEX.test(hash)) || isName;
  }

  return { valid, isName, prefix, hash };
}
```

peronczyk (changes requested):

peronczyk (approved):

PR #1711, opened 2022-11-21
[SW-423] Set verify button on seed phrase screen to fixed position

pstachel:


src/popup/pages/SeedPhraseVerifySettings.vue
```diff
@@ -1,113 +1,93 @@
...
+            :text="mnemonicShuffled[id]"
+            editable
+            @click.native="selectedWordIds.splice(index, 1)"
+          />
+        </template>
+      </div>
     </div>
-
-    <div class="footer">
+    <ModalFooter>
```
peronczyk, line 52: This page is not a Modal so this component's name makes no sense here.

src/popup/components/SeedPhraseNotification.vue
```diff
@@ -0,0 +1,77 @@
...
+</script>
+
+<style lang="scss" scoped>
+@use '../../styles/variables';
+@use '../../styles/typography';
+
+.seed-phrase-notification {
+  min-height: 176px;
+  border: 2px solid variables.$color-success-dark;
+  border-radius: 16px;
```
peronczyk, line 42: I suggest to create variable for this as this is the same value as the AccountCard has. I talked with Tsvetan and he told me that this radius is a common thing for large cards.

src/popup/components/SeedPhraseNotification.vue
```diff
@@ -0,0 +1,77 @@
...
+        }}
+      </div>
+    </div>
+  </div>
+</template>
+
+<script lang="ts">
+import { defineComponent } from '@vue/composition-api/dist/vue-composition-api';
+import CheckCircle from '../../icons/check-circle.svg?vue-component';
+import Alert from '../../icons/alert.svg?vue-component';
```
peronczyk, line 22: I suggest to add `...Icon` to icon components as in Vue 3 there is no way of having one-word component names.

Same for the `CheckCircle` above.

peronczyk (changes requested):

src/popup/components/ModalFooter.vue
```diff
@@ -0,0 +1,54 @@
...
+  },
+});
+
+</script>
+
+<style lang="scss" scoped>
+.modal-footer {
+  position: sticky;
+  bottom: 0;
+  margin: auto 0 0 0; // Move the footer to the bottom of the container
```
konradodo, line 25: margin auto 0 0

src/popup/components/SeedPhraseNotification.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <div :class="['seed-phrase-notification', { error: hasError }]">
+    <div class="icon-wrapper">
+      <Alert v-if="hasError" />
+      <CheckCircle v-else />
+    </div>
+    <div class="text">
+      <div>
+        {{
+          hasError
```
konradodo, line 10: move that to computed -> logic should be in script

src/styles/variables.scss
```diff
@@ -15,6 +15,7 @@ $z-index-modal: $z-index-header + 2;
 $z-index-loader: $z-index-modal - 1; // Modals should always be above the loader
 
 $bg-blur-radius: 4px;
+$card-border-radius: 16px;
```
peronczyk, line 18: On line 5 and 6 we have variables related to border radiuses with different naming pattern.

src/popup/components/InputSelectAsset.vue
```diff
@@ -64,7 +64,7 @@ export default {
 
   padding: 2px 12px;
   background-color: rgba(variables.$color-black, 0.3);
-  border-radius: 16px;
+  border-radius: variables.$card-border-radius;
```
peronczyk, line 67: In my opinion that is an accidental coincidence the value is similar. But the scope is different - this input has nothing to do with the cards.

peronczyk (changes requested):

src/popup/components/SeedPhraseBadge.vue
```diff
@@ -0,0 +1,70 @@
...
+
+export default defineComponent({
+  components: {
+    BtnPlain,
+    CloseIcon,
+  },
+  props: {
+    selected: Boolean,
+    editable: Boolean,
+    text: { type: String, required: true },
```
subhod-i, line 24: Is `required` needed? Isn't it required by default.
The first two props should be `selected: Boolean?` and `editable:Boolean?`
Language is typescript then we can use `?` feature for specifying the Maybe null type.

src/popup/pages/SeedPhraseVerifySettings.vue
```diff
@@ -1,113 +1,93 @@
...
-        </div>
-      </div>
-    </div>
+        :has-error="hasError"
+      />
+    </FixedScreenFooter>
   </div>
 </template>
 
 <script>
```
subhod-i, line 68: lang='ts'??

src/popup/components/SeedPhraseBadge.vue
pstachel: Yes, we always need to have the text

peronczyk (approved):

subhod-i (approved):

PR #1717, opened 2022-11-24
[SW-458] URL validation lib causes permission issues

### LINKED ISSUE #1714: URL validation lib causes permission issues
**Scroll down for screencast of this issue 👇**

Currently, in the permissions view vee-validate is used for URL validation: 
https://github.com/aeternity/superhero-wallet/blob/3fd92a936f330e2c72a00ebf45522763503b8fa4/src/popup/pages/PermissionManager.vue#L25

It's validation rules require the presence of a protocol and www prefix, as one can see here:
https://vee-validate.logaretm.com/v2/guide/rules.html#size

This behaviour seems to be tweaked in this codebase in some way, but shows strange issues still:
`studio.aepps.com` passes (good)
`studio.aepps:4200` **passes (not so good**, except if you've set something for `aepps` in your `hosts`
`localhost:4200` **fails (bad, should pass)**
`http://localhost:4200` **fails** (weird, should pass)
`http://www.localhost:4200` passes (weird, but okay)
`foo.localhost:4200` passes (weird, but okay)

This leads to the following chain of bugs: Using the wallet on `localhost:4200` leads to _this exact_ URL being saved in the permissions, but if you try editing the permissions, the URL will be displayed as invalid and you can't save the settings for this URL:

image
  
This results for the wallet to ask for permissions on every site reload.

I suggest using a different lightweight validator library or using regex(es) 😏 it would also make the codebase more future proof, as `url` was deprecated some while ago in vee-validate.

Screencast of this issue: 

https://www.youtube.com/watch?v=J1pBJzKHXQo

pstachel:
fixes #1714

### DIFF
--- a/src/store/plugins/veeValidate.js
+++ b/src/store/plugins/veeValidate.js
@@ -1,6 +1,6 @@
 import Vue from 'vue';
 import { Validator, ErrorBag, install as VeeValidate } from 'vee-validate/dist/vee-validate.minimal.esm';
-import { required, url } from 'vee-validate/dist/rules.esm';
+import { required } from 'vee-validate/dist/rules.esm';
 import BigNumber from 'bignumber.js';
 import { debounce } from 'lodash-es';
 import { Crypto } from '@aeternity/aepp-sdk';
@@ -9,7 +9,7 @@ import {
   isNotFoundError,
   getAddressByNameEntry,
   checkAensName,
-  validateTipUrl,
+  validateTipUrl, isValidURL,
 } from '../../popup/utils/helper';
 import { AENS_DOMAIN } from '../../popup/utils/constants';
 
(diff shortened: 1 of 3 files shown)


konradodo (approved):

subhod-i (approved):

src/store/plugins/veeValidate.js
```diff
@@ -37,7 +37,16 @@ Object.assign(ErrorBag.prototype, {
...
+Validator.extend('url', (str) => {
+  const pattern = new RegExp('^(https?:\\/\\/)?' // protocol
+      + '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' // domain name
+      + '((\\d{1,3}\\.){3}\\d{1,3}))|' // OR ip (v4) address
+      + 'localhost' // OR localhost
+      + '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' // port and path
+      + '(\\?[;&a-z\\d%_.~+=-]*)?' // query string
+      + '(\\#[-a-z\\d_]*)?$', 'i'); // fragment locator
+  return !!pattern.test(str);
+});
```
nikita-fuchs, line 49: 😁👌

konradodo (approved):

PR #1723, opened 2022-11-29
Fix claim tip transaction info

CedrikNikita:
This PR:
- adding missing calim tip transaction info: `nonce`, `block height`
- using a proper type for a `to` prop in `LinkButton` using `tipUrl`

- Also handling a long tipUrl correcrtly 
  |Before|After|
  |--|--|
  |image|image|


Liubov-crypto:
LGTM.

pstachel (approved):

PR #1706, opened 2022-11-18
[SW-307] composition api transactionToken and vuex

konradodo:

PR #1725, opened 2022-11-30
Fix issues related to transaction displaying 

CedrikNikita:
This PR is fixing three different problems:
- two of them were missed in #1706  (one of them was previously fixed in #1658)
  - transactions with `AE` token symbol was showed as AE coin
  - there were no labels for token `transfer` transaction
- there we no labels for the **pending** `transfer` transactions

### DIFF
--- a/src/composables/transactionToken.ts
+++ b/src/composables/transactionToken.ts
@@ -32,6 +32,7 @@ export const useTransactionToken = (
   const availableTokens = computed<ITokenList>(() => fungibleTokens.value.availableTokens);
 
   const getTxAmountTotal = useGetter('getTxAmountTotal');
+  const isTxAex9 = useGetter('isTxAex9');
 
   const getTxSymbol = useGetter('getTxSymbol');
   const getTxType = useGetter('getTxType');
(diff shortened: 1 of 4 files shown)


pstachel (approved):

src/composables/transactionToken.ts
```diff
@@ -60,7 +61,9 @@ export const useTransactionToken = (
         : getTxAmountTotal.value(transaction),
       symbol: isAllowance.value ? AETERNITY_SYMBOL : getTxSymbol.value(transaction),
       isReceived: getTxDirection.value(transaction) === 'received',
-      isAe: isAllowance.value || getTxSymbol.value(transaction) === AETERNITY_SYMBOL,
+      isAe: isAllowance.value
```
peronczyk, line 64: This conditional has both && and or. I suggest to group them to make other developer's life easier.
Also in case if the conditions should be put into separate lines I suggest to put all of them in separate lines. Eg.:

```
something: (
  condition1
  && condition2
  && condition3
)
```

konradodo (approved):

Liubov-crypto:
LGTM.

PR #1728, opened 2022-12-06
[SW-316, SW-456] permissions and retip limit

ifaouibadi:


src/popup/pages/Popups/Connect.vue
```diff
@@ -89,7 +89,15 @@ export default {
...
         };
       },
+      permission(state) {
+        return state.permissions[this.app.host];
+      },
     }),
+    appName() {
+      if (!this.permission) return this.app.name;
+
+      return this.permission.name;
```
peronczyk, line 99:
```
return this.permission?.name || this.app.name;
```

src/popup/pages/PermissionManager.vue
```diff
@@ -19,10 +19,14 @@
...
         />
       </div>
+
       <div class="permission-row">
         <InputField
           v-model="permission.host"
-          v-validate="{ required: true, url: true }"
+          v-validate="{
+            required: true,
+            url: !permission.host || !permission.host.includes('localhost')
```
peronczyk, line 28: I suggest to move this test into computed property.

src/popup/pages/Popups/Connect.vue
```diff
@@ -6,28 +6,28 @@
...
       data-cy="aepp"
     >
-      <span class="app-name">{{ app.name }}</span>
+      <span class="app-name">{{ appName }}</span>
       ({{ app.host }}) {{ $t('pages.connectConfirm.websiteRequestconnect') }}
     </div>
 
     <div class="permissions">
-      <template v-if="permissions.includes('address')">
+      <template v-if="access.includes('address')">
```
peronczyk, line 22: I suggest to store the 'address' in an constant.

src/popup/pages/Popups/Connect.vue
```diff
@@ -6,28 +6,28 @@
...
         <span class="title">
           <CheckMark class="icon" /> {{ $t('pages.connectConfirm.addressLabel') }}
         </span>
         <span class="description">
           {{ $t('pages.connectConfirm.addressRequest') }}
         </span>
       </template>
-      <template v-if="permissions.includes('transactions')">
+      <template v-if="access.includes('transactions')">
```
peronczyk, line 30: IDEM

src/popup/pages/Popups/Connect.vue
```diff
@@ -74,7 +74,7 @@ export default {
   mixins: [mixin],
   props: {
     app: { type: Object, required: true },
-    permissions: { type: Array, default: () => (['address', 'transactions']) },
+    access: { type: Array, default: () => (['address', 'transactions']) },
```
peronczyk, line 77: Also here we should use constants.

src/popup/pages/Address.vue
```diff
@@ -3,7 +3,7 @@
     :app="app"
     :resolve="onResolve"
     :reject="onReject"
-    :permissions="['address']"
+    :access="['address']"
```
peronczyk, line 6: We should use a constant for the `'address'`.

peronczyk (changes requested):

src/popup/utils/constants.ts
```diff
@@ -428,3 +428,8 @@ export const POPUP_TYPES = [
   POPUP_TYPE_MESSAGE_SIGN,
   POPUP_TYPE_RAW_SIGN,
 ] as const;
+
+export const POPUP_CONNECT_ACCESS = {
+  address: 'address',
+  transactions: 'transactions',
+};
```
peronczyk, line 435: Having an object of settings is not a bad thing. But for me the question is "why we store things as constants?". And there are 2 reasons:

1. Having a bond - one place to store things.
2. Avoiding errors. So when we use a constant and it's removed or renamed the build process will complain about this and notify us where's the problem. But when we use an object, and it's contents gets changed we don't get any info about this. So imagine a situation where someone has changed `POPUP_CONNECT_ACCESS.address` to `POPUP_CONNECT_ACCESS.hash`. We won't get any errors, because `POPUP_CONNECT_ACCESS.address` will still be valid as `undefined` is a value.

This is why I suggest to create separate constants for each of values. I know this makes the imports more complex, but keeps the app more error-proof.

peronczyk (changes requested):

peronczyk (approved):

Issue #452, opened 2020-08-09
Don't copy English wording to other locale files


davidyuk:
I'm considering this a bad practice because that wording may become outdated and a translator will waste his time working with that.

So, I'm proposing to remove untranslated content from locale files.

mradkov:
How do you propose keeping track of all locale tokens?

e.g. If we remove let's say this from the Spanish translation:

https://github.com/aeternity/superhero-wallet/blob/6a4779c14450073c21e3892007793b92493ee2b1/src/popup/locales/es.json#L878

`vue-i18n` will fall back to the English version of the value for that token, but how is the translator later supposed to know where and how to translate this (keep in mind translators does not need to be aware of JSON and to be able to structure the locales JSON files properly).

davidyuk:
I have two ideas regarding this:
- to write a script that will merge English translation to another one on demand
- to use third-party service to manage translations (I have seen once that is free for open-source projects)

PR #1730, opened 2022-12-07
[SW-42] Wallet notifications

### LINKED ISSUE #1008: Display name and avatar are missed in wallet notifications
image

pstachel:
fixes #1008


src/composables/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+  const notificationSettingsCheckedTypes = computed<string[]>(
+    () => notificationSettings.value.filter((s) => s.checked).map((s) => s.type),
+  );
+
+  const filteredNotifications = computed<INotification[]>(
+    () => [...superheroNotifications.value, ...walletNotifications.value]
+      .filter(
+        ({ status, type }) => (
+          status === NOTIFICATION_STATUS_READ
+                  || notificationSettingsCheckedTypes.value.includes(type)
```
peronczyk, line 42: Wrong indentation.

src/composables/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+import { useGetter, useState } from './vuex';
+import { getAllNotifications, modifyNotifications, normalizeNotification } from '../popup/utils/notifications';
+
+const CHUNK_SIZE = 20;
+
+const superheroNotifications = ref<INotification[]>([]);
+const walletNotifications = ref<INotification[]>([]);
+const pollingInterval = ref<NodeJS.Timer | null>(null);
+
+export function useNotifications(root: any) {
```
peronczyk, line 21: I suggest not to pass whole root object as it is unpredictable and there is no proper way to type it.

src/composables/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+  const sdk = useState('sdk');
```
peronczyk, line 22: Actually the true SDK state is in the plugin `sdkPlaugin`. So the line should look like this:
```
const sdk = useGetter('sdkPlugin/sdk');
```

src/composables/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+  const activeNetwork = useGetter('activeNetwork');
+  const account = useGetter('account');
```
peronczyk, line 24: I suggest to add typing `IAccount` here.

src/composables/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  function loadMore() {
+    canLoadMore.value = false;
+    offset.value += 1;
+    if (notificationsToShow.value.length < filteredNotifications.value.length) {
+      canLoadMore.value = true;
+    }
+  }
+
+  function handleSetPollingInterval() {
```
peronczyk, line 86: The word `handle` in my opinion is not necessary. I suggest to rename this and the following methods to:
`startNotificationsPolling`
and
`stopNotificationsPolling`

src/composables/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+  watchUntilTruthy,
+} from '../popup/utils';
+import { useGetter, useState } from './vuex';
+import { getAllNotifications, modifyNotifications, normalizeNotification } from '../popup/utils/notifications';
+
+const CHUNK_SIZE = 20;
+
+const superheroNotifications = ref<INotification[]>([]);
+const walletNotifications = ref<INotification[]>([]);
+const pollingInterval = ref<NodeJS.Timer | null>(null);
```
peronczyk, line 19: This shouldn't be a `ref`. Regular `let` would do the job.

src/composables/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+  function loadMore() {
+    canLoadMore.value = false;
+    offset.value += 1;
+    if (notificationsToShow.value.length < filteredNotifications.value.length) {
+      canLoadMore.value = true;
+    }
+  }
+
+  function handleSetPollingInterval() {
+    pollingInterval.value = executeAndSetInterval(() => getSuperheroNotifications(), 30000);
```
peronczyk, line 87: I suggest to be sure the polling is not turned on before executing new polling. Maybe by firing the `stopNotificationsPolling` at the beginning of this function.

src/composables/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  function handleClearPollingInterval() {
+    if (pollingInterval.value) {
+      clearInterval(pollingInterval.value);
+    }
+  }
+
+  async function handleMarkAsReadSuperhero() {
```
peronczyk, line 96: In this case also the word 'handle' is obsolete.

src/composables/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+      if (notification.type === NOTIFICATION_TYPE_WALLET) {
+        setWalletNotificationsStatus(
+          notification.createdAt,
+          NOTIFICATION_STATUS_READ,
+        );
+      }
+    });
+  }
+
+  function handleMarkAsRead() {
```
peronczyk, line 130: And again I suggest to remove the 'handle' word :)
Maybe the name would make more sense as `markAllAsRead`?

peronczyk (changes requested):

src/popup/utils/notifications.ts
```diff
@@ -0,0 +1,64 @@
+import { fetchJson, postJson } from './helper';
```
peronczyk, line 1: After spending some time with this PR I feel that the contents of this file should be moved to the composable. I know that wouldn't be the perfect idea, but I think still better than storing these functions here. Most of these functions requires a lot of arguments, which are present within the notifications composable. Like for example the SDK.

src/popup/utils/notifications.ts
```diff
@@ -0,0 +1,64 @@
...
+import {
+  IChainNames, INetwork, INotification, NotificationStatus,
+} from '../../types';
+import { AGGREGATOR_URL } from './constants';
+
+// TODO - type sdk
+export const getAllNotifications = async (sdk: any, activeNetwork: INetwork, address: string) => {
```
peronczyk, line 8: Word `get` in programming means it's a getter - synchronous thing. For the sake of connecting with the API I suggest to use `fetch...` to indicate what actually this method do.

src/popup/utils/utils.ts
```diff
@@ -136,17 +138,24 @@ export function watchUntilTruthy<T>(getter: () => T): Promise<NonNullable<T>> {
 export function rxJsObservableToVueState(
   observable: Observable<any>,
   defaultState: any = null,
-): Ref<any> {
+): Ref<any> | any[] {
```
peronczyk, line 141: This function returns reference. From what I see there's no way that this function could return `any[]`.

peronczyk (changes requested):

src/styles/variables.scss
```diff
@@ -25,6 +25,7 @@ $color-grey-light: #babac0;
 $color-grey-medium: #373737;
 $color-grey-dark: #787878;
 $color-grey-border: #5e5e5e;
+$color-grey-border-avatar: #5a5a5a;
```
peronczyk, line 28: This color is in 99% similar to the above one. I guess this is a mistake on the UX level. I suggest to align with Tsvetan.

src/composables/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+} from '../popup/utils';
+import { getAllNotifications, modifyNotifications, normalizeNotification } from '../popup/utils/notifications';
+
+const CHUNK_SIZE = 20;
+
+const superheroNotifications = ref<INotification[]>([]);
+const walletNotifications = ref<INotification[]>([]);
+let pollingInterval: NodeJS.Timer | null = null;
+
+export function useNotifications(notificationSettings: INotificationSetting[]) {
```
peronczyk, line 25: I guess the `notificationSettings` can be taken from the Vuex store.

peronczyk (changes requested):

src/styles/variables.scss
```diff
@@ -45,6 +46,7 @@ $color-success-dark: #00d3a1;
 $color-success-hover: #00dc87;
 $color-warning: #ffb422;
 $color-danger: #ff2b5e;
+$color-special: #ff4746;
```
peronczyk, line 49: I talked with Tsvetan about this and the outcome of the conversation is as follows:

1. The color you used here should be used as the `$color-danger`. So from now on all "danger" places will use it.
2. The color that we actually use as the "danger" (`#ff2b5e`) should be moved to the new variable `$color-secondary` and should be used for the purpose you created `$color-special`.
3. `$color-special` should be removed.
4. The `BtnMain` variant `secondary` should use instead of gray the color `$color-secondary` (the pink).
5. The old `secondary` variant of `BtnMain` should be replaced with `muted`. Please remember to align every occurence of the BtnMain with `variant="secondary"` prop.

src/popup/locales/cn.json
```diff
@@ -208,10 +208,9 @@
       "noTransactionsFound": "未发现转帐！",
       "balance": "余额",
       "copied": "已复制!",
-      "seedNotification": "你需要 {0} 你的助记词",
+      "seedNotification": "Welcome to Superhero wallet! Please backup your seed phrase!",
```
peronczyk, line 211: I suggest to remove the entries, that are using english within non-english translation files. This makes no sense to bloat files.

src/popup/locales/de.json
```diff
@@ -149,10 +149,9 @@
       "noTransactionsFound": "Keine Transaktionen gefunden!",
       "balance": "Balance",
       "copied": "Copied!",
-      "seedNotification": "You need to {0} your seed phrase",
+      "seedNotification": "Welcome to Superhero wallet! Please backup your seed phrase!",
```
peronczyk, line 152: IDEM

src/popup/locales/es.json
```diff
@@ -205,10 +205,9 @@
       "noTransactionsFound": "¡No se encontraron transacciones!",
       "balance": "Saldo",
       "copied": "¡Copiado!",
-      "seedNotification": "Necesitas {0} tu frase semilla",
+      "seedNotification": "Welcome to Superhero wallet! Please backup your seed phrase!",
```
peronczyk, line 208: IDEM

src/popup/locales/fr.json
```diff
@@ -149,10 +149,9 @@
       "noTransactionsFound": "Aucune transaction trouvée!",
       "balance": "Balance",
       "copied": "Copied!",
-      "seedNotification": "You need to {0} your seed phrase",
+      "seedNotification": "Welcome to Superhero wallet! Please backup your seed phrase!",
```
peronczyk, line 152: IDEM

src/popup/locales/it.json
```diff
@@ -149,10 +149,9 @@
       "noTransactionsFound": "Nessuna transazione trovato!",
       "balance": "Balance",
       "copied": "Copied!",
-      "seedNotification": "You need to {0} your seed phrase",
+      "seedNotification": "Welcome to Superhero wallet! Please backup your seed phrase!",
```
peronczyk, line 152: IDEM

src/popup/pages/Notifications.vue
```diff
@@ -1,207 +1,66 @@
...
+      notificationsToShow,
+      canLoadMore,
+      loadMore,
+      markAsReadAll,
+    } = useNotifications(notificationSettings.value);
+
+    onMounted(() => {
+      loadMore();
+      if (IS_EXTENSION) {
+        root.$nextTick(() => {
```
peronczyk, line 47: `nextTick` can be imported from the Composition Api to be typed properly.

peronczyk (changes requested):

src/composables/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+        ),
+      )
+      .sort((a, b) => new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime()),
+  );
+
+  const notificationsToShow = computed<INotification[]>(
+    () => filteredNotifications.value.slice(0, offset.value * CHUNK_SIZE),
+  );
+
+  async function getSuperheroNotifications() {
```
konradodo, line 52: in other composables we used arrow function

src/popup/components/InfiniteScroll.vue
```diff
@@ -0,0 +1,54 @@
...
+  props: {
+    isMoreData: { type: Boolean, required: true },
+  },
+  setup(props, { emit }) {
+    const appInner = document.querySelector('.app-inner')!;
+
+    const onScroll = () => {
+      const isDesktop = document.documentElement.clientWidth > 480 || process.env.IS_EXTENSION;
+      const { scrollHeight, scrollTop, clientHeight } = isDesktop
+        ? appInner : document.documentElement;
```
konradodo, line 27: const { scrollHeight, scrollTop, clientHeight } = isDesktop
        ? appInner 
        : document.documentElement;

src/popup/pages/Notifications.vue
konradodo: yes, and  please use await not callBack

src/composables/notifications.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  function stopNotificationsPolling() {
+    if (pollingInterval) {
+      clearInterval(pollingInterval);
+    }
+  }
+
+  function startNotificationsPolling() {
+    stopNotificationsPolling();
+    pollingInterval = executeAndSetInterval(() => getSuperheroNotifications(), 30000);
```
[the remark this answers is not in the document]
pstachel: What's the point of moving interval to const?

src/popup/locales/cn.json
CedrikNikita: https://github.com/aeternity/superhero-wallet/issues/452

src/composables/notifications.ts
pstachel: I think in all of the components (and composables as well) we try to use 'function'

src/popup/components/NotifyBell.vue
```diff
@@ -29,15 +29,16 @@ export default defineComponent({
     BellIcon,
   },
   setup(props, { root }) {
-    const superheroNotifications = rxJsObservableToVueState(
-      root.$store.state.observables.notifications,
-      [],
-    ) as Ref<INotification[]>;
-    const notifications = computed<INotification[]>(() => root.$store.state.notifications);
+    const notificationSettings = computed(() => root.$store.state.notificationSettings);
```
peronczyk, line 32: `useState` can be used here.

src/popup/components/InfiniteScroll.vue
```diff
@@ -0,0 +1,54 @@
...
+      onScroll,
+    };
+  },
+});
+</script>
+
+<style scoped>
+  .infinite-scroll {
+    height: 100%;
+  }
```
peronczyk, line 53: Unnecessary indentation.

src/popup/components/InfiniteScroll.vue
```diff
@@ -0,0 +1,54 @@
...
+export default defineComponent({
+  name: 'InfiniteScroll',
+  props: {
+    isMoreData: { type: Boolean, required: true },
+  },
+  setup(props, { emit }) {
+    const appInner = document.querySelector('.app-inner')!;
+
+    const onScroll = () => {
+      const isDesktop = document.documentElement.clientWidth > 480 || process.env.IS_EXTENSION;
```
peronczyk, line 25: 480 - magic number.

src/popup/components/InfiniteScroll.vue
```diff
@@ -0,0 +1,54 @@
...
+  defineComponent, onMounted, onUnmounted,
+} from '@vue/composition-api';
+
+export default defineComponent({
+  name: 'InfiniteScroll',
+  props: {
+    isMoreData: { type: Boolean, required: true },
+  },
+  setup(props, { emit }) {
+    const appInner = document.querySelector('.app-inner')!;
```
peronczyk, line 22: I know that this idea of selecting the `app-inner` is used in few places, but I consider this as a really bad pattern.
What I suggest is to create a composable that would provide a bridge between the app viewport and other components.

Only an idea:
```
// App.vue
setup(props) {
  // Template Refs: https://github.com/vuejs/composition-api#limitations
  const innerElem = ref(null);

  const { initViewport } = useViewport();
  
  onMounted(() => {
    initViewport(innerElem);
  });
}
```

```
// Any other component, like the current one:
setup() {
  const { onViewportScroll } = useViewport();
  onViewportScroll(({ isDesktop ) => {
    // Some logic
  });
}
```

peronczyk (changes requested):

src/popup/App.vue
```diff
@@ -132,10 +157,16 @@ export default defineComponent({
       await watchUntilTruthy(() => isRestored.value);
 
       setNotificationSettings();
+      initNotifications();
```
peronczyk, line 160: Maybe instead of always turning on the notifications polling the composable can decide on this. There are screens that don't need the notifications (like for example the web iframe popups).

The composable can track how many components are using them at a moment and when the number goes to 0 the polling can be turned off. This way the App should not import the notifications composable.

src/styles/variables.scss
pstachel: Resolved in #1741

src/composables/notifications.ts
```diff
@@ -0,0 +1,227 @@
...
+      )
+      .sort((a, b) => new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime()),
+  );
+
+  const notificationsToShow = computed<INotification[]>(
+    () => notificationsFiltered.value
+      .slice(0, fetchedNotificationsOffset.value * fetchedNotificationsLimit),
+  );
+
+  async function fetchAllNotifications() {
```
peronczyk, line 62: Return value should be typed.

src/composables/notifications.ts
```diff
@@ -0,0 +1,227 @@
...
+} from '../popup/utils';
+
+const fetchedNotificationsLimit = 20;
+
+const mountedComponentsCounter = ref<number>(0);
+const notificationsSuperhero = ref<INotification[]>([]);
+const notificationsWallet = ref<INotification[]>([]);
+let pollingInterval: NodeJS.Timer | null = null;
+
+export function useNotifications(preventMountedIncrement = false) {
```
peronczyk, line 27: I think this composable's settings can grow over time so I suggest to use the common pattern of providing the settings as an object. You can see this here: src/composables/tokensList.ts

Also I suggest to inverse the logic of this setting. For example:
```
requirePolling: false // Default value
```
So if a component needs the actual notifications this setting should be set to `true`.
What do you think?

src/composables/notifications.ts
```diff
@@ -0,0 +1,227 @@
...
+  NOTIFICATION_STATUS_CREATED,
+  NOTIFICATION_STATUS_READ,
+  NOTIFICATION_TYPE_WALLET,
+  executeAndSetInterval,
+  watchUntilTruthy, fetchJson, postJson, NOTIFICATION_ENTITY_TYPE_TIP, AGGREGATOR_URL,
+} from '../popup/utils';
+
+const fetchedNotificationsLimit = 20;
+
+const mountedComponentsCounter = ref<number>(0);
```
peronczyk, line 22: This counter does not count the mounted components, but components that require polling. So the name is kind of misleading.

src/composables/notifications.ts
```diff
@@ -0,0 +1,227 @@
...
+  INotificationSetting,
+  NotificationStatus,
+} from '../types';
+import { useGetter, useState } from './vuex';
+import {
+  NOTIFICATION_STATUS_CREATED,
+  NOTIFICATION_STATUS_READ,
+  NOTIFICATION_TYPE_WALLET,
+  executeAndSetInterval,
+  watchUntilTruthy, fetchJson, postJson, NOTIFICATION_ENTITY_TYPE_TIP, AGGREGATOR_URL,
```
peronczyk, line 17: I suggest to split this line same as the previous ones with the constants on top.

src/composables/notifications.ts
```diff
@@ -0,0 +1,227 @@
...
+  const activeNetwork = useGetter<INetwork>('activeNetwork');
+  const account = useGetter<IAccount>('account');
+  const notificationSettings = useState<INotificationSetting[]>('notificationSettings');
+
+  const canLoadMore = ref(true);
+  const fetchedNotificationsOffset = ref(0);
+
+  const notificationSettingsCheckedTypes = computed<string[]>(
+    () => notificationSettings.value
+      .filter((setting: INotificationSetting) => setting.checked)
```
peronczyk, line 38: No need to use `INotificationSetting` as the `setting`'s type is inherited from the `notificationSettings`.

You can also go with:
```
.filter(({ checked }) => checked)
```

peronczyk (changes requested):

src/composables/notifications.ts
```diff
@@ -0,0 +1,227 @@
...
+  executeAndSetInterval,
+  watchUntilTruthy, fetchJson, postJson, NOTIFICATION_ENTITY_TYPE_TIP, AGGREGATOR_URL,
+} from '../popup/utils';
+
+const fetchedNotificationsLimit = 20;
+
+const mountedComponentsCounter = ref<number>(0);
+const notificationsSuperhero = ref<INotification[]>([]);
+const notificationsWallet = ref<INotification[]>([]);
+let pollingInterval: NodeJS.Timer | null = null;
```
peronczyk, line 25: `setTimeout` return "timer" or so called "intervalId". So I suggest to name this variable `pollingTimer` or (my choice) `pollingIntervalId`.

src/composables/notifications.ts
```diff
@@ -0,0 +1,227 @@
...
+      }
+    });
+  }
+
+  function markAsReadAll() {
+    markAsReadWallet();
+    markAsReadSuperhero();
+  }
+
+  async function initNotifications() {
```
peronczyk, line 194: As I mentioned on our call: there's too many nested functions within this file.

src/composables/notifications.ts
```diff
@@ -0,0 +1,227 @@
...
+        : `${AGGREGATOR_URL}tip/${sourceId}/comment/${entityId}`,
+    });
+  }
+
+  async function getSuperheroNotifications() {
+    const chainNames = useState('chainNames');
+
+    notificationsSuperhero.value = (
+      await fetchAllNotifications()
+    ).map(normalizeNotification(chainNames.value));
```
peronczyk, line 122: `normalizeNotification` is not used anywhere so I suggest to use it's code directly here.

src/composables/notifications.ts
```diff
@@ -0,0 +1,227 @@
...
+      ...other,
+      sender,
+      chainName: sender && chainNames?.[sender],
+      path: entityType === NOTIFICATION_ENTITY_TYPE_TIP
+        ? `${AGGREGATOR_URL}tip/${entityId}`
+        : `${AGGREGATOR_URL}tip/${sourceId}/comment/${entityId}`,
+    });
+  }
+
+  async function getSuperheroNotifications() {
```
peronczyk, line 117: This function is not a getter.

peronczyk (changes requested):

konradodo (approved):

src/composables/viewport.ts
```diff
@@ -0,0 +1,44 @@
...
+    callback({ isOutsideOfViewport });
+  }
+
+  function initViewport(scrollableElement: HTMLDivElement | null) {
+    innerElement.value = scrollableElement;
+  }
+
+  function addViewportListener(onScrollMethod: Function) {
+    innerElement.value.addEventListener('scroll', () => onViewportScroll(onScrollMethod));
+    window.addEventListener('scroll', () => onViewportScroll(onScrollMethod));
```
peronczyk, line 30: I suggest to add debounce to this event handlers.

peronczyk (changes requested):

src/composables/viewport.ts
```diff
@@ -0,0 +1,51 @@
...
+    callback({ isOutsideOfViewport });
+  }
+
+  function initViewport(scrollableElement: HTMLDivElement | null) {
+    viewportElement.value = scrollableElement;
+  }
+
+  function debouncedOnScrollAction(onScrollMethod: Function) {
+    return debounce(() => onViewportScroll(onScrollMethod), 100);
+  }
```
peronczyk, line 33: In my opinion there is no need to use additional function and pass the arguments to it when attaching the event listeners. You can just create debounced function:

```
const onViewportScroll = debounce(() => {
  // Body of the function
}, 100);

<Elem>.addEventListener('scroll', () => onViewportScroll(callback));
```

Also the `addViewportListener` and `clearViewportListener` can be run inside of this composable to avoid having to run the `onMounted`and `onUnmounted` in a place we want to use it:

So here it can look this way:
```
function onViewportScroll(callback: Function) {
  onMounted(() => addViewportListener(callback));
  onBeforeUnmount(() => clearViewportListener(callback));
}
```

so then in the component running it will look this way:
```
const { onViewportScroll } = useViewport();

onViewportScroll(() => {
  // Local component's code
});
```

Not tested, just an idea.

src/popup/components/InfiniteScroll.vue
```diff
@@ -0,0 +1,49 @@
+<template>
+  <div
+    class="infinite-scroll"
+    @scroll="onViewportScroll"
```
peronczyk, line 4: In my opinion this is wrong - we should not watch the scroll here. We have the scroll events attached to the viewport with the composable.

src/popup/components/InfiniteScroll.vue
```diff
@@ -0,0 +1,49 @@
...
+      }
+    }
+
+    onMounted(() => {
+      addViewportListener(onScroll);
+    });
+
+    onUnmounted(() => {
+      clearViewportListener(onScroll);
+    });
```
peronczyk, line 36: As mentioned - this can be replaced with one method on the composabel level.

peronczyk (changes requested):

src/composables/notifications.ts
```diff
@@ -0,0 +1,227 @@
...
+
+  const notificationsToShow = computed<INotification[]>(
+    () => notificationsFiltered.value
+      .slice(0, fetchedNotificationsOffset.value * FETCHED_NOTIFICATIONS_LIMIT),
+  );
+
+  async function fetchAllNotifications(): Promise<INotification[]> {
+    const fetchUrl = `${activeNetwork.value.backendUrl}/notification/user/${account.value.address}`;
+    const responseChallenge = await fetchJson(fetchUrl);
+    const sdk = await getSdk();
```
peronczyk, line 80: `Promise.all`

src/composables/viewport.ts
```diff
@@ -0,0 +1,46 @@
...
+      || process.env.IS_EXTENSION
+    );
+
+    const { scrollHeight, scrollTop, clientHeight } = isDesktop
+      ? viewportElement.value
+      : document.documentElement;
+
+    const isOutsideOfViewport = scrollHeight - scrollTop <= clientHeight + 100;
+
+    callback({ isOutsideOfViewport });
```
peronczyk, line 24:
```suggestion
    callback({
      isOutsideOfViewport: scrollHeight - scrollTop <= clientHeight + 100
    });
```

src/popup/components/AddressTruncated.vue
```diff
@@ -0,0 +1,48 @@
...
+  @extend %face-mono-15-regular;
+
+  letter-spacing: -0.25em;
+  text-align: center;
+  margin-left: -1px;
+  margin-right: 3px;
+
+  &.medium {
+    @extend %face-mono-15-medium;
+  }
```
peronczyk, line 46: I think the difference between the medium and regular is negligible so I suggest to use only one variant.

src/popup/pages/Dashboard.vue
```diff
@@ -42,7 +42,7 @@
           </template>
           <BtnMain
             class="card-button"
-            variant="secondary"
+            variant="danger"
```
peronczyk, line 45: I think this is wrong. Please align with Tsvetan.

peronczyk (changes requested):

peronczyk (approved):

PR #1736, opened 2022-12-12
[SW-494] Refactor: Migrate router to TypeScript

peronczyk:


ifaouibadi (approved):

pstachel (approved):

src/popup/components/Truncate.vue
```diff
@@ -4,6 +4,7 @@
     :class="{ right }"
   >
     <span
+      ref="container"
```
CedrikNikita, line 7: Does `Truncate` change relly blong to this commit?

Liubov-crypto:
LGTM

src/popup/components/Truncate.vue
peronczyk: No, but after the changes in the App and Header components there were issues with rendering this component caused by the lack of the `parentElement` so I needed to fix this within this PR.

PR #1739, opened 2022-12-14
[496] Features/push notification

ifaouibadi:


src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,93 @@
...
+  const getTxSymbol = useGetter<any>('getTxSymbol');
+
+  /**
+   * This function is used to format the given address.
+   * @param {string} address - the address to format.
+   * @returns {string} - returns the account's name if it exists,
+   * otherwise it returns the translated string "Account" + the account's index +1.
+  */
+  function formatAddress(address: string): string {
+    const account: IAccount | undefined = accounts.value.find(
```
konradodo, line 21: you no need ( Account | undefined ) ts automatically give that type

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    return `${i18n.t('pages.account.heading')} ${account.idx + 1}`;
+  }
+
+  /**
+   * @param {string} address - the address to check if it's exists in the existing accounts list.
+   * @returns {boolean} - indicating whether or not the address is found in the accounts list.
+   */
+  function isAddressFromExistingWalletAccounts(address?: string): boolean {
+    if (!address || !accounts.value) return false;
```
konradodo, line 37:
```suggestion
    if (!(address && accounts.value) {
      return false;
    }
```

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /**
+   * This function starts listening for incoming transactions on the 'Transactions' channel.
+   */
+  function startListeningForIncomingTransactions() {
+    const receivedTransactions: Set<string> = new Set();
+
+    channelSubscribe('Transactions', (transaction: ITransaction) => {
+      if (!transaction || !transaction.tx) return;
```
konradodo, line 60:
```suggestion
      if (!transaction?.tx) return;
```

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,93 @@
...
+        && !receivedTransactions.has(hash)
+      ) {
+        receivedTransactions.add(hash);
+
+        const amount = amountRounded(getTxAmountTotal.value(transaction));
+        const title = i18n.t('pages.notifications.incomingTransaction');
+        const text = `${formatAddress(recipientId)} ${i18n.t('pages.transactions.received')} ${amount} ${getTxSymbol.value(transaction)}`;
+
+        addWalletNotification({
+          id: hash as any,
```
konradodo, line 76: can we type hash correctly?

src/composables/websocket.ts
```diff
@@ -0,0 +1,145 @@
...
+   */
+  function channelSubscribe(
+    channel: WEBSOCKET_CHANNEL_TYPES,
+    // eslint-disable-next-line no-unused-vars
+    cb: (payload: any) => void = (payload: any) => null,
+  ) {
+    // subscribe to channel
+    subscribeToWebsocketChannel(channel);
+    // register the subscriber
+    if (!subscribers.value[channel]) subscribers.value[channel] = [];
```
konradodo, line 128:
```suggestion
    if (!subscribers.value[channel]) { 
      subscribers.value[channel] = [];
    }
```

src/composables/incomingTransactions.ts
ifaouibadi: 5090A497-0F41-4AC1-B9B3-7DCCDE63FC91

package.json
```diff
@@ -52,6 +52,9 @@
     "dex-contracts-v2": "github:aeternity/dex-contracts-v2",
     "lodash-es": "^4.17.21",
     "qr-code-styling": "github:aeternity/qr-code-styling",
+    "rxjs": "^6.6.7",
+    "rxjs-etc": "^10.6.2",
```
CedrikNikita, line 56: That's a rebase mistake.

CedrikNikita (changes requested):

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * @param {string} address - the address to check if it's exists in the existing accounts list.
+   * @returns {boolean} - indicating whether or not the address is found in the accounts list.
+   */
+  function isAddressFromExistingWalletAccounts(address?: string): boolean {
+    if (!(address && accounts.value)) {
+      return false;
+    }
+    return accounts.value.some((account: IAccount) => account.address === address);
```
konradodo, line 40:
```suggestion
  function isAddressFromExistingWalletAccounts(address?: string) {
   return 
   (
     address 
     && accounts.value
     && accounts.value.some((account: IAccount) => account.address === address)
     );
```

src/store/modules/notifications.js
```diff
@@ -0,0 +1,28 @@
+import { NOTIFICATION_STATUS_READ } from '../../popup/utils';
+
```
konradodo, line 2: please use composable for this

src/composables/websocket.ts
```diff
@@ -0,0 +1,147 @@
...
+    // eslint-disable-next-line no-unused-vars
+    cb: (payload: any) => void = (payload: any) => null,
+  ) {
+    // subscribe to channel
+    subscribeToWebsocketChannel(channel);
+    // register the subscriber
+    if (!subscribers.value[channel]) {
+      subscribers.value[channel] = [];
+    }
+    subscribers.value[channel].push(cb);
```
konradodo, line 131:
```suggestion
    subscribers.value[channel]) 
     ?  subscribers.value[channel].push(cb);
     :  subscribers.value[channel] = [];
   
   
```

src/composables/incomingTransactions.ts
ifaouibadi: Also, there's a chance that the account cannot be found with `Array.find`.
and `.find` returns T | undefined


DDAD6382-D845-4BC0-92D2-B6034ABFDB28

src/composables/websocket.ts
ifaouibadi: 6481EE4D-44F9-49D9-99C2-C1C80C4D29F3

peronczyk (changes requested):

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,96 @@
...
+  const getTxAmountTotal = useGetter<any>('getTxAmountTotal');
+  const getTxSymbol = useGetter<any>('getTxSymbol');
+
+  /**
+   * This function is used to format the given address.
+   * @param {string} address - the address to format.
+   * @returns {string} - returns the account's name if it exists,
+   * otherwise it returns the translated string "Account" + the account's index +1.
+  */
+  function getAccountNameByAddress(address: string): string {
```
peronczyk, line 20: This function should be moved to `accounts` composable in my opinion.

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (account.name) return account.name;
+
+    return `${i18n.t('pages.account.heading')} ${account.idx + 1}`;
+  }
+
+  /**
+   * @param {string} address - the address to check if it's exists in the existing accounts list.
+   * @returns {boolean} - indicating whether or not the address is found in the accounts list.
+   */
+  function isAddressFromExistingWalletAccounts(address?: string): boolean {
```
peronczyk, line 36: There is already the same function in the `accounts` composable: `isLocalAccountAddress`.

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (!amount || Number.isNaN(amount)) return amount;
+    if (Number.isInteger(amount) || amount === 0) return amount;
+    return parseFloat(amount.toFixed(precision));
+  }
+
+  /**
+   * This function starts listening for incoming transactions on the 'Transactions' channel.
+   */
+  function startListeningForIncomingTransactions() {
+    const receivedTransactions: Set<string> = new Set();
```
peronczyk, line 60:
```suggestion
    const receivedTransactions = new Set<string>();
```

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,96 @@
...
+    );
+  }
+
+  /**
+   * This function rounds a given number to a specified precision.
+   * @param {number} amount - The amount to round
+   * @param {number} precision - The number of decimal places to round to (default 2)
+   * @returns {number} - The rounded amount
+   */
+  function amountRounded(amount: number, precision: number = 2): number {
```
peronczyk, line 50: There is already `amountRounded` function in the `utils.ts` file. Maybe it can be used? If the logic is different please name this function differently.
Also I think this a utility function so it should be moved to our global utils.

peronczyk (changes requested):

src/lib/sendPushNotification.ts
```diff
@@ -0,0 +1,57 @@
...
+import { INotification } from '../types';
+import { IS_CORDOVA, IS_EXTENSION, IS_WEB } from './environment';
+
+/**
+ * Schedules a web notification.
+ *
+ * @param {Partial<INotification>} notification - The notification to schedule.
+ */
+export function scheduleWebNotification(notification: Partial<INotification>) {
+  try {
```
peronczyk, line 11: I think there is no need to wrap this block with `try...catch` because both `showNotification` and `requestPermission` are not throwing anything.

peronczyk (changes requested):

src/composables/accounts.ts
```diff
@@ -26,11 +27,30 @@ export function useAccounts({ store }: IDefaultComposableOptions) {
...
   }
 
+  /**
+   * This function is used to format the given address.
+   * @param {string} address - the address to format.
+   * @returns {string} - returns the account's name if it exists,
+   * otherwise it returns the translated string "Account" + the account's index +1.
+  */
+  function getAccountNameByAddress(address: string): string {
+    const _account = accounts.value.find(
```
peronczyk, line 37: I strongly suggest to avoid using underscores as this is less readable. If there is a name collision there are other options. Like for example `accountFound`. Or even better:

```
const { idx, name } = accounts.value.find((acc) => acc.address === address) || {};
```

src/composables/accounts.ts
```diff
@@ -26,11 +27,30 @@ export function useAccounts({ store }: IDefaultComposableOptions) {
...
+      (acc: IAccount) => acc.address === address,
+    );
+
+    if (!_account?.idx) return address;
+
+    if (_account.name) return _account.name;
+
+    return `${i18n.t('pages.account.heading')} ${_account.idx + 1}`;
```
peronczyk, line 45: If you test if the `idx` exists the return should have something to do with the idx. Now the if statements are messed up between each other.

Maybe this would be better:
```
    if (name) {
      return name;
    }
    if (idx) {
      return `${i18n.t('pages.account.heading')} ${idx + 1}`;
    }
    return address;
```

peronczyk (changes requested):

src/composables/websocket.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  onNetworkChange(store, () => {
+    if (activeNetwork.value.websocketUrl) {
+      connectToWebsocketServer(activeNetwork.value.websocketUrl);
+    }
+  });
+
+  return {
+    wsClient,
```
peronczyk, line 152: This is not used in any place that uses the composable. And I don't think it's a good idea to return the client. This composable should be the final layer of abstraction without any side entrances :)

src/composables/websocket.ts
```diff
@@ -0,0 +1,155 @@
...
+});
+
+export function useWebsocket({ store }: IDefaultComposableOptions) {
+  const activeNetwork = useGetter<INetwork>('activeNetwork');
+
+  /**
+   * This function is used to subscribe to a specific WebSocket channel.
+   * @param {WebSocketChannelTypes} channelName - The name of the channel to subscribe to
+   */
+  function subscribeToWebsocketChannel(channelName: WebSocketChannelTypes) {
```
peronczyk, line 29: As the native class has the "Socket" name uppercased I suggest to follow this naming: `subscribeToWebSocketChannel`.

src/composables/websocket.ts
```diff
@@ -0,0 +1,155 @@
...
+  MicroBlocks: [],
+  KeyBlocks: [],
+});
+const channels = ref<Record<WebSocketChannelTypes, boolean>>({
+  Transactions: false,
+  MicroBlocks: false,
+  KeyBlocks: false,
+});
+
+export function useWebsocket({ store }: IDefaultComposableOptions) {
```
peronczyk, line 22: As the native class has the "Socket" name uppercased I suggest to follow this naming: `useWebSocket`.

peronczyk (changes requested):

src/composables/websocket.ts
```diff
@@ -0,0 +1,155 @@
...
+    // eslint-disable-next-line no-unused-vars
+    cb: (payload: any) => void = (payload: any) => null,
+  ) {
+    // subscribe to channel
+    subscribeToWebsocketChannel(channel);
+    // register the subscriber
+    if (!subscribers.value[channel]) {
+      subscribers.value[channel] = [];
+    }
+    subscribers.value[channel].push(cb);
```
peronczyk, line 142: There is a default value for the `cb` which is an empty function. Is this really required to have empty functions added to this array? Maybe we should have the default value set to null and add the callback only if it's not null?

src/composables/websocket.ts
```diff
@@ -0,0 +1,155 @@
...
+   *    - The function to be called when a message is received on the specified channel.
+   */
+  function channelSubscribe(
+    channel: WebSocketChannelTypes,
+    // eslint-disable-next-line no-unused-vars
+    cb: (payload: any) => void = (payload: any) => null,
+  ) {
+    // subscribe to channel
+    subscribeToWebsocketChannel(channel);
+    // register the subscriber
```
peronczyk, line 138: I suggest to add empty lines before comments.

src/composables/websocket.ts
```diff
@@ -0,0 +1,155 @@
+import { ref } from '@vue/composition-api';
+import camelcaseKeysDeep from 'camelcase-keys-deep';
+import { IDefaultComposableOptions, INetwork, WebSocketChannelTypes } from '../types';
+import { createNetworkWatcher } from './composablesHelpers';
+import { useGetter } from './vuex';
+
+const { onNetworkChange } = createNetworkWatcher();
+
+const wsClient = ref<WebSocket>();
```
peronczyk, line 9: After spending some time on reviewing the code I feel that this doesn't need to be a Vue ref. Instead just regular `let`.

peronczyk (changes requested):

peronczyk (changes requested):

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,56 @@
...
+  const getTxAmountTotal = useGetter<any>('getTxAmountTotal');
+  const getTxSymbol = useGetter<any>('getTxSymbol');
+
+  /**
+   * This function starts listening for incoming transactions on the 'Transactions' channel.
+   */
+  function startListeningForIncomingTransactions() {
+    const receivedTransactions: Set<string> = new Set<string>();
+
+    channelSubscribe('Transactions', (transaction: ITransaction) => {
```
pstachel, line 22: What do you think of storing the channel names as a constant?

src/composables/websocket.ts
```diff
@@ -0,0 +1,155 @@
...
+      !channelName
+      || channels.value[channelName]
+      || !isConnected.value
+      || !wsClient.value
+    ) return;
+
+    wsClient.value.send((JSON.stringify({
+      op: 'Subscribe',
+      payload: channelName,
+    })));
```
peronczyk, line 40: I suggest to create an interface `IMiddlewareWebSocketSubscriptionMessage` in our types directory and comment it with the reference link to https://github.com/aeternity/ae_mdw#websocket-interface

peronczyk (changes requested):

src/popup/utils/utils.ts
```diff
@@ -409,6 +409,19 @@ export function amountRounded(rawAmount: number | BigNumberPublic): string {
...
+   * This function rounds a given number to a specified precision.
+   * @param {number} amount - The amount to round
+   * @param {number} precision - The number of decimal places to round to (default 2)
+   * @returns {number} - The rounded amount
+   */
+export function roundAmountToPrecision(amount: number, precision: number = 2): number {
+  if (!amount || Number.isNaN(amount) || Number.isInteger(amount) || amount === 0) {
+    return amount;
+  }
+  return parseFloat(amount.toFixed(precision));
```
peronczyk, line 422:
```
return (test)
  ? valueIfTrue
  : valueIfFalse;
```

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,56 @@
...
+        const amount = roundAmountToPrecision(getTxAmountTotal.value(transaction));
+        const title = i18n.t('pages.notifications.incomingTransaction');
+        const text = `${getAccountNameByAddress(recipientId)} ${i18n.t('pages.transactions.received')} ${amount} ${getTxSymbol.value(transaction)}`;
+
+        addWalletNotification({
+          id: hash,
+          text,
+          title,
+          buttonLabel: i18n.t('pages.notifications.viewTransaction'),
+          path: `/account-details/transactions/${hash}`,
```
peronczyk, line 43: I suggest to use route names instead of direct path. In this case to generate path (as far as I remember) you can use:
```
router.resolve({ name: ROUTE_TX_DETAILS, params: { hash } });
```

peronczyk (changes requested):

peronczyk (changes requested):

src/composables/accounts.ts
```diff
@@ -26,11 +27,33 @@ export function useAccounts({ store }: IDefaultComposableOptions) {
...
 
+  /**
+   * @param {string} address - the address to format.
+   * @returns {string} - returns the account's name if it exists,
+   * otherwise it returns the translated string "Account" + the account's index +1.
+  */
+  function getAccountNameByAddress(address: string): string {
+    const { idx, name } = accounts.value.find(
+      (acc) => acc.address === address,
+    ) || {};
```
peronczyk, line 38: I suggest to create separate function for this, eg.: `getAccountByAddress`. I'm asking for this because I would use it in my current branch.

src/composables/websocket.ts
```diff
@@ -0,0 +1,172 @@
...
+      || channels[channelName]
+      || !isWsConnected.value
+    ) {
+      return;
+    }
+
+    wsClient.send((JSON.stringify({
+      op: 'Subscribe',
+      payload: channelName,
+    } as IMiddlewareWebSocketSubscriptionMessage)));
```
peronczyk, line 48: Using `as` forces TS to think that the data has the defined format. So sometimes wrong formats passes the test. To have the object properly typed I suggest to use this code:
```
const message: IMiddlewareWebSocketSubscriptionMessage = {
      op: 'Subscribe',
      payload: channelName,
};
wsClient.send((JSON.stringify(message));
```

peronczyk (changes requested):

src/composables/composablesHelpers.ts
```diff
@@ -15,14 +15,21 @@ export function createNetworkWatcher() {
   let currentNetworkId: string;
 
   return {
-    onNetworkChange: (store: Store<any>, callback: () => void) => {
+    onNetworkChange: (
+      store: Store<any>,
+      // eslint-disable-next-line no-unused-vars
+      callback: (activeNetwork: any) => void,
+      immediate = false,
```
peronczyk, line 22: I suggest to pass this in a similar way as the Vue `watch` does - as an object of options.

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,74 @@
...
+  router: VueRouter
+}
+
+export function useIncomingTransactions({ store, router }: IIncomingTransactionsOptions) {
+  const { watchChannelUpdates } = useWebSocket({ store });
+  const { addWalletNotification } = useNotifications({ store });
+  const { isLocalAccountAddress, getAccountNameByAddress } = useAccounts({ store });
+  const getTxAmountTotal = useGetter<any>('getTxAmountTotal');
+  const getTxSymbol = useGetter<any>('getTxSymbol');
+  const unWatchChannelUpdates = ref();
```
peronczyk, line 22: No need to use ref. Just set this as a `let`.

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * This function starts listening for incoming transactions on the 'Transactions' channel.
+   */
+  function startListeningForIncomingTransactions() {
+    const receivedTransactions: Set<string> = new Set<string>();
```
peronczyk, line 28: No need to set the type twice.
```suggestion
    const receivedTransactions = new Set<string>();
```

src/composables/websocket.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    // eslint-disable-next-line consistent-return
+    return () => {
+      delete subscribers[channel][subscriberId];
+    };
+  }
+
+  watch(
+    () => activeNetwork.value,
+    (_activeNetwork: INetwork) => {
```
peronczyk, line 161: It's rather subjective, but I suggest to not use variable names prefixed by underscore as it means it's a temporary thing. If there's a name collision you can use `val` here.

src/composables/websocket.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  /**
+   * Subscribes to a specific websocket channel
+   * @param {WebSocketChannelTypes} channel - The websocket channel to subscribe to.
+   * @param {Function} [callback=(payload: any) => payload]
+   *    - The function to be called when a message is received on the specified channel.
+   * @returns {Function}
+   *    - A function that can be called to unsubscribe the specified callback from the channel.
+   */
+  function watchChannelUpdates(
```
peronczyk, line 137: What do you think about creating 3 more functions that would serve as a shortcuts to specific channels? For example:
```
const watchTransactionsUpdates = (payload) => watchChannelUpdates('Transactions', payload);
```
So whenever any of the components wants to watch a specific channel it would import this shortcut function instead of the `watchChannelUpdates` and the constant with the channel name.

src/lib/pushNotification.ts
peronczyk, line 1: In general I like the idea of adding comments to clarify what the specific block of code does. But in this file the comments brings no value. The title of the comment is the function name converted from camelCase to regular text, and the params description is the copy of the type.

peronczyk (changes requested):

src/composables/websocket.ts
```diff
@@ -0,0 +1,188 @@
...
+      const data = camelcaseKeysDeep(JSON.parse(message.data));
+      if (
+        !data.subscription
+        || !subscribers[data.subscription as WebSocketChannelTypes]
+      ) {
+        return;
+      }
+      // Call all subscribers for the channel
+      Object.values(subscribers[data.subscription as WebSocketChannelTypes]).forEach(
+        (subscriberCb: any) => subscriberCb(data.payload),
```
peronczyk, line 72: `subscriberCb` is not `any`. Just remove `as any` and proper type will be inherited from the type assigned in the line 16.

src/types/index.ts
```diff
@@ -611,3 +614,12 @@ export interface IDefaultComposableOptions {
    */
   store: Store<any>
 }
+
+export type WebSocketChannelTypes = 'Transactions' | 'MicroBlocks' | 'KeyBlocks';
```
peronczyk, line 618: For the type that has the union of possible values please use singular name (same as for example `CurrencyCode`.
```suggestion
export type WebSocketChannelType = 'Transactions' | 'MicroBlocks' | 'KeyBlocks';
```

peronczyk (changes requested):

src/composables/websocket.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    // Subscribe to all the channels
+    (Object.keys(channels) as WebSocketChannelTypes[])
+      .forEach(subscribeToWebSocketChannel);
+  }
+
+  /**
+   * UnSubscribes websocket channels.
+   */
+  function unsubscribeFromWebSocketChannels() {
```
peronczyk, line 94: Maybe this function should also remove all callbacks that are set on `subscribers` ?

src/composables/websocket.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  /**
+   * Subscribes to websocket channels.
+   */
+  function subscribeToWebSocketChannels() {
+    // Add an event listener for messages
+    wsClient.addEventListener('message', handleWebsocketMessage);
+
+    // Subscribe to all the channels
+    (Object.keys(channels) as WebSocketChannelTypes[])
```
peronczyk, line 87: This line does the same as the line 99 but is written in a different way. I suggest to maintain consistency.

src/popup/App.vue
```diff
@@ -163,9 +170,18 @@ export default defineComponent({
...
       await watchUntilTruthy(() => isRestored.value);
       setNotificationSettings();
+      startListeningForIncomingTransactions();
 
       loadAeternityData();
       root.$store.commit('setChainNames', await root.$store.dispatch('getCacheChainNames'));
+
+      if (IS_WEB) {
+        webNotification.requestPermission();
+      }
```
peronczyk, line 180: In my opinion this piece of code should be put closer to the code that actually displays the notification to the user. I see few solutions for this:
1. In the `pushNotification` file instead of having functions create a singleton JS class with the public method that displays the notification and the constructor asks for the permission.
2. Encapsulate both asking for permission and sending the notification in the `notifications.ts` composable.

If I may choose I would go with the 1st one as it follows similar pattern as the `Logger`.

src/popup/utils/constants.ts
```diff
@@ -570,3 +573,9 @@ export const DASHBOARD_CARD_ID = {
 } as const;
 
 export const SUPPORTED_MULTISIG_CONTRACT_VERSION = '2.0.0';
+
+export const WEB_SOCKET_CHANNEL_TYPES: Record<WebSocketChannelTypes, WebSocketChannelTypes> = {
```
peronczyk, line 577: Maybe instead of `TYPES` better name would be `NAMES`?

src/composables/websocket.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    // register the subscriber
+    if (!subscribers[channel]) {
+      subscribers[channel] = {};
+    }
+
+    const subscriberId = genUuid();
+    subscribers[channel][subscriberId] = callback;
+
+    // eslint-disable-next-line consistent-return
```
peronczyk, line 155: I think this can be removed.

src/types/index.ts
```diff
@@ -232,8 +233,10 @@ export interface INotification {
   type: string
   updatedAt?: string
   isSeedBackup?: boolean
+  isIncomingTransaction?: boolean,
```
peronczyk, line 236: IDEM

src/types/index.ts
```diff
@@ -221,7 +222,7 @@ export interface INotification {
   createdAt: string
   entityId?: string
   entityType?: string
-  id?: number
+  id?: number | string
```
peronczyk, line 225: I suggest to end all interface properties with semicolon like the TS docs are doing: https://www.typescriptlang.org/docs/handbook/2/objects.html#property-modifiers

src/types/index.ts
```diff
@@ -611,3 +614,12 @@ export interface IDefaultComposableOptions {
    */
   store: Store<any>
 }
+
+export type WebSocketChannelTypes = 'Transactions' | 'MicroBlocks' | 'KeyBlocks';
+
+// https://github.com/aeternity/ae_mdw#websocket-interface
+export interface IMiddlewareWebSocketSubscriptionMessage {
+  op: 'Subscribe' | 'Unsubscribe',
```
peronczyk, line 622: I suggest to end all interface properties with semicolon like the TS docs are doing: https://www.typescriptlang.org/docs/handbook/2/objects.html#property-modifiers

peronczyk (changes requested):

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,72 @@
...
+  router: VueRouter
+}
+
+export function useIncomingTransactions({ store, router }: IIncomingTransactionsOptions) {
+  const { watchTransactionsUpdates } = useWebSocket({ store });
+  const { addWalletNotification } = useNotifications({ store });
+  const { isLocalAccountAddress, getAccountNameByAddress } = useAccounts({ store });
+  const getTxAmountTotal = useGetter<any>('getTxAmountTotal');
+  const getTxSymbol = useGetter<any>('getTxSymbol');
+  let unWatchTransactionsUpdates : () => void;
```
peronczyk, line 21:
```suggestion
  let unWatchTransactionsUpdates: () => void;
```

src/popup/utils/utils.ts
```diff
@@ -410,6 +410,18 @@ export function amountRounded(rawAmount: number | BigNumberPublic): string {
...
 }
 
+/**
+   * This function rounds a given number to a specified precision.
+   * @param {number} amount - The amount to round
+   * @param {number} precision - The number of decimal places to round to (default 2)
+   * @returns {number} - The rounded amount
+   */
+export function roundAmountToPrecision(amount: number, precision: number = 2): number {
+  return (!amount || Number.isNaN(amount) || Number.isInteger(amount) || amount === 0)
```
peronczyk, line 420: No need to use `amount === 0` as the `!amount` checks this.

peronczyk (changes requested):

src/popup/App.vue
ifaouibadi: if you request notification permission while the user is idle in the app the browser will automatically reject it.
cause browsers expects permission request after user click or page reload

src/composables/websocket.ts
peronczyk: OK, makes sense. Closing.

src/popup/App.vue
peronczyk: OK, so the singleton could have the `init` method which could be run in the App.vue file.
What I want is to have a clear abstraction encapsulated in one file that handles all the logic related with native notifications.

src/composables/websocket.ts
```diff
@@ -0,0 +1,192 @@
...
+   *    - The function to be called when a message is received on the specified channel.
+   * @returns {Function}
+   *    - A function that can be called to unsubscribe the specified callback from the channel.
+   */
+  function watchChannelUpdates(
+    channel: WebSocketChannelName,
+    // eslint-disable-next-line no-unused-vars
+    callback: (payload: any) => void,
+  ): () => void {
+    // subscribe to channel
```
peronczyk, line 149: I think it's obvious what the below line does - I suggest to remove this comment.

src/composables/websocket.ts
```diff
@@ -0,0 +1,192 @@
...
+    (Object.keys(channelsSubscribed) as WebSocketChannelName[]).forEach(
+      (channelName) => subscribeToWebSocketChannel(channelName),
+    );
+  }
+
+  /**
+   * UnSubscribes websocket channels.
+   */
+  function unsubscribeFromWebSocketChannels() {
+    // Remove event listener for messages
```
peronczyk, line 100: I think it's obvious what the below line does - I suggest to remove this comment.

src/types/index.ts
```diff
@@ -615,3 +619,13 @@ export interface IDefaultComposableOptions {
    */
   store: Store<any>
 }
+
+// export type WebSocketChannelName = 'Transactions' | 'MicroBlocks' | 'KeyBlocks';
```
peronczyk, line 623: This can be removed I guess.

peronczyk (approved):

src/composables/incomingTransactions.ts
```diff
@@ -0,0 +1,72 @@
...
+export interface IIncomingTransactionsOptions extends IDefaultComposableOptions {
+  router: VueRouter
+}
+
+export function useIncomingTransactions({ store, router }: IIncomingTransactionsOptions) {
+  const { watchTransactionsUpdates } = useWebSocket({ store });
+  const { addWalletNotification } = useNotifications({ store });
+  const { isLocalAccountAddress, getAccountNameByAddress } = useAccounts({ store });
+  const getTxAmountTotal = useGetter<any>('getTxAmountTotal');
+  const getTxSymbol = useGetter<any>('getTxSymbol');
```
pstachel, line 20: I think both of those could be typed

src/composables/websocket.ts
```diff
@@ -0,0 +1,192 @@
...
+  KeyBlocks: {},
+};
+
+const channelsSubscribed: Record<WebSocketChannelName, boolean> = {
+  Transactions: false,
+  MicroBlocks: false,
+  KeyBlocks: false,
+};
+
+// eslint-disable-next-line no-unused-vars
```
pstachel, line 33: Is it still unused?

src/composables/websocket.ts
```diff
@@ -0,0 +1,192 @@
...
+    // Subscribe to all the channels
+    (Object.keys(channelsSubscribed) as WebSocketChannelName[]).forEach(
+      (channelName) => subscribeToWebSocketChannel(channelName),
+    );
+  }
+
+  /**
+   * UnSubscribes websocket channels.
+   */
+  function unsubscribeFromWebSocketChannels() {
```
pstachel, line 99: What are your thoughts of removing the comments in favour of adding the word "All" to function names - it would be self explanatory then

src/composables/websocket.ts
peronczyk, line 1: After conversation with @subhod-i we came up with an idea that this file should not be a 'composable' as it doesn't hold that much of a reactive 'state'. Only the `isConnected` which could be a regular `let`. Also the watcher for network could be moved to `App.vue`.

So I think this whole logic should be a TS class.

peronczyk (changes requested):

src/composables/incomingTransactions.ts
peronczyk: No need for now. We are using this methods in many places and we didn't type them. But in the future... :)

subhod-i (changes requested):

subhod-i:
https://github.com/aeternity/superhero-wallet/blob/2518e011400cb3558216757ae8ae8f9856474dff/src/composables/websocket.ts#L165-L167

@ifaouibadi  Listening to each and every transaction in the Aeternity chain is not scalable and not correct. Hardly transactions are relevant to the account. Please make use of the `Object` payload type to listen to events for only required local accounts. 

From Middleware documentation 
```
Object, which takes a further field, target - can be any æternity entity. So you may subscribe to any æternity
 object type, and be sent all transactions which reference the object. For instance, if you have an oracle 
ok_JcUaMCu9FzTwonCZkFE5BXsgxueagub9fVzywuQRDiCogTzse you may subscribe to this object and 
be notified of any events which relate to it - presumable you would be interested in queries, to which you 
would respond. Of course you can also subscribe to accounts, contracts, names, whatever you like.
```

https://github.com/aeternity/ae_mdw#supported-payloads

Perhaps we can discuss this in a meeting.

ifaouibadi:
> @ifaouibadi Listening to each and every transaction in the Aeternity chain is not scalable and not correct. Hardly transactions are relevant to the account. Please make use of the `Object` payload type to listen to events for only required local accounts.

fixed

src/composables/accounts.ts
```diff
@@ -47,6 +48,25 @@ export function useAccounts({ store }: IDefaultComposableOptions) {
...
+  */
+  function getAccountNameByAddress(address: string): string {
+    const { idx, name } = getAccountByAddress(address) || {};
+
+    if (name) {
+      return name;
+    }
+
+    if (idx) {
+      return `${i18n.t('pages.account.heading')} ${idx + 1}`;
```
CedrikNikita, line 64: This field was removed in #2036. It is now in `common`.

ifaouibadi: didn't find any changes to the pages account heading in #2036  and the translations still there https://github.com/aeternity/superhero-wallet/blob/develop/src/popup/locales/en.json#L500

src/lib/webSocketClient.ts
```diff
@@ -0,0 +1,138 @@
...
+    callback: (payload: any) => void,
+  ) {
+    if (isWsConnected) {
+      wsClient.send(JSON.stringify(message));
+    }
+
+    subscribersQueue.push(message);
+
+    const uuid = genUuid();
+    // const uuid = message.target || genUuid();
```
CedrikNikita, line 53:
```suggestion
```

src/popup/pages/NetworkForm.vue
subhod-i, line 1: How can a user enter a WebSocket URL when adding a custom network?
 Is this something the user has to enter or can it be derived from the base middleware URL?

subhod-i (commented):
I have 2 major considerations regarding this PR
1. Multiple WebSocket connections are active upon changing the network back and forth between mainnet and testnet. Although the WebSocket methods are static it looks like there are many parallel connections that keep on getting added as we switch networks. If you observe the notifications are still received from the first connection made. Here is the screenshot 
Screenshot 2023-05-04 at 1 07 51 PM

2. Notifications are received only from the active network. I believe notifications should be network agnostic with each notification indicating the network.

subhod-i:
I remember @thepiwo once said we will introduce a web socket for multisig backend too to detect new multisig vaults. 
@ifaouibadi Please take this into consideration we should be able to use `WebsocketClient.ts` file even for other web socket connections.

src/lib/webSocketClient.ts
```diff
@@ -0,0 +1,138 @@
+import { v4 as genUuid } from 'uuid';
+import camelcaseKeysDeep from 'camelcase-keys-deep';
+import { IMiddlewareWebSocketSubscriptionMessage, WebSocketChannelName } from '../types';
+import {
+  WEB_SOCKET_CHANNEL_KEY_BLOCKS,
+  WEB_SOCKET_CHANNEL_KEY_OBJECT,
+  WEB_SOCKET_CHANNEL_MICRO_BLOCKS,
+  WEB_SOCKET_CHANNEL_TRANSACTIONS,
+  WEB_SOCKET_SUBSCRIBE,
```
subhod-i, line 9: Perhaps you can make these as part one object `WEB_SCOKET_CHANNEL`

src/lib/webSocketClient.ts
```diff
@@ -0,0 +1,138 @@
...
+      );
+    } catch (error) {
+      handleUnknownError(error);
+    }
+  }
+
+  static disconnect() {
+    wsClient.addEventListener('open', WebSocketClient.handleWebsocketOpen);
+    wsClient.addEventListener('close', WebSocketClient.handleWebsocketClose);
+    wsClient.addEventListener('message', WebSocketClient.handleWebsocketClose);
```
subhod-i, line 124: Is this needed?

src/lib/webSocketClient.ts
```diff
@@ -0,0 +1,138 @@
...
+  static subscribeForChannel(
+    message: IMiddlewareWebSocketSubscriptionMessage,
+    // eslint-disable-next-line no-unused-vars
+    callback: (payload: any) => void,
+  ) {
+    if (isWsConnected) {
+      wsClient.send(JSON.stringify(message));
+    }
+
+    subscribersQueue.push(message);
```
subhod-i, line 50: If web socket is connected then there is no need to add the new subscription to the cache queue.
```suggestion
    if (isWsConnected) {
      wsClient.send(JSON.stringify(message));
    } else {
      subscribersQueue.push(message);
    }
```

thepiwo:
I started work on websockets for the ga-multisig backend, but we never followed up with an actual plan on integration, so I abandoned the approach for the time https://github.com/aeternity/ga-multisig-backend/commit/515b598c47c541c2ef4739ae5f21acf625f8aaed

let's have a meeting to pick it up again

subhod-i:
> I started work on websockets for the ga-multisig backend, but we never followed up with an actual plan on integration, so I abandoned the approach for the time [aeternity/ga-multisig-backend@515b598](https://github.com/aeternity/ga-multisig-backend/commit/515b598c47c541c2ef4739ae5f21acf625f8aaed)
> 
> let's have a meeting to pick it up again

This is important to optimize the performance of the SH wallet, especially to eliminate polling to the multisig backend. Let's talk soon ✌️

src/lib/webSocketClient.ts
ifaouibadi: we need to keep the list of subscribers for network change event

src/composables/composablesHelpers.ts
```diff
@@ -15,14 +19,21 @@ export function createNetworkWatcher() {
...
-
       if (!currentNetworkId) {
         currentNetworkId = activeNetwork.networkId;
+        if (options?.immediate) {
+          callback(activeNetwork);
+        }
       } else if (currentNetworkId !== activeNetwork.networkId) {
         currentNetworkId = activeNetwork.networkId;
-        callback();
+        callback(activeNetwork);
```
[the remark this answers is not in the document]
ifaouibadi: if the callback is optional then this composable is not needed cc @peronczyk

src/composables/composablesHelpers.ts
```diff
@@ -7,6 +7,10 @@ import { Store } from 'vuex';
 import { INetwork } from '../types';
 import { useConnection } from './connection';
 
+export interface INetworkWatcherOptions {
+  immediate?: boolean;
```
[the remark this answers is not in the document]
ifaouibadi: it's better to follow Vue watch naming

ifaouibadi:
> 1. Although the WebSocket methods are static it looks like there are many parallel connections that keep on getting added as we switch networks. If you observe the notifications are still received from the first connection made. Here is the screenshot

Improved network unsubscribe from previous when it's been changed


> 2. Notifications are received only from the active network. I believe notifications should be network agnostic with each notification indicating the network.

it's not a good idea to listen for all networks been added in the app as it can cause notifications duplicate or receiving unexpected notification from non active network

src/composables/composablesHelpers.ts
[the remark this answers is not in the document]
subhod-i: but `immediate` word is very abstract

subhod-i:
> it's not a good idea to listen for all networks been added in the app as it can cause notifications duplicate or receiving unexpected notification from non active network

Fine.

src/background/popupHandler.ts
```diff
@@ -17,7 +17,7 @@ export const getAeppUrl = (v: any) => new URL(v.connection.port.sender.url);
 export const showPopup = async (aepp: any, type: string, params?: any) => {
   const id = uuid();
   const { href, protocol, host } = typeof aepp === 'object' ? getAeppUrl(aepp) : new URL(aepp);
-  const tabs = await browser.tabs.query({ active: true });
+  const tabs: any[] = await browser.tabs.query({ active: true });
```
konradodo, line 20: we can create type for that tabs.
under of this line, tabs have two keys. url -> string, and id -> number

subhod-i (approved):

peronczyk (approved):

CedrikNikita:
I suggest to add a redirect from transaction notification to open this transaction details.

src/lib/webSocketClient.ts
```diff
@@ -0,0 +1,141 @@
...
+  WEB_SOCKET_UN_SUBSCRIBE,
+  handleUnknownError,
+} from '../popup/utils';
+
+let wsClient: WebSocket;
+let isWsConnected: boolean;
+
+const subscribersQueue: IMiddlewareWebSocketSubscriptionMessage[] = [];
+// eslint-disable-next-line no-unused-vars
+const subscribers:Record<WebSocketChannelName, Record<string, (payload: any) => void>> = {
```
peronczyk, line 16: Missing space
```suggestion
const subscribers: Record<WebSocketChannelName, Record<string, (payload: any) => void>> = {
```

src/lib/webSocketClient.ts
```diff
@@ -0,0 +1,141 @@
...
+    subscribersQueue.push(message);
+
+    const uuid = genUuid();
+    subscribers[message.payload][uuid] = callback;
+    return () => {
+      delete subscribers[message.payload][uuid];
+    };
+  }
+
+  // eslint-disable-next-line no-unused-vars
```
peronczyk, line 57: This is not required anymore - rebase develop.

src/lib/webSocketClient.ts
```diff
@@ -0,0 +1,141 @@
...
+      {
+        op: WEB_SOCKET_SUBSCRIBE,
+        payload: WEB_SOCKET_CHANNELS.Object,
+        target: address,
+      },
+      callback,
+    );
+  }
+
+  // eslint-disable-next-line no-unused-vars
```
peronczyk, line 69: IDEM

src/lib/webSocketClient.ts
```diff
@@ -0,0 +1,141 @@
...
+    return WebSocketClient.subscribeForChannel(
+      {
+        op: WEB_SOCKET_SUBSCRIBE,
+        payload: WEB_SOCKET_CHANNELS.Transactions,
+      },
+      callback,
+    );
+  }
+
+  // eslint-disable-next-line no-unused-vars
```
peronczyk, line 80: IDEM

src/lib/webSocketClient.ts
```diff
@@ -0,0 +1,141 @@
...
+    return WebSocketClient.subscribeForChannel(
+      {
+        op: WEB_SOCKET_SUBSCRIBE,
+        payload: WEB_SOCKET_CHANNELS.MicroBlocks,
+      },
+      callback,
+    );
+  }
+
+  // eslint-disable-next-line no-unused-vars
```
peronczyk, line 91: IDEM

src/lib/webSocketClient.ts
```diff
@@ -0,0 +1,141 @@
...
+const subscribersQueue: IMiddlewareWebSocketSubscriptionMessage[] = [];
+// eslint-disable-next-line no-unused-vars
+const subscribers:Record<WebSocketChannelName, Record<string, (payload: any) => void>> = {
+  [WEB_SOCKET_CHANNELS.Transactions]: {},
+  [WEB_SOCKET_CHANNELS.MicroBlocks]: {},
+  [WEB_SOCKET_CHANNELS.KeyBlocks]: {},
+  [WEB_SOCKET_CHANNELS.Object]: {},
+};
+
+export default class WebSocketClient {
```
peronczyk, line 23: This class has only static methods so it is a static class. Static classes should not operate on any state, but actually there is a state: `wsClient` and `isWsConnected`. This is really bad pattern to mix the concept of regular classes (which should encapsulate state and operations) with the static/singleton. So please use encapsulated state or convert the class into singleton.

peronczyk (changes requested):

src/composables/composablesHelpers.ts
```diff
@@ -25,15 +29,22 @@ export function createNetworkWatcher() {
   let currentNetworkId: string;
 
   return {
-    onNetworkChange: async (store: Store<any>, callback: () => void) => {
+    onNetworkChange: async (
+      store: Store<any>,
+      // eslint-disable-next-line no-unused-vars
```
peronczyk, line 34: Can be removed due to recent changes in ESLint

src/popup/utils/constants.ts
```diff
@@ -594,3 +596,22 @@ export const ALLOWED_ICON_STATUSES = [
...
+  Transactions: 'Transactions',
+  MicroBlocks: 'MicroBlocks',
+  KeyBlocks: 'KeyBlocks',
+  Object: 'Object',
+};
+
+export const WEB_SOCKET_SUBSCRIBE = 'Subscribe';
+export const WEB_SOCKET_UN_SUBSCRIBE = 'Unsubscribe';
+
+export const WEB_SOCKET_CHANNEL_NAMES = [
```
peronczyk, line 610: This can be removed as it's only used within the type definitions. See my further comment I did there on how you can actually create the type for the channel name.

src/types/index.ts
```diff
@@ -671,3 +675,12 @@ export interface IDefaultComposableOptions {
 }
 
 export type StatusIconType = typeof ALLOWED_ICON_STATUSES[number];
+
+export type WebSocketChannelName = typeof WEB_SOCKET_CHANNEL_NAMES[number];
```
peronczyk, line 679: Instead of having to create the array and import it you can just use `ObjectValues<typeof WEB_SOCKET_CHANNELS>`

src/types/index.ts
```diff
@@ -233,25 +235,27 @@ export interface IPermission {
...
+  receiver?: string;
+  sender?: string;
+  sourceId?: string;
+  sourceType?: string;
+  status: NotificationStatus;
+  text?: string | TranslateResult;
+  type: string;
+  updatedAt?: string;
+  isSeedBackup?: boolean;
+  isIncomingTransaction?: boolean;
```
peronczyk, line 255: I suggest to rename it to `incomingTransactionExists` or `hasIncomingTransaction`.

peronczyk (changes requested):

src/lib/webSocketClient.ts
```diff
@@ -0,0 +1,150 @@
...
+  WEB_SOCKET_UN_SUBSCRIBE,
+  handleUnknownError,
+} from '../popup/utils';
+
+let wsClient: WebSocket;
+let isWsConnected: boolean;
+
+const subscribersQueue: IMiddlewareWebSocketSubscriptionMessage[] = [];
+
+const subscribers: Record<WebSocketChannelName, Record<string, (payload: any) => void>> = {
```
peronczyk, line 17: `(payload: any) => void` is used many times in this file. Maybe consider creating a type for this, like for example `WebSocketCallback` or something similar.

src/lib/WebSocketClient.ts
```diff
@@ -0,0 +1,169 @@
...
+
+const subscribers: Record<WebSocketChannelName, Record<string, (
+  payload: ITransaction | ITopHeader
+) => void>> = {
+  [WEB_SOCKET_CHANNELS.Transactions]: {},
+  [WEB_SOCKET_CHANNELS.MicroBlocks]: {},
+  [WEB_SOCKET_CHANNELS.KeyBlocks]: {},
+  [WEB_SOCKET_CHANNELS.Object]: {},
+};
+
```
peronczyk, line 31: You choose to use a Singleton class so the states used within the class should be encapsulated inside of the class.

src/popup/utils/constants.ts
```diff
@@ -594,3 +596,16 @@ export const ALLOWED_ICON_STATUSES = [
   'success',
   'warning',
 ] as const;
+
+export const WEB_SOCKET_CHANNELS = {
+  Transactions: 'Transactions',
+  MicroBlocks: 'MicroBlocks',
+  KeyBlocks: 'KeyBlocks',
+  Object: 'Object',
```
peronczyk, line 604: This is actually not an channel name. Also `Object` is already used in the `IMiddlewareWebSocketSubscriptionMessage` `payload` prop.

peronczyk (changes requested):

src/popup/utils/constants.ts
ifaouibadi: it's a channel name :D https://github.com/aeternity/ae_mdw#publishing-message-format

used to subscribe for specific address

peronczyk (approved):

src/composables/composablesHelpers.ts
peronczyk, line 1: Changes done in this file are not necessary as non of the files are using them.

subhod-i (changes requested):
We will have a re-review. Multiple updates

PR #1742, opened 2022-12-16
fix: coins & tokens router path

ifaouibadi:


src/popup/pages/FungibleTokens/TokenContainer.vue
```diff
@@ -123,16 +123,17 @@ export default defineComponent({
...
     const tabs = [
       {
         text: root.$t('pages.transactionDetails.transactions'),
-        routeName: 'token-transactions',
+        routeName: `${tokenType}-transactions`,
       },
       {
-        text: root.$t(`pages.token-details.${root.$route.params.type}-details`),
-        routeName: 'token-details',
+        text: root.$t(`pages.token-details.${tokenType}-details`),
```
peronczyk, line 135: As far as I know the string interpolation in terms of the translation paths is considered as a bad practice. Mostly because the i18n scrapers - tools that analyses the rpository and checks which translations are not used. Those tools can't run the JS code, so they are not able to check interpolated paths.
Reference: https://medium.com/doctolib/keys-to-efficient-i18n-ec38fd461d7d#e344

src/popup/pages/FungibleTokens/TokenContainer.vue
```diff
@@ -123,16 +123,17 @@ export default defineComponent({
   setup(props, { root }) {
     const { getSdk } = useSdk();
 
+    const tokenType = root.$route.meta?.isCoinDetails ? 'coin' : 'token';
```
peronczyk, line 126: Also I suggest to avoid using magic strings.

peronczyk (changes requested):

src/popup/pages/FungibleTokens/TokenContainer.vue
CedrikNikita: I think @peronczyk suggested to use constants defined in `routeNames`.

peronczyk (approved):

src/popup/pages/FungibleTokens/TokenContainer.vue
```diff
@@ -184,20 +191,15 @@ export default defineComponent({
...
       if (isContract(contractId) && !isAe) {
         await getSdk();
         tokenPairs.value = await root.$store.dispatch('fungibleTokens/getContractTokenPairs', contractId);
       }
       loading.value = false;
-    });
-    onBeforeUnmount(() => {
-      root.$store.commit('setPageTitle', null);
+
+      console.info('========================');
```
CedrikNikita, line 200: This should be removed

CedrikNikita (changes requested):

CedrikNikita (approved):

PR #1743, opened 2022-12-16
[SW-512] Header route meta fix and general route meta simplification

peronczyk:


src/popup/components/Header.vue
```diff
@@ -85,60 +91,46 @@ export default {
...
+    const showHeaderNavigation = computed(() => !!routeMeta.value?.showHeaderNavigation);
+    const isLogoDisabled = computed(() => root.$route.name === ROUTE_ACCOUNT);
+    const titleTruncated = computed(
+      () => routeMeta.value?.title ? root.$t(`pages.titles.${routeMeta.value.title}`) : '',
+    );
+
+    function back() {
+      if (root.$store.state.route.from.name === ROUTE_INDEX) {
+        return root.$router.push({
+          name: (isLoggedIn.value) ? ROUTE_ACCOUNT : ROUTE_INDEX,
```
pstachel, line 106:
```suggestion
          name: isLoggedIn.value ? ROUTE_ACCOUNT : ROUTE_INDEX,
```

src/popup/components/Header.vue
```diff
@@ -85,60 +91,46 @@ export default {
...
-        this.isLoggedIn ? '/account' : '/',
-      );
-    },
+      return root.$router.back();
+    }
+
+    function close() {
+      root.$router.replace({
+        name: (isLoggedIn.value) ? ROUTE_ACCOUNT : ROUTE_INDEX,
+      });
```
pstachel, line 119: this code is used above - maybe you could extract it to a computed?

pstachel (changes requested):

src/popup/components/Header.vue
peronczyk: OK, changed.

src/popup/components/Header.vue
peronczyk: Actually some wise developer once said to me that if I wrap the test in ternary operation in braces it will be easier and faster to analyse the code we see. After years of coding I'm sure that's true.

Because you follow the code from top to bottom, from left to right. So when you see something like this:
```
const a = something...
```
you assume that this is a regular assignment. But When your eyes rich something like this:
```
const a = (something...
```
you immediately know that there will be some logic without seeing whole line.

pstachel (approved):

src/popup/components/Header.vue
```diff
@@ -85,60 +91,43 @@ export default {
...
-  methods: {
-    back() {
-      if (this.$store.state.route.from.path === '/') {
-        this.$router.push(this.isLoggedIn ? '/account' : '/');
-        return;
+  setup(props, { root }) {
+    const isLoggedIn = useGetter('isLoggedIn');
+    const currentHomeRouteName = computed(() => (isLoggedIn.value) ? ROUTE_ACCOUNT : ROUTE_INDEX);
+    const routeMeta = computed(() => root.$route.meta as WalletRouteMeta);
+    const showHeaderNavigation = computed(() => !!routeMeta.value?.showHeaderNavigation);
```
konradodo, line 98:
```suggestion
    const showHeaderNavigation = computed<boolean>(() => !!routeMeta.value?.showHeaderNavigation);
```

src/popup/components/Header.vue
```diff
@@ -85,60 +91,43 @@ export default {
...
+    const isLogoDisabled = computed(() => root.$route.name === ROUTE_ACCOUNT);
```
konradodo, line 99:
```suggestion
    const isLogoDisabled = computed<boolean>(() => root.$route.name === ROUTE_ACCOUNT);
```

src/popup/components/Header.vue
```diff
@@ -85,60 +91,43 @@ export default {
...
+    const titleTruncated = computed(
+      () => routeMeta.value?.title ? root.$t(`pages.titles.${routeMeta.value.title}`) : '',
+    );
```
konradodo, line 102:
```suggestion
    const titleTruncated = computed<string>(
      () => routeMeta.value?.title ? root.$t(`pages.titles.${routeMeta.value.title}`) : '',
    );
```

konradodo (approved):

src/popup/components/Header.vue
peronczyk: The type for this computed property is automatically inferred so there is no need to do force it. I mean: without adding `<boolean>` the TS engine knows that the `showHeaderNavigation` is a boolean.

src/popup/components/Header.vue
peronczyk: IDEM

src/popup/components/Header.vue
peronczyk: Actually it's not string but `TranslateResult`. Forcing it to be a `<string>` creates an error. But in practice the type is inferred correctly.

CedrikNikita:
This PR introducing new header logic to the `Auctions` page, which is incorrect.

PR #1746, opened 2022-12-19
[SW-495] show active network on header

ifaouibadi:


src/popup/components/Header.vue
```diff
@@ -93,12 +93,13 @@ export default defineComponent({
   },
   setup(props, { root }) {
     const isLoggedIn = useGetter('isLoggedIn');
+    const activeNetwork = useGetter('activeNetwork');
```
peronczyk, line 96:
```suggestion
    const activeNetwork = useGetter<INetwork>('activeNetwork');
```

src/popup/components/Header.vue
```diff
@@ -93,12 +93,13 @@ export default defineComponent({
...
     const currentHomeRouteName = computed(() => (isLoggedIn.value) ? ROUTE_ACCOUNT : ROUTE_INDEX);
     const routeMeta = computed(() => root.$route.meta as WalletRouteMeta);
     const showHeaderNavigation = computed(() => !!routeMeta.value?.showHeaderNavigation);
     const isLogoDisabled = computed(() => root.$route.name === ROUTE_ACCOUNT);
     const titleTruncated = computed(
-      () => routeMeta.value?.title ? root.$t(`pages.titles.${routeMeta.value.title}`) : '',
+      () => routeMeta.value?.title ? root.$t(`pages.titles.${routeMeta.value.title}`) : activeNetwork.value?.name,
```
peronczyk, line 102: I'm not sure about this logic. Now the network name will be displayed everywhere we don't have the title. But maybe the network name should only be visible on the Dashboard?

peronczyk (changes requested):

src/popup/components/Header.vue
```diff
@@ -26,7 +26,22 @@
       </Component>
     </div>
 
+    <BtnPlain
```
peronczyk, line 29: I see little problem with this - this button takes too much space in the header. User can click on empty space after the button which is not intuitive in my opinion.

src/popup/components/Header.vue
```diff
@@ -26,7 +26,22 @@
...
+      v-if="showActiveNetwork"
+      :to="{ name: 'network-settings' }"
+      extend
+      class="network-btn"
+    >
+      <div
+        class="circle"
+        :class="{
+          [nodeStatus]: true
+        }"
```
peronczyk, line 39:
```suggestion
        :class="[ nodeStatus ]"
```

src/popup/components/Header.vue
```diff
@@ -26,7 +26,22 @@
       </Component>
     </div>
 
+    <BtnPlain
+      v-if="showActiveNetwork"
```
peronczyk, line 30: I suggest to use the `showHeaderNavigation` instead of being dependant on the route name.

src/popup/components/Header.vue
```diff
@@ -87,6 +104,9 @@ export default defineComponent({
   },
   setup(props, { root }) {
     const isLoggedIn = useGetter('isLoggedIn');
+    const nodeStatus = useState('nodeStatus');
+    const activeNetwork = useGetter<INetwork>('activeNetwork');
+    const showActiveNetwork = computed<boolean>(() => root.$route.name === ROUTE_ACCOUNT);
```
peronczyk, line 109: Can be replaced with `showHeaderNavigation` located below.

src/popup/components/Header.vue
```diff
@@ -173,6 +196,35 @@ export default defineComponent({
...
+
+    display: inline-flex;
+    align-items: center;
+    justify-content: flex-start;
+    color: rgba(variables.$color-white, 0.75);
+
+    .circle {
+      width: 6px;
+      height: 6px;
+      border-radius: 6px;
```
peronczyk, line 210:
```suggestion
      border-radius: 100%;
```

peronczyk (changes requested):

src/popup/components/Header.vue
ifaouibadi: But that button is only related to the current Active Network and it only navigates the networks page

src/popup/components/Header.vue
ifaouibadi: `nodeStatus `  is needed to decide what color state should be used for 
07DCC4A2-C54B-432E-BD52-5D13E9097885

`activeNetwork ` needed to display which network is active 

`showActiveNetwork ` is to only display that block when we're on the dashboard screen

peronczyk: I'm referring only to `showActiveNetwork`. In my opinion it's obsolete as we should not base on the route name, but on the route meta option.

peronczyk (approved):

pstachel (approved):

paolomolo:
LGTM

PR #1751, opened 2022-12-20
[SW-485] transactions labels

ifaouibadi:


src/composables/transactionToken.ts
```diff
@@ -99,9 +105,70 @@ export const useTransactionToken = (
     || getDexContracts.value.wae.includes(transaction.value.tx.contractId)
   ));
 
+  const labels = computed(() => { //
```
peronczyk, line 108: 1. I suggest to type the returned value as there are multiple returns and there is no easy way to establish this.

2. Unnecessary comment start at the beginning.
3. Rather subjective, but maybe having longer name would be meaningful for this computed property?

src/popup/components/TransactionItem.vue
```diff
@@ -80,78 +77,17 @@ export default defineComponent({
...
-          root.$t('transaction.type.spendTx'),
-          props.transaction.tx.callerId === account.value.address
-            ? root.$t('transaction.spendType.out')
-            : root.$t('transaction.spendType.in'),
-        ];
-      }
-
-      return props.transaction.pending ? [] : [transactionTypes[txType.value]];
-    });
+    // const labels = computed(() => ([]));
```
peronczyk, line 90: I guess this can be removed.

src/popup/components/TransactionOverview.vue
```diff
@@ -103,6 +103,13 @@ export default {
     if (this.tx.recipientId?.startsWith('nm_')) {
       this.name = (await this.$store.state.middleware.getNameById(this.tx.recipientId)).name;
     }
+    if (!this.transaction.function) {
+      this.transaction.function = 'deposit';
+    }
+    console.info('========================');
```
peronczyk, line 109: :)

src/popup/components/TransactionOverview.vue
```diff
@@ -103,6 +103,13 @@ export default {
     if (this.tx.recipientId?.startsWith('nm_')) {
       this.name = (await this.$store.state.middleware.getNameById(this.tx.recipientId)).name;
     }
+    if (!this.transaction.function) {
+      this.transaction.function = 'deposit';
```
peronczyk, line 107: I suggest to avoid using magic strings.

peronczyk (changes requested):

src/composables/transactionLabels.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    const transactionTypes = i18n.t('transaction.type') as Record<TransactionType, any>;
+
+    if (txType.value?.startsWith('name')) {
+      return [AENS, transactionTypes[txType.value]];
+    }
+    if (txType.value === SCHEMA.TX_TYPE.spend) {
+      return [
+        i18n.t('transaction.type.spendTx'),
+        getTxDirection.value(transaction) === 'sent'
```
peronczyk, line 51: I suggest to avoid using magic strings.

src/composables/transactionLabels.ts
```diff
@@ -0,0 +1,102 @@
...
+        i18n.t('pages.token-details.tip'),
+        transaction.claim
+          ? i18n.t('transaction.spendType.in')
+          : i18n.t('transaction.spendType.out'),
+      ];
+    }
+    if (
+      txType.value === SCHEMA.TX_TYPE.contractCall
+      && availableTokens.value[transaction.tx.contractId]
+      && (transaction.tx.function === 'transfer' || transaction.incomplete)
```
peronczyk, line 87: I suggest to avoid using magic strings.

src/composables/transactionLabels.ts
```diff
@@ -0,0 +1,102 @@
...
+  DEX,
+} from '../popup/utils';
+import { useGetter, useState } from './vuex';
+
+export const useTransactionLabels = (
+  transaction: ITransaction | undefined,
+  defaultLabels: string[] = [],
+) => {
+  const activeNetwork = useGetter('activeNetwork');
+  const account = useGetter('account');
```
peronczyk, line 20: I suggest to type this with the `IAccount` interface.

src/composables/transactionLabels.ts
```diff
@@ -0,0 +1,102 @@
...
+  AENS,
+  DEX,
+} from '../popup/utils';
+import { useGetter, useState } from './vuex';
+
+export const useTransactionLabels = (
+  transaction: ITransaction | undefined,
+  defaultLabels: string[] = [],
+) => {
+  const activeNetwork = useGetter('activeNetwork');
```
peronczyk, line 19: I suggest to type this with `INetwork` interface.

src/popup/components/TransactionInfo.vue
```diff
@@ -75,20 +76,19 @@ export default {
     txFunction: { type: String, default: null },
     sender: { type: Object, required: true },
     recipient: { type: Object, required: true },
+    tx: { type: Object, default: null },
```
peronczyk, line 79:
```suggestion
    tx: { type: Object as PropType<ITransaction>, default: null },
```

src/popup/components/TransactionOverview.vue
```diff
@@ -5,106 +5,125 @@
...
-      switch (this.txType) {
+  setup(props, { root }) {
+    const name = ref<string>();
+    const getTxType = useGetter('getTxType');
+    const getTxDirection = useGetter('getTxDirection');
+    const getExplorerPath = useGetter('getExplorerPath');
+    const getDexContracts = useGetter('getDexContracts');
+    const getPreferred = computed(() => root.$store.getters['names/getPreferred']);
+    const _account = useGetter<IAccount>('account');
+    const account: any = computed<any>(() => ({
```
peronczyk, line 41: I suggest to avoid using `any` to type things that can be typed. Also if this computed is not an `IAccount` I suggest to rename it to something like `parsedAccount` or `computedAccount` or `accountFor...`.

src/popup/components/TransactionOverview.vue
```diff
@@ -5,106 +5,125 @@
...
-    }),
-    transaction() {
-      switch (this.txType) {
+  setup(props, { root }) {
+    const name = ref<string>();
+    const getTxType = useGetter('getTxType');
+    const getTxDirection = useGetter('getTxDirection');
+    const getExplorerPath = useGetter('getExplorerPath');
+    const getDexContracts = useGetter('getDexContracts');
+    const getPreferred = computed(() => root.$store.getters['names/getPreferred']);
```
peronczyk, line 39: As this is not an actual function I suggest to rename it to `prefered...` (remove "get").

peronczyk (changes requested):

src/composables/transactionLabels.ts
```diff
@@ -0,0 +1,114 @@
...
+  TX_FUNCTION_TRANSFER,
+  TX_FUNCTION_SENT,
+} from '../popup/utils';
+import { useGetter, useState } from './vuex';
+
+export const useTransactionLabels = (
+  transaction: ITransaction | undefined,
+  defaultLabels: string[] = [],
+) => {
+  const activeNetwork = useGetter<INetwork>('activeNetwork');
```
konradodo, line 28: After RxJs @ifaouibadi  you cant use useGetter and useState in composable

src/composables/transactionLabels.ts
```diff
@@ -0,0 +1,114 @@
...
+  DEX,
+  TX_FUNCTION_TIP,
+  TX_FUNCTION_RETIP,
+  TX_FUNCTION_TRANSFER,
+  TX_FUNCTION_SENT,
+} from '../popup/utils';
+import { useGetter, useState } from './vuex';
+
+export const useTransactionLabels = (
+  transaction: ITransaction | undefined,
```
konradodo, line 25:
```suggestion
  transaction?: ITransaction,
```

src/composables/transactionLabels.ts
```diff
@@ -0,0 +1,114 @@
...
+    && availableTokens.value[transaction.tx.contractId]);
+
+  const isDex = computed(() => transaction
+    && getDexContracts.value && transaction.tx.contractId && (
+    getDexContracts.value.router.includes(transaction.tx.contractId)
+    || getDexContracts.value.wae.includes(transaction.tx.contractId)
+  ));
+
+  const labels = computed<string[]>(() => {
+    if (!transaction || !transaction.tx) return defaultLabels;
```
konradodo, line 50:
```suggestion
    if (!(transaction && transaction.tx)) {
     return defaultLabels;
     }
```

src/popup/components/TransactionInfo.vue
```diff
@@ -75,20 +80,19 @@ export default {
...
-        : '';
-    },
-    getTitle() {
-      return FUNCTION_TYPE_DEX.pool.includes(this.txFunction)
-        ? this.$t('transaction.dexType.pool')
-        : this.title;
-    },
+  setup(props) {
+    const { labels } = useTransactionLabels(
+      { tx: props.tx } as any,
```
pstachel, line 87:
```suggestion
      { tx: props.tx } as ITx,
```

src/popup/components/TransactionOverview.vue
```diff
@@ -5,106 +5,130 @@
...
-    }),
-    transaction() {
-      switch (this.txType) {
+  setup(props, { root }) {
+    const name = ref<string>();
+    const getTxType = useGetter('getTxType');
+    const getTxDirection = useGetter('getTxDirection');
+    const getExplorerPath = useGetter('getExplorerPath');
+    const getDexContracts = useGetter('getDexContracts');
+    const getPreferredName = computed(() => root.$store.getters['names/getPreferred']);
```
pstachel, line 44:
```suggestion
    const getPreferredName = useGetter('names/getPreferred');
```

src/popup/components/TransactionOverview.vue
```diff
@@ -5,106 +5,130 @@
...
-  },
-  async mounted() {
-    await watchUntilTruthy(() => this.$store.state.middleware);
-    if (this.tx.recipientId?.startsWith('nm_')) {
-      this.name = (await this.$store.state.middleware.getNameById(this.tx.recipientId)).name;
-    }
+    });
+
+    onMounted(async () => {
+      await watchUntilTruthy(() => root.$store.state.middleware);
```
pstachel, line 117: you could use `const middleware = useState('middleware')`

pstachel (changes requested):

pstachel (approved):

peronczyk (changes requested):

src/popup/components/TransactionItem.vue
```diff
@@ -52,24 +52,22 @@
 </template>
```
peronczyk, line 52: Is `toUpperCase()` still required in line 45?

src/popup/components/TransactionOverview.vue
```diff
@@ -5,106 +5,131 @@
...
+import { IAccount } from '../../types';
+
 import TransactionInfo from './TransactionInfo.vue';
 
-export default {
+export default defineComponent({
+  name: 'TransactionOverview',
   components: { TransactionInfo },
   props: {
     tx: { type: Object, required: true },
```
peronczyk, line 35: After switching to TypeScript I suggest to use `ITx` PropType.

src/popup/components/TransactionOverview.vue
```diff
@@ -5,106 +5,131 @@
...
+  setup(props, { root }) {
+    const name = ref<string>();
+    const middleware = useState('middleware');
+    const getTxType = useGetter('getTxType');
+    const getTxDirection = useGetter('getTxDirection');
+    const getExplorerPath = useGetter('getExplorerPath');
+    const getDexContracts = useGetter('getDexContracts');
+    const getPreferredName = useGetter('names/getPreferred');
+    const account = useGetter<IAccount>('account');
+    const parsedAccount = computed<IAccount>(() => ({
```
peronczyk, line 47: For the computed properties I suggest to use the type for the return. So in this case:
```
computed((): IAccount => ...
```
But in this case you won't be able to use the `IAccount` because of missing `label` and `url`. So new (eg.: IAccountExtended`) type should be created.

src/popup/components/TransactionOverview.vue
```diff
@@ -5,106 +5,131 @@
...
+      url: getExplorerPath.value(account.value?.address),
+    }));
+    const isDexRecipient = computed(() => (
+      [...getDexContracts.value.router, ...getDexContracts.value.wae]
+        .includes(props.tx?.contractId)
+    ));
+
+    const txType = computed(() => getTxType.value({ tx: props.tx }));
+
+    const transaction = computed(() => {
```
peronczyk, line 59: I suggest to add proper type to the function return.

src/popup/utils/constants.ts
```diff
@@ -447,3 +447,10 @@ export const INPUT_MESSAGE_STATUSES = {
   warning: 'warning',
   error: 'error',
 } as const;
+
+export const TX_FUNCTION_TIP = 'tip';
+export const TX_FUNCTION_RETIP = 'retip';
+export const TX_FUNCTION_TRANSFER = 'transfer';
+export const TX_FUNCTION_SENT = 'sent';
+export const TX_FUNCTION_RECEIVED = 'received';
+export const TX_FUNCTION_DEPOSIT = 'deposit';
```
peronczyk, line 456: It's really nice that we have TypeScript right now, because some inconsistencies can be noticed. Like for example the possible values for the `transaction.tx.function`. Unitil now it should use the `TxFunction` type located in `src/types/index.ts`.
But here you listed other options, in which some of them are the same.
Please establish the correct values for the `transaction.tx.function`, update the `TxFunction` type and use it to create the list of constants. Or (which I suggest) do the opposite:

```
export const TX_FUNCTION = {
  tip: 'tip',
  retip: 'retip',
  ...
} as const;
```

and then in the types file:
```
import { TX_FUNCTION } ...

export type TxFunction = ObjectValues<typeof TX_FUNCTION>;
```

src/popup/components/TransactionOverview.vue
```diff
@@ -5,106 +5,131 @@
...
-  async mounted() {
-    await watchUntilTruthy(() => this.$store.state.middleware);
-    if (this.tx.recipientId?.startsWith('nm_')) {
-      this.name = (await this.$store.state.middleware.getNameById(this.tx.recipientId)).name;
-    }
+    });
+
+    onMounted(async () => {
+      await watchUntilTruthy(() => middleware);
+      if (props.tx.recipientId?.startsWith('nm_')) {
```
peronczyk, line 119: I suggest to use `validateHash` util function:
```
const { valid, prefix } = validateHash(props.tx.recipientId);
if (valid && prefix === HASH_PREFIX_NAME) {
  ...
```

(prefered) Or create separate util function similar to `isContract` that would wrap above logic.

peronczyk (changes requested):

src/popup/components/TransactionItem.vue
ifaouibadi: replaced with CSS style

subhod-i:
@ifaouibadi the transaction labels are handled differently in the new SDK version. It's now an Enum of its own. We should probably wait for the SDK PR to get merged before merging this

src/popup/components/TransactionInfoDetailsParty.vue
```diff
@@ -10,6 +10,7 @@
       class="name"
     >
       <Truncate
+        v-if="txParty.name || txParty.label"
```
pstachel, line 13: It can't be resolved like that - the design will break. Check out the solution from SW-493.
https://github.com/aeternity/superhero-wallet/pull/1763/files#diff-81e706951f062c4bd205ee4449b419270ae942e6161be2d6936446a537c0da64R14

src/composables/transactionToken.ts
```diff
@@ -102,11 +118,73 @@ export function useTransactionToken({
...
 
-  const isDex = computed(() => transaction.value
+  const isDex = computed<boolean>(() => !!(transaction.value
     && getDexContracts.value && transaction.value.tx.contractId && (
     getDexContracts.value.router.includes(transaction.value.tx.contractId)
     || getDexContracts.value.wae.includes(transaction.value.tx.contractId)
-  ));
+  )));
+
+  const labels = computed<TranslateResult[]>(() => {
```
peronczyk, line 127: If the type `TranslateResult[]` is put in square brackets after the `computed` it means that we are just guessing what will be returned. So this code for example will be correct:
```
const a = computed<number>(() => 'lorem ipsum');
```

In order to properly type this I suggest to type the returned function. This way the returned body will also by validated:
```
computed((): (string | TranslateResult)[] => ...
```

peronczyk (changes requested):

src/types/index.ts
```diff
@@ -16,6 +21,8 @@ export type ObjectValues<T> = T[keyof T];
 
 export type InputMessageStatus = ObjectValues<typeof INPUT_MESSAGE_STATUSES>;
 
+export type TxFunctionTypes = ObjectValues<typeof TX_FUNCTIONS>;
```
peronczyk, line 24: This is not that simple. In some places we are using `snake_case` names and in other `camelCase`. I had a problem with this in one of my incoming PRs.

I solved this by using this types:
```
/**
 * TxFunction names coming directly from the API or ready to be sent.
 */
export type TxFunctionRaw = ObjectValues<typeof TX_FUNCTIONS>;

/**
 * TxFunctions used internally by the app.
 */
export type TxFunctionParsed = keyof typeof TX_FUNCTIONS;

export type TxFunction = TxFunctionRaw | TxFunctionParsed;
```

Here's the change I've made:
https://github.com/aeternity/superhero-wallet/pull/1780/files#diff-a9ba9cbedd19c9f66d564d2e89912890209b98f0a7ef19187877d2587300e476

peronczyk (changes requested):

peronczyk (approved):

PR #1754, opened 2022-12-20
[SW-519] Buttons and table rows color unification

peronczyk:
List of changes: https://aeternity-blockchain.atlassian.net/browse/SW-519?focusedCommentId=12472


pstachel (approved):

src/popup/components/buttons/BtnMain.vue
```diff
@@ -80,6 +80,7 @@ export default defineComponent({
   @extend %face-sans-16-regular;
 
   position: relative;
+  z-index: 1;
```
CedrikNikita, line 83: Why `z-index` was introduced?

peronczyk: Because the `before` pseudo element has the z-index `-1` so if the parent has no positive z-index the `before` would not be visible (will be placed "under" the body.

src/popup/components/PanelItem.vue
```diff
@@ -56,36 +56,22 @@ export default {
...
+  justify-content: space-between;
   text-decoration: none;
   text-align: left;
-  background: rgba(variables.$color-white, 0.08);
-  color: rgba(variables.$color-white, 0.95);
-  -webkit-transition: 0.4s;
-  transition: 0.4s;
-  opacity: 0.75;
+  color: rgba(variables.$color-white, 0.75);
+  transition: variables.$transition-interactive;
```
pstachel, line 72: If you touched imports refactoring in that PR maybe you could also remove `variables.` here?

src/popup/components/buttons/BtnBox.vue
```diff
@@ -1,98 +1,103 @@
...
+      v-if="icon"
+      class="icon"
+      :icon="icon"
+      :is-boxed="isBig"
+    />
+
+    <span
+      v-if="text"
+      class="text"
+    >{{ text }}</span>
```
pstachel, line 23:
```suggestion
    <span
      v-if="text"
      class="text"
    >
      {{ text }}
    </span>
```

src/popup/components/buttons/BtnBox.vue
```diff
@@ -1,98 +1,103 @@
...
+
+    <span
+      v-if="subtitle"
+      class="subtitle"
+    >{{ subtitle }}</span>
```
pstachel, line 28:
```suggestion
    <span
      v-if="subtitle"
      class="subtitle"
    >
      {{ subtitle }}
    </span>
```

pstachel (changes requested):

src/popup/components/buttons/BtnBase.vue
```diff
@@ -0,0 +1,138 @@
+<template>
+  <Component
+    :is="component"
+    :to="to"
+    :href="href"
+    :target="(href) ? '_blank' : null"
```
pstachel, line 6:
```suggestion
    :target="href ? '_blank' : null"
```

pstachel (approved):

src/popup/components/PanelItem.vue
peronczyk: OK, done.

src/popup/components/buttons/BtnBox.vue
peronczyk: I did different approach with the `v-text` because I didn't want to introduce additional spaces to the rendered code.
This code:
```
<span>
  {{ text }}
</span>
```
is transpiled to:
```
<span> {{ text }} </span>
```
From my experience this can lead to some problems. Are you OK with this?

src/popup/components/buttons/BtnBox.vue
peronczyk: Same here.

src/popup/components/buttons/BtnBox.vue
pstachel: Sure

src/popup/components/NotifyBell.vue
```diff
@@ -2,9 +2,9 @@
   <BtnIcon
```
CedrikNikita, line 2: Now `BtnIcon` doesn't have a default slot, so the inner element won't be visible.

CedrikNikita (changes requested):

src/popup/components/NotifyBell.vue
peronczyk: Fixed.

CedrikNikita (approved):
Overall, the code changes seem reasonable. I have not tested it. Additional thanks for removing the dashboard folder.

PR #1757, opened 2022-12-22
[SW-527] Fix scrollbars visibility in firefox

peronczyk:


src/popup/components/Scrollable.vue
```diff
@@ -7,7 +7,7 @@
 </template>
 
 <style lang="scss" scoped>
-@use '../../styles/variables';
+@use '../../styles/variables' as *;
```
pstachel, line 10: If the whole project we use variables as `variable.$color-some-color` should we diverge from it in favour of using just a variable name?

pstachel (approved):

src/popup/components/Scrollable.vue
peronczyk: We are not using any 3rd party frameworks so in my opinion it's easier to not use the `viariables` prefix. We can gradually migrate into this approach.

konradodo (approved):

CedrikNikita (dismissed):
Test ability to merge by adding a commit as a request change.

src/popup/App.vue
```diff
@@ -17,7 +17,7 @@
       v-show="!qrScannerOpen"
       ref="innerElement"
       class="app-inner"
-      :class="{ 'show-scrollbar': $route.meta.showScrollbar }"
+      :class="{ 'show-scrollbar': showScrollbar }"
```
CedrikNikita, line 20: Test ability to merge by adding a commit as a request change.

CedrikNikita (changes requested):

PR #1762, opened 2022-12-29
[SW-536] Convert components using balances to Composition API

peronczyk:


src/popup/components/BalanceInfo.vue
```diff
@@ -3,34 +3,48 @@
...
+
+    const balances = rxJsObservableToVueState<BigNumber[]>(
+      (root.$store.state as any).observables.balances,
+    );
+
+    const idx = computed(() => props.accountIdx === -1 ? activeIdx.value : props.accountIdx);
+    const balance = computed(() => balances.value[idx.value].toNumber());
+    const formattedCurrency = computed(
+      () => formatCurrency.value(balance.value * currentCurrencyRate.value),
+    );
```
pstachel, line 40: you can consider using `convertToCurrencyFormatted` instead of `formatCurrency` and `currentCurrencyRate`

src/popup/components/BalanceInfo.vue
```diff
@@ -3,34 +3,48 @@
...
+    const activeIdx = useState('accounts', 'activeIdx');
+    const formatCurrency = useGetter('formatCurrency');
+    const currentCurrencyRate = useGetter('currentCurrencyRate');
+
+    const balances = rxJsObservableToVueState<BigNumber[]>(
+      (root.$store.state as any).observables.balances,
+    );
+
+    const idx = computed(() => props.accountIdx === -1 ? activeIdx.value : props.accountIdx);
+    const balance = computed(() => balances.value[idx.value].toNumber());
```
pstachel, line 37: If I'm not mistaken we do have a `balance` state in observables

src/popup/components/InputAmount.vue
```diff
@@ -1,82 +1,88 @@
...
 import InputField from './InputField.vue';
 
-export default {
+export default defineComponent({
   components: {
     InputField,
   },
   props: {
+    value: { type: [String, Number], default: '' },
     noToken: { type: Boolean },
```
pstachel, line 51:
```suggestion
    noToken: Boolean,
```

src/popup/components/InputAmount.vue
```diff
@@ -1,82 +1,88 @@
...
+      (root.$store.state as any).observables.balance,
+    );
+
+    const formatCurrency = useGetter('formatCurrency');
+    const currentCurrencyRate = useGetter('currentCurrencyRate');
+    const hasError = computed(() => (root as any).$validator.errors.has('amount'));
+    const max = computed(() => balance.value.minus(fee.value).toNumber());
+    const currencyAmount = computed(
+      () => formatCurrency.value((+(props.value || 0) * currentCurrencyRate.value).toFixed(2)),
+    );
```
pstachel, line 68: As mentioned before - you could use `convertToCurrencyFormatted`

src/popup/components/InputAmountV2.vue
```diff
@@ -61,76 +61,81 @@
...
+    // eslint-disable-next-line no-unused-vars
+    const getAeternityToken = useGetter<(o: any) => IAsset>('fungibleTokens/getAeternityToken');
+    const formatCurrency = useGetter('formatCurrency');
+
+    const balance = rxJsObservableToVueState<BigNumber>(
+      (root.$store.state as any).observables.balance,
+    );
+    const balanceCurrency = rxJsObservableToVueState<number>(
+      (root.$store.state as any).observables.balance,
+    );
```
pstachel, line 101: `balance` and `balanceCurrency` are exactly the same. I think it should look like this

```suggestion
    const balance = rxJsObservableToVueState<BigNumber>(
      (root.$store.state as any).observables.balance,
    );
    const balanceCurrency = rxJsObservableToVueState<number>(
      (root.$store.state as any).observables.balanceCurrency,
    );
```

pstachel (changes requested):

src/popup/components/BalanceInfo.vue
peronczyk: Yes, I thought about the same but this component is not showing the balance of the current account but the balance based on the IDX. So we are picking the balance from the list.

src/popup/components/BalanceInfo.vue
peronczyk: OK, makes sense. Thanks, changed.

src/popup/components/InputAmount.vue
peronczyk: Thanks, makes sense. Changed.

src/popup/components/InputAmountV2.vue
peronczyk: Uuuu, my bad. Thanks for finding this out.

src/popup/components/InputAmount.vue
```diff
@@ -1,82 +1,84 @@
 <template>
   <InputField
-    v-validate="ownValidation ? validation : {
+    v-validate="{
       required: true,
       min_value_exclusive: 0,
-      ...+balance.minus(fee) > 0 ? { max_value: max } : {},
       enough_ae: fee.toString(),
-      ...validation,
+      ...max > 0 ? { max_value: max } : {},
```
konradodo, line 7:
```suggestion
      max_value: max > 0 ? max : undefined,
```

src/popup/components/InputAmount.vue
peronczyk: Makes sense. Changed.

src/popup/components/InputAmount.vue
peronczyk: OK

konradodo (approved):

pstachel (approved):

PR #1765, opened 2023-01-05
[SW-539] Use transaction tags in transaction items

peronczyk:


src/popup/components/TransactionItem.vue
```diff
@@ -38,14 +38,16 @@
...
-            {{ label.toUpperCase() }}
-          </label>
+            v-text="label"
+          />
         </div>
-        <span v-if="fiatAmount">{{ fiatAmount }}</span>
+        <span
+          v-if="fiatAmount"
+          class="fiat-amount"
+        >{{ fiatAmount }}</span>
```
konradodo, line 50:
```suggestion
        <span
          v-if="fiatAmount"
          class="fiat-amount"
        >
        {{ fiatAmount }}
        </span>
```

konradodo (approved):

src/popup/components/TransactionItem.vue
peronczyk: Why? :)

PR #1767, opened 2023-01-09
[SW-259] Refactor: replace RxJS with Vue Composables

peronczyk:


CedrikNikita:
For the new account we are getting such errors, which suppose to be handled as a valid behaviour.
`Uncaught (in promise) Error: v3/accounts/ak_Kqz3NbLUHsmxSB7tAqh4ixq44Hq6rAfSjXNCLP226zABFErdh error: Account not found
    at http (index.js?4f8d:103:1)
    at async request (swagger.mjs?0a14:148:1)
    at async Object.balance (node.mjs?63e1:138:1)
    at async Promise.all (index 0)
`

CedrikNikita (changes requested):
For the new account we are getting such errors, which suppose to be handled as a valid behaviour.
`Uncaught (in promise) Error: v3/accounts/ak_Kqz3NbLUHsmxSB7tAqh4ixq44Hq6rAfSjXNCLP226zABFErdh error: Account not found at http (index.js?4f8d:103:1) at async request (swagger.mjs?0a14:148:1) at async Object.balance (node.mjs?63e1:138:1) at async Promise.all (index 0)
`

src/popup/components/TotalWalletAmount.vue
```diff
@@ -10,27 +10,27 @@
...
         (previousValue, currentValue) => previousValue.plus(currentValue),
         new BigNumber(0),
       );
-      return convertToCurrencyFormatted.value(total);
+      return convertToCurrencyFormatted.value(total.toNumber());
     });
 
     return {
+      balances,
+      balaX,
```
CedrikNikita, line 33: Is`balaX` variable actually used?

peronczyk: My mistake. I was using it for debugging purposes. I'll remove it.

peronczyk:
> For the new account we are getting such errors, which suppose to be handled as a valid behaviour. `Uncaught (in promise) Error: v3/accounts/ak_Kqz3NbLUHsmxSB7tAqh4ixq44Hq6rAfSjXNCLP226zABFErdh error: Account not found at http (index.js?4f8d:103:1) at async request (swagger.mjs?0a14:148:1) at async Object.balance (node.mjs?63e1:138:1) at async Promise.all (index 0) `

True, I fixed this. Please see now @CedrikNikita .

CedrikNikita (changes requested):
Changes made in `package-lock.json` also need to be committed.

peronczyk:
> Changes made in `package-lock.json` also need to be committed.

Done

pstachel (approved):

src/composables/balances.ts
```diff
@@ -0,0 +1,93 @@
...
+  isNotFoundError,
+  handleUnknownError,
+} from '../popup/utils';
+import { useSdk } from './sdk';
+import { createPollingBasedOnMountedComponents } from './composablesHelpers';
+
+type Balances = Record<string, BigNumber>;
+type BalancesRaw = Record<string, string>;
+
+interface IUseBalancesOptions {
```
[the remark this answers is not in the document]
peronczyk: Yeah, you have right. I'll change this.

src/composables/topHeader.ts
```diff
@@ -0,0 +1,35 @@
+import { computed, ref } from '@vue/composition-api';
+import type { Store } from 'vuex';
+import type { ITopHeader } from '../types';
+import { createPollingBasedOnMountedComponents } from './composablesHelpers';
+import { useSdk } from './sdk';
+
+interface UseTopHeaderOptions {
+  /**
+   * TODO: Temporary solution to avoid dependency circle
```
davidyuk, line 9: Recently I had a dependency cycle in types, and I think it is ok to import such dependency using `import type` then linter won't blame it. This would be a cycle dependency only on the stage of type checking. Would it be ok or would it complicate unit testing or so?

src/popup/components/TransferSendForm.vue
```diff
@@ -250,7 +252,7 @@ export default defineComponent({
       () => formModel.value.selectedAsset?.contractId !== AETERNITY_CONTRACT_ID,
     );
     const isMaxValue = computed<boolean>(
-      () => isNumbersEqual(formModel.value.amount, max.value),
+      () => isNumbersEqual(+formModel.value.amount, +max.value),
```
davidyuk, line 255:
```suggestion
      () => +formModel.value.amount === +max.value,
```
🤔

peronczyk: Makes sense as those 2 are not BigNumbers. Changed...

CedrikNikita (commented):
Why do you replace `useGetter` in some places with `computed` and in other places you do opposite?
It seems that you decided to replace them in `TransactionDetails`, because it is harder to unit test such things.

CedrikNikita (approved):
In general, code changes look resonable.

peronczyk:
> Why do you replace `useGetter` in some places with `computed` and in other places you do opposite? It seems that you decided to replace them in `TransactionDetails`, because it is harder to unit test such things.

Yes, in some places I removed it because it was causing dependency cycle. The file that defines `useGetter` and `useState` imports the store. So when I use them in a composable and then import this composable in any non Vue file ESLint is telling me that there is a dependency cycle.

Also for the `TransactionDetails` i replaced the `useGetter` with `computed` to make the unit test file much simpler and avoid mocking functions.

src/composables/topHeader.ts
davidyuk: It is possible to import not the whole store but only it's type using `import type`?

PR #1769, opened 2023-01-10
feat: composable for lastTransactions

konradodo:


src/composables/lastTransactions.ts
```diff
@@ -0,0 +1,21 @@
+import { computed } from '@vue/composition-api';
+import { ITransaction } from '../types';
+import store from '../store';
+import {
+  COUNT_OF_DASHBOARD_TRANSACTION,
+  sortTransaction,
+} from '../popup/utils';
+
+export const useLastTransaction = () => {
```
pstachel, line 9:
```suggestion
export const useCombinedLatestTransactions = () => {
```

src/composables/lastTransactions.ts
```diff
@@ -0,0 +1,21 @@
...
+  const availableTokens = computed<Record<string, ITransaction[]>>(
+    () => (store as any).state?.fungibleTokens?.transactions,
+  );
```
pstachel, line 12: you have set either wrong name of variable or have wrong getter

pstachel: Also please use useState here

src/composables/lastTransactions.ts
```diff
@@ -0,0 +1,21 @@
...
+  const lastTransactions = computed(() => {
```
pstachel, line 13:
```suggestion
  const latestTransactions = computed(() => {
```

src/composables/lastTransactions.ts
```diff
@@ -0,0 +1,21 @@
...
+    const allTransaction = Object.values(availableTokens.value).flatMap((t) => t);
```
pstachel, line 14:
```suggestion
    const allTransactions = Object.values(availableTokens.value).flatMap((t) => t);
```

src/popup/App.vue
```diff
@@ -20,7 +20,7 @@
       :class="{ 'show-scrollbar': $route.meta.showScrollbar }"
     >
       <Header v-if="showHeader" />
-
+      {{ lastTransactions }}
```
pstachel, line 23: shouldn't be here

pstachel (changes requested):

pstachel (commented):
Also I feel that there's no need for a composable here - combined transaction list will be used only as a separate component on dashboard, so whole logic can be encapsulated there

src/composables/lastTransactions.ts
konradodo: i can't, after @peronczyk  change we cant use useState, useGetter in composables

PR #1772, opened 2023-01-11
[SW-561] Input Field message property interface

peronczyk:


src/popup/components/InputField.vue
```diff
@@ -128,59 +148,62 @@ export default {
...
-  methods: {
-    checkIfNumber(event) {
+  setup(props, { emit }) {
+    const focused = ref(false);
+
+    const inputId = `input-${getCurrentInstance()?.uid}`;
+    const inputMode = computed(() => props.type === 'number' ? 'decimal' : 'text');
+    const hasError = computed(
+      () => (typeof props.message === 'object')
+        ? props.message?.status === 'error'
```
pstachel, line 158: You have declared a variable for statuses here
https://github.com/aeternity/superhero-wallet/pull/1772/files#diff-d176ac880b039692a793823c27562dd25951237c941c78490de5fc64e8eb6f82R114

src/popup/components/InputField.vue
```diff
@@ -128,59 +148,62 @@ export default {
...
+        : !!props.message,
+    );
+    const hasWarning = computed(
+      () => (typeof props.message === 'object')
+        ? props.message?.status === 'warning'
```
pstachel, line 163: You have declared a variable for statuses here
https://github.com/aeternity/superhero-wallet/pull/1772/files#diff-d176ac880b039692a793823c27562dd25951237c941c78490de5fc64e8eb6f82R114

src/popup/components/InputField.vue
```diff
@@ -128,59 +148,62 @@ export default {
...
+        : false,
+    );
+    const showMessage = computed(
+      () => (typeof props.message === 'object')
+        ? !props.message?.hideMessage
+        : !!props.message,
```
pstachel, line 169: Just a suggestion, but `typeof props.message === 'object'` is used three times in highlighted scope - what do you think of making a computed property that will check that?

pstachel (changes requested):

src/popup/components/InputField.vue
peronczyk: Yeah, that's true. Unfortunately TypeScript is complaining if I move the test into computed property because TS is not aware of the reactive nature of the computed properties and how this affects if statements :/
But I simplified those conditionals.

peronczyk: I tried again and I came up with a solution. Please see how I did it.

src/popup/components/InputField.vue
peronczyk: True, changed.

src/popup/components/InputField.vue
peronczyk: True, changed.

pstachel (approved):

ifaouibadi (approved):

PR #1779, opened 2023-01-16
[SW-585] fix(transactions): apply correct filter logic for pending transactions

CedrikNikita:


peronczyk (approved):

src/popup/components/TransactionList.vue
```diff
@@ -154,7 +154,9 @@ export default defineComponent({
         )
         .sort((a, b) => {
           const [aMicroTime, bMicroTime] = [a, b].map((tr) => (new Date(tr.microTime)).getTime());
-          return a.pending || (bMicroTime - aMicroTime);
+          return (a.pending && !b.pending && -1)
+            || (b.pending && !a.pending && 1)
+            || bMicroTime - aMicroTime;
```
peronczyk, line 159: If the condition is multiline I suggest to put all lines as separate ones. Eg.:
```
return (
  a
  || b
  || c
);
```

Liubov-crypto:
LGTM

PR #1780, opened 2023-01-16
[SW-565] Replace popups mixin with composition api

peronczyk:
… transaction resolvers


src/popup/pages/Popups/Connect.vue
```diff
@@ -56,78 +56,77 @@
...
+    const permission = computed<IPermission>(() => root.$store.state.permissions[props.app.host]);
+    const appName = computed(() => permission.value?.name || props.app.name);
+    const isConnected = useGetter('isConnected');
+    const getExplorerPath = useGetter('getExplorerPath');
+    const account = useGetter<IAccount>('account');
+    const accountExtended = computed(() => ({
+      ...account.value,
+      label: root.$t('transaction.overview.accountAddress'),
+      url: getExplorerPath.value(account.value.address),
+    }));
```
pstachel, line 104: Please group useGetters and computed properties together

src/popup/pages/Popups/MessageSign.vue
```diff
@@ -46,44 +46,52 @@
...
-    }),
+  setup(props, { root }) {
+    const isConnected = useGetter('isConnected');
+    const getExplorerPath = useGetter('getExplorerPath');
+    const account = useGetter<IAccount>('account');
+    const accountExtended = computed(() => ({
+      ...account.value,
+      label: root.$t('transaction.overview.accountAddress'),
+      url: getExplorerPath.value(account.value.address),
+    }));
```
pstachel, line 82: I remember that this exact same object is also used int transaction.ts if fetchOwnershipAccount function - maybe it could be extracted somewhere?

src/popup/utils/transactionTokenInfoResolvers.ts
```diff
@@ -1,24 +1,65 @@
...
+  tokens: ITokenList | null
+) => TransactionResolverReturnData;
+
+type TransactionResolverGenerator = (
+  tokenAMapper: (t: ITransaction) => any,
+  tokenBMapper: (t: ITransaction) => any,
+  poolTokenAmountMapper?: (t: ITransaction) => any,
+  liquidityMethod?: 'add' | 'remove'
+) => TransactionResolver;
+/* eslint-enable no-unused-vars */
```
pstachel, line 35: is it needed here?

pstachel (changes requested):

src/popup/utils/transactionTokenInfoResolvers.ts
peronczyk: You mean the eslint enable? It enables again what was disabled in line 11.

src/popup/pages/Popups/MessageSign.vue
peronczyk: I confirm that similar code exists in few places. I tried to create utility function, but besides the repeated `label` there was no common data to return. It was always merging in the account and getting explorer path from provided address. So I decided to use your interface to type those places for now. In the future we could think about something better.

src/popup/utils/transactionTokenInfoResolvers.ts
pstachel: Sorry, I misread that. No issues then

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -153,153 +183,190 @@ export default {
...
+      amount: getTxAmountTotal.value(props.transaction),
+      symbol: getTxSymbol.value(props.transaction),
+    }));
+
+    const filteredTxFields = computed(
+      () => TX_FIELDS_TO_DISPLAY.filter((field) => !!props.transaction[field]),
+    );
+
+    const swapTokenAmountData = computed(
+      (): ITokenResolved => swapDirection.value === 'maxSpent' ? tokenList.value[0] : tokenList.value[1],
```
konradodo, line 251: 'maxSpent' -> constant?

src/popup/pages/Popups/Connect.vue
```diff
@@ -55,78 +55,83 @@
...
+    reject: { type: Function as PropType<(e: Error) => void>, required: true },
   },
-  data() {
+  setup(props, { root }) {
+    const isConnected = useGetter('isConnected');
+    const getExplorerPath = useGetter('getExplorerPath');
+    const account = useGetter<IAccount>('account');
+
+    const permission = computed<IPermission>(() => root.$store.state.permissions[props.app.host]);
+    const appName = computed(() => permission.value?.name || props.app.name);
```
konradodo, line 104:
```suggestion
    const permissions = useState('permissions', props.app.host); 
    const appName = computed(() => permission.value?.name || props.app.name);
```

konradodo (approved):

src/popup/pages/Popups/Connect.vue
peronczyk: OK

src/popup/components/Modals/ConfirmTransactionSign.vue
peronczyk: I'm leaving this for now as this is a key used to grab translation. I need to dive deeper with this, like for example get rid of parsing of the translation paths.

pstachel (approved):

Issue #1782, opened 2023-01-17
Error coming from `api.coingecko` is breaking `balances` composable


CedrikNikita:
The account balance is shown as 0, in case the currencies is not loaded, which shouldn't be the case.
If an error is coming from coingecko, our app is handling it incorrectly and breaking with such issue:
```
popup.e0e6fee5.js:1 TypeError: Cannot read property 'usd' of undefined
    at currentCurrencyRate (VM24 popup.e0e6fee5.js:1)
    at t._wrappedGetters.<computed>.t._wrappedGetters.
```

PR #1727, opened 2022-12-05
[SW-89] Update  SDK to v12 (typescript version)

subhod-i:

PR #1783, opened 2023-01-17
[SW-586] Enable withdraw test

CedrikNikita:
- on reviewing #1727, found out an error that could've been intercepted by this disabled test,
- our automated tests were overloading `api.coingecko` and it was throwing `429 (Too Many Requests)`, which resulting in a broken `balances` composable (#1782)

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -10,7 +10,6 @@
   "ignoreTestFiles": [
     "tip.js",
     "transaction-details.js",
-    "transactions.js",
-    "withdraw.js"
+    "transactions.js"
   ]
 }
--- a/src/popup/components/Modals/SpendSuccess.vue
+++ b/src/popup/components/Modals/SpendSuccess.vue
@@ -5,6 +5,7 @@
     from-bottom
     centered
     :body-without-padding-bottom="hasPayload"
+    data-cy="spend-success"
     @close="resolve"
   >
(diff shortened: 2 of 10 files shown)


src/popup/utils/constants.ts
```diff
@@ -48,6 +48,9 @@ export const STUB_CALLDATA = 'cb_AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACDJfUr
 export const STUB_NONCE = 10000;
 export const MAX_UINT256 = new BigNumber(2).exponentiatedBy(256).minus(1);
 export const STUB_TOKEN_CONTRACT_ADDRESS = 'ct_T6MWNrowGVC9dyTDksCBrCCSaeK3hzBMMY5hhMKwvwr8wJvM8';
+export const STUB_CURRENCY = [{
```
peronczyk, line 51: In my opinion we should keep the mocked data required for the tests in separate constants file located somewhere inside of the `tests/` directory.

Or maybe we can move this into the `src/popup/utils/config.js` file which is mostly used by the test files.

peronczyk (changes requested):

peronczyk (approved):

pstachel (approved):

PR #1788, opened 2023-01-25
SW-600 fix error handler

ifaouibadi:


src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -4,7 +4,7 @@
     from-bottom
     has-close-button
     class="multisig-vault-create"
-    @close="reject()"
+    @close="reject"
```
peronczyk, line 7: I know that it's possible to not use the parentheses but actually it's a bad pattern, as this kind of changes ends up in passing events into functions as arguments.
So for example when you do this:
```
@click="someAction"
```
what actually happens is
```
@click="(event) => someAction(event)"
```
and for the `@click` the event is `MouseEvent`. This is why I always suggest to leave empty paren to block the possibility of passing unintended arguments.

peronczyk (changes requested):

peronczyk (approved):

konradodo (approved):

PR #1790, opened 2023-01-25
[SW-601] Custom form select input

peronczyk:


src/popup/components/Modals/FormSelectOptions.vue
```diff
@@ -0,0 +1,102 @@
...
+    options: { type: Array as PropType<FormSelectOption[]>, default: () => [] },
+  },
+  setup(props) {
+    const searchTerm = ref('');
+    const optionsFiltered = computed((): FormSelectOption[] => {
+      if (searchTerm.value) {
+        const searchTermLower = searchTerm.value.toLowerCase();
+        return props.options.filter(({ text, value }) => (
+          text.toLowerCase().includes(searchTermLower)
+          || value.toString().includes(searchTermLower)
```
pstachel, line 68: We have a function called `compareCaseInsensitive` - maybe it could be extended for not exact comparison?

pstachel (approved):

ifaouibadi (approved):

src/popup/components/Modals/FormSelectOptions.vue
peronczyk: Mentioned function is lowering both strings each time its fired. Here I'm storing the parsed search term for the performance reasons.

peronczyk: So actually I see that Konrad's code needs to be refactored :)

PR #1792, opened 2023-01-27
feat(multisig): add additional functionality to composable

CedrikNikita:


src/composables/multisigAccounts.ts
```diff
@@ -0,0 +1,294 @@
...
+import type {
+  IAccount,
+  INetwork,
+  IMultisigCreationStep,
+  IMultisigAccountBase,
+  IMultisigAccount,
+} from '../types';
+import { useSdk } from './sdk';
+
+interface UseMultisigOptions {
```
peronczyk, line 27: `UseMultisigAccountsOptions`

src/composables/multisigAccounts.ts
```diff
@@ -0,0 +1,294 @@
...
+
+const initPollingWatcher = createPollingBasedOnMountedComponents();
+
+export function useMultisigAccounts({ store }: UseMultisigOptions) {
+  const { getDrySdk, getSdk } = useSdk({ store });
+
+  const accounts = computed<IAccount[]>(() => store.getters.accounts);
+  const activeNetwork = computed<INetwork>(() => store.getters.activeNetwork);
+
+  multisigAccounts.value = getStoredMultisigAccounts(activeNetwork.value.networkId);
```
peronczyk, line 61: I'm afraid of performance impact of this action. The `multisigAccounts` composable could be used within many components at the same time. So every time the component fires `useMultisigAccounts` it will try to get the data from local storage and parse it.

I suggest to replace this with logic that would check if the data for the currently selected network is empty and only then retrieve it from the local storage.

Or modify the `getStoredMultisigAccounts` to cache the data already retrieved for the selected network.

What do you think?

src/composables/multisigAccounts.ts
```diff
@@ -0,0 +1,294 @@
...
+      gaAccountId: gaAccount.publicKey,
+    };
+  }
+
+  async function getTransactionByHash(txHash: string) {
+    return fetchJson(`${activeNetwork.value.multisigBackendUrl}/tx/${txHash}`);
+  }
+
+  function setActiveGaAccountId(gaAccountId: string) {
+    activeMultisigAccountId.value = gaAccountId;
```
peronczyk, line 125: Maybe it would make sense to stick with one name: `...gaAccount` or `...multisigAccount`.

src/composables/multisigAccounts.ts
```diff
@@ -0,0 +1,294 @@
...
+      recipientId,
+      amount,
+    });
+  }
+
+  async function postSpendTx(tx: string, txHash: string) {
+    return postJson(`${activeNetwork.value.multisigBackendUrl}/tx`, { body: { hash: txHash, tx } });
+  }
+
+  async function proposeTx(spendTx: string, contractId: string) {
```
peronczyk, line 213: This and few other methods are strictly related to the multisig transactions - nothing to do with multisig accounts. So maybe creating separate composable for the `multisigTransactions` would be a good idea.

peronczyk (changes requested):

pstachel (approved):

peronczyk (approved):

PR #1795, opened 2023-01-31
[SW-554, SW-590] Multisig tx detail Screen

ifaouibadi:

Issue #939, opened 2021-03-29
Avoid "Tx" term in UI


davidyuk:
Is it clear for everybody that "Tx" stands for "transaction"? How to translate it to other languages?

Currently, in the sidebar menu we have "Tx history", maybe we can rename it to "Transactions"?

cc: @onvisions

subhod-i:
Closing this for now. Please reopen if the problem still persists.

PR #1797, opened 2023-01-31
[SW-555, SW-572, SW-610] UI for TX list on Multisig Account page & pending multisig dashboard widget

pstachel:


src/composables/multisigTransactions.ts
```diff
@@ -46,7 +47,7 @@ export function useMultisigTransactions({ store }: UseMultisigTransactionsOption
   async function fetchTransactionByHash(txHash: string): Promise<IRawMultisigTx | null> {
     return fetchJson(`${activeNetwork.value.multisigBackendUrl}/tx/${txHash}`)
       .then((res) => res)
-      .catch(null);
+      .catch(() => {});
```
konradodo, line 50: handleUnknownError
```suggestion
      .catch(handleUnknownError);
```

src/popup/components/TransactionList.vue
```diff
@@ -10,6 +10,10 @@
         v-for="transaction in filteredTransactions"
         :key="transaction.hash"
         :transaction="transaction"
+        :confirmed-by="getConsensusInfo(transaction).confirmedBy"
+        :confirmations-required="getConsensusInfo(transaction).confirmationsRequired"
+        :total-signers="getConsensusInfo(transaction).totalSigners"
+        :is-multisig-transaction="getConsensusInfo(transaction).isMultisigTransaction"
         :data-cy="transaction.pending && 'pending-txs'"
```
konradodo, line 17: getConsensusInfo(transaction) -> to computed

pstachel: how?

konradodo (approved):

src/popup/components/TransactionList.vue
konradodo: into TransactionItem

src/popup/components/TransactionList.vue
```diff
@@ -49,7 +53,7 @@ import {
   watch,
 } from '@vue/composition-api';
 import type {
-  IAccount,
+  IAccount, IActiveMultisigTx, IMultisigAccount,
   INetwork,
   ITokenList,
```
konradodo, line 58:
```suggestion
  IAccount,
   IActiveMultisigTx,
    IMultisigAccount,
  INetwork,
  ITokenList,
```

src/composables/transaction.ts
```diff
@@ -13,7 +13,7 @@ import type {
   TransactionType,
   INetwork,
   TxFunctionRaw,
-  TxFunctionParsed,
+  TxFunctionParsed, IActiveMultisigTx,
```
konradodo, line 16:
```suggestion
  TxFunctionParsed,
   IActiveMultisigTx,
```

src/composables/transaction.ts
```diff
@@ -142,19 +145,23 @@ export function useTransaction({
...
-  )));
+  const isDex = computed((): boolean => !!(
+    transaction.value?.tx
+    && getDexContracts.value
+    && transaction.value.tx.contractId
+    && (
+      getDexContracts.value.router.includes(transaction.value.tx.contractId)
+      || getDexContracts.value.wae.includes(transaction.value.tx.contractId)
+    )
+  ));
```
konradodo, line 164:
```suggestion

 const { wae, router } = getDexContracts.value
  const isDex = computed((): boolean => !!(
    transaction.value?.tx
    && getDexContracts.value
    && transaction.value.tx.contractId
    && (
      [...wae, ...router ].includes(transaction.value.tx.contractId)
    )
  ));
```

src/popup/locales/en.json
```diff
@@ -1351,5 +1355,6 @@
   "total": "Total",
   "totalMultisig": "Total in multisig vaults",
   "priceNotAvailable": "price: n/a",
-  "search": "Search"
+  "search": "Search",
+  "tx": "Tx"
```
CedrikNikita, line 1359: https://github.com/aeternity/superhero-wallet/issues/939

src/composables/transaction.ts
```diff
@@ -142,19 +146,20 @@ export function useTransaction({
...
 
-  const isDex = computed((): boolean => !!(transaction.value
-    && getDexContracts.value && transaction.value.tx.contractId && (
-    getDexContracts.value.router.includes(transaction.value.tx.contractId)
-    || getDexContracts.value.wae.includes(transaction.value.tx.contractId)
-  )));
+  const { wae, router } = getDexContracts.value;
+  const isDex = computed((): boolean => !!(
+    transaction.value?.tx?.contractId
+    && getDexContracts.value
```
peronczyk, line 160: This test makes no sense as in line 157 the logic assumes that the value of `getDexContracts.value` is an object.

src/composables/transaction.ts
```diff
@@ -35,7 +36,7 @@ interface UseTransactionOptions {
    * TODO: Temporary solution to avoid dependency circle
    */
   store: Store<any>
-  initTransaction?: ITransaction
+  initTransaction?: ITransaction | IActiveMultisigTx
```
peronczyk, line 39: I strongly recommend to not mix two totally different types in one composable. I understand the common part is the `tx`. So maybe we should split the logic that parses the `ITx` into new composable?

peronczyk (changes requested):

src/popup/components/TransactionList.vue
```diff
@@ -108,59 +119,83 @@ export default defineComponent({
...
+                throw new Error(`${root.$t('pages.recentTransactions.unknownMode')} ${displayMode.value.key}`);
+            }
+          })
+          .filter(
+            (tr) => !searchPhrase.value
+                    || getTxSymbol.value(tr)
+                      .toLocaleLowerCase()
+                      .includes(
+                        searchPhrase.value.toLocaleLowerCase(),
+                      ),
```
peronczyk, line 175: IDEM

src/popup/components/TransactionList.vue
```diff
@@ -197,17 +232,30 @@ export default defineComponent({
...
+
     watch(displayMode, () => {
       checkLoadMore();
     });
 
-    onMounted(() => {
+    onMounted(async () => {
+      if (isMultisigDashboard.value) {
+        loading.value = true;
+        await watchUntilTruthy(() => activeMultisigAccount.value);
```
peronczyk, line 256: I suggest to avoid using the `watchUntilTruthy` function as I think we need to get rid of it. Also this line blocks running the code that is placed after the `if`.

Maybe using Vue watcher before the onMounted would be better?

```
watch(activeMultisigAccount, (acc) => {
  ...
});
```

src/popup/components/TransactionList.vue
```diff
@@ -197,17 +232,30 @@ export default defineComponent({
...
+      }
+    }
+
     watch(displayMode, () => {
       checkLoadMore();
     });
 
-    onMounted(() => {
+    onMounted(async () => {
+      if (isMultisigDashboard.value) {
```
peronczyk, line 254: The `loadMore` and setInterval logic should be wrapped in `else` block` as fetching the regular transactions is not needed when browsing the multisig ones.

peronczyk (changes requested):

src/popup/components/TransactionList.vue
pstachel: It should wait for the account as it needs it's address in loadMore function

src/popup/components/PendingMultisigTransactionCard.vue
```diff
@@ -0,0 +1,67 @@
...
+export default defineComponent({
+  components: { TransactionItem },
+  setup(props, { root }) {
+    const { activeMultisigAccount } = useMultisigAccounts({ store: root.$store });
+    const { fetchActiveMultisigTx } = useMultisigTransactions({ store: root.$store });
+
+    const pendingMultisigTransaction = ref();
+
+    watch(
+      () => activeMultisigAccount.value,
```
peronczyk, line 31:
```
watch(activeMultisigAccount, async () => ...
```

src/popup/components/TransactionList.vue
pstachel: It shouldn't, we should fetch all of the transactions

src/popup/components/PendingMultisigTransactionCard.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <div
+    v-if="pendingMultisigTransaction && pendingMultisigTransaction.tx"
+    class="pending-multisig-transaction-card"
+  >
+    <div class="title">
+      {{ $t('dashboard.pendingMultisigCard.title' ) }}
```
peronczyk, line 7: Unnecessary space before ending bracket.

peronczyk (changes requested):

src/popup/components/PendingMultisigTransactionCard.vue
pstachel: Impossible to fit it in one line

src/popup/components/TransactionItem.vue
```diff
@@ -68,45 +78,77 @@ import {
...
+      isAllowance: isAllowance.value,
+      // TODO - refactor useTransactionTokens to use only tx
+      transaction: (props.multisigTransaction || props.transaction) as unknown as ITransaction,
     });
 
+    const redirectRoute = computed(() => {
+      if (props.multisigTransaction) {
+        return { name: 'multisig-proposal-details' };
+      }
+      const { hash } = props.multisigTransaction || props.transaction;
```
peronczyk, line 134: There is already if that checks if this is a multisig transaction, so passing the `props.multisigTransaction` here makes no sense.

src/popup/components/TransactionItem.vue
```diff
@@ -68,45 +78,77 @@ import {
...
+    const getConsensusInfo = computed(() => ({
+      confirmedBy: props.multisigTransaction.confirmedBy?.length || 0,
+      totalSigners: props.multisigTransaction.signers?.length || 0,
+      confirmationsRequired: props.multisigTransaction.confirmationsRequired?.toString(),
+    }));
+
     const fiatAmount = computed(() => {
-      // TODO add type to tokens
-      const aeToken = tokens.value?.find((t: any) => t?.isAe);
+      const aeToken = tokens.value?.find((t: ITokenResolved) => t?.isAe);
```
peronczyk, line 145: `tokens` should have the types assigned, so there is no necessity to set type `ITokenResolved` to `t`.

src/popup/components/TransactionOverview.vue
```diff
@@ -59,26 +61,28 @@ export default defineComponent({
 
     const { getSdk } = useSdk({ store: root.$store });
 
+    const { tx } = props;
```
peronczyk, line 64: Destructuring props is a bad thing because the reactivity is lost. I see that in this case this is not a problem, but still... I think we should not do it for the sake of having some common rules.

https://stackoverflow.com/questions/64926354/vue-3-watch-doesn-t-work-if-i-watch-a-destructured-prop

In the future we need to add ESLint rule for this.

src/popup/components/TransactionTagList.vue
```diff
@@ -0,0 +1,157 @@
...
+
+    const account = useGetter<IAccount>('account');
+    const activeNetwork = useGetter<INetwork>('activeNetwork');
+    const availableTokens = useState<ITokenList>('fungibleTokens', 'availableTokens');
+    const getTxDirection = useGetter('getTxDirection');
+
+    const labels = computed<(string | TranslateResult)[]>(() => {
+      if (!tx) return [];
+
+      const transactionTypes = i18n.t('transaction.type') as Record<TransactionType, TranslateResult>;
```
peronczyk, line 67: The types here are wrong. The `transaction.type` translation keys are actually camelCase: "spendTx", "contractCreateTx", "contractCallTx"... and the `TransactionType` are PascalCase: "SpendTx", "ContractCreateTx", "ContractCallTx".

src/popup/components/TransactionTagList.vue
```diff
@@ -0,0 +1,157 @@
...
+          )
+      ) {
+        return [
+          i18n.t('transaction.type.spendTx'),
+          tx.callerId === account.value.address
+            ? i18n.t('transaction.spendType.out')
+            : i18n.t('transaction.spendType.in'),
+        ];
+      }
+
```
peronczyk, line 127: Unnecessary empty line.

src/popup/components/TransactionTagList.vue
```diff
@@ -0,0 +1,157 @@
...
+      if (txType.value === TX_TYPE_MDW.PayingForTx) {
+        return [
+          i18n.t('transaction.type.payingForTx'),
+        ];
+      }
+      return props.isPending ? [] : [transactionTypes[txType.value]];
```
peronczyk, line 133: Before this return please add this additional if: https://github.com/aeternity/superhero-wallet/pull/1795/files#diff-6f2892100eb0e98f7e0a50d24ad5e09400985127894b862678518d394878e403

src/popup/components/TransactionTagList.vue
```diff
@@ -0,0 +1,157 @@
...
+  components: { TransactionTag },
+  props: {
+    tx: { type: Object as PropType<ITx>, default: null },
+    isIncomplete: Boolean,
+    isPending: Boolean,
+    isClaim: Boolean,
+    dense: Boolean,
+  },
+  setup(props, { root }) {
+    const { tx } = props;
```
peronczyk, line 51: Please avoid destructuring the props as this makes the `tx` to loose the reactivity.
https://stackoverflow.com/questions/64926354/vue-3-watch-doesn-t-work-if-i-watch-a-destructured-prop

src/popup/components/TransactionTagList.vue
```diff
@@ -0,0 +1,157 @@
...
+            ? i18n.t('transaction.spendType.in')
+            : i18n.t('transaction.spendType.out'),
+        ];
+      }
+      if (
+        txType.value === SCHEMA.TX_TYPE.contractCall
+          && availableTokens.value[tx.contractId]
+          && (
+            tx.function === TX_FUNCTIONS.transfer
+              || props.isIncomplete
```
peronczyk, line 117: Wrong indentation.

src/popup/components/TransactionTagList.vue
```diff
@@ -0,0 +1,157 @@
...
+      if (
+        (
+          tx.contractId
+              && (
+                activeNetwork.value.tipContractV1 === tx.contractId
+                  || activeNetwork.value.tipContractV2 === tx.contractId
+              )
+              && (
+                tx.function === TX_FUNCTIONS.tip
+                  || tx.function === TX_FUNCTIONS.retip
```
peronczyk, line 100: Wrong indentation.

src/popup/components/TransactionTagList.vue
```diff
@@ -0,0 +1,157 @@
...
+            ? i18n.t('transaction.dexType.pool')
+            : i18n.t('transaction.dexType.swap'),
+        ];
+      }
+      if (
+        (
+          tx.contractId
+              && (
+                activeNetwork.value.tipContractV1 === tx.contractId
+                  || activeNetwork.value.tipContractV2 === tx.contractId
```
peronczyk, line 96: Wrong indentation.

peronczyk (changes requested):

peronczyk (changes requested):

src/popup/components/TransactionItem.vue
```diff
@@ -68,45 +78,75 @@ import {
...
+    const { tokens } = useTransactionTokens({
       store: root.$store,
-      initTransaction: props.transaction,
+      direction: direction.value,
+      isAllowance: isAllowance.value,
+      // TODO - refactor useTransactionTokens to use only tx
+      transaction: (props.multisigTransaction || props.transaction) as unknown as ITransaction,
     });
 
+    const redirectRoute = computed(() => {
```
peronczyk, line 129: `(): Location =>`

Location is imported from the Vuex.

peronczyk (approved):

CedrikNikita (approved):

PR #1801, opened 2023-02-01
Refactor: Add types to categorizeContractCallTxObject util function

peronczyk:


src/popup/utils/utils.ts
```diff
@@ -190,3 +192,48 @@ export function compareCaseInsensitive(
...
+export function categorizeContractCallTxObject(transaction: ITransaction): {
+  amount: number
+  to?: string
+  token: string
+  url?: string
+  note?: string
+} | null {
+  if (transaction.tx.type?.toLowerCase() !== SCHEMA.TX_TYPE.contractCall.toLowerCase()) {
+    return null;
+  }
```
pstachel, line 205: check out `compareCaseInsensitive` function

pstachel (approved):

CedrikNikita (changes requested):

konradodo (approved):

src/popup/utils/utils.ts
```diff
@@ -188,5 +190,50 @@ export function compareCaseInsensitive(
...
+  token: string
+  url?: string
+  note?: string
+} | null {
+  if (!compareCaseInsensitive(transaction.tx.type, SCHEMA.TX_TYPE.contractCall)) {
+    return null;
+  }
+  if (transaction.incomplete || transaction.pending) {
+    return {
+      amount: transaction.tx.amount,
```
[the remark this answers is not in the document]
peronczyk: OK, fixed. It turns out that this function can receive both `ITransaction` and `IPendingTransaction`. I needed to align some types to handle this situation. Please see now.

src/popup/components/TransferReview.vue
```diff
@@ -224,8 +224,9 @@ export default defineComponent({
             amount,
             recipient,
             hash: actionResult.hash,
-            type: 'spendToken',
             pendingTokenTx: true,
+            pending: true,
```
CedrikNikita, line 228: `pending` field is being added by the `addPendingTransaction` action itselft

peronczyk: `addPendingTransaction` is a store function, so there's no types there. I added it here as the `pending` property is required by the `IPendingTransaction`. In the future, when we will start to remove the Vuex we can try to do it better.

CedrikNikita (approved):
Code changes look reasonable. Did not test it.

PR #1804, opened 2023-02-02
Use cached multisig account data incase of error while polling new data

subhod-i:


src/composables/multisigAccounts.ts
```diff
@@ -44,6 +45,20 @@ function getStoredMultisigAccounts(networkId: string): IMultisigAccount[] {
...
+
+const getVersion = async (instance: any) => (await instance.methods.get_version()).decodedResult;
+
+const getConsensus = async (instance: any) => (await instance.methods.get_consensus_info())
+  .decodedResult;
+
+const getContractInstance = async (contractId: string, sdk: ISdk) => sdk.getContractInstance({
+  aci: SimpleGAMultiSigAci,
+  contractAddress: contractId,
+});
```
peronczyk, line 60: In my opinion this method gives no value - it's just a shortcut used once. Using `sdk.getContractInstance` directly is a better option I think.

src/composables/multisigAccounts.ts
```diff
@@ -149,36 +164,36 @@ export function useMultisigAccounts({ store }: UseMultisigAccountsOptions) {
...
+            version,
+            contractId,
+            balance: (gaAccountId && (await sdk.balance(gaAccountId))) || 0,
+            address: gaAccountId,
+            multisigAccountId: gaAccountId,
+          };
+        } catch (e) {
+          /* Node might throw nonce mismatch error, skip the current account update
+          * return the existing data and account details will be updated in the next poll.
+          */
```
peronczyk, line 194: Please use correct DocBlock.
```
/**
 *
 */
```
This VS Code plugin is really helpful in this field: https://marketplace.visualstudio.com/items?itemName=aaron-bond.better-comments

peronczyk (changes requested):

konradodo (approved):

pstachel (approved):

Liubov-crypto:
Checked, didn't find any nonce errors in console. LGTM.

src/composables/multisigAccounts.ts
```diff
@@ -149,36 +158,39 @@ export function useMultisigAccounts({ store }: UseMultisigAccountsOptions) {
...
+            version,
+            contractId,
+            balance: (gaAccountId && (await sdk.balance(gaAccountId))) || 0,
+            address: gaAccountId,
+            multisigAccountId: gaAccountId,
+          };
+        } catch (e) {
+          /** Node might throw nonce mismatch error, skip the current account update
+           * return the existing data and account details will be updated in the next poll.
+           */
```
peronczyk, line 191: https://en.wikipedia.org/wiki/JSDoc

In case you want to make a comment don't add the text in first line.
This is bad:
```
/** Lorem
 * Ipsum
 */
```

This is OK:
```
/**
 * Lorem
 * Ipsum
 */
```

🙂

subhod-i:
This needs @CedrikNikita 's approval to merge to develop

src/composables/multisigAccounts.ts
```diff
@@ -149,36 +158,40 @@ export function useMultisigAccounts({ store }: UseMultisigAccountsOptions) {
...
+            ...otherMultisig,
+            nonce: Number(nonce),
+            signers,
+            version,
+            contractId,
+            balance: (gaAccountId && (await sdk.balance(gaAccountId))) || 0,
+            address: gaAccountId,
+            multisigAccountId: gaAccountId,
+          };
+        } catch (e) {
```
CedrikNikita, line 188: I propose to not omit an important Errors that differs from the `nonce` ones.

subhod-i: Hmm. this is a great catch @CedrikNikita

src/composables/multisigAccounts.ts
```diff
@@ -149,36 +159,43 @@ export function useMultisigAccounts({ store }: UseMultisigAccountsOptions) {
...
+          };
+        } catch (error) {
+          /**
+           * Node might throw nonce mismatch error, skip the current account update
+           * return the existing data and account details will be updated in the next poll.
+           */
+          if (error instanceof DryRunError) {
+            return multisigAccounts.value.find((account) => account.contractId === contractId);
+          }
+          return handleUnknownError(error);
```
CedrikNikita, line 197:
```suggestion
          if (!(error instanceof DryRunError)) {
            handleUnknownError(error);
          }
          return multisigAccounts.value.find((account) => account.contractId === contractId);
```

CedrikNikita (changes requested):
In case the `localStorage` is empty and an error appears already, we would get a nullish element of array, and then when we would try to render such accounts we would get issues like.

```
TypeError: Cannot read properties of undefined (reading 'contractId')
```

To reproduce such behaviour, you can create a vault of the version of contract we are currently not supporting by precompiled aci and bytecode.

src/composables/multisigAccounts.ts
```diff
@@ -149,38 +159,48 @@ export function useMultisigAccounts({ store }: UseMultisigAccountsOptions) {
...
+        } catch (error) {
+          /**
+           * Node might throw nonce mismatch error, skip the current account update
+           * return the existing data and account details will be updated in the next poll.
+           */
+          if (!(error instanceof DryRunError)) {
+            handleUnknownError(error);
+          }
+          return multisigAccounts.value
+            .filter(Boolean)
```
CedrikNikita, line 198:
```suggestion
```
There is no need to filter it.

subhod-i: I had some errors while testing this but once cleaning the localstorage it works without this filter.

CedrikNikita (approved):
Code changes look reasonable, i didn't test it well.

subhod-i:
Merging this. Entire feature wise need to be tested for better resilience.

PR #1805, opened 2023-02-02
[SW-607] Multisig create vault confirmation screen

peronczyk:


src/popup/components/Modals/ConfirmRawSign.vue
```diff
@@ -28,7 +28,7 @@
     </div>
 
     <DetailsItem
-      :label="$t('modals.confirm-transaction-sign.data-sign')"
+      :label="$t('modals.confirmTransactionSign.data-sign')"
```
pstachel, line 31: why part of the translation key is in camelCase and part in kebab-case?

src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -6,126 +6,153 @@
...
+        <InputField
+          v-model.trim="signer.address"
+          v-validate="{
+            required: true,
+            name_registered_address_or_url: true,
+          }"
+          :name="`signer-address-${index}`"
+          :label="getSignerLabel(index)"
+          :placeholder="$t('modals.createMultisigAccount.signerInputPlaceholder')"
+          :message="errors.first(`signer-address-${index}`)"
```
pstachel, line 33: I don't like that you are using almost the same template string three times here - maybe it should be wrapped in a function to return `signer-address-${index}`?

src/popup/components/MultisigVaultCreateReview.vue
```diff
@@ -0,0 +1,232 @@
...
+                v-text="$t('common.you')"
+              />
+            </div>
+          </div>
+        </template>
+      </DetailsItem>
+
+      <DetailsItem
+        label="Consensus"
+        value="3/3 out of 3"
```
pstachel, line 54: I think the value shouldn't be static here

src/popup/components/MultisigVaultCreateReview.vue
```diff
@@ -0,0 +1,232 @@
...
+          <Avatar :address="creatorAccount.address" />
+          <div class="creator-data">
+            <BtnPill
+              class="account-select"
+              dense
+            >
+              <FormSelect
+                v-model="creatorAddress"
+                unstyled
+                :default-text="'Choose account'"
```
pstachel, line 19: please put it in translation file

src/popup/components/MultisigVaultCreateReview.vue
```diff
@@ -0,0 +1,232 @@
...
+    const creatorAccount = computed(
+      () => accounts.value.find(({ address }) => address === creatorAddress.value),
+    );
+
+    (async () => {
+      loading.value = true;
+      const rawGasFee = await estimateMultisigAccountDeployGasFee(
+        null,
+        props.confirmationsRequired,
+        [...props.signers.map(({ address }) => address)],
```
pstachel, line 170:
```suggestion
        props.signers.map(({ address }) => address),
```

pstachel (changes requested):

src/popup/components/Modals/ConfirmRawSign.vue
peronczyk: Guess what, someone did interpolation:
```
:label="$t('modals.confirm-transaction-sign')[key]"
```
For now I don't want to mess this up. But in the future we definitely need to refactor this.

src/popup/components/Modals/MultisigVaultCreate.vue
peronczyk: This is used twice. And it's normal to do it this way when binding the v-validate rules with collecting the errors for dynamic fields.

src/popup/components/MultisigVaultCreateReview.vue
peronczyk: This is an leftover which is replaced with the slot content. I removed this. Thanks for noticing this.

src/popup/components/MultisigVaultCreateReview.vue
peronczyk: Hahaha, that's a good one :) Totally right... fixing this.

src/popup/components/MultisigVaultCreateReview.vue
```diff
@@ -0,0 +1,229 @@
...
+                v-if="isLocalAccountAddress(signer.address)"
+                dense
+                v-text="$t('common.you')"
+              />
+            </div>
+          </div>
+        </template>
+      </DetailsItem>
+
+      <DetailsItem label="Consensus">
```
pstachel, line 52: `Consensus` should be in the translation file

pstachel (approved):

src/popup/components/MultisigVaultCreateReview.vue
peronczyk: Fixed

src/popup/components/MultisigVaultCreateReview.vue
```diff
@@ -0,0 +1,229 @@
...
+                dense
+                v-text="$t('common.you')"
+              />
+            </div>
+          </div>
+        </template>
+      </DetailsItem>
+
+      <DetailsItem :label="$t('multisig.consensus')">
+        {{ confirmationsRequired }}/{{ confirmationsRequired }} {{ $t('common.of') }} {{ signers.length }}
```
konradodo, line 53: that string you can get from composables. Props name is "consensulLabel" You can check multisigDetails.vue

peronczyk: Unfortunately not in this case because this is a new account, not a one that was retrieved. And here you can see that both numbers before and after the slash are the same.

src/composables/multisigAccountCreate.ts
```diff
@@ -0,0 +1,146 @@
...
+  IMultisigCreationPhase,
+} from '../types';
+import { useSdk } from './sdk';
+import { MULTISIG_CREATION_PHASES, MULTISIG_SIMPLE_GA_BYTECODE } from '../popup/utils';
+import SimpleGAMultiSigAci from '../lib/contracts/SimpleGAMultiSigACI.json';
+
+export function useMultisigAccountCreate({ store }: IDefaultComposableOptions) {
+  const { getDrySdk, getSdk } = useSdk({ store });
+
+  let rawTx: string;
```
konradodo, line 17: why you declare this variables as "let" not as "ref"

src/composables/multisigAccountCreate.ts
```diff
@@ -0,0 +1,146 @@
...
+  let accountId: string;
```
konradodo, line 18: why you declare this variables as "let" not as "ref"

peronczyk: Because the reactivity is not needed here. This is a state shared between two function and it's not passed to any template/watcher.

konradodo (approved):

src/composables/multisigAccountCreate.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  const multisigAccount = ref<IMultisigAccountBase | null>(null);
+  const multisigAccountCreationPhase = ref<IMultisigCreationPhase>(null);
+  const multisigAccountCreationEncodedCallData = ref<string>();
+  const multisigAccountCreationFee = ref<BigNumber>(new BigNumber(0));
```
subhod-i, line 23: We could maintain just one `multisigAccountCreationFeeAe` state variable.

subhod-i (approved):

src/composables/multisigAccounts.ts
```diff
@@ -62,45 +62,68 @@ export function useMultisigAccounts({ store }: UseMultisigAccountsOptions) {
...
+  }
+
   /**
    * @param {number} noOfConfirmations no of minimum confirmation needed to process a transaction
    * @param {string[]} signers
    */
   async function deployMultisigAccount(
     noOfConfirmations: number,
-    signersAddresses: [string, string, ...string[]],
+    signersAddresses: string[],
```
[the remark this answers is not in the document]
peronczyk: Yes, but this requirement we need to test in the logic, not on the code validation level. So we should be able to add only one signer, but when creating the vault we need to validate this. This is why we have the `MULTISIG_VAULT_MIN_NUM_OF_SIGNERS` constant.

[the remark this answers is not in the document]
subhod-i: I agree with that but since this code in the composables and it mostly acts like a library. In the future this function may be reused by other developers hence I guess this type definition is also important along with the logic condition too.

PR #1763, opened 2022-12-29
[SW-493] Transaction details for all subaccounts

pstachel:

PR #1808, opened 2023-02-03
Fix transaction displaying issues

CedrikNikita:
missed in #1763,
missed in #1751

### DIFF
--- a/src/composables/transaction.ts
+++ b/src/composables/transaction.ts
@@ -167,7 +167,7 @@ export function useTransaction({
     if (txType.value === SCHEMA.TX_TYPE.spend) {
       return [
         i18n.t('transaction.type.spendTx'),
-        getTxDirection.value(transaction) === TX_FUNCTIONS.received
+        getTxDirection.value(transaction.value) === TX_FUNCTIONS.received
           ? i18n.t('transaction.spendType.in')
           : i18n.t('transaction.spendType.out'),
       ];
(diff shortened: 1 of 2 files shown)


src/store/getters.js
```diff
@@ -122,9 +122,9 @@ export default {
   ),
   getTxDirection: (_, { account: { address }, getTxType }) => ({ tx }, externalAddress) => {
     if (getTxType({ tx }) === SCHEMA.TX_TYPE.spend) {
-      return tx.recipientId === address
-        ? TX_FUNCTIONS.received
-        : TX_FUNCTIONS.sent;
+      return tx.senderId === address
+        ? TX_FUNCTIONS.sent
+        : TX_FUNCTIONS.received;
```
pstachel, line 127: With that change you'll see that transaction as received
https://feature-fix-transaction-displaying.wallet.z52da5wt.xyz/account-details/transactions/th_2JwMpEBBQP5DgE75MhRaTAtXMmJwriRZDpsZUV1qyoGiy2myvg

pstachel (changes requested):

src/store/getters.js
CedrikNikita: I will omit this change from the PR in preference to more general issue #1809

pstachel (approved):

Liubov-crypto:
There is one issue with FT labelling. I added it to our Jira backlog. The rest LGTM.

peronczyk (approved):

PR #1811, opened 2023-02-03
[SW-616] Propose transaction wording fixes

peronczyk:


src/popup/components/Modals/TransferSend.vue
```diff
@@ -34,8 +34,12 @@
...
-        :text="showSendButton ? $t('pages.send.send') : $t('modals.send.next')"
+        :icon="(showSendButton && !isMultisig) ? ArrowSendIcon : null"
+        :text="showSendButton
+          ? isMultisig
+            ? $t('modals.multisigTxProposal.proposeAndApprove')
+            : $t('pages.send.send')
+          : $t('modals.send.next')"
         data-cy="next-step-button"
         @click="proceedToNextStep"
       />
```
pstachel, line 45: If I'm not mistaken we do have two "Send Buttons" - I'd suggest wrapping the whole logic in a component and use it in both places (Account Details & Dashboard)

src/popup/components/Modals/TransferSend.vue
```diff
@@ -34,8 +34,12 @@
...
         class="button-action-primary"
         :disabled="error || !isConnected || !transferData.address || !transferData.amount"
-        :icon="showSendButton ? ArrowSendIcon : null"
-        :text="showSendButton ? $t('pages.send.send') : $t('modals.send.next')"
+        :icon="(showSendButton && !isMultisig) ? ArrowSendIcon : null"
+        :text="showSendButton
+          ? isMultisig
+            ? $t('modals.multisigTxProposal.proposeAndApprove')
+            : $t('pages.send.send')
+          : $t('modals.send.next')"
```
pstachel, line 42: Double ternary operator should be avoided, but if you are in need you should wrap it in a computed property

pstachel (changes requested):

src/popup/components/Modals/TransferSend.vue
pstachel: Or maybe you could use two separate components for it

peronczyk: I created computed property for this. Please check if it's OK.

pstachel (approved):

ifaouibadi (approved):

PR #1812, opened 2023-02-06
[SW-607] [Tech enabler] Expandable details item

peronczyk:


src/popup/components/DetailsItem.vue
```diff
@@ -1,47 +1,89 @@
...
+        <slot>
+          {{ value }}
+          <span
+            v-if="secondary"
+            class="secondary"
+          >
+            {{ secondary }}
+          </span>
+          <slot name="value" />
+        </slot>
```
pstachel, line 43: Should there be a slot inside a slot?

pstachel (approved):

src/popup/components/DetailsItem.vue
peronczyk: In general: yes, it is normal to have slot inside of an slot. Maybe not the best solution, but commonly used in some frameworks.
In our case my goal is to migrate the way how we pass the content to `DetailsItem`. Currently we need to use named slot, but in my opinion this should be the default one. So instead having to:
```
<DetailsItem label="Test">
  <template #value>
    Some content
  </template>
</DetailsItem>
```
we should be able to:
```
<DetailsItem label="Test">
    Some content
</DetailsItem>
```

I didn't want to refactor whole app, so for now I'm leaving the named slot, but we will gradually migrate the existing code.

ifaouibadi (approved):

PR #1819, opened 2023-02-08
[SW-620] Separate multisig views from standard accounts on routes

pstachel:


src/composables/accountSwitcher.ts
```diff
@@ -0,0 +1,87 @@
...
+    activeIdx: multisigAccountIdx,
+    selectAccount,
+    getKeyValue,
+  };
+}
+function useStandardAccountSwitcher(
+  { store }: UseAccountSwitcherOptions,
+): IStandardAccountSwitcher {
+  const activeIdx = useState<number>('accounts', 'activeIdx');
+  const accounts = useGetter<IAccount[]>('accounts');
```
konradodo, line 65: we cannot use useState and useGetter, but wen do refactor in vuex.ts, we can resolve that problem. Let's sync up with me

peronczyk (changes requested):

src/composables/homeRoute.ts
```diff
@@ -0,0 +1,27 @@
+import { computed, ref, watch } from '@vue/composition-api';
```
peronczyk, line 1: I really don't like the design of this composable. The `homeRouteName` should not be an reactive thing. This should be a computed prop based on the environment (isLoggedIn / route / store state). But to be honest I would vote for removing this composable and maybe replacing it with a utility function `getHomePageRouteName(route, store)` or something similar.

src/popup/components/AccountCardWrapper.vue
```diff
@@ -0,0 +1,79 @@
+<template>
```
peronczyk, line 1: This component actually is not a wrapper as it doesn't wrap anything. It's the card itself.

src/popup/components/AccountCardWrapper.vue
```diff
@@ -0,0 +1,79 @@
...
+        name="top"
+        :color="color"
+      />
+    </div>
+
+    <div class="middle">
+      <slot name="middle" />
+    </div>
+
+    <slot name="bottom" />
```
peronczyk, line 20: I suggest to wrap this with additional div to avoid further problems and allow to see the structure of the component in the DevTools.

src/popup/components/BalanceInfo.vue
```diff
@@ -22,21 +22,22 @@ export default defineComponent({
...
+    const isMultisigPage = computed(() => !!root.$route.meta?.isMultisigPage);
+    const currentAddress = computed(() => props.address || account.value.address);
 
     const balance = computed(
-      () => (isMultisigDashboard.value
-        ? currentAccount.value.balance
-        : balances.value[currentAccount.value.address])?.toNumber() || 0,
+      () => isMultisigPage.value
+        ? props.multisigBalance
+        : balances.value[currentAddress.value]?.toNumber() || 0,
```
peronczyk, line 40: In my opinion this component needs to be refactored to receive one prop: `balance` as a number and that's all. Because now it looks really off to have `multisigBalance` and `account` props.

peronczyk (changes requested):

konradodo (approved):

src/popup/components/DashboardHeader.vue
```diff
@@ -1,6 +1,6 @@
 <template>
```
pstachel, line 1: Split this component between multisig and standard accounts

src/popup/components/AccountCardBase.vue
```diff
@@ -0,0 +1,89 @@
...
+    </div>
+
+    <div class="bottom">
+      <slot name="bottom" />
+    </div>
+  </BtnBase>
+</template>
+
+<script lang="ts">
+import { computed, defineComponent, PropType } from '@vue/composition-api';
```
konradodo, line 27: split this line

src/popup/components/AccountCardConsensus.vue
```diff
@@ -9,9 +9,7 @@
       <PendingIcon class="icon pending" />
       {{ $t('multisig.consensus') }}
       <span class="highlighted">
-        {{ currentAccount.confirmedBy.length }}/{{ currentAccount.confirmationsRequired }}
-        {{ $t('common.of') }}
-        {{ totalSignatures }}
+        {{ multisigAccount.consensusLabel }}
```
konradodo, line 12: we cannot use multisig Account.consensus Label bart found big problem with that. (this is not reloaded after changed languages)

src/popup/components/DashboardWrapper.vue
```diff
@@ -0,0 +1,96 @@
...
+    BtnMain,
+  },
+  setup(props, { root }) {
+    const backedUpSeed = useState('backedUpSeed');
+
+    const query = computed(() => root.$route.query);
+
+    watch(query, (value) => {
+      if (!isEmpty(value)) {
+        root.$store.dispatch('modals/open', { name: MODAL_TRANSFER_SEND });
```
konradodo, line 61: useDispatch

konradodo (approved):

src/composables/ui.ts
```diff
@@ -0,0 +1,19 @@
+import { ref } from '@vue/composition-api';
+import { ROUTE_ACCOUNT } from '../popup/router/routeNames';
+import type { IDefaultComposableOptions } from '../types';
+
+const defaultHomeRouteName = ref(ROUTE_ACCOUNT);
```
peronczyk, line 5: Why "default"? It's the home route :)

src/popup/components/AccountCard.vue
```diff
@@ -1,99 +1,64 @@
...
         :color="color"
       />
-    </div>
+    </template>
 
-    <div class="middle">
-      <BalanceInfo :account="account" />
-    </div>
+    <template #middle>
+      <BalanceInfo :balance="balance.toNumber()" />
```
peronczyk, line 17: I suggest to move this into computed property to allow it to be typed.

src/popup/components/DashboardWrapper.vue
```diff
@@ -0,0 +1,96 @@
+<template>
+  <div class="dashboard">
+    <slot name="header" />
+
+    <div class="dashboard-cards">
+      <div class="buttons-row">
```
peronczyk, line 6: I suggest to add `v-if="$slots.buttons"`

src/popup/utils/utils.ts
```diff
@@ -301,3 +301,8 @@ export function isContainingNestedTx(tx: ITx): boolean {
 export function getInnerTransaction(tx: ITx): any {
   return isContainingNestedTx(tx) ? tx.tx?.tx : tx;
 }
+
+export function getFromLocalStorage(key: string, defaultValue: any = null) {
```
peronczyk, line 305: In my opinion providing additional argument that is an alternative value is rather bed pattern. Without that you always can write the code this way:
```
const a = getFromLocalStorage('foo') || defaultValue;
```

src/popup/pages/Retip.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="retip">
-    <BalanceInfo />
+    <BalanceInfo :balance="balance.toNumber()" />
```
peronczyk, line 3: I suggest to move this to computed property.

src/popup/components/AccountDetailsBase.vue
```diff
@@ -0,0 +1,228 @@
...
+        return;
+      }
+      const newClientHeight = entries[0].target.clientHeight;
+      if (
+        newClientHeight && (
+          (
+            clientHeight.value + BALANCE_AND_ACTIONS_HEIGHT + ACCOUNT_INFO_HEIGHT
+          ) < newClientHeight
+              || newClientHeight <= initialClientHeight.value
+        )
```
peronczyk, line 104: This condition looks messed up.
1. In case of more than one `and` the `&&` signs should be put in next line.
2. Long comparisions looks better when stored in constants.

peronczyk (approved):

PR #1822, opened 2023-02-09
[SW-562, SW-632] Display "PayingForTX" and "GAMetaTx" transactions correctly

ifaouibadi:


src/popup/pages/TransactionDetails.vue
```diff
@@ -236,10 +274,15 @@ export default defineComponent({
       isErrorTransaction,
       isAllowance,
       isDex,
+      isMultisig,
```
CedrikNikita, line 277: In case of `payForTx` type of transaction fee should be a summary of both fees (inner and external).

src/popup/components/TransactionOverview.vue
```diff
@@ -136,8 +138,23 @@ export default defineComponent({
             },
             title: transactionTypes[txType.value],
           };
+        case TX_TYPE_MDW.PayingForTx: {
```
CedrikNikita, line 141: Actually `PayingForTx` is a wrapper on top of a transaction so, the inner transaction should be on top of this switch, because our plan is just to add additional details row to a properly parsed inner transaction.

CedrikNikita (changes requested):

src/popup/components/TransactionTagList.vue
```diff
@@ -127,6 +128,13 @@ export default defineComponent({
       }
       if (txType.value === TX_TYPE_MDW.PayingForTx) {
         return [
+          i18n.t('transaction.type.createMultisigVault'),
```
CedrikNikita, line 131: This is the case only if inner transaction is `GAAttachTx` and we are actually using this multisig contract

CedrikNikita (changes requested):

src/popup/components/TransactionTagList.vue
```diff
@@ -50,11 +50,18 @@ export default defineComponent({
     dense: Boolean,
   },
   setup(props, { root }) {
+    const getTxType = computed(() => root.$store.getters.getTxType);
+    const tx = (
+      getTxType.value(props.tx) === TX_TYPE_MDW.GAMetaTx
+      || getTxType.value(props.tx) === TX_TYPE_MDW.PayingForTx
+    )
+      ? props.tx.tx?.tx as any : props.tx as any;
```
peronczyk, line 58: The same code was used in [TransactionOverview.vue](https://github.com/aeternity/superhero-wallet/pull/1822/files#diff-7555767fde85bbc840175b6d9e1eaa7e9459580fe4cc884750809f322c9986e2)
So please consider the comment I've made there. And also if the same code is there maybe having an utility function would be better?

src/popup/components/TransactionTagList.vue
```diff
@@ -50,11 +50,18 @@ export default defineComponent({
     dense: Boolean,
   },
   setup(props, { root }) {
+    const getTxType = computed(() => root.$store.getters.getTxType);
```
peronczyk, line 53: I'm not 100% sure, but this can be removed. You are passing the `tx` to the `useTransactionTx` composable and there is `txType` taken from the `tx`.

src/popup/components/TransactionTagList.vue
```diff
@@ -63,77 +70,88 @@ export default defineComponent({
...
       if (!props.tx) return [];
+      const externalLabels = [];
+      let innerLabels = [];
+
+      if (getTxType.value(props.tx) === TX_TYPE_MDW.GAMetaTx) {
+        externalLabels.push(i18n.t('transaction.type.gaMetaTx'));
+      }
+      if (getTxType.value(props.tx) === TX_TYPE_MDW.PayingForTx) {
+        externalLabels.push(i18n.t('transaction.type.payingForTx'));
+      }
```
peronczyk, line 81: This computed property is already really messed up. Adding additional logic and two separate arrays of labels makes it almost unreadable. But first of all no need to use additional `getTxType`. You can use the `txType.value` that already exists. And then just build one array of labels with a list of ifs.

peronczyk (changes requested):

src/popup/components/TransferReview.vue
```diff
@@ -343,6 +343,7 @@ export default defineComponent({
         emit('success');
         loading.value = true;
       }
+      loading.value = true;
```
peronczyk, line 346: Something is wrong here - the same line is in `finally` block.

src/popup/components/TransactionTagList.vue
CedrikNikita: There are two transactions within new type of tx we are trying to support `innerTx` and `externalTx`. We are trying to get labels for the both of them. That is why we are getting type for both of them.

src/popup/components/TransactionTagList.vue
CedrikNikita: IDEM

peronczyk: OK, so what I suggest is to use the `useTransactionTx` twice: for the `props.tx` and for the innerTx. Or extend this composable to return itself again for the inner tx:
```
// This goes inside of the useTransactionTx
// And there is a naming collision because `innerTx` is already taken, so it should be replaced with just `tx`.
const innerTx = computed(() => useTransactionTx({
  store,
  tx: <the_check_if_the_tx_has_innerTx> ? tx.value.tx?.tx : undefined,
}));
...
return {
  ...
  innerTx,
}
```

this way here we would have:
```
const {
  txType,
  isAllowance,
  isDex,
  innerTx,
} = useTransactionTx({ store: root.$store, tx });

const {
  txType: innerTxType,
} = innerTx.value
```

src/popup/pages/TransactionDetails.vue
```diff
@@ -85,6 +85,64 @@
             </template>
           </DetailsItem>
 
+          <DetailsItem
+            v-if="multisigTransactionFeePaidBy"
+            :label="$t('pages.transactionDetails.feePaidBy')"
+            small
+          >
+            <template #value>
```
peronczyk, line 93: I changed the `<DetailsItem>` component that we don't need to use the `#value` slot as now it's the default one. So in most cases it's:
```
<DetailsItem label="...">
  Stuff goes here
</DetailsItem>
```

src/popup/pages/TransactionDetails.vue
```diff
@@ -85,6 +85,64 @@
...
+              </div>
+            </template>
+          </DetailsItem>
+
+          <DetailsItem
+            v-if="multisigContractId"
+            :label="$t('pages.transactionDetails.vaultContractId')"
+            small
+          >
+            <template #value>
```
peronczyk, line 127: IDEM

src/popup/pages/TransactionDetails.vue
```diff
@@ -275,6 +384,18 @@ export default defineComponent({
       if (transaction.value?.tx) {
         setTransactionTx(transaction.value.tx);
       }
+
+      if (getTxType.value(transaction.value?.tx) === TX_TYPE_MDW.GAMetaTx) {
+        try {
+          const { contract_id: contractId = null } = await fetchJson(
+            `https://testnet.aeternity.io/v3/accounts/${transaction.value?.tx?.gaId}`,
```
peronczyk, line 391: I think we need to use the current network settings here.

peronczyk (changes requested):

src/popup/components/TransactionTagList.vue
CedrikNikita: Well, ofc we can do that. But all the info i want to get from `useTransactionTx` composable related only to an innerTx, not an external one.

src/popup/pages/TransactionDetails.vue
```diff
@@ -260,6 +328,44 @@ export default defineComponent({
...
+
+    const multisigTransactionFeePaidBy = computed((): string | null => {
+      if (getTxType.value(transaction.value?.tx) !== TX_TYPE_MDW.PayingForTx) return null;
+      return transaction.value?.tx?.payerId ?? null;
+    });
+
+    const isContainingInternalTx = computed((): boolean => [
+      TX_TYPE_MDW.GAMetaTx,
+      TX_TYPE_MDW.PayingForTx,
+    ].includes(getTxType.value(transaction.value?.tx)));
```
peronczyk, line 355: This is already used twice. I suggest to reuse an utility function for this.

src/popup/pages/TransactionDetails.vue
```diff
@@ -236,18 +297,25 @@ export default defineComponent({
...
     const transaction = ref<ITransaction>();
+    const multisigContractId = ref<string>();
 
     const getTx = computed(() => root.$store.getters.getTx);
     const getTxTipUrl = computed(() => root.$store.getters.getTxTipUrl);
     const getExplorerPath = computed(() => root.$store.getters.getExplorerPath);
     const isTxAex9 = computed(() => root.$store.getters.isTxAex9);
     const getTxSymbol = computed(() => root.$store.getters.getTxSymbol);
     const getTxAmountTotal = computed(() => root.$store.getters.getTxAmountTotal);
+    const activeNetwork = computed(() => root.$store.getters.activeNetwork);
```
peronczyk, line 318: I suggest to add `INetwork` typing here.

src/store/getters.js
```diff
@@ -94,6 +94,12 @@ export default {
     && (TX_TYPE_MDW[tx.type]
       || SCHEMA.OBJECT_ID_TX_TYPE[tx.tag]
       || (Object.values(SCHEMA.TX_TYPE).includes(tx.type) && tx.type)),
+
+  getInnerTransaction: (_, { getTxType }) => (tx) => [
+    TX_TYPE_MDW.GAMetaTx,
+    TX_TYPE_MDW.PayingForTx,
+  ].includes(getTxType(tx)) ? tx.tx?.tx : tx,
```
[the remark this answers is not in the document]
peronczyk: Here's a PR that moves the `getTxType` from the store to the utils as it is not using the store state at all: https://github.com/aeternity/superhero-wallet/pull/1827

So if the `getTxType` is not using the store state same the `getInnerTransaction` can be moved into the `utils` file.

src/popup/components/TransactionTagList.vue
peronczyk: Sure, if you think my propositions are bad leave it as it is. I just don't like how this logic is becoming an unreadable behemoth. Previously it was rather bad and now it's even worse.

peronczyk: Another idea: maybe the `transactionTx` composable could return computed property `innerTxType`?

src/store/getters.js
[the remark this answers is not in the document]
peronczyk: Another thing that caught my eye is the name of the function - it suggest it will return the inner transaction. But:
1. It's inner tx
2. If the test return fail the regular tx is returned, so the function actually return inner or regular tx. My suggestion is to return `null` instead of the `tx` in this case.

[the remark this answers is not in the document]
CedrikNikita: This function suppose to return the innerTx for the last nested transaction, maybe i should rename it to something more meaningful, `lastNestedInnerTx`.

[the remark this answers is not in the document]
peronczyk: Sure, makes sense.

peronczyk (approved):

pstachel (approved):

CedrikNikita (approved):

PR #1823, opened 2023-02-13
(TMP) Refactor fungible tokens

ifaouibadi:


src/types/index.ts
```diff
@@ -348,6 +347,19 @@ export interface IDexContracts {
...
+  amount: number
+  reserve: number
+}
+
+export interface IDexTokenPairs {
+  token0: IDexToken
+  token1: IDexToken
+  totalSupply: number,
+  balance: number
+  balances: any
```
subhod-i, line 360: Any or `Record<string, number>`?

subhod-i (approved):
LGTM.

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,456 @@
...
+  }
+
+  function getTxSymbol(tx?: ITx): string {
+    if (!tx) {
+      return AETERNITY_SYMBOL;
+    }
+    if (tx?.contractId) {
+      return availableTokens.value[tx.contractId]?.symbol;
+    }
+    const contractCallData = categorizeContractCallTxObject({ tx });
```
CedrikNikita, line 379: This will return incorrect results in several cases, the parameter suppose to be a proper transaction not just an inner one. The fields `incomplete`, `pending` and `amount` are also required.

CedrikNikita (changes requested):
1) **Changing to per account tokens loading must be agreed**
2) I suggest to not do it in a refactoring PR
3) commit name should be changed

Currrently watching for loading of tokens per account change is not wokring correctly.
The assets page in this particular example will also show assets for a different account.
Assets not working

src/composables/transactions.ts
```diff
@@ -0,0 +1,68 @@
...
+
+interface UseTransactionsOptions {
+  /**
+   * TODO: Temporary solution to avoid dependency circle
+   * TODO: refactor all transaction related action to this composable.
+   */
+  store: Store<any>,
+}
+
+export function useTransactions({ store }: UseTransactionsOptions) {
```
CedrikNikita, line 17: We can use `IDefaultComposableOptions` here.

src/popup/components/TransferReview.vue
```diff
@@ -291,14 +301,14 @@ export default defineComponent({
         }
         const transaction: IPendingTransaction = {
           hash: txResult.hash,
-          amount,
```
CedrikNikita, line 294: You need to rework `categorizeContractCallTxObject` in order to change the structure of pending transaction.

src/store/plugins/pendingTransactionHandler.js
```diff
@@ -24,7 +24,6 @@ export default async (store) => {
           showSpendModal();
           break;
         case 'spendToken':
-          store.dispatch('fungibleTokens/loadTokenBalances');
```
CedrikNikita, line 27: After this change in the worst case the token balance will be updated in 10 seconds (current polling time).

tests/e2e/integration/other.js
```diff
@@ -1,52 +1,3 @@
...
-    time: Date.now(),
-    tx: {
-      callerId: ACCOUNT_ADDRESS,
-      contractId: STUB_TOKEN_CONTRACT_ADDRESS,
-      type: SCHEMA.TX_TYPE.contractCall,
-      function: 'transfer',
-    },
-  },
-];
-
```
CedrikNikita, line 49: Why have you decide to completly remove this test?

ifaouibadi: fixed

src/store/plugins/pendingTransactionHandler.js
ifaouibadi: replaced with https://github.com/aeternity/superhero-wallet/pull/1823/files#diff-72b4f519bcdf7b7f49059fbea39c36b6c760d84199e160db539b0879f115d8f3R126

the issue with that is, it loads the token balances instantly after the transaction has been sent and MDW needs a few seconds to sync updates.

So that line was not updating the balances properly

src/composables/accounts.ts
```diff
@@ -1,12 +1,23 @@
...
   // TODO in th future the state of the accounts should be stored in this composable
   const accounts = computed((): IAccount[] => store.getters.accounts);
   const account = computed((): IAccount => store.getters.account);
   const accountsAddressList = computed(() => accounts.value.map((acc) => acc.address));
 
+  const activeAccountAddress = computed((): string => (
+    isMultisigDashboard.value && activeMultisigAccount.value
+      ? activeMultisigAccount.value.address
+      : account.value.address
+  ));
```
peronczyk, line 19: I strongly suggest to not mix the regular accounts with multisig vaults. These are 2 separate features. Also there is no such a thing as `multisigAccount.address`.

src/types/index.ts
```diff
@@ -157,6 +157,7 @@ export interface IMultisigConsensus {
 
 export interface IMultisigAccountRaw {
   contractId: string;
+  address: string;
```
peronczyk, line 160: This is a duplicate of `gaAccountId`. Recently I removed this duplicate. Please avoid adding it again as the multisig vaults is a totally different thing than regular accounts.

peronczyk (changes requested):

src/composables/balances.ts
```diff
@@ -49,18 +50,22 @@ const initPollingWatcher = createPollingBasedOnMountedComponents();
...
+  const balance = computed(() => balances.value[activeAccountAddress.value] || new BigNumber(0));
   const balanceCurrency = computed(() => balance.value.toNumber() * currentCurrencyRate.value);
 
   async function updateBalances() {
     const sdk = await getSdk();
-    const balancesPromises = accounts.value.map(
-      ({ address }) => sdk.balance(address).catch((error) => {
+    const addresses: string[] = [
+      ...accounts.value.map((acc) => acc.address),
+      ...multisigAccounts.value.map((acc) => acc.address),
```
peronczyk, line 65: I strongly suggest to decouple multisig feature from the regular accounts. If you want to have the balances polling I suggest to create separate composable `multisigBalances`. Why? because if user is switched to the regular accounts view there is no need to poll the multisig balances and vice versa.

Also there is no such a thing as multisig address. It is `gaAccountId`.

peronczyk (changes requested):

src/popup/components/SwapRates.vue
```diff
@@ -37,39 +37,44 @@
...
 import TokenAmount from './TokenAmount.vue';
 
-export default {
+export default defineComponent({
   components: {
     Tokens,
     TokenAmount,
   },
   props: {
     transaction: { type: Object, required: true },
```
peronczyk, line 56: As this file is now using TypeScript I suggest to type the prop (probably `ITransaction`).

src/popup/components/SwapRates.vue
```diff
@@ -37,39 +37,44 @@
...
+      ...FUNCTION_TYPE_DEX.pool,
+    ].includes(props.transaction.tx.function)));
 
-      const resolver = transactionTokenInfoResolvers[camelCase(this.transaction.tx.function)];
+    const rates = computed(() => {
+      if (!isSwapTx.value) {
+        return [];
+      }
+
+      const resolver = (transactionTokenInfoResolvers as any)[
```
peronczyk, line 71: Please don't add `as any` to `transactionTokenInfoResolvers`. I've made big effort to type this. If you have issues with typing it I suggest to use Paweł's solution done in `transactionTokens.ts` file:
```
const functionName = camelCase(transaction.tx.function) as TxFunctionParsed;
```

src/popup/components/SwapRoute.vue
```diff
@@ -30,67 +30,78 @@
...
-    tokens() {
-      if (!FUNCTION_TYPE_DEX.swap.includes(this.transaction.tx.function)) return [];
-      const resolver = transactionTokenInfoResolvers[camelCase(this.transaction.tx.function)];
+  setup(props, { root }) {
+    const activeNetwork = useGetter<INetwork>('activeNetwork');
+    const { availableTokens } = useFungibleTokens({ store: root.$store });
+
+    const tokens = computed(() => {
+      if (!FUNCTION_TYPE_DEX.swap.includes(props.transaction.tx.function)) return [];
+      const resolver = (transactionTokenInfoResolvers as any)[
```
peronczyk, line 58: IDEM

peronczyk (changes requested):

CedrikNikita:
image
AE suppose to be 1 in this case

subhod-i (commented):
More updates

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,454 @@
...
+        return addressTransactions.slice(0, 10);
+      }
+    } else {
+      rawTransactions = await fetchJson(`${activeNetwork.value.middlewareUrl}/aex9/transfers/to/${address}`);
+    }
+
+    const newTransactions = rawTransactions
+      .filter((tx: any) => !getDexContracts.value.router.includes(tx.contract_id))
+      .map((tx: any) => ({
+        ...tx,
```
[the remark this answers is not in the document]
ifaouibadi: it's a different type actually added new interface for it `IRawTokenTransaction`

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,454 @@
...
+        tx: {
+          contractId: tx.contract_id,
+          senderId: tx.sender,
+          recipientId: tx.recipient,
```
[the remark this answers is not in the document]
ifaouibadi: it's used to map token transactions (`IRawTokenTransaction`) into a normal transaction (`ITx`) so it can be rendered with the same component

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,491 @@
...
+}
+
+export function useFungibleTokens({ store }: IDefaultComposableOptions) {
+  const activeNetwork = computed<INetwork>(() => store.getters.activeNetwork);
+  const getDexContracts = computed<IDexContracts>(() => store.getters.getDexContracts);
+  const middleware = computed(() => store.state.middleware);
+  const activeCurrency = computed(() => store.state.current.currency);
+  const { getSdk } = useSdk({ store });
+
+  function getTokenBalance(address: string) {
```
peronczyk, line 77: It returns an array, so the name should be plural: "...Balances".

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,491 @@
...
+   * state. If `recent` is false or `transactions.value[address]` is empty, the function fetches all
+   * transfer transactions for the address. The resulting array of transactions is stored in the
+   * `transactions` state for the address.
+   *
+   * @param {string} options.address - The address for which to fetch transfer transactions.
+   * @param {boolean} options.recent - If true, fetches only the latest transactions.
+   * @returns {Promise<ITransaction[]>} - An array of transactions for the specified address.
+   * @throws {Error} - If an unknown error occurs during the request.
+   */
+  async function getTokensHistory(address: string, recent = false): Promise<ITransaction[]> {
```
peronczyk, line 341: Little suggestion about naming pattern I would like to have in our app. Functions that are asynchronous (fetches something) should start with `fetch...`. Functions that parses existing data should start with `get...`. So in this case i suggest to use `fetch...`.

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,491 @@
...
+  }: { balanceCurrency: BigNumber | number, tokenBalance: BigNumber | Balance }): IToken | IAsset {
+    const aePublicDataExists = aePublicData.value && Object.keys(aePublicData.value).length > 0;
+
+    return {
+      ...(aePublicDataExists ? aePublicData : {} as IAsset),
+      convertedBalance: tokenBalance,
+      symbol: AETERNITY_SYMBOL,
+      balanceCurrency,
+      contractId: AETERNITY_CONTRACT_ID,
+    } as IToken | IAsset;
```
peronczyk, line 118: `as IToken | IAsset` is forcing the type here. So it makes false claim.

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,491 @@
...
+   * @returns {Object} - An object representing the Aeternity token, with the following properties:
+   *  - convertedBalance: The balance of the token.
+   *  - symbol: The symbol of the token (i.e. "AE").
+   *  - balanceCurrency: The balance currency of the token.
+   *  - contractId: The contract ID of the Aeternity token contract.
+   */
+  function getAeternityToken({
+    balanceCurrency,
+    tokenBalance,
+  }: { balanceCurrency: BigNumber | number, tokenBalance: BigNumber | Balance }): IToken | IAsset {
```
peronczyk, line 109: This function returns one structure so the return type `IToken | IAsset` is not true.

peronczyk (changes requested):

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,491 @@
...
+        const objectStructure = {
+          ...token,
+          value: contractId,
+          text: `${convertedBalance} ${token.symbol}`,
+          contractId,
+          balance,
+          convertedBalance,
+        };
+
+        return objectStructure as any;
```
peronczyk, line 206: I know that I'm annoying, but as always: I suggest to avoid using `as any`.
Also the object can be returned directly without creating `objectStructure` constant.

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,491 @@
...
+        newTokens[address].tokenBalances = balances;
+      }
+
+      tokens.value = { ...newTokens };
+    } catch (e) {
+      handleUnknownError(e);
+    }
+  }
+
+  async function getAeternityData() {
```
peronczyk, line 225: Getters (`get...`) are synchronous functions that should return something. You already used `load...` functions to update internal states so I suggest to rename this function to `load...`.

peronczyk (changes requested):

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,511 @@
...
+  ITransaction,
+  IAsset,
+  Balance,
+  IAex9Balance,
+  ITokenBalance,
+} from '../types';
+import { createPollingBasedOnMountedComponents } from './composablesHelpers';
+import { useSdk } from './sdk';
+
+const activeAddress = ref<string | null>(null);
```
peronczyk, line 45: Active account address in my opinion should be stored inside of the composable function and provided as the function argument to avoid multiple state changes when multiple components mounted at the same time is trying to use the composable with different addresses.

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,511 @@
...
+const availableTokens = ref<ITokenList>({});
```
peronczyk, line 46: I think the available tokens should be cached for each of the addresses. So here it should be:
```
const availableTokens = ref<Record<string, ITokenList>({});
```
where the `string` is in account address.

peronczyk (changes requested):

src/types/index.ts
```diff
@@ -86,7 +105,8 @@ export interface IAsset {
   atl: number
   atl_change_percentage: number
   atl_date: string
-  balanceCurrency: number
+  balanceCurrency: any
+  // balanceCurrency: number | Balance
```
peronczyk, line 109: This needs a final value.

src/composables/fungibleTokens.ts
ifaouibadi: `export type ITokenList = Record<string, IToken>`

peronczyk (approved):

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,486 @@
...
+    } catch (error) {
+      handleUnknownError(error);
+    }
+  }
+
+  /**
+   * Loads the address token balances.
+   */
+  async function loadTokenBalances() {
+    if (!accountAddress) return;
```
pstachel, line 158:
```suggestion
    if (!accountAddress) {
        return;
    }
```

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,486 @@
...
+        `${CURRENCY_URL}${activeCurrency.value}`,
+      );
+      aePublicData.value = aeternityData;
+    } catch (error) {
+      handleUnknownError(error);
+    }
+  }
+
+  async function createOrChangeAllowance([contractId, amount]: any) {
+    if (!accountAddress) return;
```
pstachel, line 225:
```suggestion
    if (!accountAddress) {
        return;
     }
```

src/composables/transactions.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    await watchUntilTruthy(() => middleware.value);
+
+    const txs: ITransaction[] = await Promise.all([
+      (recent || transactions.value.nextPageUrl === ''
+        ? middleware.value.getTxByAccount(address, limit, 1)
+        : fetchJson(`${activeNetwork.value.middlewareUrl}${transactions.value.nextPageUrl}`))
+        .then(({ data, next }: any) => {
+          const result = recent || transactions.value.nextPageUrl === '' ? data : camelcaseKeysDeep(data);
+          if (!recent) store.commit('setTransactionsNextPage', next);
```
pstachel, line 37:
```suggestion
          if (!recent) {
              store.commit('setTransactionsNextPage', next);
          }    
```

pstachel (changes requested):

src/popup/components/Modals/SpendSuccess.vue
```diff
@@ -88,45 +102,61 @@ export default {
...
+    } = useFungibleTokens({
+      store: root.$store,
+      accountAddress: props.transaction.tx.callerId,
+    });
+    const middleware = useState('middleware');
+    const getExplorerPath = useGetter('getExplorerPath');
+    const getPreferred = useGetter('names/getPreferred');
+
+    const hideAvatar = ref<boolean>(false);
+    const nameRecipient = ref<any>(null);
```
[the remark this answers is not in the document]
peronczyk: I would go further and write this as follows:
```
const nameRecipient = ref<string>();
```
This way the value would be `string | undefined`.

src/composables/fungibleTokens.ts
```diff
@@ -0,0 +1,506 @@
...
+
+    return +aettosToAe(
+      new BigNumber(
+        transaction.tx?.amount
+        || (transaction.tx?.tx?.tx as ITx)?.amount
+        || transaction.tx?.nameFee || 0,
+      )
+        .plus(isSent ? transaction.tx?.fee || 0 : 0)
+        .plus(isSent ? transaction.tx?.tx?.tx?.fee || 0 : 0)
+        .toString(),
```
CedrikNikita, line 461: Won't it work perfectly without `toString`?

src/popup/components/AccountCard.vue
```diff
@@ -54,10 +55,18 @@ export default defineComponent({
...
+    const { loadTokenBalances } = useFungibleTokens({
+      store: root.$store,
+      accountAddress: props.account.address,
+    });
     const { balance } = useBalances({ store: root.$store });
 
     const numericBalance = computed<number>(() => balance.value.toNumber());
 
+    onMounted(() => {
+      loadTokenBalances();
```
CedrikNikita, line 67: This will result in `accounts.length` number of loading started simultaneiously on each `/account` page openning.

src/popup/components/AccountCardMultisig.vue
```diff
@@ -49,7 +51,14 @@ export default defineComponent({
...
     selected: Boolean,
   },
-  setup() {
+  setup(props, { root }) {
+    const { loadTokenBalances } = useFungibleTokens({
+      store: root.$store,
+      accountAddress: props.account.gaAccountId,
+    });
+
+    onMounted(() => loadTokenBalances());
```
CedrikNikita, line 60: Same as above.

src/popup/components/Modals/SpendSuccess.vue
```diff
@@ -88,45 +102,61 @@ export default {
...
+
+    const isAe = computed((): boolean => (
+      getTxSymbol(props.transaction) === AETERNITY_SYMBOL
+    ));
+
+    const payload = computed(() => getPayload(props.transaction));
+    const hasPayload = computed((): boolean => payload.value?.length);
+
+    function closeModal() {
+      loadTokenBalances();
```
CedrikNikita, line 132: Since you yourself mentioned that token balances are loading not so fast, i think we can omit this `loadTokenBalances`. The updated amount will be uploaded in a general poll interval.

CedrikNikita (commented):
- My suggestion regarding performance improvement would be, to load token balances for each account after account discovery (once!). Then to load token balances only for a chosen account. 
- Maybe also we should not load token balances for a multisig at all for now (since we are not supporting any token related transactions, but this can be up to @paolomolo ).
- fungibleTokens plugin should be deleted

CedrikNikita:
Amount of `/aex9/balances/` endpoint calls
|Usage in 1 minute|Before|After|
|---|---|---|
|Switching accounts/multisig and switching networks|18| 43|

In **new** approach, upon using if you don't switch the account the polling is not happening at all
image

PR #1831, opened 2023-02-17
[SW-636] Improve multisig performance

CedrikNikita:
Table of dry runs called every 7 seconds
|Multisig info|Before|After|
|---|---|---|
|2 visible vaults (14 invisible/old version)|**132** dry-runs | **12** dry-runs|
|1 visible vault (0 invisible/old version)|**16** dry-runs | **6** dry-runs|


src/composables/multisigAccounts.ts
```diff
@@ -86,63 +87,62 @@ export function useMultisigAccounts({ store }: IDefaultComposableOptions) {
...
+              nonce,
+              signers,
+              consensusResult,
+              balance,
+            ] = await Promise.all([
+              contractInstance.methods.get_nonce(),
+              contractInstance.methods.get_signers(),
+              contractInstance.methods.get_consensus_info(),
+              gaAccountId ? sdk.balance(gaAccountId) : 0,
+            ]);
```
subhod-i, line 109: This is what I was talking about earlier. This information is not needed in the background. `signers, nonce, activeTx` data is needed only to display when the multisigAccount card is active. We could run the polls for these details only on the Active multisig tx review page( on demand)

subhod-i (approved):

subhod-i:
LGTM.

src/composables/multisigAccounts.ts
CedrikNikita: With your approach when you will swtich between multisig accounts you will not see an info instantly.

subhod-i: It's a trade-off. You would otherwise simply call the dry run a hundred times and this data is not needed. 

Maybe we can fetch once in the beginning and later update on demand.

CedrikNikita: I agree about `signers` and `nonce`, but `activeTx` is still needed on dashboard.

subhod-i: Cool. May not be important for the MVP. PR looks good to me as of now.

Liubov-crypto:
Tested, LGTM

pstachel (approved):

PR #1833, opened 2023-02-17
Upgrade to sdk v13.0.0-alpha.1

subhod-i:


src/popup/pages/Names/AuctionBid.vue
```diff
@@ -21,7 +21,7 @@
         <DetailsItem :label="$t('total')">
           <template #value>
             <TokenAmount
-              :amount="+totalsAmount"
+              :amount="+amountTotal"
```
CedrikNikita, line 24: I suggest to move it to the separate PR, in order to faster introduce this fix.

PR #1835, opened 2023-02-17
[SW-631] Allow to go to multisig vault details after creating it

peronczyk:


src/composables/multisigAccounts.ts
```diff
@@ -107,31 +117,32 @@ export function useMultisigAccounts({ store }: IDefaultComposableOptions) {
...
             version: version.decodedResult,
-            consensusLabel,
-            contractId,
             balance: convertToken(balance, -MAGNITUDE),
-            address: gaAccountId,
-            multisigAccountId: gaAccountId,
+            refusedBy: [],
           };
+
+          return account;
```
pstachel, line 145: Why do you assign a value to the account variable and then return it if it's not used anywhere?

src/popup/components/AccountCard.vue
```diff
@@ -57,7 +57,9 @@ export default defineComponent({
   setup(props, { root }) {
     const { isMultisigDashboard } = useMultisigAccounts({ store: root.$store });
 
-    const color = computed(() => getAddressColor(props.account.address));
+    const color = computed(() => getAddressColor(isMultisigDashboard.value
+      ? (props.account as IMultisigAccount).gaAccountId
+      : (props.account as IAccount).address));
```
pstachel, line 62: Could you please leave a TODO here for a navigation refactor?

src/popup/components/AccountInfo.vue
```diff
@@ -66,14 +67,16 @@ export default defineComponent({
...
-    account: { type: Object, required: true },
+    account: { type: Object as PropType<IAccount | IMultisigAccount>, required: true },
   },
   setup(props, { root }) {
     const { isMultisigDashboard } = useMultisigAccounts({ store: root.$store });
 
     const activeNetwork = useGetter('activeNetwork');
 
-    const address = computed(() => props.account.address);
+    const address = computed(() => isMultisigDashboard.value
```
pstachel, line 77: Please leave a comment here for navigation refactor

pstachel (approved):

src/composables/multisigAccounts.ts
peronczyk: OK, I'll use the type on the function return.

src/composables/multisigAccounts.ts
```diff
@@ -107,30 +121,29 @@ export function useMultisigAccounts({ store }: IDefaultComposableOptions) {
             gaAccountId ? sdk.balance(gaAccountId) : 0,
           ]);
 
-          const consensus = consensusResult.decodedResult as any;
+          const consensus: IMultisigConsensus = camelCaseKeysDeep(consensusResult.decodedResult);
```
CedrikNikita, line 124: This will lead to an error, `txHash` is breaking in this case.

CedrikNikita (changes requested):

src/composables/multisigAccounts.ts
CedrikNikita: image
First initial value, second is after `camelcaseKeysDeep`

peronczyk: Thanks for the clarification. Please take a look on my change. I console.log'ged it and looks as it should be. But maybe you see more issues?

src/composables/multisigAccounts.ts
```diff
@@ -108,40 +122,41 @@ export function useMultisigAccounts({ store }: IDefaultComposableOptions) {
...
               consensusLabel,
               contractId,
+              gaAccountId,
+              nonce: Number(nonce.decodedResult),
+              signers: signers.decodedResult,
               balance: convertToken(balance, -MAGNITUDE),
-              address: gaAccountId,
-              multisigAccountId: gaAccountId,
-              hasPendingTransaction: consensus.tx_hash && !consensus.expired,
+              refusedBy: [],
```
CedrikNikita, line 145:
```suggestion
```
This is one of the consensus fields which is being rewritten.

src/composables/multisigAccounts.ts
```diff
@@ -159,13 +174,13 @@ export function useMultisigAccounts({ store }: IDefaultComposableOptions) {
...
         if (!b.balance.minus(a.balance).isZero()) {
-          return b.balance.minus(a.balance);
+          return b.balance.minus(a.balance).toNumber();
         }
         return new Date(a.createdAt).getTime() - new Date(b.createdAt).getTime();
       });
 
     if (!activeMultisigAccountId.value) {
-      setActiveMultisigAccountId(result[0]?.multisigAccountId);
+      setActiveMultisigAccountId(result[0]?.gaAccountId);
```
CedrikNikita, line 183: You were suggesting to stick either to `ga` notation or to `multisig` ones.

CedrikNikita (changes requested):

src/composables/multisigAccounts.ts
peronczyk: `gaAccountId` is a thing that comes from the API, so here I'm sticking with the root naming to avoid unnecessary data properties transformation.

src/composables/multisigAccounts.ts
peronczyk: OK, removed.

CedrikNikita (approved):
Code changes look reasonable.

PR #1836, opened 2023-02-17
[SW-651] fix: change drySdk network on network change

CedrikNikita:


src/types/index.ts
```diff
@@ -400,6 +400,7 @@ export interface ITopHeader {
 export type ISignMessage = (m: any) => Promise<any>
 
 export interface ISdk {
+  addNode: (a: string, b: any, c: boolean) => void
```
pstachel, line 403: I'd suggest sticking to meaningful names

pstachel (approved):

ifaouibadi (approved):

subhod-i:
This is not needed. `drySdk` is never attached to a node or it is not meant to make any calls to the node because we are already using `ACI` to get the contract instance at https://github.com/aeternity/superhero-wallet/blob/3c852d9d6496c7e9647c4c7dfd95c39ffcd8cf47/src/composables/multisigAccountCreate.ts#L27-L30

Also, we shall be able to send the confirmed Tx to the chain using regular SDK itself at https://github.com/aeternity/superhero-wallet/blob/3c852d9d6496c7e9647c4c7dfd95c39ffcd8cf47/src/composables/multisigTransactions.ts#L136-L140

subhod-i:
> This is not needed. `drySdk` is never attached to a node or it is not meant to make any calls to the node. We are already using `ACI` to get the contract instance at
> 
> https://github.com/aeternity/superhero-wallet/blob/3c852d9d6496c7e9647c4c7dfd95c39ffcd8cf47/src/composables/multisigAccountCreate.ts#L27-L30
> 
> Also, we shall be able to send the confirmed Tx to the chain using regular SDK itself at
> 
> https://github.com/aeternity/superhero-wallet/blob/3c852d9d6496c7e9647c4c7dfd95c39ffcd8cf47/src/composables/multisigTransactions.ts#L136-L140

If we achieve this then this PR could be discarded

davidyuk:
As I understand if we doing dry run of contract call that refers to an existing contract deployed on a specific network, then dryRunSdk should know about the current network

Liubov-crypto:
The same behaviour and view as in develop branch, excluding obsolete transactions.

PR #1842, opened 2023-02-20
[SW-384] Multisig account address is missing on Propose and approve screen

konradodo:


pstachel (approved):

src/popup/components/TransferReview.vue
```diff
@@ -389,13 +398,22 @@ export default defineComponent({
...
+      activeMultisigAccount,
     };
   },
 });
 </script>
 
 <style scoped lang="scss">
-.details-item {
-  margin-top: 16px;
+.review-wrapper {
```
CedrikNikita, line 408: I suggest to rename the root class of the component to it's name, but in kebab-case as in every other place in the project.

src/popup/components/TransferReview.vue
```diff
@@ -389,13 +398,22 @@ export default defineComponent({
...
+  .details-item {
+    margin-top: 16px;
+  }
+
+  .multisig-account {
+    display: flex;
+    justify-content: center;
+  }
 }
+
```
CedrikNikita, line 418:
```suggestion
```

src/popup/components/TransferReview.vue
```diff
@@ -1,10 +1,17 @@
 <template>
-  <div class="review-wrapper">
+  <div class="transfer-review">
     <ModalHeader
       :title="isMultisig ? $t('modals.multisigTxProposal.title') : $t('pages.send.reviewtx')"
       :subtitle="isMultisig ? null : $t('pages.send.checkalert')"
     />
 
+    <div
+      v-if="isMultisig"
```
subhod-i, line 9: `v-show` instead of `v-if`

subhod-i (approved):

src/popup/components/TransferReview.vue
konradodo: hmm i always prefer v-if if i not need render something

PR #1850, opened 2023-02-21
[SW-654] cleaning transaction after change multisig account

konradodo:


ifaouibadi (approved):

src/popup/components/TransactionList.vue
```diff
@@ -238,6 +238,7 @@ export default defineComponent({
     });
 
     onMounted(async () => {
+      root.$store.commit('initTransactions');
```
pstachel, line 241: I think that initTransaction is handled on account change - you could work on that or wait for SW-620 to be merged

src/popup/components/TransactionList.vue
```diff
@@ -214,42 +215,28 @@ export default defineComponent({
       }
     }
 
-    async function loadMore() {
+    async function loadMore(recent: boolean = false) {
       if (loading.value) return;
```
pstachel, line 219: if (loading.value) {
    return;
}

src/popup/components/TransactionList.vue
```diff
@@ -214,42 +215,28 @@ export default defineComponent({
...
       checkLoadMore();
     });
 
     onMounted(async () => {
-      loadMore();
-      const polling = setInterval(() => getLatest(), 10000);
-      root.$once('hook:destroyed', () => {
-        clearInterval(polling);
-        isDestroyed.value = true;
-      });
```
pstachel, line 247: why did we get rid of transaction polling?

pstachel (changes requested):

pstachel (approved):

src/popup/components/TransactionList.vue
```diff
@@ -168,13 +170,13 @@ export default defineComponent({
...
-                      tr.tx.recipientId === currentAddress.value
-                      || (
-                        tr.tx.senderId !== currentAddress.value
-                          && tr.tx.recipientId.startsWith('nm_')
-                      )
+                  && (
+                    tr.tx.recipientId === currentAddress.value
+                    || (
+                      tr.tx.senderId !== currentAddress.value
+                      && tr.tx.recipientId.startsWith('nm_')
```
peronczyk, line 177: By the way you can replace this with `isAensName` utility function.

peronczyk (approved):

CedrikNikita (changes requested):
This PR is not fixing 654 completly, it is done in #1863.
I suggest to leave only a truncateWidth fix here.

PR #1856, opened 2023-02-22
[SW-694] latest transaction correctly showing, transaction item fixes

konradodo:


src/popup/components/LatestTransactionsCard.vue
```diff
@@ -104,10 +104,10 @@ export default defineComponent({
 
       const allTransactions = await Promise.all(allTransactionsPromises);
 
-      latestTransactions.value = uniqBy(
```
konradodo, line 107: uniqBy -> here i whait for Paulo decision, either it will be removed or not

pstachel (approved):

src/store/getters.js
```diff
@@ -118,14 +118,17 @@ export default {
     new BigNumber(transaction.fee || transaction.tx?.fee || 0),
   ),
   getTxDirection: (_, { account: { address } }) => (tx, externalAddress) => {
+    const currentAddress = externalAddress || address;
+
     if (getTxType(tx) === SCHEMA.TX_TYPE.spend) {
-      return tx.recipientId === address
+      return tx.recipientId === currentAddress
```
CedrikNikita, line 124: @pstachel and @Liubov-crypto had some concerns about this changes previously.

pstachel: Yes, I did, but you were right that it should look like that

src/popup/components/LatestTransactionsCard.vue
```diff
@@ -30,7 +30,6 @@ import {
   ref,
   watch,
```
CedrikNikita, line 31:
```
Duplicate keys detected: 'ak_2AVeRypSdS4ZosdKWW1C4avWU4eeC2Yq7oP7guBGy8jkxdYVUy-th_2bVpja3SK2YAffwFrteGPpXxA9epcbsu5VxoXKbf9EPKqwaT5R-0'. This may cause an update error.
```

In case one subaccount sending to another latest transaction.
https://github.com/aeternity/superhero-wallet/blob/c0b57b68835d070d8af194bcbe3049459b458378/src/popup/components/LatestTransactionsCard.vue#L19

CedrikNikita (changes requested):

src/popup/components/LatestTransactionsCard.vue
konradodo: i was unable to reproduce it log

konradodo: before that pr, problem can be exist, because "direction" not worked properly and code has wrong transaction Owner

CedrikNikita (changes requested):
> ```
> Duplicate keys detected: 'ak_2AVeRypSdS4ZosdKWW1C4avWU4eeC2Yq7oP7guBGy8jkxdYVUy-th_2bVpja3SK2YAffwFrteGPpXxA9epcbsu5VxoXKbf9EPKqwaT5R-0'. This may cause an update error.
> ```

This problem will appear if you will send fungible token to an another subaccount, in this case transaction would appear twice from `getTokensHistory` and from `getTxByAccountAddress`.

CedrikNikita:
```
popup.9c70b4d7.js:1 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'slice')
```

src/composables/transactionTx.ts
```diff
@@ -98,8 +100,9 @@ export function useTransactionTx({
       ? TX_FUNCTIONS.received
       : getTxDirection.value(
         innerTx.value,
-        ownershipStatus.value !== TRANSACTION_OWNERSHIP_STATUS.current
-        && txOwnerAddress.value,
+        externalAddress
+        || (ownershipStatus.value !== TRANSACTION_OWNERSHIP_STATUS.current
+        && txOwnerAddress.value),
```
pstachel, line 105: That's a mistake - || and && are not on the same level

src/popup/components/TransactionLabel.vue
```diff
@@ -222,6 +221,12 @@ export default defineComponent({
         };
       }
 
+      if (txType.value === TX_TYPE_MDW.GAAttachTx
+        && externalTxType === TX_TYPE_MDW.PayingForTx
+      ) {
```
pstachel, line 226: if you are splitting the condition you should either drop whole condition or nothing

src/popup/utils/utils.ts
```diff
@@ -315,7 +315,14 @@ export function defaultTransactionSortingCallback(
     if (withoutTimeIndex === 1) {
       return 1;
     }
-    return bMicroTime - aMicroTime;
+    const direction = bMicroTime - aMicroTime;
+    // Workaround to display received transaction after send (they have the same time)
+    if (direction === 0) {
+      // eslint-disable-next-line no-shadow
```
pstachel, line 321: that's a mistake

pstachel (changes requested):

pstachel (approved):

src/popup/locales/en.json
```diff
@@ -381,7 +381,8 @@
       "propose": "Propose Tx",
       "revoke": "Revoke Tx",
       "refuse": "Disapprove Tx",
-      "confirm": "Approve Tx"
+      "confirm": "Approve Tx",
+      "vaultCreated": "Vault created"
```
CedrikNikita, line 385:
```suggestion
```
Is not used.

CedrikNikita (changes requested):
```
TypeError: Cannot read properties of undefined (reading 'clientWidth')
```
Still persist.

src/types/index.ts
```diff
@@ -358,6 +358,11 @@ export interface ITransaction {
   url?: string;
 }
 
+export interface IDashboardTransaction extends ITransaction {
+  transactionOwner?: string,
```
CedrikNikita, line 362: `ITx` already includes `transactionOwner`.

src/popup/components/TransactionLabel.vue
```diff
@@ -35,7 +35,6 @@
           v-if="externalLabel.length"
         >
```
CedrikNikita, line 36: Currently in case the type failed to be found (another discussion, should be fixed on sdk update), the` txType.value` will return `false`, so next line should be adjust
https://github.com/aeternity/superhero-wallet/blob/fd2a6d7067aabb6cae32f45a07c53868ad5e85d1/src/popup/components/TransactionLabel.vue#L234
to
`if(txType.value && txType.value.includes('name'))`

CedrikNikita (changes requested):

src/popup/components/TransactionLabel.vue
```diff
@@ -230,7 +229,7 @@ export default defineComponent({
 
       const translation = transactionListTypes[txType.value!] || transactionTypes[txType.value!];
 
-      if (txType.value && txType?.value?.includes('name')) {
+      if (txType.value && txType.value?.includes('name')) {
```
pstachel, line 232: Wasn't that fixed in your other PR? If not - why `txType.value` is doubled here?

konradodo: must be this way, because txType.value sometimes can be "false"

pstachel: yes and that's solved by the question mark after it

CedrikNikita: image

pstachel: Ah yes I missed that it would have wrong type, never mind then

ifaouibadi (approved):

peronczyk (approved):

src/popup/components/LatestTransactionsCard.vue
```diff
@@ -82,6 +84,7 @@ export default defineComponent({
             (transaction: ITransaction) => ({
               ...transaction,
               transactionOwner: address,
+              direction: getTxDirection.value(transaction.tx, address),
```
CedrikNikita, line 87: Usage of `getTxDirection` is incorrect here, transactions may have innerTx inside. I would suggest to use `direction` from `transactionTx` composable.

CedrikNikita (changes requested):

CedrikNikita (approved):
Code changes look reasonable. I didn' test it.

PR #1858, opened 2023-02-22
[SW-667] Use composition api in avatar component and refactor some store utilities

peronczyk:


src/popup/components/Avatar.vue
```diff
@@ -8,50 +8,46 @@
   >
 </template>
 
-<script>
-import { mapState } from 'vuex';
+<script lang="ts">
+import { computed, defineComponent, ref } from '@vue/composition-api';
+import { useGetter } from '../../composables/vuex';
+import { INetwork } from '../../types';
```
pstachel, line 14: import type

pstachel (approved):

src/popup/components/Avatar.vue
peronczyk: Done

ifaouibadi (approved):

PR #1859, opened 2023-02-22
use chosen account to pay for vault creation

subhod-i:
Set active account to the chosen one and then rollback the active account index once `payForTransaction` is built.

### DIFF
--- a/src/popup/components/MultisigVaultCreateReview.vue
+++ b/src/popup/components/MultisigVaultCreateReview.vue
@@ -110,8 +110,8 @@ import {
   ICreateMultisigAccount,
   IMultisigCreationPhase,
 } from '../../types';
-import { AETERNITY_SYMBOL } from '../utils';
-import { useAccounts, useSdk } from '../../composables';
+import { AETERNITY_SYMBOL, handleUnknownError } from '../utils';
+import { useAccounts, useMultisigAccountCreate, useSdk } from '../../composables';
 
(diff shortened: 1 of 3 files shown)


src/popup/components/MultisigVaultCreateReview.vue
```diff
@@ -146,12 +148,30 @@ export default defineComponent({
...
 
     watch(creatorAddress, async (val, oldVal) => {
       if (val !== oldVal) {
         creatorAccountFetched.value = undefined;
         const sdk = await getSdk();
         creatorAccountFetched.value = await sdk.api.getAccountByPubkey(val) as IAccountFetched;
+        const { idx: activeAccountIdx } = account.value;
+        root.$store.commit(
+          'accounts/setActiveIdx',
+          accounts.value.find(({ address }) => address === val)?.idx,
```
pstachel, line 162: Shouldn't we break this function execution in case there is no matching account found?

subhod-i: It is guranteed to be found because we are watching here for account selection and  that list is from `acccounts` state itself.

pstachel (approved):

konradodo (approved):

pstachel (approved):

subhod-i:
Closing in favor of https://github.com/aeternity/superhero-wallet/pull/1881

PR #1867, opened 2023-02-27
[SW-673] fix max amount & fee calculation

ifaouibadi:


src/composables/maxAmount.ts
```diff
@@ -49,89 +49,93 @@ export function useMaxAmount({ store, formModel }: MaxAmountOptions) {
...
-          || tokenInstance.value.deployInfo.address !== val.selectedAsset.contractId
-        ) {
-          tokenInstance.value = await sdk.getContractInstance({
-            source: FUNGIBLE_TOKEN_CONTRACT,
-            contractAddress: val.selectedAsset.contractId,
-          });
-        }
-        selectedAssetDecimals.value = val.selectedAsset.decimals!;
-      }
+    const _max = selectedAssetBalance.minus(fee.value);
```
konradodo, line 65:
```suggestion
    const maxValue = selectedAssetBalance.minus(fee.value);
```

PR #1871, opened 2023-02-27
[SW-615] fix: do not show all WAE transactions as DEX

CedrikNikita:


src/composables/transactionTx.ts
```diff
@@ -89,6 +90,11 @@ export function useTransactionTx({
 
     return !!(
       innerTx.value?.contractId
+      && innerTx.value?.function
+      && (
+        Object.values(flatten(Object.values(FUNCTION_TYPE_DEX)))
+          .includes(innerTx.value?.function as TxFunctionRaw)
+      )
```
pstachel, line 97: I don't see the need for the first `Object.values()` call. Also we could reduce our dependency on external libraries and use plain js (but that's just a suggestion).
```suggestion
      && (
        Object.values(FUNCTION_TYPE_DEX).flat()
          .includes(innerTx.value?.function as TxFunctionRaw)
      )
```

src/composables/transactionTx.ts
```diff
@@ -89,6 +89,11 @@ export function useTransactionTx({
 
     return !!(
       innerTx.value?.contractId
+      && innerTx.value?.function
+      && (
+        Object.values(FUNCTION_TYPE_DEX).flat()
```
peronczyk, line 94: I suggest to use `Object.values(TX_FUNCTIONS)` here.

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -222,13 +219,10 @@ export default defineComponent({
       return 'total';
     });
 
-    const totalAmount = computed(() => getTxAmountTotal.value(
-      txWrapped.value,
-      getTxDirection.value(props.transaction),
-    ));
+    const totalAmount = computed(() => getTxAmountTotal.value(txWrapped.value, direction.value));
```
peronczyk, line 222: This change makes the code less readable in my opinion. Less lines != better code :)

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -222,13 +219,10 @@ export default defineComponent({
...
 
     const singleToken = computed((): ITokenResolved => ({
-      isReceived: getTxDirection.value(props.transaction) === TX_FUNCTIONS.received,
+      isReceived: direction.value === TX_FUNCTIONS.received,
```
peronczyk, line 225: Side note: isn't this a mistake that the direction values are stored in the `TX_FUNCTION` collection?

src/composables/transactionTx.ts
CedrikNikita: We are searching only for DEXfunctions, but `TX_FUNCTIONS` contain all function not only DEX.

peronczyk (approved):

src/popup/components/Modals/ConfirmTransactionSign.vue
CedrikNikita: Looks like a mistake to me.

src/composables/transactionTx.ts
peronczyk: Uhhh, OK. Thanks for clarification.

pstachel (approved):

PR #1874, opened 2023-02-28
[SW-690] Move the currencies logic from store to a composable

peronczyk:


src/composables/currencies.ts
```diff
@@ -0,0 +1,89 @@
...
+const LOCAL_STORAGE_CURRENCY_KEY = 'currency';
+const APP_CURRENCY_CODES = CURRENCIES.map(({ code }) => code).join(',');
+
+const currencyRates = ref<Record<string, number>>({});
+const currentCurrencyCode = ref<CurrencyCode>(getLocalStorageItem<CurrencyCode>([LOCAL_STORAGE_CURRENCY_KEY]) || 'usd');
+
+const initPollingWatcher = createPollingBasedOnMountedComponents();
+
+export function useCurrencies() {
+  const minTipAmount = computed(() => 0.01 / currencyRates.value.usd);
```
pstachel, line 23: wouldn't the initial value for `currencyRates` be 0?

src/popup/components/InputAmountV2.vue
```diff
@@ -101,12 +101,15 @@ export default defineComponent({
 
     const currentAsset = computed((): IAsset => props.selectedAsset || aeToken);
     const hasError = computed(() => (root as any).$validator.errors.has('amount'));
-    const totalAmount = computed(
-      () => (currentAsset.value?.current_price)
-        ? ((+props.value || 0) * currentAsset.value.current_price).toFixed(2)
+    const totalAmountFormatted = computed(() => formatCurrency(
+      (currentAsset.value?.current_price)
```
pstachel, line 105:
```suggestion
      currentAsset.value?.current_price
```

src/popup/components/TokenAmount.vue
```diff
@@ -59,23 +60,30 @@ export default {
...
+
+    const amountRounded = computed(() => {
+      if (Number.isInteger(props.amount) || props.amount === 0) {
+        return props.amount;
+      }
+      return props.amount.toFixed((props.highPrecision || props.amount < 0.01) ? 9 : 2);
+    });
+
+    const amountFiat = computed((): string => {
+      if (props.hideFiat || props.aex9) return '';
```
pstachel, line 74:
```suggestion
      if (props.hideFiat || props.aex9) {
          return '';
      }
```

src/popup/components/TokenAmount.vue
```diff
@@ -59,23 +60,30 @@ export default {
...
+      const converted = convertToCurrency(props.amount);
+      if (props.amount === 0) return formatCurrency(0);
+      if (converted < 0.01) return `<${formatCurrency(0.01)}`;
+      return `≈${formatCurrency(converted)}`;
+    });
```
pstachel, line 79: Please wrap the returns with braces

pstachel (approved):

src/popup/components/TokenAmount.vue
peronczyk: Sure. I also see that this piece of code already exists. So I'll reuse it as an method.

src/popup/components/InputAmountV2.vue
peronczyk: I already mentioned few times that wrapping the condition with braces in one liners makes the code more readable. At least in my opinion.

src/composables/currencies.ts
peronczyk: image
This is how it looked before. I just moved the code from one place to another. I'll try to address your concerns on the chat.

peronczyk: I checked few things and the rate for euro is `0.070233`. So `0.01 / 0.070233 = 0,1428` - I think it's OK.

davidyuk:
```suggestion
  const minTipAmount = computed(() => currencyRates.value.usd && (0.01 / currencyRates.value.usd));
```
🤷‍♀️  or to ensure that you have some initial ae/usd price

peronczyk: Ups, my mistake - I misunderstood your question. `currencyRates.value.usd` is undefined in the initial state, so `0.01 / undefined = NaN`. It's not causing error, but still is a mistake. I added a fallback to `1` in case of falsy value.

subhod-i:
This refactoring looks good. We need to discuss heavy polling in our app. There shall be some way to optimize the polling. My first thought is we shall move to websockets for accounts, balances, and currencies data sources.

src/composables/currencies.ts
```diff
@@ -0,0 +1,102 @@
...
+  setLocalStorageItem,
+} from '../popup/utils';
+import { createPollingBasedOnMountedComponents } from './composablesHelpers';
+
+const POLLING_INTERVAL = 3600000;
+const LOCAL_STORAGE_CURRENCY_KEY = 'currency';
+const APP_CURRENCY_CODES = CURRENCIES.map(({ code }) => code).join(',');
+
+const currencyRates = ref<Record<string, number>>({});
+const currentCurrencyCode = ref<CurrencyCode>(getLocalStorageItem<CurrencyCode>([LOCAL_STORAGE_CURRENCY_KEY]) || 'usd');
```
konradodo, line 18: usd -> constans

konradodo (approved):

src/composables/currencies.ts
peronczyk: OK

PR #1875, opened 2023-02-28
[SW-707] refactor: exclude DEX transactions from Out

CedrikNikita:


onvisions:
As far as the DEX transactions (swap, provide liquidity) are both OUT and IN (you send token and receive token) I would suggest that at the moment filtering logic should be:

ALL: show all transactions;
OUT: show Spend Out + DEX transactions;
IN: show all Spend In + DEX transactions;
DEX: show only the DEX transactions.

On the other hand from user's perspective it might be expected and more useful if the DEX transactions are excluded from IN and OUT filtering lists. We need to make some extra research about this one, @paolomolo.

subhod-i (approved):

ifaouibadi (approved):

subhod-i (approved):
If the logic is right then PR changes LGTM.

konradodo (approved):

PR #1877, opened 2023-02-28
[SW-684] Disable URL tipping for URLs inside wallet

konradodo:


ifaouibadi (approved):

src/popup/components/TransferSendForm.vue
```diff
@@ -53,7 +53,7 @@
       v-validate="{
         required: true,
         not_same_as: isMultisig? multisigVaultAddress : account.address,
-        name_registered_address_or_url: true,
+        name_registered_address: true,
```
pstachel, line 56: Shouldn't we still be able to tip superhero.com related urls?

CedrikNikita (changes requested):
It is not disabled fully at this point. Would be nice to have @paolomolo vision on that.

1) It is not a disabling, but a removing of a feature
2) We can still be redirected to this page (upon sharing functionality) from cordova, expecting to be able to tip on the url

src/popup/components/TransferSendForm.vue
peronczyk: From the UI: no. We are only leaving some logic to allow tipping of superhero URLs done on the superhero.com side.

PR #1879, opened 2023-03-01
[SW-626] fix: multisig assets list

ifaouibadi:


src/popup/pages/AccountDetailsMultisigTokens.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+  <div class="token-list-wrapper">
```
CedrikNikita, line 2: I suggest to name a root class as a component name, but in kebab-case as every other component had.

CedrikNikita (approved):
In general, fairly simple short term solution, we were searching for. I didn't test it well enough.

pstachel (approved):

src/popup/pages/FungibleTokens/TokenContainer.vue
```diff
@@ -155,12 +164,17 @@ export default defineComponent({
     const simplexLink = computed(() => buildSimplexLink(account.value.address));
     const showFilterBar = computed(() => !!root.$route?.meta?.showFilterBar);
 
+    const tokenBalance = computed(() => (
+      isMultisig.value
+        ? activeMultisigAccount.value?.balance || new BigNumber(0)
+        : balance.value || new BigNumber(0)
+    ));
```
pstachel, line 171: I think you could export this one from tokenList.ts

PR #1880, opened 2023-03-01
[SW-673] max amount & fee calculator

ifaouibadi:


src/composables/maxAmount.ts
```diff
@@ -45,111 +41,93 @@ export interface MaxAmountOptions extends IDefaultComposableOptions {
...
-        selectedAssetDecimals.value = val.selectedAsset.decimals!;
-      }
+    const maxValue = selectedAssetBalance.minus(fee.value);
+    return (maxValue.isPositive() ? maxValue : 0).toString();
+  });
 
-      const numericAmount = (val.amount && val.amount > 0) ? val.amount : 0;
-      const amount = new BigNumber(numericAmount).shiftedBy(MAGNITUDE);
+  async function updateCalculatedFee() {
+    if (!formModel.value?.selectedAsset) return;
```
pstachel, line 69:
```suggestion
    if (!formModel.value?.selectedAsset) {
       return;
     }
```

src/composables/maxAmount.ts
```diff
@@ -45,111 +41,93 @@ export interface MaxAmountOptions extends IDefaultComposableOptions {
...
+    const sdk = await getSdk();
+    const numericAmount = (
+      formModel.value.amount
+      && formModel.value.amount > 0
```
pstachel, line 73:
```suggestion
      formModel.value?.amount > 0
```

pstachel, line 73:
```suggestion
      formModel.value?.amount > 0
```

src/composables/maxAmount.ts
```diff
@@ -45,111 +41,93 @@ export interface MaxAmountOptions extends IDefaultComposableOptions {
...
-              : val.selectedAsset.contractId,
-          },
-        );
-        return;
-      }
+    try {
+      nonce.value = (await sdk.api
+        .getAccountByPubkey(account.value.address))?.nonce;
+    } catch (error: any) {
+      if (!error.message.includes('Account not found')) handleUnknownError(error);
```
pstachel, line 81:
```suggestion
      if (!error.message.includes('Account not found')) {
          handleUnknownError(error);
      }
```

src/composables/maxAmount.ts
```diff
@@ -45,111 +41,93 @@ export interface MaxAmountOptions extends IDefaultComposableOptions {
...
+        ...sdk.Ae.defaults,
+        senderId: account.value.address,
+        recipientId: account.value.address,
+        amount,
+        ttl: 0,
+        nonce: nonce.value + 1,
+        payload: formModel.value.payload,
+      },
+    })).shiftedBy(-MAGNITUDE);
+    if (!minFee.isEqualTo(fee.value)) fee.value = minFee;
```
pstachel, line 119:
```suggestion
    if (!minFee.isEqualTo(fee.value)) {
        fee.value = minFee;
    }
```

pstachel (approved):

src/composables/maxAmount.ts
```diff
@@ -45,111 +41,96 @@ export interface MaxAmountOptions extends IDefaultComposableOptions {
...
-          || tokenInstance.value.deployInfo.address !== val.selectedAsset.contractId
-        ) {
-          tokenInstance.value = await sdk.getContractInstance({
-            source: FUNGIBLE_TOKEN_CONTRACT,
-            contractAddress: val.selectedAsset.contractId,
-          });
-        }
-        selectedAssetDecimals.value = val.selectedAsset.decimals!;
-      }
+    const maxValue = selectedAssetBalance.minus(fee.value);
```
CedrikNikita, line 64: There are no need to minus fee if you are sending token.

CedrikNikita (changes requested):

CedrikNikita (changes requested):
- The reason for the constant updating of the token balance is that middleware is syncing slower than contract updates it's balance.
So sending two transactions in a row: first with amount and second one with MAX amount would fail in this case.
- The reason for fetching nonce is to calculate fee correctly (because at some point nonce will be high enough to impact the fee)

PR #1886, opened 2023-03-01
[SW-702] Move the middleware initialization logic from the store to a composable

peronczyk:
…mposable


src/popup/pages/Names/Auction.vue
```diff
@@ -35,44 +43,51 @@ export default {
...
+    async function updateAuctionEntry() {
+      loading.value = true;
+      const middleware = await getMiddleware();
       try {
-        const res = await this.$store.state.middleware.getNameById(this.name);
+        const res = await middleware.getNameById(props.name);
         const { auctionEnd, bids } = res.info;
-        const loadedBids = await Promise.all(bids.map(async (txId) => {
-          const { tx } = await this.$store.state.middleware.getTxByIndex(txId);
+        const loadedBids = await Promise.all(bids.map(async (txId: any) => {
```
pstachel, line 57: If it's `txId` aren't we able to retrieve it's type (number)?

src/store/plugins/names.js
```diff
@@ -224,7 +231,7 @@ export default (store) => {
   store.watch(
     ({ accounts: { hdWallet: { nextAccountIdx } } }) => nextAccountIdx,
     async () => {
-      if (!store.state.middleware) return;
+      if (!isMiddlewareReady.value) return;
```
pstachel, line 234:
```suggestion
      if (!isMiddlewareReady.value) {
          return;
      }
```

pstachel (approved):

src/popup/pages/Names/Auction.vue
peronczyk: Sure, makes sense.

src/composables/middleware.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    // Old swagger.json file is not maintained anymore so we need to tweak this
+    // entry to align with the current v1 middleware.
+    // TODO Not applicable when we move to newer version of the MDW.
+    spec.paths['/name/pointees/{id}'] = spec.paths['/names/pointees/{id}'];
+    delete spec.paths['/names/pointees/{id}'];
+
+    middleware.value = mapObject(
+      (await genSwaggerClient(middlewareUrl, { spec })).api,
+      ([k, v]: any[]) => [camelCase(k), v],
```
konradodo, line 85: for me better is written full name for variables

src/popup/utils/utils.ts
```diff
@@ -199,14 +199,15 @@ export function buildSimplexLink(address: string) {
...
   return new Promise((resolve) => {
     const unwatch = watch(
       getter,
       (value) => {
-        if (!value) return;
-        resolve(getter() as NonNullable<T>);
-        defer(() => unwatch());
+        if (value) {
+          resolve(value as NonNullable<T>);
+          defer(() => unwatch());
```
konradodo, line 209:
```suggestion
          defer(unwatch);
```

src/store/actions.js
```diff
@@ -91,17 +91,21 @@ export default {
...
   },
-  async fetchTransactions({
-    state, getters, dispatch, commit,
-  }, { limit, recent, address }) {
-    if (!state.middleware || (state.transactions.nextPageUrl === null && !recent)) {
+  async fetchTransactions(context, { limit, recent, address }) {
+    const {
+      state, getters, dispatch, commit,
+    } = context;
+    if (state.transactions.nextPageUrl === null && !recent) {
```
konradodo, line 98:
```suggestion
    if (!!state.transactions.nextPageUrl && !recent) {
```

konradodo (approved):

src/popup/utils/utils.ts
peronczyk: I would prefer to stick to what we have for the security reasons. Explained on private.

src/composables/middleware.ts
peronczyk: Sure, makes 100% sense. Fixed.

PR #1888, opened 2023-03-02
[SW-680] prevent creating vault with account that has no balance

ifaouibadi:


src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -266,10 +274,15 @@ export default defineComponent({
         ),
       ).length,
     );
+    const noEnoughBalanceToCreateMultisig = computed((): boolean => (
+      !!signers.value[0].address
+      && balances.value[signers.value[0].address]?.isLessThan(1)
```
CedrikNikita, line 279: Obviously `isLessThan(1)` is incorrect condition.

src/popup/locales/en.json
```diff
@@ -127,12 +127,13 @@
       "consensusRequiredHelpTitle": "Consensus required",
       "consensusRequiredHelpMsg": "As the creator of this multisig vault you can determine the authorized signers and set the consensus needed for a transaction confirmation (how many signatures are required to confirm each transaction).",
       "createMultisigVault": "Create Vault",
-      "errorDuplicatingSigner": "Cannot add the same address as a signer twice"
+      "errorDuplicatingSigner": "Cannot add the same address as a signer twice",
+      "errorNoEnoughBalanceToCreateVault": "No enough balance to create multisig vault."
```
CedrikNikita, line 131: We can use on of the existing `not enough balance` wordings
https://github.com/aeternity/superhero-wallet/blob/b5dccbc6c706b7be44298cc5d566275fac2c3214/src/popup/locales/en.json#L747
https://github.com/aeternity/superhero-wallet/blob/b5dccbc6c706b7be44298cc5d566275fac2c3214/src/popup/locales/en.json#L1043
https://github.com/aeternity/superhero-wallet/blob/b5dccbc6c706b7be44298cc5d566275fac2c3214/src/popup/locales/en.json#L1414

or at least adjust the message to be "No**t** enough..."

src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -331,11 +345,20 @@ export default defineComponent({
     }
 
     async function openReviewStep() {
-      currentStep.value = STEPS.review;
-      await multisigAccountPrepare(
-        confirmationsRequired.value,
-        signers.value.map((signer) => signer.address),
-      );
+      try {
+        loading.value = true;
```
CedrikNikita, line 349: Would be nice for the user to be able to see loading state.

src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -266,6 +275,11 @@ export default defineComponent({
         ),
       ).length,
     );
+    const noEnoughBalanceToCreateMultisig = computed((): boolean => (
```
peronczyk, line 278: I think it should be `notEnough...` instead of `noEnough...`.

peronczyk (changes requested):

src/popup/components/Modals/MultisigVaultCreate.vue
ifaouibadi: added
```suggestion
        <Loader v-if="loading" />
```
in the template

peronczyk (approved):

CedrikNikita:
In general would be way better to see how much you actually need to be able to create a vault, so i would rather introduce this issue in the review page, also because you can chose a different account to pay.

CedrikNikita (changes requested):
It will not prevent user to change the payed in the next page. And will not explain him that there is not enought balance, on the new payer account.

src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -249,6 +262,7 @@ export default defineComponent({
       deployMultisigAccount,
     } = useMultisigAccountCreate({ store: root.$store });
 
+    const loading = ref<boolean>(false);
```
subhod-i, line 265: this variable is declared but never reassigned.

src/composables/multisigAccountCreate.ts
```diff
@@ -135,6 +138,9 @@ export function useMultisigAccountCreate({ store }: IDefaultComposableOptions) {
     const creationFeeUnformatted = new BigNumber(outerFee).plus(innerFee).toFixed();
     multisigAccountCreationFee.value = Number(AmountFormatter.toAe(creationFeeUnformatted));
     multisigAccountCreationPhase.value = MULTISIG_CREATION_PHASES.signed;
+    notEnoughBalanceToCreateMultisig.value = (
+      balances.value[payerId]?.isLessThan(multisigAccountCreationFee.value)
+    );
```
subhod-i, line 143: This calculation could be done as `computed` in the respective view components.

subhod-i (approved):
Approved with one optimization comment

src/composables/multisigAccountCreate.ts
ifaouibadi: Because in the new update, the payer account is isolated from the current active account and we can't access the payerId in `[MultisigVaultCreate.vue](https://github.com/aeternity/superhero-wallet/pull/1888/files#diff-6342e78347a20aa189ed26026bf0d0cb81995b9cf436ac0fe4a50b47799449a1)` had to perform this calculation here.  also, we are doing the fee calculation in the function.

subhod-i: I see. sounds good!

PR #1890, opened 2023-03-02
Refactor: use topHeader composable data in multisigTransactions

peronczyk:
…tch the top block height

### DIFF
--- a/src/composables/multisigTransactions.ts
+++ b/src/composables/multisigTransactions.ts
@@ -12,6 +12,7 @@ import {
 } from '../popup/utils';
 import { useSdk } from './sdk';
 import { useMultisigAccounts } from './multisigAccounts';
+import { useTopHeaderData } from './topHeader';
 import type {
   IActiveMultisigTx,
   IDefaultComposableOptions,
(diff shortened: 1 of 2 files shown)


pstachel (approved):

src/composables/multisigTransactions.ts
```diff
@@ -12,6 +12,7 @@ import {
 } from '../popup/utils';
 import { useSdk } from './sdk';
 import { useMultisigAccounts } from './multisigAccounts';
+import { useTopHeaderData } from './topHeader';
```
subhod-i, line 15: `topHeader` term is a bit confusing as it has a meaning in the UX space.

src/composables/topHeader.ts
```diff
@@ -17,12 +17,22 @@ export function useTopHeaderData({ store }: IDefaultComposableOptions) {
 
   const topBlockHeight = computed(() => topHeaderData.value?.height || 0);
 
-  initPollingWatcher(async () => {
+  async function updateTopHeaderData() {
     const sdk = await getSdk();
     topHeaderData.value = await sdk.api.getTopHeader();
```
subhod-i, line 22: @davidyuk Shouldn't we use `getCurrentKeyBlock` instead of `getTopHeader`

subhod-i (approved):
LGTM.

ifaouibadi (approved):

src/composables/multisigTransactions.ts
peronczyk: This composable is written around this call `sdk.api.getTopHeader()`. But I'm open for suggestions about better name. Can you suggest one?

subhod-i: Ideally, it could be renamed to `useHeadBlock` or `useTopBlock`. but renaming this depends on https://github.com/aeternity/superhero-wallet/pull/1890#discussion_r1124096202.

Right now it looks fine as it is a wrapper around `sdk.api.getTopHeader()`

PR #1891, opened 2023-03-03
[SW-704] show proper transaction for transaction owner

ifaouibadi:


CedrikNikita (changes requested):
This would lead us to a completly wrong information for transactions that are not related to our accounts.

src/popup/components/TransactionItem.vue
```diff
@@ -136,6 +140,11 @@ export default defineComponent({
           hash: props.transaction.hash,
           transactionOwner: props.transaction.transactionOwner,
         },
+        query: {
+          address: props.isMultisig
```
CedrikNikita, line 144: Shall we maybe rename the query parameter to a `owner` for more clarity? It's just a suggestion.

CedrikNikita (approved):
Code changes look reasonable.

pstachel (approved):

src/popup/pages/TransactionDetails.vue
```diff
@@ -313,7 +313,7 @@ export default defineComponent({
       outerTxType,
     } = useTransactionTx({
       store: root.$store,
-      externalAddress: props.transactionOwner,
+      externalAddress: props.transactionOwner || root.$route?.query?.owner as string,
```
peronczyk, line 316: I think instead of passing this alternative value by query the logic that was set in the `TransactionItem` file can be moved before the `useTransactionTx` in this file.

peronczyk (changes requested):

src/popup/pages/TransactionDetails.vue
CedrikNikita: Without query on refresh we wont be able to know which we have loaded.

peronczyk: Yeah, I get this. But doesn't that create a vulnerability? I can send you a link to my transaction where you will be shown as the owner. So I still think the logic for determining who is the owner should be encapsulated within this page.
Or I'm missing something?

peronczyk (approved):

PR #1899, opened 2023-03-06
[SW-718] fix(multisig): show correct copied text

CedrikNikita:


pstachel (approved):

src/popup/pages/MultisigProposalDetails.vue
```diff
@@ -52,7 +52,7 @@
                 <CopyText
                   hide-icon
                   :value="multisigTx.recipientId"
-                  :copied-text="$t('hashCopied')"
+                  :copied-text="$t('addressCopied')"
```
peronczyk, line 55: I think we should not keep any common translation in the root of the translation file. I suggest to move it to `common.addressCopied`.

peronczyk (approved):

src/popup/pages/MultisigProposalDetails.vue
CedrikNikita: I think we can create a separate task for that.

PR #1904, opened 2023-03-07
Refactor: Common interface for storing data on user device

peronczyk:


konradodo (approved):

davidyuk:
If `localStorage` works in web extensions I would use it everywhere because it works in sync and it would simplify code, in some cases significantly. But probably `localStorage` won't work correctly in scripts injected into page.

peronczyk:
> If localStorage works in web extensions I would use it everywhere because it works in sync and it would simplify code, in some cases significantly. But probably localStorage won't work correctly in scripts injected into page.

@davidyuk It's kind of possible to access localStorage but it is suggested to avoid this :(

image

peronczyk:
This feature was re-introduced within this PR: https://github.com/aeternity/superhero-wallet/pull/2237
It was required to use the browser storage instead of local storage to sync the state between the app and the extension background.

PR #1905, opened 2023-03-08
[SW-445, SW-576] remove default browser input validation on amounts

ifaouibadi:


src/popup/components/InputField.vue
```diff
@@ -59,7 +59,6 @@
             autocomplete="off"
             step="any"
             data-cy="input"
-            :type="type"
```
peronczyk, line 62: Are you sure? The input should have the type property. Here we are using the component's property `type`.

peronczyk, line 62: Are you sure? The input should have the type property. Here we are using the component's property type.

src/popup/components/TransferSendForm.vue
```diff
@@ -85,7 +85,7 @@
     </div>
 
     <InputAmount
-      v-model="formModel.amount"
+      v-model.number="formModel.amount"
```
CedrikNikita, line 88: Even if you want to stick to this approach there are ~10 places where `InputAmount` is used, and where this line should be used in this case.

src/popup/components/InputField.vue
ifaouibadi: Reverted it as `v-model.number` resolves both issues reported on SW-576

src/popup/components/TransferSendForm.vue
ifaouibadi: updated for other components too

src/popup/components/InviteItem.vue
```diff
@@ -38,7 +38,7 @@
     </div>
     <template v-else>
       <InputAmount
-        v-model="topUpAmount"
+        v-model.number="topUpAmount"
```
CedrikNikita, line 41: The number approach will break the long numbers logic we have, and we will lose precision.

CedrikNikita (changes requested):

CedrikNikita (approved):

peronczyk (approved):

PR #1906, opened 2023-03-08
[SW-723] fix: set default name only if it points to current account

CedrikNikita:


konradodo (approved):

src/popup/components/NameItem.vue
```diff
@@ -185,7 +185,10 @@ export default defineComponent({
     const account = useGetter<IAccount>('account');
 
     const nameEntry = computed<IName | null>(() => root.$store.getters['names/get'](props.name));
-    const hasPointer = computed(() => nameEntry.value?.pointers?.accountPubkey);
+    const hasPointer = computed<boolean>(() => !!nameEntry.value?.pointers?.accountPubkey);
```
peronczyk, line 188: When typing computed properties it is better to type the return value of the callback as it will better deal with checking the logic. Now it just informs the parser what would be the returned value.

So instead of
```
computed<Something>(() => ...
```
it's better to
```
```
computed((): Something => ...
```

File: src/popup/components/NameItem.vue [hunk 2]
```diff
@@ -185,7 +185,10 @@ export default defineComponent({
...
+    const isAbleToBeDefault = computed<boolean>(
```
peronczyk, line 189: IDEM

peronczyk (approved):

subhod-i (approved):

src/popup/components/NameItem.vue
```diff
@@ -24,14 +24,14 @@
           class="buttons"
         >
           <button
-            v-show="hasPointer"
+            v-show="isAbleToBeDefault"
```
pstachel, line 27: Maybe instead of `isAbleToBeDefault` you could use `canBeDefault`?

pstachel (approved):

PR #1911, opened 2023-03-09
[SW-725] Select assets menu is not scrollable

ifaouibadi:


src/popup/components/ListItemWrapper.vue
```diff
@@ -28,9 +28,7 @@ export default defineComponent({
 
 .list-item-wrapper {
   display: block;
-  padding: 8px var(--screen-padding-x);
-  margin-left: calc(-1 * var(--screen-padding-x));
-  margin-right: calc(-1 * var(--screen-padding-x));
```
CedrikNikita, line 33: Why do you remove `calc`s here?

CedrikNikita (approved):
In general, code changes look good. In FF browser look fine for me, but i didn't test it heavily.

src/popup/components/ListItemWrapper.vue
ifaouibadi: this was adding -16px margins to left & right
and for padding the `var(--screen-padding-x)`is not working

CedrikNikita: `var(--screen-padding-x)` is heavily used across our project, we shall adjust it everywhere.

PR #1913, opened 2023-03-09
[SW-645] Improve UX regarding 30 sec. delay until vault is active in the wallet

pstachel:


src/types/index.ts
```diff
@@ -178,6 +173,7 @@ export interface IMultisigAccount extends IMultisigConsensus, IMultisigAccountRe
   signers: string[];
   consensusLabel?: string;
   hasPendingTransaction: boolean;
+
```
CedrikNikita, line 176:
```suggestion
```

src/composables/multisigAccountCreate.ts
```diff
@@ -163,17 +182,43 @@ export function useMultisigAccountCreate({ store }: IDefaultComposableOptions) {
...
     }
 
-    // Wait for the account to be discovered by the wallet to allow to open it's details.
-    // TODO in the future we could try to add the new account to the list of the accounts.
+    addPendingMultisigAccount(multisigAccount.value);
+    setActiveMultisigAccountId(multisigAccount.value.gaAccountId);
+
+    const delay = () => new Promise(() => setTimeout(() => {}, 10000));
+
+    await delay();
```
peronczyk, line 220: This looks rather bad. At least add comment that would describe why we need to wait 10s. Also the delay function should be moved to utils in my opinion.

src/composables/multisigAccountCreate.ts
```diff
@@ -163,17 +182,43 @@ export function useMultisigAccountCreate({ store }: IDefaultComposableOptions) {
...
         contractId: gaContract.contractId,
-        multisigAccountId: accountId,
+        balance: new BigNumber(gaContract.balance),
+        gaAccountId: accountId,
+        signers,
+        confirmationsRequired,
+        consensusLabel: prepareConsensusLabel(0, confirmationsRequired, signers.length),
+        nonce: 1, // Default value for freshly created account
+        createdAt: dayjs().toISOString(),
+        updatedAt: dayjs().toISOString(),
```
peronczyk, line 195: The `dayjs()` function is fired twice - I suggest to create const for this before the assignment.

src/composables/multisigAccountCreate.ts
```diff
@@ -1,15 +1,21 @@
 import { computed, ref } from '@vue/composition-api';
 import BigNumber from 'bignumber.js';
 import { AmountFormatter, Crypto, TxBuilder } from '@aeternity/aepp-sdk';
+import dayjs from 'dayjs';
 import type {
   IDefaultComposableOptions,
-  IKeyPair,
-  IMultisigAccountBase,
+  IKeyPair, IMultisigAccount,
```
peronczyk, line 7: Please split this line.

src/composables/multisigAccounts.ts
```diff
@@ -148,10 +182,10 @@ export function useMultisigAccounts({ store }: IDefaultComposableOptions) {
...
             const hasPendingTransaction = !!txHash && !consensus.expired;
 
-            const consensusLabel = (
-              `${hasPendingTransaction ? consensus?.confirmedBy?.length : 0}`
-              + `/${consensus.confirmationsRequired} ${i18n.t('common.of')} `
-              + `${signers.decodedResult?.length}`
+            const consensusLabel = prepareConsensusLabel(
+              hasPendingTransaction ? consensus?.confirmedBy?.length : 0,
+              consensus.confirmationsRequired,
+                signers.decodedResult?.length,
```
peronczyk, line 188: From the beginning this should not be placed here as it saves the outcome of the translation function to the object property. So when user changes the language this label won't change in some places.
So the `prepareConsensusLabel` function should be used directly in the templates.

peronczyk (changes requested):

src/composables/multisigAccountCreate.ts
pstachel: Ooops, forgot to delete it

src/popup/components/AccountCardMultisig.vue
```diff
@@ -32,9 +34,12 @@ import AccountCardConsensus from './AccountCardConsensus.vue';
 import AccountCardBase from './AccountCardBase.vue';
 
 import type { IMultisigAccount } from '../../types';
+import AccountCardSyncing from './AccountCardSyncing.vue';
```
peronczyk, line 37: Please put this at the end.

src/popup/components/MultisigVaultCreateProgress.vue
```diff
@@ -132,6 +145,14 @@ export default defineComponent({
       return index === currentPhaseIndex.value + 1;
     }
 
+    function currentPhaseText(index: number, localPhase: PhaseLabel) {
```
peronczyk, line 148: In my opinion this function is not necessary as this logic would still makes sense on the template side:

```
<div v-if="isPhaseCurrent(index)">
  {{ localPhase.pendingText || localPhase.text }}&hellip;
</div>
<div v-else>
  {{ localPhase.text }}
</div>
```
What do you think?

src/popup/utils/utils.ts
```diff
@@ -443,3 +443,11 @@ export function getAeFee(value: number | string) {
...
   window.open(url, '_blank');
 }
+
+export function prepareConsensusLabel(
+  confirmedBy: number,
+  confirmationsRequired: number,
+  totalConfirmations: number,
+) {
+  return `${confirmedBy}/${confirmationsRequired} ${i18n.t('common.of')} ${totalConfirmations}`;
+}
```
peronczyk, line 453: In my opinion this should be a component.

src/popup/components/AccountCardMultisig.vue
```diff
@@ -15,12 +15,14 @@
       <AccountCardConsensus
         :multisig-account="account"
       />
+      <AccountCardSyncing v-if="isPendingAccount" />
```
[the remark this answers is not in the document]
pstachel: They won't overlap but I think your suggestion will make it obvious - good catch

src/popup/components/MultisigVaultCreateProgress.vue
pstachel: of course, I think that's just a cosmetic change - I prefer less elements in the template

src/composables/multisigAccounts.ts
```diff
@@ -49,8 +50,14 @@ export function useMultisigAccounts({ store }: IDefaultComposableOptions) {
 
   const accounts = computed<IAccount[]>(() => store.getters.accounts);
   const activeNetwork = computed<INetwork>(() => store.getters.activeNetwork);
-  const activeMultisigAccount = computed<IMultisigAccount | undefined>(() => multisigAccounts.value
-    .find((account) => account.gaAccountId === activeMultisigAccountId.value));
+  const allMultisigAccounts = computed<IMultisigAccount[]>(
+    () => [...multisigAccounts.value, ...pendingMultisigAccounts.value],
+  );
+
+  const activeMultisigAccount = computed<IMultisigAccount | undefined>(
```
konradodo, line 57: activeMultisigAccount can be undefined? if not you can change te type and add on end of "find" "!"

src/composables/multisigAccounts.ts
```diff
@@ -60,14 +67,41 @@ export function useMultisigAccounts({ store }: IDefaultComposableOptions) {
...
+    pendingMultisigAccounts.value = pendingMultisigAccounts.value.map((account) => {
+      if (account.gaAccountId === gaAccountId) {
+        return {
+          ...account,
+          txHash,
+          hasPendingTransaction: true,
+        };
+      }
+      return account;
+    });
```
konradodo, line 92:
```suggestion

      return account.gaAccountId === gaAccountId
        ? {
          ...account,
          txHash,
          hasPendingTransaction: true,
        }
       : account
```

src/popup/components/AccountCardMultisig.vue
```diff
@@ -44,8 +49,17 @@ export default defineComponent({
...
   },
-  setup() {
+  setup(props, { root }) {
+    const { pendingMultisigAccounts } = useMultisigAccounts({ store: root.$store });
+
+    const isPendingAccount = computed(
+      () => pendingMultisigAccounts.value.find(
+        ({ gaAccountId }) => gaAccountId === props.account.gaAccountId,
+      ),
+    );
```
konradodo, line 59: isPendingAccount -> this return account, not boolean

src/popup/components/AccountCardSyncing.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+  <div class="account-card-syncing">
+    <PendingIcon class="pending-icon" />
+    <span class="message">{{ $t('modals.creatingMultisigAccount.syncing') }}</span>
```
konradodo, line 4: split line

konradodo (approved):

src/composables/multisigAccounts.ts
pstachel: yes

src/composables/multisigAccounts.ts
pstachel: I'll create a task for it as it already existed in our current code

src/popup/utils/utils.ts
pstachel: I'll create a task for it with the refactor (which I tried to avoid in this PR) of all places where it exists

peronczyk (approved):

PR #1915, opened 2023-03-10
[SW-744] fix: token list margins

ifaouibadi:


src/popup/pages/AccountDetailsMultisigTokens.vue
```diff
@@ -35,6 +35,8 @@ export default defineComponent({
 .multisig-token-list-wrapper {
   position: relative;
   padding-top: 4px;
+  margin-left: calc(-1 * var(--screen-padding-x));
+  margin-right: calc(-1 * var(--screen-padding-x));
```
peronczyk, line 39: This fixes visually the issues, but I feel this needs further changes as in my opinion.

Mostly because previously one element (ListItemWrapper) had both the negative margins and the positive paddings. So the logic was encapsulated in one place. And both were dynamic.

The current fix is actually bringing the negative margins back, which you mentioned was breaking few things, but the paddings are now set to fixed amount of 12px (inside of the `ListItemWrapper`). So they will not align with pages that has different viewport padding than 12px.

src/popup/pages/AccountDetailsTokens.vue
```diff
@@ -34,6 +34,8 @@ export default defineComponent({
 .token-list-wrapper {
   position: relative;
   padding-top: 4px;
+  margin-left: calc(-1 * var(--screen-padding-x));
+  margin-right: calc(-1 * var(--screen-padding-x));
```
peronczyk, line 38: IDEM

peronczyk (changes requested):

src/popup/pages/AccountDetailsMultisigTokens.vue
ifaouibadi: Previously negative margins were in `ListItemWrapper`and it was hard to control as it's been imported in so many places, impacting all the screens that use it and forcing you to add extra padding-x to resolve the padding.

now we only add it to the main screens that needs full width display

peronczyk: > it's been imported in so many places impacting all the screens that use it
Only 2 places:
image

> forcing you to add extra padding-x to resolve the padding
not forcing, but allowing to use screen's padding variable, which is set for all the screens we use. So no matter how big the screen padding is, the list items would fit. And now this approach is "forcing" devs to reset the side padding by using negative margins on the wrapping element whenever the ListItemWrapper is used. So what actually is the difference?

> now we only add it to the main screens that needs full width display
Yeah, now you need to think about resetting the paddings by using negative margins **each** time you use the ListItemWrapper. And also the `ListItemWrapper` has fixed inline padding equal to `12px` so there is no way of using it on screens with different paddings - the content of the list item won't align other elements on the screen.

src/popup/pages/AccountDetailsTokens.vue
```diff
@@ -34,12 +34,12 @@ export default defineComponent({
...
   position: relative;
   padding-top: 4px;
+  margin-inline: calc(-1 * var(--screen-padding-x));
 }
 
 .search-bar-wrapper {
   background: var(--screen-bg-color);
-  margin-left: calc(-1 * var(--screen-padding-x));
-  margin-right: calc(-1 * var(--screen-padding-x));
+  margin-inline: calc(-1 * var(--screen-padding-x));
```
peronczyk, line 42: Is this `.search-bar-wrapper` even usable? I mean it's in the scoped styles block and this class does not exists in the current scope. So I think it could be removed.

peronczyk, line 42: Is this .search-bar-wrapper even usable? I mean it's in the scoped styles block and this class does not exists in the current scope and event in the app. So I think it could be removed.
Also can be removed from the AccountDetailsMultisigTokens.

peronczyk (changes requested):

Liubov-crypto:
lgtm

peronczyk (approved):

PR #1918, opened 2023-03-14
[SW-719] firefox & safari input validation

ifaouibadi:
Needed to remove the native browser input validation cause it's causing issues for Safari & Firefox

we already have custom logic for validating number withing `<InputField` `checkIfNumber` so this removal is safe

### DIFF
--- a/src/popup/components/InputField.vue
+++ b/src/popup/components/InputField.vue
@@ -59,7 +59,6 @@
             autocomplete="off"
             step="any"
             data-cy="input"
-            :type="type"
             :value="value"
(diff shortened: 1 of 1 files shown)


pstachel (approved):

src/popup/components/InputField.vue
```diff
@@ -59,7 +59,6 @@
             autocomplete="off"
             step="any"
             data-cy="input"
-            :type="type"
```
peronczyk, line 62: I suggest to set the type to `text` instead of removing this attribute.

PR #1925, opened 2023-03-15
[SW-761] Fix unnecessary fetching of latest transactions

peronczyk:


src/popup/components/LatestTransactionsCard.vue
```diff
@@ -113,21 +112,28 @@ export default defineComponent({
...
-    });
+    }
+
+    // To avoid unnecessary data transfers instead of constant polling
+    // we are fetching the transactions only if the total balance of the accounts changes.
+    watch(
+      balancesTotal,
+      (val, oldVal) => {
+        if (!latestTransactions.value.length || val !== oldVal) {
+          setTimeout(() => updateData(), 100);
```
CedrikNikita, line 125: Transaction appearing on the middleware slower than node updates the balance. So the latest transaction will not be fetched in most of cases.

src/popup/components/LatestTransactionsCard.vue
```diff
@@ -113,21 +112,28 @@ export default defineComponent({
...
-    watch(balances, updateData);
 
-    onMounted(async () => {
-      await updateData();
       isLoading.value = false;
-    });
+    }
+
+    // To avoid unnecessary data transfers instead of constant polling
+    // we are fetching the transactions only if the total balance of the accounts changes.
```
CedrikNikita, line 120: We need also to watch changes of token balances for every account, in order to cover also fungible token transactions.

CedrikNikita (changes requested):

src/popup/components/LatestTransactionsCard.vue
CedrikNikita: #1928

pstachel (approved):

src/popup/components/LatestTransactionsCard.vue
peronczyk: Your PR is fixing this: https://github.com/aeternity/superhero-wallet/pull/1928/files
We'll merge both of them in order to fix the whole problem.

src/popup/components/LatestTransactionsCard.vue
peronczyk: As discussed on priv I increased the delay. In the future we need to replace this with websockets.

CedrikNikita (approved):

PR #1931, opened 2023-03-16
[SW-752] Initialize sdk13 instance and use it for sending AE

subhod-i:


src/lib/wallet.js
```diff
@@ -39,7 +39,9 @@ export default async function initSdk() {
   store.commit('setNodeStatus', NODE_STATUS_CONNECTING);
   try {
     await store.dispatch('sdkPlugin/initialize');
+    await store.dispatch('sdk13Plugin/initialize');
     await watchUntilTruthy(() => store.getters['sdkPlugin/sdk']);
+    await watchUntilTruthy(() => store.getters['sdk13Plugin/sdk']);
```
pstachel, line 44: Maybe Promise.all (or two) should be considered here?

pstachel, line 44: Maybe Promise.all (or two) should be considered here?

src/store/plugins/sdk13.js
pstachel, line 1: I think we decided to divert from using the store in favour of composables?

src/store/plugins/sdk13.js
```diff
@@ -0,0 +1,205 @@
...
+      sdk: ({ ready }) => (ready ? sdk : undefined),
+    },
+    mutations: {
+      setSdkReady(state) {
+        state.ready = true;
+      },
+    },
+    actions: {
+      async initialize({ commit }) {
+        if (sdk) return;
```
pstachel, line 36:
```suggestion
        if (sdk) {
            return;
        }
```

src/store/plugins/sdk13.js
```diff
@@ -0,0 +1,205 @@
...
+        const signCb = async (payload, options, origin = null) => {
+          try {
+            const host = origin ? new URL(origin).host : null;
+            const permission = (!host && IS_CORDOVA)
+              ? true
+              : await store.dispatch('permissions/checkPermissions', {
+                host,
+                method: 'sign',
+                params: payload,
+              });
```
pstachel, line 54: I think you could simplify it to
```suggestion
            const permission = (!host && IS_CORDOVA) || await store.dispatch('permissions/checkPermissions', {
                host,
                method: 'sign',
                params: payload,
              });
```

src/store/plugins/sdk13.js
```diff
@@ -0,0 +1,205 @@
...
+          const aepp = aeppInfo[options.aeppRpcClientId];
+          if (
+            (await store.dispatch('permissions/checkPermissions', {
+              host: new URL(aepp.origin).host,
+              method,
+              params: options,
+            }))
+            || (await showPopup(
+              aepp.origin,
+              method === 'transaction.sign' ? 'sign' : 'messageSign',
```
pstachel, line 79: You could use POPUP_TYPES constant here

src/store/plugins/sdk13.js
```diff
@@ -0,0 +1,205 @@
...
+              { ...options, ...(method === 'transaction.sign') ? { tx: payload, txObject: unpackTx(payload) } : { message: payload } },
```
pstachel, line 80: the comparison `method === 'transaction.sign'` occurs twice here - maybe it could be moved  to a variable?

src/store/plugins/sdk13.js
```diff
@@ -0,0 +1,205 @@
...
+        };
+
+        const signCbBackground = async (method, payload, options) => {
+          const aepp = aeppInfo[options.aeppRpcClientId];
+          if (
+            (await store.dispatch('permissions/checkPermissions', {
+              host: new URL(aepp.origin).host,
+              method,
+              params: options,
+            }))
```
pstachel, line 76: please adjust the formatting here
```suggestion
          if (
            (await store.dispatch('permissions/checkPermissions', {
              host: new URL(aepp.origin).host,
              method,
              params: options,
            })
          )
```

src/store/plugins/sdk13.js
```diff
@@ -0,0 +1,205 @@
...
+            || (await showPopup(
+              aepp.origin,
+              method === 'transaction.sign' ? 'sign' : 'messageSign',
+              { ...options, ...(method === 'transaction.sign') ? { tx: payload, txObject: unpackTx(payload) } : { message: payload } },
+            ).then(
+              () => true,
+              () => false,
+            ))
+          ) {
```
pstachel, line 85: Maybe Promise.some could speed it up a bit?

src/store/plugins/sdk13.js
```diff
@@ -0,0 +1,205 @@
...
+        });
+        commit('setSdkReady');
+      },
+    },
+  });
+
+  store.watch(
+    (state, getters) => getters.activeNetwork,
+    async (network, oldNetwork) => {
+      if (isEqual(network, oldNetwork)) return;
```
pstachel, line 180:
```suggestion
      if (isEqual(network, oldNetwork)) {
          return;
      }
```

src/store/plugins/sdk13.js
subhod-i: Yeah, But that would again increate the PR size. Too many refactors!

src/store/plugins/sdk13.js
subhod-i: this is good logic. We check the permissions database first to see if no user confirmation is needed. If needed then we call the `showPopup` method to again manually get the confirmations from user.

src/store/plugins/sdk13.js
```diff
@@ -0,0 +1,223 @@
...
+        }
+        await watchUntilTruthy(
+          () => store.state.isRestored && !isEmpty(store.getters.account),
+        );
+
+        const { activeNetwork } = store.getters;
+
+        // TODO: use this method for `onMessageSign` replacement.
+        // eslint-disable-next-line no-unused-vars
+        const signCb = async (payload, options, origin = null) => {
```
peronczyk, line 52: This function is not used, but you commented that it should be.

src/store/plugins/sdk13.js
```diff
@@ -0,0 +1,223 @@
...
+                ...options,
+                modal: false,
+                host: aepp.origin,
+              },
+            });
+          }
+          throw new RpcRejectedByUserError('Rejected by user');
+        };
+
+        // TODO: move this to a seperate file with typescript and combine the custom type
```
peronczyk, line 117: Why not just do it now? :)

peronczyk (changes requested):

src/store/plugins/sdk13.js
peronczyk: But this class is new code - it's not refactor.

peronczyk: My comment was referring only to the `SuperheroSdk` class, so it's around of 20 lines. I'm not forcing you to rewrite whole file.
The reason I'd like this code to be moved into separate TS file is that after checking it TS is complaining about this piece of code. And I really don't want us to add new problematic code into repository. There is already too much of "temporary hacks".

peronczyk: image

src/store/plugins/sdk13.js
subhod-i: Using it now

src/types/@aeternity/aepp-sdk.d.ts
```diff
@@ -0,0 +1,20 @@
...
+
+declare module '@aeternity/aepp-sdk-13' {
+  interface AeSdkWallet {
+    async spend(
+      amount: Parameters<typeof spend>[0],
+      recipient: Parameters<typeof spend>[1],
+      opt: ISpendOptions,
+    ) : ReturnType<spend>;
+  }
+}
```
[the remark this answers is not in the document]
subhod-i: Added.

src/store/plugins/sdk13.js
subhod-i: Define appropriate files.

src/popup/utils/utils.ts
```diff
@@ -491,3 +495,53 @@ export function calculateFontSize(amountValue: BigNumber | number) {
   }
   return '12px';
 }
+
+export function connectFrames(sdk: ISdk | AeSdkWallet) {
```
subhod-i, line 499: Helper method borrowed from https://github.com/aeternity/superhero-wallet/blob/3ccdb5241b29f808581071bea3d4ecb9b7c3b232/src/popup/utils/utils.ts#L464

PR: https://github.com/aeternity/superhero-wallet/pull/1731

src/composables/sdk13.ts
subhod-i, line 1: Composable borrowed from @peronczyk PR: https://github.com/aeternity/superhero-wallet/pull/1731

src/store/plugins/sdk13FrameReset.js
subhod-i, line 1: Borrowed from @peronczyk PR: https://github.com/aeternity/superhero-wallet/pull/1731

src/lib/accounts/index.ts
```diff
@@ -0,0 +1,122 @@
...
+import { Store } from 'vuex';
+import { useAccounts } from '../../composables/accounts';
+import { IS_CORDOVA, IS_EXTENSION_BACKGROUND } from '../environment';
+import { POPUP_TYPE_MESSAGE_SIGN, POPUP_TYPE_SIGN, POPUP_TYPE_TX_SIGN } from '../../popup/utils';
+import { showPopup } from '../../background/popupHandler';
+
+export class AccountSuperhero extends AccountBase {
+  address: Encoded.AccountAddress;
+
+  store:Store<any>;
```
pstachel, line 18:
```suggestion
  store: Store<any>;
```

src/lib/accounts/index.ts
```diff
@@ -0,0 +1,122 @@
...
+import { Encoded } from '@aeternity/aepp-sdk-13/src/utils/encoder';
+import { Store } from 'vuex';
+import { useAccounts } from '../../composables/accounts';
+import { IS_CORDOVA, IS_EXTENSION_BACKGROUND } from '../environment';
+import { POPUP_TYPE_MESSAGE_SIGN, POPUP_TYPE_SIGN, POPUP_TYPE_TX_SIGN } from '../../popup/utils';
+import { showPopup } from '../../background/popupHandler';
+
+export class AccountSuperhero extends AccountBase {
+  address: Encoded.AccountAddress;
+
```
pstachel, line 17:
```suggestion
```

src/lib/accounts/index.ts
```diff
@@ -0,0 +1,122 @@
...
+      opt: {
+        ...opt,
+        modal: !permission,
+        host,
+      },
+    });
+  } catch (error) {
+    throw new RpcRejectedByUserError(error.message);
+  }
+};
```
pstachel, line 77: Aren't you missing a two spaces at the beginning here?

src/composables/sdk13.ts
```diff
@@ -0,0 +1,176 @@
...
+                name: MODAL_CONFIRM_CONNECT,
+                app: {
+                  name: app.host.hostname,
+                  icons: [],
+                  protocol: app.host.protocol,
+                  host: app.host.host,
+                },
+              })),
+          }))
+        ) return Promise.reject(new RpcRejectedByUserError('Rejected by user'));
```
pstachel, line 115: I'd suggest adding braces here as the return may blend in with if's conditions
```suggestion
        ) {
          return Promise.reject(new RpcRejectedByUserError('Rejected by user'));
        }
```

src/lib/accounts/index.ts
subhod-i: Good catch.

pstachel (approved):

src/lib/accounts/index.ts
```diff
@@ -0,0 +1,122 @@
...
+  aeppInfo: Record<string, any> = {};
+
+  constructor(store: Store<any>) {
+    super();
+    this.store = store;
+    const { account } = useAccounts({ store: this.store });
+    this.address = account.value.address as Encoded.AccountAddress;
+  }
+
+  signTransaction = (txBase64: Encoded.Transaction, opt: any): Promise<Encoded.Transaction> => {
```
peronczyk, line 29: I suggest to define class methods as regular function (as described in the documentation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#class_body). In general it's more common.

src/lib/accounts/index.ts
```diff
@@ -0,0 +1,122 @@
...
+      return this.bgPermissionCheckAndSign(
+        POPUP_TYPE_TX_SIGN,
+        txBase64,
+        { ...opt, origin: opt.aeppOrigin },
+      );
+    }
+    return this.store.dispatch('accounts/signTransaction', { txBase64, opt: { ...opt, onAccount: this.address } });
+  }
+
+  signMessage = (message: string, opt: any): Promise<Uint8Array> => {
```
peronczyk, line 43: IDEM

src/lib/accounts/index.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (IS_CORDOVA) {
+      return this.fgPermissionCheckAndSign(messageToHash(message), opt, opt.aeppOrigin);
+    }
+    if (IS_EXTENSION_BACKGROUND) {
+      return this.bgPermissionCheckAndSign('message.sign', message, { ...opt, origin: opt.aeppOrigin });
+    }
+    return this.sign(messageToHash(message), opt);
+  }
+
+  sign = (data: string | Uint8Array, opt: any): Promise<Uint8Array> => IS_EXTENSION_BACKGROUND
```
peronczyk, line 53: IDEM

src/lib/accounts/index.ts
```diff
@@ -0,0 +1,122 @@
...
+    ? sign(data, this.store.getters.account.secretKey) as any
+    : this.store.dispatch('accounts/sign', data, opt)
+
+  fgPermissionCheckAndSign = async (payload: any, opt: any, origin: any = null) => {
```
peronczyk, line 57: IDEM

src/lib/accounts/index.ts
```diff
@@ -0,0 +1,122 @@
...
+          modal: !permission,
+          host,
+        },
+      });
+    } catch (error) {
+      throw new RpcRejectedByUserError(error.message);
+    }
+  };
+
+  bgPermissionCheckAndSign = async (
```
peronczyk, line 79: IDEM

src/lib/accounts/index.ts
peronczyk, line 1: I strongly suggest to name the file the same way as the exported class.

src/lib/accounts/index.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  store: Store<any>;
+
+  aeppInfo: Record<string, any> = {};
+
+  constructor(store: Store<any>) {
+    super();
+    this.store = store;
+    const { account } = useAccounts({ store: this.store });
+    this.address = account.value.address as Encoded.AccountAddress;
```
peronczyk, line 26: Maybe it would be better to store whole `account` in the scope of the class to allow using it instead of `store.getters.account` ? This would help to migrate from Vuex.

src/lib/accounts/index.ts
```diff
@@ -0,0 +1,122 @@
...
+      return this.fgPermissionCheckAndSign(messageToHash(message), opt, opt.aeppOrigin);
+    }
+    if (IS_EXTENSION_BACKGROUND) {
+      return this.bgPermissionCheckAndSign('message.sign', message, { ...opt, origin: opt.aeppOrigin });
+    }
+    return this.sign(messageToHash(message), opt);
+  }
+
+  sign = (data: string | Uint8Array, opt: any): Promise<Uint8Array> => IS_EXTENSION_BACKGROUND
+    ? sign(data, this.store.getters.account.secretKey) as any
```
peronczyk, line 54: I suggest to pick the account from the composable instead of the Vuex.

peronczyk (changes requested):

peronczyk (changes requested):

src/store/plugins/sdk13FrameReset.js
peronczyk, line 1: I suggest to move this logic from Vuex modules.

src/types/index.ts
peronczyk, line 1: The `ISdk` interface is a legacy code for the old SDK. We should not use it for the SDK 13 because the actual types  comes from the `AeSdkWallet` class.

src/composables/sdk13.ts
```diff
@@ -0,0 +1,178 @@
...
+import {
+  MODAL_CONFIRM_CONNECT,
+  POPUP_TYPE_CONNECT,
+  connectFrames,
+  watchUntilTruthy,
+} from '../popup/utils';
+import { showPopup } from '../background/popupHandler';
+import { useAccounts } from './accounts';
+
+let sdk: AeSdkWallet;
```
peronczyk, line 28: Instead of `AeSdkWallet` it should be `ShSdkWallet`.

src/types/@aeternity/aepp-sdk.d.ts
```diff
@@ -0,0 +1,25 @@
...
+
+declare module '@aeternity/aepp-sdk-13' {
+  class AeSdkWallet {
+    async spend(
+      amount: Parameters<typeof spend>[0],
+      recipient: Parameters<typeof spend>[1],
+      opt: ISpendOptions,
+    ) : ReturnType<spend>;
+  }
+}
```
peronczyk, line 25: I guess if you replace the type for the `sdk` in the `sdk13.ts` composable from `AeSdkWallet` to `ShSdkWallet` this code wouldn't be needed anymore.

peronczyk (changes requested):

src/lib/accounts/index.ts
subhod-i: since the account is used for only one function at the `sign` message. I am using the accounts composables at the sign method itself and not at the class level.

src/types/index.ts
subhod-i: Got it. Not using ISdk type for SDK13 instances

src/store/plugins/sdk13FrameReset.js
subhod-i: @peronczyk This is from your PR https://github.com/aeternity/superhero-wallet/pull/1731/files#diff-57b34ec02ec268601a1b59610d1efab9aab3df62fb7bef9fa08a80713f9ff7c0

Do you want me to move  this file to libs?

src/lib/accounts/index.ts
peronczyk: OK

peronczyk (approved):

Liubov-crypto:
LGTM

PR #1940, opened 2023-03-21
[SW 747✈️] Use SDK13 in extension and Aepp wallet communication

subhod-i:


src/store/modules/accounts/hdWallet.js
```diff
@@ -94,31 +94,31 @@ export default {
       txBase64,
       opt: { modal = true, host = null },
     }) {
-      const sdk = rootGetters['sdkPlugin/sdk'];
+      const networkId = await (rootGetters['sdkPlugin/sdk'] ?? rootGetters['sdk13Plugin/sdk'].api).getNetworkId();
```
pstachel, line 97: This code is repeated [here](https://github.com/aeternity/superhero-wallet/pull/1940/files#diff-37085670e4cc4d0853535bb598e3c36b31ffb49af1badbebe2018c1591d51b9cR112) - maybe it could be extracted to a function or stored above?

peronczyk (changes requested):

src/store/modules/accounts/hdWallet.js
peronczyk, line 97: My draft PR actually removes the SDK Vuex module so I strongly suggest to use the composable. Please see how it's done here: https://github.com/aeternity/superhero-wallet/pull/1731/files#diff-37085670e4cc4d0853535bb598e3c36b31ffb49af1badbebe2018c1591d51b9c

On the other hand side in one of my recent PRs Nikita suggested to not use the `sdk.getNetworkId()` but the getter `activeNetwork.networkId` which is more safe for us.

src/store/modules/accounts/hdWallet.js
```diff
@@ -94,31 +94,31 @@ export default {
...
       );
       return TxBuilder.buildTx({ encodedTx, signatures: [signature] }, SCHEMA.TX_TYPE.signed).tx;
     },
     async signTransactionFromAccount({ dispatch, rootGetters }, {
       txBase64,
-      opt: { modal = true, host = null, fromAccount },
+      opt: { modal = true, host = null, onAccount },
     }) {
-      const sdk = rootGetters['sdkPlugin/sdk'];
+      const networkId = await (rootGetters['sdkPlugin/sdk'] ?? rootGetters['sdk13Plugin/sdk'].api).getNetworkId();
```
peronczyk, line 112: IDEM

peronczyk (changes requested):

src/store/plugins/sdk13.js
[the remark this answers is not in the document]
subhod-i: I am avoiding heavy refactoring in SDK PRs. This could be a separate task and a PR accordingly.

src/store/modules/accounts/hdWallet.js
subhod-i: I am using getters for choosing network Id, instead of `sdk.getNetworkId()`.
As I said, I am reducing refactoring changes with the SDK task as that is not the motive of the task.

src/store/plugins/sdk13.js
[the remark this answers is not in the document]
subhod-i: For this new file, it makes sense to write a composable. Thanks

Liubov-crypto:
@subhod-i  If I choose  Don't ask for login requests and Auto-sign messages in superhero wallet permissions there will be no auto-sign of messages, I will receive popups to sign the message. 

https://user-images.githubusercontent.com/69896204/228876073-fe98f3c8-5dbb-41ad-a093-235f67c5cf54.mov

the same for DEX: 

https://user-images.githubusercontent.com/69896204/228878305-3f8f15b0-ce65-4478-a28a-4de72422e499.mov

subhod-i:
> @subhod-i If I choose Don't ask for login requests and Auto-sign messages in superhero wallet permissions there will be no auto-sign of messages, I will receive popups to sign the message.
> 
> the same for DEX:

Okay. Let me check

Liubov-crypto:
@subhod-i fixed, LGTM

src/store/plugins/sdk13.js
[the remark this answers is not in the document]
subhod-i: Moved to composabels already.

src/store/modules/accounts/hdWallet.js
subhod-i: Using `activeNetwork` getter

src/lib/wallet.js
```diff
@@ -1,27 +1,30 @@
...
 let initSdkRunning = false;
 
 if (IN_FRAME) {
   store.registerModule('sdk-frame-reset', {
     actions: {
-      async reset({ rootGetters }) {
-        Object.values(rootGetters['sdkPlugin/sdk'].rpcClients).forEach((aepp) => {
+      async reset() {
+        const { getSdk } = useSdk13({ store });
+        const sdk = await getSdk();
```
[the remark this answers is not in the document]
subhod-i: Fixed

src/store/modules/permissions.js
```diff
@@ -78,7 +78,7 @@ export default {
       const permissionsMethods = {
         'connection.open': 'address',
         'address.subscribe': 'address',
-        'message.sign': 'messageSign',
+        messageSign: 'messageSign',
```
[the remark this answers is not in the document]
subhod-i: I am using the existing POPUP_CONSTs

src/store/plugins/sdk.js
```diff
@@ -33,144 +31,28 @@ export default (store) => {
...
+                : store.dispatch('accounts/signTransaction', { txBase64, opt }),
             }),
           },
         })({
           nodes: [{ name: activeNetwork.name, instance: await Node({ url: activeNetwork.url }) }],
           compilerUrl: activeNetwork.compilerUrl,
           name: 'Superhero',
-          onConnection: cbAccept,
-          onDisconnect(_, client) {
-            client.disconnect();
```
CedrikNikita, line 147: This is not longer being propagated.

Liubov-crypto:
It seems that permission settings are not working for DEX and for superhero.com: 
https://github.com/aeternity/superhero-wallet/assets/69896204/8a930db8-7083-45a2-9294-c7c5a94537f4

The limits that I set do not work.

src/store/plugins/sdk.js
subhod-i: With the new SDK version, there is no need to propagate disconnect. @davidyuk

CedrikNikita: Well, the `superhero-ui` now doesn't reload the state if the wallet is resetting.

subhod-i: Good catch. Let me check.

Liubov-crypto:
> It seems that permission settings are not working for DEX and for superhero.com: https://github.com/aeternity/superhero-wallet/assets/69896204/8a930db8-7083-45a2-9294-c7c5a94537f4
> 
> The limits that I set do not work.

Retested, the same behaviour on dev branch.

src/composables/sdk13.ts
```diff
@@ -164,6 +153,31 @@ export function useSdk13({ store }: IDefaultComposableOptions) {
...
 
+  watch(
+    activeAccount,
+    (oldVal, newVal) => {
+      if (!isEqual(oldVal, newVal) && sdk) {
+        sdk._pushAccountsToApps();
+      }
+    },
+    { deep: true },
+  );
```
peronczyk, line 164: I don't think it's a good pattern that we watch the accounts and perform actions as this composable is not always loaded, so the watcher can miss some updates.

src/composables/sdk13.ts
```diff
@@ -164,6 +153,31 @@ export function useSdk13({ store }: IDefaultComposableOptions) {
...
+        sdkBlocked = true;
+        sdk.pool.delete(sdkCurrentNetwork.name);
+        const nodeInstance = await createNodeInstance(activeNetwork.value.url);
+        sdk.addNode(activeNetwork.value.name, nodeInstance!, true);
+        sdkCurrentNetwork = activeNetwork.value;
+        sdkBlocked = false;
+      }
+    },
+    { deep: true },
+  );
```
peronczyk, line 179: I consider using this code as a step back from what we have and than we refactored.
1. As this is a composable we can't be sure if it's loaded at the moment of changing the network. So this watcher may not run.
2. We should not watch the network - we need to reload the SDK only if the app needs to use the SDK and it detects that the network has changed.

src/background/wallet.js
```diff
@@ -52,22 +53,22 @@ export async function init() {
...
   connectionsQueue.forEach(addAeppConnection);
   connectionsQueue = [];
 }
 
-export function disconnect() {
-  Object.values(store.getters['sdkPlugin/sdk'].rpcClients).forEach((aepp) => {
+export async function disconnect() {
+  const { getSdk } = useSdk13({ store });
+  const sdk = await getSdk();
+  Object.values(sdk._clients).forEach((aepp) => {
```
peronczyk, line 71: Properties starting from an underscore in many cases are considered private ones. Maybe the SDK has some kind of getter for this value?
@davidyuk FYI

src/content-scripts/inject.js
peronczyk, line 1: Maybe that's a good time to refactor this file into TypeScript?

src/store/modules/permissions.js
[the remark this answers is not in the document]
peronczyk: I agree with Denis that this looks weird that other keys has a dot a this one not.

peronczyk (changes requested):

src/background/wallet.js
peronczyk, line 71: Properties starting from an underscore in many cases are considered private ones. Maybe the SDK has some kind of getter for this value?
@davidyuk FYI

src/store/plugins/sdk.js
subhod-i: @CedrikNikita Fixed it.

davidyuk: > there is no need to propagate disconnect

This part should be working the same as in sdk@12

src/composables/sdk13.ts
peronczyk: As we discussed my suggestion is to bring back the `resetSdkNode` method and temporarily use it within the scope of the store when running the  `switchNetwork` action. After migration from the Vuex we would run it within the scope of the networks composable.

src/composables/sdk13.ts
peronczyk: Regarding our discussion I think we can replace it with some kind of `onActiveAccountChange` hook returned by the accounts composable.

src/composables/sdk13.ts
subhod-i: `resetSdkNode` would create on-demand SDK instances and thus will not notify the connected aepps. Plus when you re-instantiate the SDK then we would lose all connected aepps.

src/composables/sdk13.ts
subhod-i: We have parked this optimization for better research.

src/background/wallet.js
subhod-i: Looking forward to the aepp-wallet connection refactoring.

subhod-i: As discussed, SDK will undergo aepp<->wallet refactoring, once that is done, private method accessors are replaced accordingly.

Liubov-crypto:
Got some issues in console when tested dex: 
err1
err2
netw
err sdk
cons

I set on "auto sign messages" but seems it's not working on dex: 

https://github.com/aeternity/superhero-wallet/assets/69896204/ab7d583b-8318-47f4-b9d0-4b7498db558b

subhod-i:
> Got some issues in console when tested dex: err1 netw err sdk cons

All of these issues are present in the master version too. These are some validations pending from DEX side. No new error is introduced in this PR

davidyuk (approved):

src/lib/wallet.js
```diff
@@ -8,18 +8,21 @@ import { IN_FRAME } from './environment';
...
+
+        Object.values(sdk._clients).forEach((aepp) => {
           if (aepp.info.status && aepp.info.status !== 'DISCONNECTED') {
             aepp.sendMessage(
-              { method: 'connection.close', params: { reason: 'bye' }, jsonrpc: '2.0' },
+              { method: METHODS.closeConnection, params: { reason: 'bye' }, jsonrpc: '2.0' },
```
peronczyk, line 25: I don't see any import of the `METHODS` constant in this file.

peronczyk (changes requested):

CedrikNikita (approved):
Code changes look reasonable, didn't test it.

Liubov-crypto:
During my testing the wallet was very slow to update info properly (Txs list, latest Tx widget, etc.), maybe it's related to middleware. 

1. Transactions list wasn't updated properly in mainnet. I made a couple of Txs, but it took approx 1 hr to see them in the tx list. 

2. Got this issues on Dex, but maybe this is related to Dex middleware issues: 

sw
dr
netw

3. If I will use 2nd or 3rd acc for multisig creation it will fall: 
2nd

4. Set pointer name isn't working in mainnet. I tried 2 times, Tx was successful, but I wasn't able make default name after (make default button was not available). 

5. Got this issue in console, after click on Tx details in multisig, wallet stop to response: 
d

Liubov-crypto:
@subhod-i  I have the same issue on dev branch:  click on Tx details in multisig, wallet stop to response.

Liubov-crypto:
@subhod-i  I retested and now I can set pointer to chain name. Please ignore issue 3. It works today. 

1. Transactions list wasn't updated properly in mainnet. I made a couple of Txs, but it took approx 1 hr to see them in the tx list. - bad updating on dev branch too. 

4. I tried to register chain name in the mainnet 2 times, it fails. Second time the money was taken but name is absent in my wallet. - This issue is also reproducible on the dev branch. 
chr

About Dex 2nd issue: 
Some similar error  is in console but everything was working fine, maybe it's worth to recheck it:
dev

Liubov-crypto:
I've rechecked issue with registering chain name, now it's working for me here and on dev branch as well.

CharisSiarampalis:
I do not have an issue registering a chain name.
Other than that the other issues seems reproduceable on dev

subhod-i:
Fixed the duplicate node error. Rest of the issues are reproducible on develop too.

peronczyk (approved):

CharisSiarampalis:
@subhod-i Other than the issues reproducible on dev Lgtm in general

PR #1946, opened 2023-03-24
[SW-759] aens button sized on ios

ifaouibadi:


CedrikNikita (commented):
You just decreased the padding for "mobile". I guess it can be a solution if it fits most of the phones (or smallest).

pstachel (approved):

CedrikNikita (approved):
If it's fixing the problem then fine.

PR #1947, opened 2023-03-24
[SW-776] Assets are not visible on multisig vaults account dashboard when switching between networks

pstachel:


peronczyk (approved):

CedrikNikita (approved):
Code changes look reasonable.

PR #1949, opened 2023-03-24
[SW-778] Animation on the names is broken

konradodo:


pstachel (approved):

src/popup/components/Truncate.vue
```diff
@@ -125,23 +122,18 @@ export default defineComponent({
...
       border-radius: 2px;
+      margin-left: -2px;
       mask-image:
         linear-gradient(
           90deg,
           rgba(black, 0) 0%,
-          rgba(black, 1) 5%,
-          rgba(black, 1) 95%,
+          rgba(black, 1) 2px,
+          rgba(black, 1) calc(100% - 2px),
```
peronczyk, line 136: Here we are defining the size of the gradient that covers the beginning and the end of the string. In my opinion 2px is too small.

peronczyk (dismissed):

PR #1952, opened 2023-03-27
[SW-795] Refactor: Move helper functions to utils

peronczyk:


src/background/utils.js
```diff
@@ -27,6 +45,30 @@ export const getTippingContractInstance = async (tx) => {
...
+    if (e.message.indexOf('wrong_abi_version') > -1) {
+      instance.setOptions({ backend: 'aevm' });
+      return contractCall({
+        instance, method, params, decode, async,
+      });
+    }
+    throw e.message;
+  }
+
+  if (async) return decode ? call.decodedResult : call;
```
pstachel, line 68:
```suggestion
  if (async) {
    return decode ? call.decodedResult : call;
  }
```

pstachel (approved):

src/background/utils.js
peronczyk, line 1: The new methods here was just moved from the `helper.js` file. No changes was made.

src/types/index.ts
```diff
@@ -31,6 +31,13 @@ export type ObjectValues<T> = T[keyof T];
  */
 type PublicPart<T> = {[K in keyof T]: T[K]};
 
+/**
+ * Allowed options that can be passed to our fetch utility functions
+ */
+export interface IRequestInitBodyParsed extends Omit<RequestInit, 'body'> {
+  body?: object | null;
```
subhod-i, line 38: Do we need `undefined` and `null` both here?

peronczyk: This is overriding native browser interface. But you have right: for our conveniency we can remove the `null`.
Done

src/background/utils.js
```diff
@@ -1,13 +1,34 @@
...
 } from '../popup/utils';
 import store from './store';
 
+const BACKEND_AE_VM = 'aevm';
+const BACKEND_FATE = 'fate';
+
 let tippingContract;
 
+export const setContractInstance = async (tx, sdk, contractAddress = null) => {
+  let contractInstance = false;
```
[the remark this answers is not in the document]
peronczyk: This method was moved 1:1 from file `helper.js` (line 94). To be honest I would like to avoid messing around with this code within this PR. Maybe you can create a task for suggested changes?

src/types/index.ts
```diff
@@ -432,6 +444,26 @@ export interface ITopHeader {
...
 export type ISignMessage = (m: any) => Promise<any>
 
+export interface IName {
+  autoExtend: boolean;
+  createdAtHeight: number;
+  expiresAt: number;
+  hash: string;
+  name: string;
+  owner: string;
+  pointers: Record<string, any>;
```
subhod-i, line 454: I guess we do have a common type `dictionary` for this

subhod-i (approved):
Approved with one comment. PLease have a look.

src/types/index.ts
peronczyk: Sure, makes sense. Updated.

PR #1953, opened 2023-03-27
[SW-780] Keep the latest transactions state when navigating

peronczyk:


CedrikNikita (commented):
I think it's more a `perf` than a `fix` commit.

src/composables/latestTransactionList.ts
pstachel, line 1: What do you think of adding some space between functions?

pstachel (approved):

src/popup/components/LatestTransactionsCard.vue
```diff
@@ -59,99 +38,23 @@ export default defineComponent({
...
-    watch(
-      tokens,
-      (oldTokens, newTokens) => {
-        if (!isEqual(oldTokens, newTokens)) {
-          updateData();
-        }
-      },
-      { deep: true },
+    const isLoading = computed(
+      () => isTransactionListLoading.value && !transactionList.value.length,
```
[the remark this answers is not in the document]
peronczyk: Thanks for finding this out. I added new logic that monitors the network and resets the transaction list when the network changes. Please review the change.

src/composables/latestTransactionList.ts
peronczyk: OK

src/composables/latestTransactionList.ts
```diff
@@ -0,0 +1,143 @@
...
+    tokens,
+    (oldTokens, newTokens) => {
+      if (!isEqual(oldTokens, newTokens)) {
+        updateTransactionListData();
+      }
+    },
+    { deep: true },
+  );
+
+  onNetworkChange(store, () => {
```
CedrikNikita, line 133: Why can't we just watch here the `activeNetwork ` getter instead?

peronczyk: Because the component that uses the composable is not rendered at the moment of changing the network. So when we go to network settings, change network and go back the watcher won't fire.

CedrikNikita (approved):

PR #1960, opened 2023-03-29
[SW-748] Use SDK 13 types instead of cutsom defined types to determine transaction types and labels

subhod-i:


src/popup/utils/utils.ts
```diff
@@ -401,12 +400,18 @@ export function amountRounded(rawAmount: number | BigNumberPublic): string {
...
 }
 
-export function getTxType(tx: ITx): TxType {
-  return (
-    TX_TYPE_MDW[tx.type]
-    || (tx.tag && SCHEMA.OBJECT_ID_TX_TYPE[tx.tag])
-    || (Object.values(SCHEMA.TX_TYPE).includes(tx.type) && tx.type)
-  );
+export function getTxType(tx: ITx): Tag | null {
+  const typeKey = tx.type as keyof typeof Tag;
```
peronczyk, line 404: Is it possible to `tx.type` to have a proper type from the beginning?
Also this `keyof typeof Tag` type is used more than once. Maybe it would be good to create separate type from this?

src/popup/locales/de.json
```diff
@@ -74,7 +74,7 @@
       "SpendTx": "Spend",
       "ContractCreateTx": "Contract create",
       "ContractCallTx": "Contract call",
-      "NamePreclaimTx": "Preclaim name",
+      "namePreclaimTx": "Preclaim name",
```
peronczyk, line 77: I think we should not keep the untranslated string in translation files. Maybe just remove the whole untranslated block to avoid having to update all files every time something changes? Also the surrounding keys does not match the EN keys.

src/popup/locales/it.json
```diff
@@ -74,7 +74,7 @@
       "SpendTx": "Spend",
       "ContractCreateTx": "Contract create",
       "ContractCallTx": "Contract call",
-      "NamePreclaimTx": "Preclaim name",
+      "namePreclaimTx": "Preclaim name",
```
peronczyk, line 77: IDEM

src/popup/locales/ru.json
```diff
@@ -73,7 +73,7 @@
       "SpendTx": "Spend",
       "ContractCreateTx": "Contract create",
       "ContractCallTx": "Contract call",
-      "NamePreclaimTx": "Preclaim name",
+      "namePreclaimTx": "Preclaim name",
```
peronczyk, line 76: IDEM

peronczyk (commented):
Good job with the changes. Love to see it merged soon.

peronczyk (approved):

src/popup/locales/de.json
subhod-i: Good catch. This diff was by mistake. Rolling back. Thanks

src/popup/locales/it.json
subhod-i: Rolled back.

src/popup/locales/ru.json
subhod-i: Rolled back.

Liubov-crypto:
I used scan qr to propose TX from multisig acc and initial requested sum was 0.0001 AEE

token contract ct_2PREoQ49HXshAeSPTngmiT2p1Zvp7N9gUGt55Lb8MqQ5jiRMbj

it was scanned as 0.0001 AE token

The steps are recorded on video.

Seems like a bug,  cuz on the final stage we have a wrong info:

https://user-images.githubusercontent.com/69896204/229145294-2bd39923-a46c-4dfd-b751-198a15b52ef9.mov

src/popup/components/TransactionLabel.vue
```diff
@@ -203,9 +200,11 @@ export default defineComponent({
         };
       }
 
-      const translation = transactionListTypes[txType.value!] || transactionTypes[txType.value!];
+      const translation = transactionListTypes[
+        camelCase(Tag[outerTxType.value!])] || transactionTypes[camelCase(Tag[outerTxType.value!])
+      ];
```
pstachel, line 205: I'd suggest better formatting here 
```suggestion
     const selectedType = camelCase(Tag[outerTxType.value!])
      const translation = transactionListTypes[selectedType] || transactionTypes[selectedType]
```

pstachel (approved):

CharisSiarampalis:
Lgtm in general

Liubov-crypto:
Scan QR with preset amount isn't working for me.

src/popup/utils/constants.ts
```diff
@@ -1,5 +1,5 @@
 import { SCHEMA } from '@aeternity/aepp-sdk';
```
pstachel, line 1: Shouldn't it be removed?

src/types/index.ts
```diff
@@ -289,28 +289,6 @@ export type TxFunctionParsed = keyof typeof TX_FUNCTIONS;
 
 export type TxFunction = TxFunctionRaw | TxFunctionParsed;
 
-export type TxType =
```
peronczyk, line 292: Instead of removing this I suggest to set the value of `keyof typeof Tag` to allow of reusing it within the app.

subhod-i: Ideally `TxType` shouldn't be used. Only `Tag` must be used for the transaction type.

peronczyk: OK

src/popup/utils/constants.ts
subhod-i: Removing `Schema` from codebase is an independent PR

subhod-i:
> Scan QR with preset amount isn't working for me. Receiver's address will not be added.
> 
> saf un
> 
> Also got some issue when tried to swap WTT to AE, but maybe it's Dex side: errrr

Fixed it.

src/popup/components/TransactionOverview.vue
```diff
@@ -185,8 +182,28 @@ export default defineComponent({
...
+                address: recipientId,
+                name: name.value || getPreferredName.value(recipientId),
+                url: aeScan.prepareUrlByHash(recipientId),
+                label: t('transaction.overview.accountAddress'),
+              },
+              title: t('transaction.type.spendTx'),
+            };
+          }
+        }
+        // eslint-disable-next-line no-fallthrough
```
pstachel, line 204: Shouldn't we care about other innerTxTypes?

subhod-i: No, We support only SpendTx in multisig vaults

src/popup/utils/getPopupProps.ts
```diff
@@ -22,8 +21,12 @@ interface IPendingRequest {
 
 type PostMessageReturn = Promise<Partial<IPopupConfig> | null>;
 
-export function buildTx(txType: any) {
-  return TxBuilder.buildTx({ ...txParams[txType] }, txType);
+export function buildTx(txType: keyof typeof Tag) {
+  const params = {
+    ...txParams[txType],
+    tag: Tag[txType] as any,
```
pstachel, line 27: doesn't it have a type of sting?

pstachel (approved):

src/popup/utils/getPopupProps.ts
subhod-i: Added a todo. It's not the problem with tag. Made txParams as any.

PR #1963, opened 2023-03-29
[SW-788] Add colorful borders to avatars in all places where they are used 

pstachel:


ifaouibadi (approved):

src/popup/components/AccountInfo.vue
```diff
@@ -68,6 +69,7 @@ export default defineComponent({
     idx: { type: Number, default: 0 },
     canCopyAddress: Boolean,
     isMultisig: Boolean,
+    avatarWithoutBorder: Boolean,
```
peronczyk, line 72: As the `Avatar` component is using `borderless` prop I suggest to rename this prop to `avatarBorderless`.

peronczyk (approved):

PR #1964, opened 2023-03-29
[SW-525] Send modal does not open after pasting receive link

pstachel:


ifaouibadi (approved):

src/popup/components/DashboardWrapper.vue
```diff
@@ -50,12 +50,21 @@ export default defineComponent({
   setup(props, { root }) {
     const backedUpSeed = useState('backedUpSeed');
     const openModal = useDispatch('modals/open');
-    const query = computed(() => root.$route.query);
 
-    watch(query, (value) => {
```
CedrikNikita, line 55: Can't we just use `{ immediate: true }`?

CedrikNikita (commented):
In general it seems that token is not being prefilled at all, even with a fix.

CedrikNikita (approved):

PR #1967, opened 2023-03-30
[SW-790] Fix visual discrepancies in the latest Tx list

konradodo:


src/popup/components/TransactionTokenRows.vue
```diff
@@ -86,6 +86,14 @@ export default defineComponent({
...
 
     function calculateFontSize(amountValue: BigNumber | number) {
+      if (amountRounded(amountValue).length >= 14) {
+        return '14px';
+      }
+
+      if (amountRounded(amountValue).length >= 10) {
+        return '16px';
+      }
+
```
pstachel, line 96: Aren't those the exact same conditions that are below?

pstachel (approved):

src/popup/components/TransactionTokenRows.vue
```diff
@@ -86,19 +86,20 @@ export default defineComponent({
...
       }
-      if (amount.isLessThanOrEqualTo(new BigNumber(999999999))) {
+
+      if (amountLength <= 12) {
         return '16px';
       }
-      if (amount.isLessThanOrEqualTo(new BigNumber(999999999999))) {
+
+      if (amountLength <= 9) {
         return '14px';
```
pstachel, line 100: it won't work

pstachel (changes requested):

pstachel (approved):

src/popup/components/NameItem.vue
```diff
@@ -319,7 +319,7 @@ export default defineComponent({
       }
 
       .truncate {
-        @extend %face-sans-14-bold;
+        @extend %face-sans-15-medium;
```
CedrikNikita, line 322: Is this change coordinated with someone?

src/popup/components/Tokens.vue
```diff
@@ -137,7 +137,7 @@ export default defineComponent({
   }
 
   .symbol {
-    @extend %face-sans-16-medium;
+    @extend %face-sans-16-regular;
```
CedrikNikita, line 140: Same question for this change.

src/popup/components/NameItem.vue
konradodo: with tsvetan

src/popup/components/Tokens.vue
konradodo: with tsvetan

onvisions: @ Odo, @CedrikNikita
We have synced about the discrepancies observed in the font display in Figma compared to font display in the browser extension and the apps. In general the fonts used in the wallet seem to look bolder than the same font used in Figma. The idea was to make the font-weights look as much as possible alike. 

However what I see following the test link is not what the expected result was supposed to be. 
These are the correct values for the widget (and for the tx items in Transactions history list): 

image

As improvement of the font-weight display @pstachel already suggested a change in the bold text with font-weight 700 to font weight 600 (which is part of another PR) and I approved it.

src/popup/utils/utils.ts
```diff
@@ -480,16 +480,20 @@ export function openInNewWindow(url: string) {
...
+  if (amountLength <= 9) {
     return '16px';
   }
-  if (amount.isLessThanOrEqualTo(999999999999)) {
+
+  if (amountLength <= 12) {
     return '14px';
   }
+
   return '12px';
```
peronczyk, line 497: Maybe instead of returns this approach could be slimmer:
```
let size = '12px';
if ( ... ) {
  size = ...
} else if ( ... ) {
  size = ...
} else if ( ... ) {
 ...
}
return size;
```

peronczyk (approved):

PR #1970, opened 2023-04-04
[SW-532, SW-533] refactor Transaction List

konradodo:


src/popup/components/TransactionList.vue
```diff
@@ -50,165 +38,96 @@ import {
...
+      }
+
+      const {
+        scrollHeight,
+        scrollTop,
+        clientHeight,
+      } = viewportElement.value!;
+
+      if (scrollHeight - scrollTop <= clientHeight + 100) {
+        await loadMore;
```
pstachel, line 130: Does it work? And why do you need to await it here?

src/popup/utils/utils.ts
```diff
@@ -43,7 +43,7 @@ import type {
   BigNumberPublic,
   IPendingTransaction,
   IPageableResponse,
-  IDashboardTransaction,
+  IDashboardTransaction, IActiveMultisigTx,
```
pstachel, line 46:
```suggestion
  IDashboardTransaction, 
  IActiveMultisigTx,
```

src/popup/utils/utils.ts
```diff
@@ -364,10 +364,18 @@ export function getAccountNameToDisplay(acc: IAccount | undefined) {
...
 export function defaultTransactionSortingCallback(
-  a: ITransaction,
-  b: ITransaction,
+  a: ITransaction | IActiveMultisigTx,
+  b: ITransaction | IActiveMultisigTx,
 ) {
-  const [aMicroTime, bMicroTime] = [a, b].map((tr) => (new Date(tr.microTime)).getTime());
+  const [aMicroTime, bMicroTime] = [a, b].map(
+    (tr) => (new Date(
+      (tr as any).isMultisigTransaction
```
pstachel, line 372: why `tr` has the type of `any`?

src/popup/utils/utils.ts
```diff
@@ -364,10 +364,18 @@ export function getAccountNameToDisplay(acc: IAccount | undefined) {
...
 
 export function defaultTransactionSortingCallback(
-  a: ITransaction,
-  b: ITransaction,
+  a: ITransaction | IActiveMultisigTx,
+  b: ITransaction | IActiveMultisigTx,
 ) {
-  const [aMicroTime, bMicroTime] = [a, b].map((tr) => (new Date(tr.microTime)).getTime());
+  const [aMicroTime, bMicroTime] = [a, b].map(
+    (tr) => (new Date(
```
pstachel, line 371: couldn't the name of the variable be `transaction`?

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,128 @@
...
+import { i18n } from '../store/plugins/languages';
+
+interface TransactionItemList extends IDefaultComposableOptions {
+  isMultisig?: boolean,
+  token: string,
+}
+
+type ITransactionElement = ITransaction | IActiveMultisigTx
+
+export const useTransactionItemList = ({
```
pstachel, line 27: I'd suggest going with the name `useTransactionList` as you are not referring to a single element on the list but to the list as a whole

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+
+  function filterTransactionByPhrase(tr: ITransactionElement) {
+    return !searchPhrase.value || getTxSymbol.value(tr)
+      .toLocaleLowerCase()
+      .includes(searchPhrase.value.toLocaleLowerCase());
+  }
+
+  const loadedTransaction = computed((): ITransactionElement[] => {
```
pstachel, line 89: As you return multiple transactions
```suggestion
  const loadedTransactionList = computed((): ITransactionElement[] => {
```

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,128 @@
...
+  const filteredTransactions = computed(
+    () => loadedTransaction.value
+      .filter(filterProperTransactions)
+      .filter(filterTransactionByDisplayMode)
+      .filter(filterTransactionByPhrase)
+      .sort(defaultTransactionSortingCallback),
+  );
+
+  function getCommonTransaction(tx: ITransactionElement): ITransaction | null {
+    return (tx as any).isMultisigTransaction
```
pstachel, line 111: why any?

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,128 @@
...
+      ? null
+      : tx as ITransaction;
+  }
+
+  function getMultisigTransaction(tx: ITransactionElement): IActiveMultisigTx | null {
+    return (tx as any).isMultisigTransaction
+      ? tx as IActiveMultisigTx
```
pstachel, line 118: please don't mix `tx` and `transaction` as they are different to each other

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,128 @@
...
+import { usePendingMultisigTransaction } from './pendingMultisigTransaction';
+import { useTransactionAndTokenFilter } from './transactionAndTokenFilter';
+import { i18n } from '../store/plugins/languages';
+
+interface TransactionItemList extends IDefaultComposableOptions {
+  isMultisig?: boolean,
+  token: string,
+}
+
+type ITransactionElement = ITransaction | IActiveMultisigTx
```
peronczyk, line 25: In my opinion this is a really bed pattern to mix two totally different objects together. Paweł had already issues with this approach when we were trying to combine IAccount and IMultisigAccount. Here the differences are even bigger.

peronczyk (changes requested):

src/popup/components/TransactionList.vue
```diff
@@ -50,165 +38,96 @@ import {
...
-      if (props.maxLength && filteredTransactions.value.length >= props.maxLength) return;
-      if (scrollHeight - scrollTop <= clientHeight + 100) {
-        // eslint-disable-next-line no-use-before-define
-        loadMore();
-      }
-    }
-
-    async function loadMore() {
-      if (loading.value) return;
+    async function fetchData(recent?: boolean) {
```
subhod-i, line 96: `fetchData` is a generic name. We could be a bit more specific here in naming the method

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,128 @@
...
+    const innerTx = getInnerTransaction(tr.tx!);
+    return !token || (
+      token !== AETERNITY_CONTRACT_ID
+        ? innerTx?.contractId === token
+        : (!innerTx.contractId || !isFungibleTokenTx(innerTx))
+    );
+  }
+
+  function filterTransactionByDisplayMode(tr: ITransactionElement) {
+    const { direction, isDex } = useTransactionTx({
```
peronczyk, line 63: I think it's bad thing to use another composable in a loop (this function is a callback for `filter` so it will be fired multiple times.

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    return transactionList;
+  });
+
+  const filteredTransactions = computed(
+    () => loadedTransaction.value
+      .filter(filterProperTransactions)
+      .filter(filterTransactionByDisplayMode)
+      .filter(filterTransactionByPhrase)
+      .sort(defaultTransactionSortingCallback),
```
peronczyk, line 107: From my experience it's a bad pattern to create multiple callbacks for the filters and use them this way. For actions like this you can use the piping concept: https://www.freecodecamp.org/news/pipe-and-compose-in-javascript-5b04004ac937/

```
export function pipe<T = any[]>(fns: ((data: T) => T)[]) {
  return (data: T) => fns.reduce((currData, func) => func(currData), data);
}
```

Convert all the callback functions to do single filtering:
```
function filterProperTransactions(data: X) {
  return data.filter( ... );
}
```

And here use those 2 thing as follows:
```
() => pipe([
  filterProperTransactions,
  filterTransactionByDisplayMode,
  filterTransactionByPhrase,
  defaultTransactionSortingCallback,
])(loadedTransaction.value)
```

peronczyk (changes requested):

src/composables/transactionItemList.ts
konradodo: @CedrikNikita Have we better way to checking direction and isDex ?

src/composables/transactionItemList.ts
konradodo: In this case, I don't see another solution. We must handle both types in transaction list.

src/composables/transactionItemList.ts
CedrikNikita: I think we can extract the logic into separate functions.

src/popup/components/TransactionList.vue
```diff
@@ -50,165 +38,97 @@ import {
...
+import { useDispatch, useGetter } from '../../composables/vuex';
 import {
   useMultisigAccounts,
-  usePendingMultisigTransaction,
   useTransactionAndTokenFilter,
-  useTransactionTx,
+  useViewport,
 } from '../../composables';
-
+import { useTransactionList } from '../../composables/transactionItemList';
```
pstachel, line 49: Why don't you import it straight from the composables?

src/composables/directionOfTransaction.ts
pstachel, line 1: getting rid of "of" word would be nice here

src/composables/ownership.ts
```diff
@@ -0,0 +1,72 @@
...
+  IDefaultComposableOptions,
+} from '../types';
+import { i18n } from '../store/plugins/languages';
+import {
+  TRANSACTION_OWNERSHIP_STATUS,
+  getTxOwnerAddress,
+} from '../popup/utils';
+import { useSplitTransaction } from './splitTransaction';
+
+interface UseTransactionOptions extends IDefaultComposableOptions {
```
pstachel, line 15: those are not options for a transaction

src/composables/ownership.ts
```diff
@@ -0,0 +1,72 @@
...
+  tx?: ITx
+  externalAddress?: string
+}
+
+export function useOwnership({
```
pstachel, line 20: I'd suggest to state what ownership is this composable about

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  function filterProperTransactions(transactionList: ITransactionElement[]) {
+    return transactionList.filter((transaction) => {
+      const innerTx = getInnerTransaction(transaction.tx!);
+      return !token || (
+        token !== AETERNITY_CONTRACT_ID
+          ? innerTx?.contractId === token
+          : (!innerTx.contractId || !isFungibleTokenTx(innerTx))
+      );
```
pstachel, line 62: For better readability I'd suggest not to mix OR operator with the ternary operator

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,146 @@
...
+    });
+  }
+
+  function filterTransactionByDisplayMode(transactionList: ITransactionElement[]) {
+    return transactionList.filter((transaction) => {
+      const { direction } = useDirectionOfTransaction({
```
pstachel, line 68: Why do you use composable in a function?

src/composables/viewport.ts
```diff
@@ -12,13 +12,15 @@ export type OnViewportScrollCallback = (p: IScrollCallbackParams) => any;
 const viewportElement = ref<Element | undefined>();
 
 export const useViewport = () => {
-  const viewportScroll = debounce((callback: OnViewportScrollCallback) => {
-    const isDesktop = (
+  function checkIsDesktop():boolean {
```
pstachel, line 15:
```suggestion
  function checkIsDesktop(): boolean {
```

src/composables/transactionTx.ts
```diff
@@ -30,33 +27,31 @@ export function useTransactionTx({
...
-  function setExternalAddress(address: string) {
-    ownerAddress.value = address;
-  }
+  const {
+    isDex,
+  } = useDex({ tx, store });
+
+  const {
+    direction,
+  } = useDirectionOfTransaction({ store, tx, externalAddress });
```
pstachel, line 46: As mentioned above - I'm not convinced by the idea of using 4 separate composables for one thing that's always used together

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,143 @@
...
+  IActiveMultisigTx,
+  IDefaultComposableOptions, IDexContracts,
+  ITokenList,
+  ITransaction,
+  ITx,
+} from '../types';
+import {
+  AETERNITY_CONTRACT_ID,
+  defaultTransactionSortingCallback,
+  getInnerTransaction, isTxDex,
```
pstachel, line 12:
```suggestion
  getInnerTransaction, 
  isTxDex,
```

src/composables/transactionTx.ts
```diff
@@ -17,7 +17,7 @@ import {
   TX_FUNCTIONS,
   getTxType,
   isContainingNestedTx,
-  getInnerTransaction,
+  getInnerTransaction, isTxDex,
```
pstachel, line 20:
```suggestion
  getInnerTransaction, 
  isTxDex,
```

src/popup/utils/utils.ts
```diff
@@ -29,7 +29,7 @@ import {
   STUB_CALLDATA,
   STUB_NONCE,
   TX_FUNCTIONS,
-  TX_TYPE_MDW,
+  TX_TYPE_MDW, FUNCTION_TYPE_DEX,
```
pstachel, line 32:
```suggestion
  TX_TYPE_MDW, 
  FUNCTION_TYPE_DEX,
```

peronczyk (changes requested):

src/popup/utils/utils.ts
```diff
@@ -442,10 +446,18 @@ export function getAccountNameToDisplay(acc: IAccount | undefined) {
...
-  a: ITransaction,
-  b: ITransaction,
+  a: ITransaction | IActiveMultisigTx,
+  b: ITransaction | IActiveMultisigTx,
 ) {
-  const [aMicroTime, bMicroTime] = [a, b].map((tr) => (new Date(tr.microTime)).getTime());
+  const [aMicroTime, bMicroTime] = [a, b].map(
+    (transaction) => (new Date(
+      (transaction as any).isMultisigTransaction
+        ? (transaction as IActiveMultisigTx).createdAt
```
peronczyk, line 455: As far as I know `createdAt` refers to the Generalised Account, not the transaction.

src/popup/components/TransactionList.vue
```diff
@@ -50,163 +38,99 @@ import {
...
-      if (props.maxLength && filteredTransactions.value.length >= props.maxLength) return;
-      if (scrollHeight - scrollTop <= clientHeight + 100) {
-        // eslint-disable-next-line no-use-before-define
-        loadMore();
-      }
-    }
-
-    async function loadMore() {
-      if (loading.value) return;
+    async function fetchTransactionData(recent?: boolean) {
```
peronczyk, line 99: This function fetches a list, not a single transaction, so I suggest to use plural or add `List` to the name.

src/popup/utils/utils.ts
CedrikNikita: There are a `createdAt` in the response from `https://ga-multisig-backend-mainnet.prd.aepps.com/tx/...`, which is referring to the creation time of a transaction proposal.

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,145 @@
...
+    FILTER_MODE,
+  } = useTransactionAndTokenFilter();
+
+  const { pendingMultisigTransaction } = usePendingMultisigTransaction({ store });
+
+  function isFungibleTokenTx(tx: ITx) {
+    return Object.keys(availableTokens.value).includes(tx.contractId);
+  }
+
+  function filterProperTransactions(transactionList: ITransactionElement[]) {
```
peronczyk, line 58: I suggest to add a comment that would explain what "proper" means.

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,145 @@
...
+
+type ITransactionElement = ITransaction | IActiveMultisigTx
+
+export function useTransactionList({
+  store,
+  isMultisig,
+  token,
+}: TransactionItemList) {
+  const availableTokens = computed<ITokenList>(() => store.state.fungibleTokens.availableTokens);
+  const transactions = computed(() => store.state.transactions);
```
peronczyk, line 35: I suggest to type this.

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,145 @@
...
+  const getTxSymbol = computed(() => store.getters.getTxSymbol);
+  const getTxDirection = computed(() => store.getters.getTxDirection);
+  const getDexContracts = computed<IDexContracts>(() => store.getters.getDexContracts);
+
+  const getAccountPendingTransactions = computed<IActiveMultisigTx[]>(
```
peronczyk, line 40: This is not returning `IActiveMultisigTx[]` but `ITransaction[]` as far as I know.

peronczyk (changes requested):

src/popup/utils/utils.ts
CedrikNikita: But in this case the `createdAt` is related to an activeAccount
https://github.com/aeternity/superhero-wallet/blob/247e8e08368c8afca28e1dd48c124bfc2c6a9222/src/composables/multisigTransactions.ts#L78-L84

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,145 @@
...
+  pipe,
+  TX_DIRECTION,
+} from '../popup/utils';
+import { usePendingMultisigTransaction } from './pendingMultisigTransaction';
+import { useTransactionAndTokenFilter } from './transactionAndTokenFilter';
+import { i18n } from '../store/plugins/languages';
+
+interface TransactionItemList extends IDefaultComposableOptions {
+  isMultisig?: boolean,
+  token: string,
```
peronczyk, line 24: I suggest to rename this to `tokenContractId` as the current name suggest this might have something to do with the `IToken` which is totally not true.

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  function isFungibleTokenTx(tx: ITx) {
+    return Object.keys(availableTokens.value).includes(tx.contractId);
+  }
+
+  function filterProperTransactions(transactionList: ITransactionElement[]) {
+    return transactionList.filter((transaction) => {
+      if (!token) {
+        return true;
+      }
```
peronczyk, line 62: I suggest to not filter at all if there is no token provided.
```suggestion
    return (token)
      ? transactionList.filter((transaction) => { ... })
      : transactionList;
```

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,145 @@
...
+import { i18n } from '../store/plugins/languages';
+
+interface TransactionItemList extends IDefaultComposableOptions {
+  isMultisig?: boolean,
+  token: string,
+}
+
+type ITransactionElement = ITransaction | IActiveMultisigTx
+
+export function useTransactionList({
```
peronczyk, line 29: After spending some time with this code I have a conclusion that it makes no sense to move the code from the component and create separate composable for the code below.
The reason for creating composables is to have a state shared across multiple components. This composable does not have its own state and is included only in one component.

If you want to make the component slimmer you can move most of the filtering functions to utils file.

peronczyk (changes requested):

src/popup/utils/utils.ts
```diff
@@ -658,6 +670,25 @@ export function isTxOfASupportedType(encodedTx: string, isTxBase64 = false) {
...
+export function pipe<T = any[]>(fns: ((data: T) => T)[]) {
+  return (data: T) => fns.reduce((currData, func) => func(currData), data);
+}
+
+export function isTxDex(tx: ITx, dexContractx: IDexContracts) {
+  const { wae, router } = dexContractx;
+
+  return !!(
+    tx?.contractId
+    && tx?.function
```
peronczyk, line 683: No need to use question marks here as the `tx` is required (can't be nullish).

src/popup/utils/utils.ts
```diff
@@ -658,6 +670,25 @@ export function isTxOfASupportedType(encodedTx: string, isTxBase64 = false) {
...
+    && (
+      Object.values(FUNCTION_TYPE_DEX).flat()
+        .includes(tx?.function as TxFunctionRaw)
```
peronczyk, line 686: There's a check for the existence of `tx.function` so ne need to use question mark here.

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,145 @@
...
+    return Object.keys(availableTokens.value).includes(tx.contractId);
+  }
+
+  function filterProperTransactions(transactionList: ITransactionElement[]) {
+    return transactionList.filter((transaction) => {
+      if (!token) {
+        return true;
+      }
+
+      const innerTx = getInnerTransaction(transaction.tx!);
```
peronczyk, line 64: In my opinion using exclamation mark here is rather bad. You can't be sure that the `tx` is not undefined. There are some transactions without the `tx`. I suggest to remove the exclamation mark but modify the `getInnerTransaction` to have the possibility of the `tx` argument to be undefined.

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,145 @@
...
+          return direction === TX_DIRECTION.sent && !isDex;
+        case FILTER_MODE.in:
+          return direction === TX_DIRECTION.received;
+        default:
+          throw new Error(`${i18n.t('pages.recentTransactions.unknownMode')} ${displayMode.value.key}`);
+      }
+    });
+  }
+
+  function filterTransactionByPhrase(transactionList: ITransactionElement[]) {
```
peronczyk, line 93: Missing "search" word in the function name. Also plural version of "transaction" should be used as we are operating an array here.
```suggestion
  function filterTransactionsBySearchPhrase(transactionList: ITransactionElement[]) {
```

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,145 @@
...
+      const innerTx = getInnerTransaction(transaction.tx!);
+      if (token !== AETERNITY_CONTRACT_ID) {
+        return innerTx?.contractId === token;
+      }
+
+      return !innerTx.contractId || !isFungibleTokenTx(innerTx);
+    });
+  }
+
+  function filterTransactionByDisplayMode(transactionList: ITransactionElement[]) {
```
peronczyk, line 73: Plural version of "transaction" should be used as we are operating an array here.
```suggestion
  function filterTransactionsByDisplayMode(transactionList: ITransactionElement[]) {
```

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,145 @@
...
+    const transactionList: ITransactionElement[] = [
+      ...getAccountPendingTransactions.value,
+      ...(transactions.value as any).loaded as ITransaction[],
+    ];
+
+    if (isMultisig && pendingMultisigTransaction.value?.tx) {
+      transactionList.push(pendingMultisigTransaction.value);
+    }
+
+    return transactionList;
```
peronczyk, line 111: Can be written in a shorter form:
```suggestion
    return [
      ...getAccountPendingTransactions.value,
      ...(transactions.value as any).loaded as ITransaction[],
      ...((isMultisig && pendingMultisigTransaction.value?.tx)
        ? pendingMultisigTransaction.value
        : []
      ),
    ];
```

src/composables/transactionItemList.ts
peronczyk, line 73: I think this function should be moved to the utils file and change it to make the filtering based on the `tx`, not the whole `transaction`. The reason for this is that both transaction types has the `th: ITx` inside.

src/composables/transactionItemList.ts
```diff
@@ -0,0 +1,145 @@
...
+      }
+
+      return !innerTx.contractId || !isFungibleTokenTx(innerTx);
+    });
+  }
+
+  function filterTransactionByDisplayMode(transactionList: ITransactionElement[]) {
+    return transactionList.filter((transaction) => {
+      const direction = getTxDirection.value(transaction.tx);
+      const isDex = isTxDex(transaction.tx!, getDexContracts.value);
```
peronczyk, line 76: We can't be sure if the `tx` is not undefined so I suggest to not use the exclamation mark here.

peronczyk (changes requested):

src/types/index.ts
```diff
@@ -392,7 +392,7 @@ export interface ITransaction {
   tipUrl?: string;
   transactionOwner?: string;
   tx: ITx;
-  url?: string;
+  url?: number;
```
peronczyk, line 395: Url as a number? Are you sure?

src/types/index.ts
```diff
@@ -665,3 +666,12 @@ export interface IDefaultComposableOptions {
 }
 
 export type StatusIconType = typeof ALLOWED_ICON_STATUSES[number];
+
+export type ITransactionElement = ITransaction | IActiveMultisigTx
```
peronczyk, line 670: It would make more sense if the multisig transaction interface be renamed to `IMultisigTransaction`. In our terminology `TX` is something encapsulated inside the `Transaction`.

Also I suggest to rename `ITransactionElement` to something more precise. Word `Element` suggests that this interface describes sub-group of the `Transaction` which is not true. So maybe `IGeneralTransaction` or `ICommonTransaction`?

Last thing I suggest is to move this and the next interface closer to the `ITransaction` interface to have them as one group.

src/popup/utils/utils.ts
```diff
@@ -664,6 +686,22 @@ export function isTxOfASupportedType(encodedTx: string, isTxBase64 = false) {
   return SUPPORTED_TX_TYPES.includes(SCHEMA.OBJECT_ID_TX_TYPE[txObject.tag]);
 }
 
+// eslint-disable-next-line no-unused-vars
+
```
peronczyk, line 690: Something is wrong here - empty line and disable next line.

src/composables/viewport.ts
```diff
@@ -51,5 +53,6 @@ export const useViewport = () => {
     viewportElement,
     onViewportScroll,
     initViewport,
+    checkIsDesktop,
```
peronczyk, line 56: No need to add this here as it's not used in the template.

src/popup/utils/utils.ts
```diff
@@ -443,11 +453,19 @@ export function getAccountNameToDisplay(acc: IAccount | undefined) {
   return acc?.name || `${i18n.t('pages.account.heading')} ${(acc?.idx || 0) + 1}`;
 }
 
-export function defaultTransactionSortingCallback(
-  a: ITransaction,
-  b: ITransaction,
+export function sortTransactionByDate(
```
peronczyk, line 456: The function name is changed, but the logic not. So currently the name is not correct as this function is not actually sorting anything.

src/popup/components/TransactionList.vue
```diff
@@ -50,163 +40,189 @@ import {
...
+    const getTxSymbol = useGetter('getTxSymbol');
+    const getTxDirection = useGetter('getTxDirection');
+    const getDexContracts = useGetter<IDexContracts>('getDexContracts');
 
-    const { activeAccount } = useAccounts({ store: root.$store });
-    const { activeMultisigAccount } = useMultisigAccounts({ store: root.$store });
+    const getAccountPendingTransactions = useGetter<ITransaction[]>('getAccountPendingTransactions');
 
-    const { pendingMultisigTransaction } = usePendingMultisigTransaction({ store: root.$store });
+    const canLoadMore = computed(() => !!transactions.value.nextPageUrl);
```
peronczyk, line 97: Moving the getters and computed properties above the composables in my opinion is a bad thing. In most cases we the order of the logic in the setup function is:
1. Composables (use...)
2. Refs
3. Getters
4. Computed
5. Methods/Functions
6. Hooks

src/types/index.ts
```diff
@@ -636,6 +636,7 @@ export interface IActiveMultisigTx extends IMultisigAccount {
   hash?: string;
   tx?: ITx;
   isMultisigTransaction: boolean;
+  proposalCreatedAt?: number,
```
peronczyk, line 639: The interface properties should end with semicolons (`;`).

src/types/index.ts
```diff
@@ -665,3 +666,12 @@ export interface IDefaultComposableOptions {
...
 
 export type StatusIconType = typeof ALLOWED_ICON_STATUSES[number];
+
+export type ITransactionElement = ITransaction | IActiveMultisigTx
+
+export type ITransactionsState = {
+  loaded: ITransaction[],
+  nextPageUrl?: string,
+  pending: any, // TODO type that object
+  tipWithdrawnTransactions: ITransaction[],
```
peronczyk, line 676: The interface properties should end with semicolons (`;`).

peronczyk (changes requested):

peronczyk (changes requested):

src/composables/transactionItemList.ts
peronczyk: OK. Let's leave it as it is. Maybe we'll come up with something better in the future.

src/popup/utils/utils.ts
konradodo: check it now please

src/popup/utils/utils.ts
```diff
@@ -663,6 +682,37 @@ export function isTxOfASupportedType(encodedTx: string, isTxBase64 = false) {
...
+    && tx.function
+    && (
+      Object.values(FUNCTION_TYPE_DEX).flat()
+        .includes(tx.function as TxFunctionRaw)
+    )
+    && [...wae, ...router].includes(tx.contractId)
+  );
+}
+
+export const getTxOwnerAddress = (innerTx?: ITx) => innerTx?.accountId || innerTx?.callerId;
```
peronczyk, line 699: I suggest to use regular function instead of arrow to maintain consistency within this file.

src/popup/utils/utils.ts
```diff
@@ -663,6 +682,37 @@ export function isTxOfASupportedType(encodedTx: string, isTxBase64 = false) {
...
+
+export const getOwnershipStatus = (
```
peronczyk, line 701: I suggest to use regular function instead of arrow to maintain consistency within this file.

src/types/index.ts
```diff
@@ -430,7 +438,14 @@ export interface IDexContracts {
...
   wae: string[];
 }
+export type ICommonTransaction = ITransaction | IActiveMultisigTransaction
 
+export type ITransactionsState = {
+  loaded: ITransaction[];
+  nextPageUrl?: string;
+  pending: any; // TODO type that object
+  tipWithdrawnTransactions: ITransaction[];
+}
```
peronczyk, line 448: I suggest to add empty line after this one.

src/types/index.ts
```diff
@@ -430,7 +438,14 @@ export interface IDexContracts {
   router: string[];
   wae: string[];
 }
+export type ICommonTransaction = ITransaction | IActiveMultisigTransaction
```
peronczyk, line 441: I suggest to add empty line before this one.

src/popup/utils/utils.ts
peronczyk: I added new comment regarding the name so I'm closing this one.

src/popup/components/TransactionList.vue
```diff
@@ -50,179 +40,230 @@ import {
...
+    const { pendingMultisigTransaction } = usePendingMultisigTransaction({ store: root.$store });
+
+    const loading = ref(false);
+    const isDestroyed = ref(false);
+
     const availableTokens = useState<ITokenList>('fungibleTokens', 'availableTokens');
-    const transactions = useState('transactions');
-    const activeNetwork = useGetter<INetwork>('activeNetwork');
-    const getAccountPendingTransactions = useGetter('getAccountPendingTransactions');
+
```
pstachel, line 117:
```suggestion
```

src/popup/components/TransactionList.vue
```diff
@@ -50,179 +40,230 @@ import {
...
-        ];
-
-        if (props.isMultisig && pendingMultisigTransaction.value?.tx) {
-          transactionListLocal.push(pendingMultisigTransaction.value);
-        }
+    const loadedTransactionList = computed((): ICommonTransaction[] => [
+      ...getAccountPendingTransactions.value,
+      ...transactions.value.loaded,
+      ...((props.isMultisig && pendingMultisigTransaction.value?.tx)
+        ? [pendingMultisigTransaction.value!]
```
pstachel, line 136: do you need that exclamation mark at the end if you check pendingMultisigTransaction presence in a line above

src/popup/components/TransactionList.vue
```diff
@@ -50,179 +40,230 @@ import {
...
-        ? document.querySelector('.app-inner') : document.documentElement)!;
-      if (props.maxLength && filteredTransactions.value.length >= props.maxLength) return;
-      if (scrollHeight - scrollTop <= clientHeight + 100) {
-        // eslint-disable-next-line no-use-before-define
-        loadMore();
+    function narrowTransactionsToDefinedToken(transactionList: ICommonTransaction[]) {
+      if (props.token) {
+        return transactionList.filter((transaction) => {
+          const innerTx = getInnerTransaction(transaction.tx);
+          if (props.token !== AETERNITY_CONTRACT_ID) {
```
pstachel, line 149:
```suggestion
          const innerTx = getInnerTransaction(transaction.tx);
          
          if (props.token !== AETERNITY_CONTRACT_ID) {
```

src/popup/components/TransactionList.vue
```diff
@@ -50,179 +40,230 @@ import {
...
+
+        const txOwnerAddress = getTxOwnerAddress(innerTx);
+
+        const direction = getTxDirection.value(
+          outerTx.payerId ? outerTx : innerTx,
+          (transaction as ITransaction).transactionOwner
+          || (
+            getOwnershipStatus(activeAccount.value, accounts.value, innerTx)
+            !== TRANSACTION_OWNERSHIP_STATUS.current
+            && txOwnerAddress
```
pstachel, line 172: wrap the first condition with brackets

src/popup/components/TransactionList.vue
```diff
@@ -50,179 +40,230 @@ import {
...
+        }
+      });
+    }
+
+    function filterTransactionsBySearchPhrase(transactionList: ICommonTransaction[]) {
+      return transactionList.filter(
+        (transaction) => !searchPhrase.value || getTxSymbol.value(transaction)
+          .toLocaleLowerCase()
+          .includes(searchPhrase.value.toLocaleLowerCase()),
+      );
```
pstachel, line 198: If I'm not mistaken we have a function to compare lower case strings

src/popup/components/TransactionList.vue
```diff
@@ -50,179 +40,230 @@ import {
...
-    async function getLatest() {
-      try {
-        await root.$store.dispatch('fetchTransactions', { limit: 10, recent: true, address: currentAddress.value });
-      } finally {
-        loading.value = false;
+    async function loadMore() {
+      if (loading.value) {
+        return;
+      }
+      await fetchTransactionList();
```
pstachel, line 231: I don't see a point in this return 
```suggestion
      if (!loading.value) {
        await fetchTransactionList();
      }
```

src/popup/utils/utils.ts
```diff
@@ -39,6 +39,7 @@ import {
   STUB_NONCE,
   TX_FUNCTIONS,
   TX_TYPE_MDW,
+  FUNCTION_TYPE_DEX, TRANSACTION_OWNERSHIP_STATUS,
```
pstachel, line 42:
```suggestion
  FUNCTION_TYPE_DEX, 
  TRANSACTION_OWNERSHIP_STATUS,
```

src/popup/utils/utils.ts
```diff
@@ -71,6 +76,11 @@ export function includes<T, U extends T>(arr: readonly U[], elem: T): elem is U
   return arr.includes(elem as any);
 }
 
+// eslint-disable-next-line no-unused-vars
```
pstachel, line 79: check recent PR's - Bart solved this issue

src/types/index.ts
```diff
@@ -431,6 +439,15 @@ export interface IDexContracts {
   wae: string[];
 }
 
+export type ICommonTransaction = ITransaction | IActiveMultisigTransaction
+
+export type ITransactionsState = {
+  loaded: ITransaction[];
+  nextPageUrl?: string;
+  pending: any; // TODO type that object
```
pstachel, line 447: isn't that a boolean?

pstachel (changes requested):

src/types/index.ts
peronczyk: Nope, this is a list of pending transactions.

src/popup/utils/utils.ts
peronczyk: Just rebase to develop.

src/popup/components/TransactionList.vue
konradodo: no, we have utils to check equals, not for includes

peronczyk (approved):

pstachel (approved):

PR #1972, opened 2023-04-04
[SW-531] Fix error when opening the "Terms of use" from Index page and trying to go back

pstachel:


src/popup/components/Header.vue
```diff
@@ -107,6 +107,9 @@ export default defineComponent({
     );
 
     function back() {
+      if (root.$route.meta?.directBackRoute) {
```
peronczyk, line 110: I suggest to move the line 117 before this test and pick the `meta` the same way as the `fullPath` then use the `meta` here and in line 113.

peronczyk (approved):

src/popup/components/Header.vue
pstachel: good idea

src/popup/components/Header.vue
```diff
@@ -107,14 +107,18 @@ export default defineComponent({
     );
 
     function back() {
-      if (root.$route.meta?.backRoute) {
+      const { fullPath, meta } = root.$route;
+
+      if (meta?.directBackRoute) {
+        return root.$router.go(-1);
+      }
+      if (meta?.backRoute) {
```
subhod-i, line 115: `meta.backRoute` is accessed twice. Could be moved to a variable.

src/popup/components/Header.vue
```diff
@@ -107,14 +107,18 @@ export default defineComponent({
...
         // TODO: rewrite back button logic in more unified way
-        return root.$router.push(root.$route.meta?.backRoute);
+        return root.$router.push(meta?.backRoute);
       }
-      let { fullPath } = root.$route;
-      fullPath = fullPath.endsWith('/') ? fullPath.slice(0, -1) : fullPath;
+      const path = fullPath.endsWith('/') ? fullPath.slice(0, -1) : fullPath;
       return root.$router.push(
-        fullPath.substr(0, fullPath.lastIndexOf('/')) || { name: currentHomeRouteName.value },
+        path.substr(0, path.lastIndexOf('/')) || { name: currentHomeRouteName.value },
```
subhod-i, line 121: Ternary and '||" operators are used sequentially. Decreases the code readability.

Could be rewritten to
```js
    const path = fullPath.endsWith('/') ? fullPath.slice(0, -1) : fullPath;
    const lastSlashIndex = path.lastIndexOf('/');
    if (lastSlashIndex !== -1) {
      path = path.substr(0, lastSlashIndex);
    } else {
      path = { name: currentHomeRouteName };
    }
    return root.$router.push(path);
```

PR #1973, opened 2023-04-05
Feat: Solution for syncing the composables states with local storage

peronczyk:
Solution similar to https://vueuse.org/core/useStorage/ but also uses the network.

### DIFF
--- a/src/composables/composablesHelpers.ts
+++ b/src/composables/composablesHelpers.ts
@@ -1,9 +1,13 @@
 import {
+  computed,
+  getCurrentInstance,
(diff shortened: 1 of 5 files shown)


Liubov-crypto:
@peronczyk  This pr seems to reverse some of the changes that have been implemented before, for example I can reproduce the following here SW - 773 when entering the wallet for the first time: 

https://user-images.githubusercontent.com/69896204/230618393-cc7ce38c-5990-450d-ba10-7ab6cddb5cf6.mov

PR #1975, opened 2023-04-06
[SW-834] Refactor: Move isLoggedIn getter to accounts composable

peronczyk:


src/background/utils.js
```diff
@@ -28,8 +29,8 @@ export const getTippingContractInstance = async (tx) => {
 };
 
 export const contractCallStatic = async ({ tx, callType }) => {
-  const { account } = store.getters;
-  if (typeof callType !== 'undefined' && callType === 'static' && account) {
+  const { isLoggedIn } = useAccounts({ store });
+  if (typeof callType !== 'undefined' && callType === 'static' && isLoggedIn.value) {
```
pstachel, line 33: You could store `typeof callType !== 'undefined' && callType === 'static'` in a variable as it's repeated here

pstachel (approved):

src/background/utils.js
peronczyk: Makes sense. But I noticed that this is repetition makes no sense in general because the only difference between those two checks is determining if user is logged in. So I changed the structure to make it more readable.

src/composables/accounts.ts
```diff
@@ -7,6 +7,7 @@ export function useAccounts({ store }: IDefaultComposableOptions) {
   const accounts = computed((): IAccount[] => store.getters.accounts);
   const account = computed((): IAccount => store.getters.account);
   const accountsAddressList = computed(() => accounts.value.map((acc) => acc.address));
+  const isLoggedIn = computed(() => Object.keys(account.value).length > 0);
```
subhod-i, line 10: @peronczyk Could you explain this method? How it is determining if the user has logged in or not?

peronczyk: This code was just moved from the `getters.js` file - it is not my invention. But actually makes sense as the `account` is always an object, which is empty if there is no user logged in. Or you think it should be updated?

subhod-i: Okay.  We may have to revisit this when we add biometric authentication to the superhero wallet.

subhod-i (approved):

src/composables/accounts.ts
peronczyk: Definitely.

PR #1987, opened 2023-04-12
[SW-799] Remove animation on Multisig transaction proposal detail page

pstachel:


src/popup/components/Header.vue
```diff
@@ -198,12 +198,15 @@ export default defineComponent({
     display: flex;
     justify-content: flex-end;
     gap: 8px;
-    width: 20%;
+
+    .btn-close {
+      position: relative;
+      top: unset;
+      right: unset;
+    }
```
peronczyk, line 206: I think this styles should replace the styles used in the `BtnClose`. From my experience components should not position themselves.

peronczyk (changes requested):

konradodo (approved):

ifaouibadi (approved):

peronczyk (approved):

PR #1988, opened 2023-04-12
feat: fix

konradodo:


src/popup/utils/utils.ts
```diff
@@ -16,7 +16,7 @@ import {
 import {
   ADDRESS_TYPES,
   AENS_DOMAIN,
-  AETERNITY_CONTRACT_ID,
+  AETERNITY_CONTRACT_ID, DIRECTION,
```
peronczyk, line 19: Please split this line.

src/popup/utils/constants.ts
```diff
@@ -23,6 +23,11 @@ export const AETERNITY_SYMBOL = 'AE';
 export const AETERNITY_COIN_SYMBOL = 'AE Coin';
 export const AETERNITY_COIN_NAME = 'Aeternity';
 
+export const DIRECTION = {
+  sent: 'sent',
+  received: 'received',
```
peronczyk, line 28: Please also remove this values from `TX_FUNCTIONS` constant.

src/popup/components/TransactionTokenRows.vue
```diff
@@ -8,7 +8,7 @@
       :key="token.symbol"
       :class="[
         'token-row',
-        token.isReceived ? TX_FUNCTIONS.received : TX_FUNCTIONS.sent,
+        token.isReceived ? DIRECTION.received : DIRECTION.sent,
```
peronczyk, line 11: I think this was a mistake from the beginning to use the constant values here. Those are CSS class names and I think it should be placed here as strings.

src/popup/components/TransactionTokenRows.vue
```diff
@@ -48,6 +49,11 @@ import Tokens from './Tokens.vue';
 
 export default defineComponent({
   components: { Tokens },
+  computed: {
+    DIRECTION() {
+      return DIRECTION;
+    },
+  },
```
peronczyk, line 56: ?? :D

src/popup/utils/constants.ts
```diff
@@ -23,6 +23,11 @@ export const AETERNITY_SYMBOL = 'AE';
 export const AETERNITY_COIN_SYMBOL = 'AE Coin';
 export const AETERNITY_COIN_NAME = 'Aeternity';
 
+export const DIRECTION = {
```
peronczyk, line 26: I suggest to rename this constant to `TX_DIRECTION`.

peronczyk (changes requested):

src/composables/transactionTokens.ts
```diff
@@ -10,7 +10,7 @@ import type {
 import {
   AETERNITY_SYMBOL,
   MAGNITUDE,
-  TX_FUNCTIONS,
+  DIRECTION,
```
subhod-i, line 13: `TX_DIRECTION` would be even more appropriate.
`TX_DIRECTION.credit` instead of `received` and `TX_DIRECTION.debit` instead of `sent` is something you can consider too.

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -222,7 +222,7 @@ export default defineComponent({
     const totalAmount = computed(() => getTxAmountTotal.value(txWrapped.value, direction.value));
 
     const singleToken = computed((): ITokenResolved => ({
-      isReceived: direction.value === TX_FUNCTIONS.received,
+      isReceived: direction.value === DIRECTION.received,
```
subhod-i, line 225: `direction.value` IDEM

src/popup/components/TokenAmount.vue
```diff
@@ -55,10 +55,7 @@ export default defineComponent({
     hideFiat: Boolean,
     direction: {
       type: String,
-      validator: (value: any) => [
-        TX_FUNCTIONS.sent,
-        TX_FUNCTIONS.received,
-      ].includes(value),
+      validator: (value: any) => Object.values(DIRECTION).includes(value),
```
subhod-i, line 58: Perhaps we can use string union for direction instead of using `any` type here

subhod-i (approved):

src/popup/utils/constants.ts
```diff
@@ -23,14 +23,17 @@ export const AETERNITY_SYMBOL = 'AE';
 export const AETERNITY_COIN_SYMBOL = 'AE Coin';
 export const AETERNITY_COIN_NAME = 'Aeternity';
 
+export const TX_DIRECTION = {
+  credit: 'sent',
+  debit: 'received',
```
peronczyk, line 28: I spoke with Subhod and Badi and I feel that using "credit" / "debit" makes sense in general, but not in our case. The SDK is using "sent" and "receive" so we (Badi and me) think that it should be "sent" / "receive".

src/types/index.ts
```diff
@@ -611,3 +612,5 @@ export interface IDefaultComposableOptions {
    */
   store: Store<any>
 }
+
+export type ITxDirection = typeof TX_DIRECTION[keyof typeof TX_DIRECTION];
```
peronczyk, line 616: Prefix `I` should be used only for interfaces and data structures.

peronczyk (approved):

src/types/index.ts
```diff
@@ -14,6 +14,7 @@ import {
   MULTISIG_CREATION_PHASES,
   TX_FUNCTIONS,
   FUNCTION_TYPE_MULTISIG,
+  TX_DIRECTION,
```
CedrikNikita, line 17:
```suggestion
```
The import is unused.

CedrikNikita (changes requested):

CedrikNikita (approved):
Code changes look reasonable. I suggest testing team to test all the affected components.

PR #1989, opened 2023-04-12
[SW-732] Display the correct error message while attempting to set a pointer from a low balance account

subhod-i:


ifaouibadi (approved):

src/store/plugins/names.js
```diff
@@ -155,6 +156,9 @@ export default (store) => {
             { root: true },
           );
         } catch (e) {
+          if (isInsufficientBalanceError(e)) {
+            e.message = i18n.t('modals.insufficient-balance.msg');
+          }
```
peronczyk, line 161: I suggest to avoid modifying error message property. Instead I suggest to change line 162 in the following way:
```
dispatch('modals/open', {
  name: MODAL_DEFAULT,
  msg: (isInsufficientBalanceError(e))
    ? i18n.t('modals.insufficient-balance.msg')
    : e.message
}, { root: true });
```

peronczyk (changes requested):

src/store/plugins/names.js
subhod-i: Thanks

PR #1992, opened 2023-04-13
[SW-816, SW-828] Display a low balance message when sending an approved multisig transaction.

pstachel:


src/popup/utils/helper.js
pstachel, line 1: Code for helper.js borrowed from https://github.com/aeternity/superhero-wallet/pull/1989/files#diff-08d84b32fbf152c4407ebc44276a4df0362fd3c88bc8e226d6a33c9c1d15e6d7R186-R189

src/popup/pages/MultisigProposalDetails.vue
```diff
@@ -439,7 +443,11 @@ export default defineComponent({
 
         proposalCompleted.value = true;
       } catch (error) {
-        handleUnknownError(error);
+        if (isInsufficientBalanceError(error)) {
+          openModal({ name: MODAL_DEFAULT, icon: 'alert', msg: root.$t('modals.vaultLowBalance.msg') });
+        } else {
+          handleUnknownError(error);
```
subhod-i, line 449: We can notify the user if there is any other error by using a modal.
I am okay with the current approach too. Up to you @pstachel

subhod-i (approved):

src/popup/pages/MultisigProposalDetails.vue
pstachel: You're right, I'll adjust the code

src/popup/pages/MultisigProposalDetails.vue
```diff
@@ -595,7 +602,7 @@ export default defineComponent({
   }
 
   .bottom-buttons {
-    padding: 24px;
+    padding: var(--screen-padding-x);
```
CedrikNikita, line 605: The variable value and the previous value of the padding are different. Are you sure with this changes?

pstachel: That's what SW-828 was about - to align the padding of the buttons with the view

src/popup/components/Modals/Default.vue
```diff
@@ -19,7 +19,14 @@
...
 
-    <slot name="msg">
+    <TemplateRenderer
+      v-if="!$slots.msg"
+      :str="msg"
+    />
+    <slot
+      v-else
+      name="msg"
+    >
```
peronczyk, line 29: No need to use v-if here. Just do this:
```
<slot name="msg">
  <TemplateRenderer ... />
</slot>
```

src/popup/locales/en.json
```diff
@@ -119,6 +119,10 @@
     "insufficient-balance": {
       "msg": "You do not have enough balance for this transaction."
     },
+    "vaultLowBalance": {
+      "title": "Insufficient vault balance",
+      "msg": "Multisig vault balance is not enough to pay for the transaction fee. <br /><br />Please top up multisig vault."
```
peronczyk, line 124: Please remove the space before BRs :)

src/popup/utils/helper.js
```diff
@@ -181,3 +182,13 @@ export const isValidURL = (url) => {
...
+
+const hasValidationKey = (error, expectedKey) => (
+  error.validation
+  && error.validation.some(({ key }) => expectedKey === key)
+);
+
+export const isInsufficientBalanceError = (error) => (
+  error instanceof InvalidTxError
+  && hasValidationKey(error, 'InsufficientBalance')
+);
```
peronczyk, line 194: Please avoid adding new utils to non-TS files. There is pending PR for moving all stuff from this file into the utils.ts file.

peronczyk (changes requested):

src/popup/utils/utils.ts
```diff
@@ -506,3 +507,13 @@ export function isTxOfASupportedType(encodedTx: string, isTxBase64 = false) {
...
+
+const hasValidationKey = (error: any, expectedKey: string) => (
+  error.validation
+  && error.validation.some(({ key }: any) => expectedKey === key)
+);
+
+export const isInsufficientBalanceError = (error: any) => (
+  error instanceof InvalidTxError
+  && hasValidationKey(error, 'InsufficientBalance')
+);
```
peronczyk, line 519: Please use regular functions instead of arrows same as all other functions here.

peronczyk (approved):

PR #1994, opened 2023-04-14
[SW-830] perf: poll multisig accounts only if necessary

CedrikNikita:


src/composables/multisigAccounts.ts
```diff
@@ -29,6 +29,10 @@ const POLLING_INTERVAL = 7000;
 const LOCAL_STORAGE_MULTISIG_KEY = 'multisig';
 const LOCAL_STORAGE_MULTISIG_PENDING_KEY = 'multisig-pending';
 
+export interface MultisigAccountsOptions extends IDefaultComposableOptions {
+  pollOnce?: boolean;
```
CedrikNikita, line 33:
```suggestion
  pollOnce?: boolean;
```
I'm not fully satisfied with the variable name. I'm open for suggestions

pstachel: what do you think about `fetchOnce`?

pstachel (approved):

konradodo (approved):

ifaouibadi (approved):

src/composables/multisigAccounts.ts
peronczyk: Maybe follow the same pattern as here: `src/composables/notifications.ts`?

CedrikNikita: It's just a little bit different. We would still load once in case it never been loaded, if it was loaded we will use state from the localstorage.

subhod-i:
@CedrikNikita Interesting PR. Poll only if necessary but how are you determining if someone else created a vault and you are just a cosigner? Is the polling still on with this PR?

CedrikNikita:
> @CedrikNikita Interesting PR. Poll only if necessary but how are you determining if someone else created a vault and you are just a cosigner? Is the polling still on with this PR?

Yes, it is. It is off on transactionDetails screen where we need only an address of the current multisig. And where we are using only the functions that are not dependent on multisig updating.

Liubov-crypto:
LGTM

PR #1999, opened 2023-04-17
[SW-833] refactor: use one component to sign messages

CedrikNikita:


pstachel (approved):

src/popup/pages/SignMessage.vue
```diff
@@ -1,78 +1,41 @@
...
-      const signatureHex = Buffer.from(signature).toString('hex');
-      openCallbackOrGoHome(true, { signature: signatureHex });
-    };
-
-    return {
-      sendAddress,
-    };
+    onMounted(async () => {
+      const { callbackOrigin, openCallbackOrGoHome } = useDeepLinkApi({ router: root.$router });
+      const { getSdk } = useSdk({ store: root.$store });
```
peronczyk, line 11: The instantiation of the composables should be put at the beginning of the `setup` to follow common way we do it within other components.

src/popup/pages/SignMessage.vue
```diff
@@ -1,78 +1,41 @@
...
+          },
+        });
+
+        const signature = await sdk.signMessage(message);
+        const signatureHex = Buffer.from(signature).toString('hex');
+        openCallbackOrGoHome(true, { signature: signatureHex });
+      } catch (e: any) {
+        openCallbackOrGoHome(false);
+
+        if (e.message !== 'Rejected by user') handleUnknownError(e);
```
peronczyk, line 33: This comparison with `Rejected by user` or throwing error with this message is quite common in the system. I suggest to create some kind of abstraction layer for this.
For example by creating custom error:
```
class RejectedByUserError extends Error { ... }
```
and catching it this way:
```
catch(err) {
  if (err instanceof RejectedByUserError) {
    handleUnknownError(e);
  }
}
```

peronczyk (changes requested):

PR #2002, opened 2023-04-18
[SW-839] fix: set a proper url for a connected aepp

CedrikNikita:


pstachel (approved):

konradodo (approved):

Liubov-crypto:
LGTM for Dex and superhero.com

src/popup/pages/SignMessage.vue
```diff
@@ -20,7 +20,7 @@ export default defineComponent({
           app: {
             name: callbackOrigin.value?.host,
             host: callbackOrigin.value?.host,
-            url: callbackOrigin.value?.origin,
+            url: callbackOrigin.value,
```
peronczyk, line 23: Is this correct? Here the instance of `URL` (class) or null is passed. Maybe `toString` method should be applied here?

src/store/plugins/sdk.js
```diff
@@ -44,6 +44,7 @@ export default (store) => {
               name: originUrl.host,
               host: originUrl.host,
               protocol: originUrl.protocol,
+              url: originUrl,
```
peronczyk, line 47: Same here, are you sure instance of `URL` is valid value? Not a string?

peronczyk (approved):

PR #2006, opened 2023-04-19
fix: list item hover effect on mobile

ifaouibadi:


pstachel (approved):

src/styles/variables.scss
```diff
@@ -39,7 +39,7 @@ $color-bg-1-hover: #090909;
 $color-bg-2: #171717;
 $color-bg-3: #282828;
 $color-bg-4: #191919;
-$color-bg-4-hover: #121212;
```
peronczyk, line 42: We should not create colors with the opacity in the variables file.

peronczyk (changes requested):

src/styles/variables.scss
```diff
@@ -39,7 +39,7 @@ $color-bg-1-hover: #090909;
...
+$color-bg-4-hover: #12121266;
```
peronczyk, line 42: We should not create colors with the opacity in the variables file.

peronczyk (changes requested):

Liubov-crypto:
There is a little bit hover effect if you click on a transaction while scrolling. If you release it, then everything is active:

https://user-images.githubusercontent.com/69896204/233081571-90a6a953-4d85-4cc0-8202-8058c7dd71e6.MP4

PR #2014, opened 2023-04-20
[SW-448] Performance update: disable data polling when on web and the tab is not active

pstachel:


CedrikNikita (commented):
I think it is more a `perf` than a `feat` type of a commit.

src/composables/visibilityChange.ts
peronczyk, line 1: I have few things to consider regarding this file:
1. I suggest to move the logic to `ui` composable.
2. I suggest to have a state `isAppActive` (or something similar).
3. Instead of stopping and running again the polling handler I suggest to run or prevent running the callback passed to polling handler based on the mentioned state. So the interval will be running all the time, but it won;t run anything when the tab is not active.
What do you think about this isea?

src/popup/components/TransactionList.vue
```diff
@@ -71,12 +71,15 @@ import {
   useTransactionAndTokenFilter,
   useTransactionTx,
 } from '../../composables';
+import { createPollingBasedOnMountedComponents } from '../../composables/composablesHelpers';
```
peronczyk, line 74: In my opinion this is not a good change to use this composable helper inside of a component. The reason for this helper was to be used inside of the composables - functions used within many components. This is why it has the mounted components counter.

src/popup/pages/Names/Auction.vue
```diff
@@ -24,16 +24,18 @@
...
   defineComponent,
   ref,
-  onBeforeUnmount,
   watch,
 } from '@vue/composition-api';
 import BigNumber from 'bignumber.js';
-import { aettosToAe, executeAndSetInterval } from '../../utils';
+import { aettosToAe } from '../../utils';
 import { useMiddleware } from '../../../composables';
+import { createPollingBasedOnMountedComponents } from '../../../composables/composablesHelpers';
```
peronczyk, line 32: IDEM

src/popup/pages/Names/List.vue
```diff
@@ -21,37 +21,40 @@
...
 
 <script lang="ts">
-import { computed, defineComponent, onBeforeUnmount } from '@vue/composition-api';
-import type { IName } from '../../../types';
-import { useState } from '../../../composables/vuex';
+import { computed, defineComponent } from '@vue/composition-api';
+
 import { useAccounts } from '../../../composables';
+import { useDispatch, useState } from '../../../composables/vuex';
+import { createPollingBasedOnMountedComponents } from '../../../composables/composablesHelpers';
```
peronczyk, line 28: IDEM

peronczyk (changes requested):

src/composables/visibilityChange.ts
pstachel: Sure, I can proceed with that approach

CedrikNikita (approved):
Code changes look reasonable.

peronczyk (approved):

PR #2024, opened 2023-04-24
fix: Disable possibility of vault creation with chain name or an url

pstachel:


subhod-i:
Target branch not 'develop' @pstachel

PR #1976, opened 2023-04-06
[SW-835] Refactor: Move active account getter to accounts composable

peronczyk:

PR #2026, opened 2023-04-25
[SW-865] Fix name claim

CedrikNikita:
missed in #1976

### DIFF
--- a/src/popup/pages/Names/Claim.vue
+++ b/src/popup/pages/Names/Claim.vue
@@ -43,7 +43,7 @@
     <BtnMain
       class="btn-register"
       extend
-      :disabled="!sdk || !name || errors.any()"
+      :disabled="!isSdkReady || !name || errors.any()"
       @click="claim"
     >
       {{
@@ -55,7 +55,8 @@
   </div>
 </template>
 
-<script>
+<script lang="ts">
+import { defineComponent, ref, computed } from '@vue/composition-api';
 import { TxBuilderHelper } from '@aeternity/aepp-sdk';
 import {
   MAGNITUDE,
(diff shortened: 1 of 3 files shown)


pstachel (approved):

src/popup/pages/Names/Claim.vue
```diff
@@ -72,94 +73,99 @@ import CheckBox from '../../components/CheckBox.vue';
...
-            this.$store.commit('names/setPendingAutoExtendName', name);
+          const { salt } = await sdk.aensPreclaim(fullName);
+          claimTxHash = (await sdk.aensClaim(fullName, salt, { waitMined: false })).hash;
+          if (autoExtend.value) {
+            root.$store.commit('names/setPendingAutoExtendName', fullName);
           }
-          this.$router.push({ name: ROUTE_ACCOUNT_DETAILS_NAMES });
-        } catch (e) {
+          root.$router.push({ name: ROUTE_ACCOUNT_DETAILS_NAMES });
+        } catch (e: any) {
```
subhod-i, line 125:
```suggestion
        } catch (e: Exception) {
```

src/popup/pages/Names/Claim.vue
```diff
@@ -72,94 +73,99 @@ import CheckBox from '../../components/CheckBox.vue';
...
+            root.$store.dispatch('names/updatePointer', {
+              name: fullName,
               address: activeAccount.value.address,
             });
           }
-        } catch (e) {
-          openDefaultModal({
-            msg: e.message,
-          });
+        } catch (e: any) {
```
subhod-i, line 148: IDEM

subhod-i (changes requested):

src/popup/pages/Names/Claim.vue
CedrikNikita: Every other place in the project has `: any` in such case, i don't want to solve that in this particular issue.
```
Catch clause variable type annotation must be 'any' or 'unknown' if specified.Vetur(1196)
interface Error
```

subhod-i: Yeah, I see that. Perhaps we should start writing custom exceptions. This is not the right moment as too many refactorings already on it's way.

subhod-i (approved):

PR #2033, opened 2023-04-26
[SW-851] Update design of Account selection modal

pstachel:


Liubov-crypto:
sw-851

src/popup/components/FormSelectOptionsHeader.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+  <div class="form-select-options-header">
+    <span class="text-heading-3 title">{{ title }}</span>
+    <InputSearch
+      v-if="options.length > 3"
```
konradodo, line 5: 3 -> Move that values to constans

konradodo (changes requested):

src/popup/components/FormSelectOptionsHeader.vue
```diff
@@ -0,0 +1,64 @@
...
+      if (searchTerm.value) {
+        const searchTermLower = searchTerm.value.toLowerCase();
+        const result = props.options.filter(({ text, value }) => (
+          text.toLowerCase().includes(searchTermLower)
+          || value.toString().includes(searchTermLower)
+        ));
+
+        emit('search', result);
+      }
+      emit('search', props.options);
```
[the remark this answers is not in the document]
pstachel: Good catch, missed that one

konradodo (approved):

konradodo (approved):

src/popup/components/FormSelectOptionsHeader.vue
```diff
@@ -0,0 +1,70 @@
...
+  components: { InputSearch },
+  props: {
+    value: { type: String, default: '' },
+    title: { type: String, default: '' },
+    searchPlaceholder: { type: String as PropType<String | TranslateResult>, default: '' },
+    options: { type: Array as PropType<IFormSelectOption[]>, default: () => [] },
+  },
+  setup(props, { emit }) {
+    const searchTerm = ref('');
+    function filterOptions(searchPhrase: string) {
```
peronczyk, line 32: This component should not do the actual filtering. Instead it should emit the search phrase and the parent component should react on this.

peronczyk (changes requested):

src/popup/components/TokenAmount.vue
```diff
@@ -141,5 +144,13 @@ export default defineComponent({
       @extend %face-sans-18-regular;
     }
   }
+
+  &.align-right {
+    text-align: right;
+
+    .fiat {
+      @extend %face-sans-12-regular;
+    }
```
peronczyk, line 153: I think it's bad to change the font weight of elements if we are changing the alignment. Those 2 things are not connected.

src/popup/components/TokenAmount.vue
```diff
@@ -57,6 +57,7 @@ export default defineComponent({
     noSymbol: Boolean,
     highPrecision: Boolean,
     dynamicSizing: Boolean,
+    alignRight: Boolean,
```
peronczyk, line 60: Why not using the CSS text-align on the parent level as this component is not a block but inline element?

peronczyk (approved):

PR #2043, opened 2023-05-02
Refactor/migrate usage of watch when used with array

martinkaintas:
- revert some changes that were made in previous commits regarding the migration to Vue3
- fix usage of `watch` when used with an array in order for it to be compatible with Vue3


src/popup/components/ToggleMultisigButton.vue
```diff
@@ -65,7 +65,7 @@ export default defineComponent({
       if (!multisigAccounts.value?.length && props.isMultisig) {
         toggleMultisigDashboard(false);
       }
-    });
+    }, { deep: true });
```
CedrikNikita, line 68: The goal of the wathcer here is to toggle multisigDashboard back to general accounts if there are no multisig ones. So there is no need in the deep option here.

src/popup/pages/SeedPhraseVerifySettings.vue
```diff
@@ -127,7 +127,7 @@ export default defineComponent({
     watch(() => selectedWordIds.value, () => {
       showNotification.value = false;
       hasError.value = false;
-    });
+    }, { deep: true });
```
CedrikNikita, line 130: I think that the whole watcher here is obsolete, and actually doing nothing.

CedrikNikita (changes requested):

peronczyk (approved):

src/popup/components/ToggleMultisigButton.vue
peronczyk: This is related to breaking change in Vue 3: https://v3-migration.vuejs.org/breaking-changes/watch.html
So in my opinion the change is correct, but I can't be 100% sure.

peronczyk: Also I consider having this watcher and logic for switching from multisig here is a bad thing. That's a tiny component deep in the structure of the APP and it's responsible for sucha big thing. I remember first time I was searching for the logic that was redirecting me from the multisig and I didn't found it. @pstachel eventually helped me as he was the author of this. Maybe we should think about moving this somewhere else? @pstachel @CedrikNikita

src/popup/pages/SeedPhraseVerifySettings.vue
CedrikNikita: I can remove it in a separate PR, if it is needed.

src/popup/components/ToggleMultisigButton.vue
martinkaintas: @CedrikNikita you are correct, the deep option is not needed here. I've updated the PR

ZGiovas (approved):

PR #2044, opened 2023-05-02
[SW-884] Optimize sign from non active account implementation

subhod-i:


pstachel (approved):

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -172,6 +172,7 @@ export default defineComponent({
     resolve: { type: Function, required: true },
     // eslint-disable-next-line no-unused-vars
     reject: { type: Function as PropType<(e: Error) => void>, required: true },
+    fromAccount: { type: String as PropType<String>, default: null },
```
CedrikNikita, line 175: Is it actually used within this file?

subhod-i: Okay, I will fix this.

Liubov-crypto:
I sent a transaction from a newly created vault (vault was in sync process), but then I could not revoke it. There was this error in the console:

err approve

https://user-images.githubusercontent.com/69896204/237056975-259227d6-6ed0-44ad-93e1-7011e69cac3e.mov

Also several send transactions are stuck in pending status, clicking on them opens a 404 page:
st

Liubov-crypto:
The revoke error described by me before it's not reproducible now. 
I can choose another account on 2nd step when I'm creating vault. 

But got another error in console: 
errr

PR #2046, opened 2023-05-04
Feature/remove vue compat

martinkaintas:
- remove @vue/compat (migration build)
- fix issues that came up
- comment out code that is filtering errors since that is no longer possible (will be fixed in a next PR when we decide how to handle this)

### DIFF
--- a/src/lib/logger.js
+++ b/src/lib/logger.js
@@ -1,6 +1,5 @@
 /* eslint-disable no-console */
 import { pick } from 'lodash-es';
-import Vue from 'vue';
 import { detect } from 'detect-browser';
 import { getState } from '../store/plugins/persistState';
 import { useModals } from '../composables';
@@ -9,16 +8,16 @@ export default class Logger {
   static background;
 
   static init(options = {}) {
-    const { background } = options;
+    const { background, app } = options;
     Logger.background = background;
-    if (!background) {
-      Vue.config.errorHandler = (error, vm, info) => {
+    if (!background && app) {
+      app.config.errorHandler = (error, vm, info) => {
         console.error(info);
         console.error(error);
         if (error && error.message !== 'Rejected by user') Logger.write({ message: error.toString(), info, type: 'vue-error' });
       };
 
-      Vue.config.warnHandler = (message, vm, info) => {
+      app.config.warnHandler = (message, vm, info) => {
         console.error(message);
         console.error(info);
         Logger.write({
--- a/src/popup/components/Header.vue
+++ b/src/popup/components/Header.vue
@@ -41,7 +41,7 @@
         v-if="showHeaderNavigation"
         data-cy="close"
         class="btn-close"
-        @click="close"
+        @click.capture="close"
       />
       <template v-else>
         <NetworkButton />
--- a/src/popup/components/Header.vue
+++ b/src/popup/components/Header.vue
@@ -41,7 +41,7 @@
         v-if="showHeaderNavigation"
         data-cy="close"
         class="btn-close"
-        @click.capture="close"
+        @click="close"
       />
       <template v-else>
         <NetworkButton />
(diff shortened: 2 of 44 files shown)


CedrikNikita:
The reason it is not being build on side of ci is, that it is trying to get some dependencies through `ssh` instead of `https`
https://github.com/aeternity/superhero-wallet/actions/runs/4880223539/jobs/8707607063?pr=2046#step:8:8

src/popup/components/Header.vue
```diff
@@ -40,7 +40,7 @@
       <BtnClose
         v-if="showHeaderNavigation"
         data-cy="close"
-        @click="close"
+        @click.capture="close"
```
peronczyk, line 43: Instead of using `capture` I suggest to update the close button by removing the inner click event handler.

src/popup/components/Modal.vue
```diff
@@ -41,7 +43,7 @@
           <BtnClose
             v-if="hasCloseButton"
             data-cy="btn-close"
-            @click="$emit('close')"
+            @click.capture="handleClose"
```
peronczyk, line 46: Instead of using `capture` I suggest to update the plain button by removing the inner click event handler.

peronczyk (changes requested):

peronczyk (changes requested):

src/popup/components/Modal.vue
```diff
@@ -100,19 +104,27 @@ export default defineComponent({
   },
   emits: ['close', 'opened'],
   setup(props, { slots }) {
+    const isModalVisible = ref(props.isVisible);
```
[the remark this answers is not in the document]
martinkaintas: Turns out, since modals are only mounted when they are visible we do not need to pass the variable from the parent component to the modal (no more copying prop state).
We do need the v-if statement though to make sure that the modal plays the exit animation when it is closed.
[[see](https://v3-migration.vuejs.org/breaking-changes/transition-as-root.html)]

src/lib/logger.js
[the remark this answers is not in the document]
martinkaintas: I believe it is better to do the rebase on the `feature/migrate-to-vue3` branch once this PR is merged to avoid any more conflicts

src/popup/components/TransferSendForm.vue
```diff
@@ -294,14 +293,14 @@ export default defineComponent({
...
 
-      const items = errors.value[fieldName as keyof typeof errors.value];
-      if (items) {
-        // eslint-disable-next-line no-console
-        console.log(items);
-        // TODO: Vee-validate v4 does not return error codes with errors
-        // We need to find a new way to filter messages
-        throw new Error('Function needs refactoring.');
-      }
+      // const items = errors.value[fieldName as keyof typeof errors.value];
```
ZGiovas, line 296: Why do we need to keep them as commented code here?

ZGiovas (changes requested):
Minor change may needed.

src/popup/components/TransferSendForm.vue
martinkaintas: As mentioned above in the description of this PR, I've commented out code that is filtering errors since that is no longer possible and it will be fixed in a separate PR

ZGiovas (approved):
LGTM

src/popup/components/TransferSendForm.vue
ZGiovas: Ok LGTM then.

ZGiovas (approved):
LGTM

src/popup/components/Modal.vue
```diff
@@ -109,11 +111,16 @@ export default defineComponent({
       }
     });
 
+    const handleClose = () => {
```
peronczyk, line 114: We agreed within the team to use regular functions (not arrows) when setting up the functions. This makes the setup block easier to follow with eyes and distinguish functions from constants.

peronczyk (approved):

PR #2049, opened 2023-05-04
Replace vue router exact prop

ZGiovas:


subhod-i:
The pipeline wouldn't succeed if the commit message is not in Conventional commit standards

PR #2051, opened 2023-05-05
Refactor node connection

CedrikNikita:


ifaouibadi (approved):

src/popup/pages/NetworkForm.vue
```diff
@@ -149,6 +150,21 @@ export default defineComponent({
...
+        if (!(newNetwork.value as any).networkId) {
+          root.$store.commit('setUserNetwork', {
+            ...newNetwork.value,
+            index: networks.value[newNetwork.value.name].index,
+            networkId: node.nodeNetworkId,
+          });
+        }
+      } catch (e) {
+        handleUnknownError(e);
+      }
```
subhod-i, line 165: What is the motivation for defining `Node` once again? You could use the SDK API.

subhod-i (changes requested):

src/popup/pages/NetworkForm.vue
peronczyk: I also don't like this piece of code. For me updating the user networks data with data fetched from the node is rather bad thing.

peronczyk (changes requested):

src/store/plugins/sdk.js
```diff
@@ -174,19 +173,14 @@ export default (store) => {
         });
         commit('setSdkReady');
       },
+      async changeNode(_, network) {
+        await watchUntilTruthy(() => store.getters['sdkPlugin/sdk']);
+        sdk.pool.delete(network.name);
+        sdk.addNode(network.name, await Node({ url: network.url }), true);
+      },
```
subhod-i, line 180: Declared this method but called only once. Doesn't need tobe a function. Code could be inline.

src/lib/wallet.js
```diff
@@ -38,9 +38,30 @@ export default async function initSdk() {
 
   store.commit('setNodeStatus', NODE_STATUS_CONNECTING);
   try {
+    const { getMiddleware, initMiddleware } = useMiddleware({ store });
+    store.watch(
+      (state, getters) => getters.activeNetwork,
+      async (network, oldNetwork) => {
```
subhod-i, line 44: I am not sure about vue standards but naming variables `new` and `old` always seems ideal.

src/lib/wallet.js
```diff
@@ -38,9 +38,30 @@ export default async function initSdk() {
...
+        if (isEqual(network, oldNetwork)) return;
+        try {
+          if (store.state.nodeStatus === NODE_STATUS_ERROR && !store.getters['sdkPlugin/sdk']) {
```
subhod-i, line 47: @peronczyk Could `nodeStatus` be one of the composables?
I see `store.commit('setNodeStatus', NODE_STATUS_CONNECTING);` too

src/lib/wallet.js
```diff
@@ -38,9 +38,30 @@ export default async function initSdk() {
...
+            await initSdk();
+            return;
+          }
+          store.commit('setNodeStatus', NODE_STATUS_CONNECTING);
+          await Promise.all([
+            network.middlewareUrl !== oldNetwork.middlewareUrl ? initMiddleware() : null,
+            network.url !== oldNetwork.url ? store.dispatch('sdkPlugin/changeNode', network) : null,
+          ]);
```
subhod-i, line 55: `network.middlewareUrl !== oldNetwork.middlewareUrl` & `network.url !== oldNetwork.url` are unnecessary given the initial condition `if (isEqual(network, oldNetwork)) return;` in this watcher.

subhod-i (changes requested):

src/lib/wallet.js
CedrikNikita: User cancreate a new network using same url, but different middleware url.

src/lib/wallet.js
CedrikNikita: It can't. It is using the variable inside this particular file.

src/store/plugins/sdk.js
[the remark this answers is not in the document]
CedrikNikita: It can't. It is using the variable inside this particular file.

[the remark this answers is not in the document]
peronczyk: I agree with Subhod that this should not be added to the Vuex. I suggest to move it to the SDK composable. Something similar to this one: https://github.com/aeternity/superhero-wallet/pull/2069/files#diff-e6d7101251daf167a357f2c6276f092dc6b657ba30c3cbdd2082a144a124189dR43

src/lib/wallet.js
subhod-i: Oh yeah. You're right!

CedrikNikita:
the issue was fixed in #2069

PR #2060, opened 2023-05-08
fix: replace vue router exact prop

ZGiovas:


ZGiovas:
Will update it and then create a new PR

PR #2062, opened 2023-05-08
fix: replace vue router exact prop

ZGiovas:


martinkaintas (approved):
LGTM

peronczyk (approved):

PR #2082, opened 2023-05-18
[SW-900] Ability to create new network from a link

peronczyk:


src/popup/components/TransferSendForm.vue
```diff
@@ -363,12 +365,17 @@ export default defineComponent({
...
+      token,
+    }: Dictionary) {
+      formModel.value.selectedAsset = availableTokens.value[token]
         || aeternityToken.value;
-      if (query.account) formModel.value.address = query.account;
-      if (query.amount) formModel.value.amount = query.amount;
-      if (query.payload) formModel.value.payload = query.payload;
+      if (account) formModel.value.address = account;
+      if (amount) formModel.value.amount = amount;
+      if (payload) formModel.value.payload = payload;
```
pstachel, line 378: I suggest to wrap the assignments in braces
```suggestion
      if (account) {
        formModel.value.address = account;
      }
      if (amount) {
       formModel.value.amount = amount;
     }
      if (payload) {
        formModel.value.payload = payload;
      }
```

pstachel (approved):

src/popup/pages/NetworkForm.vue
```diff
@@ -149,22 +151,34 @@ export default defineComponent({
...
+        keys.forEach((key) => {
+          const val = query[key];
+
+          if (val && typeof val === 'string') {
+            set(newNetwork.value, key, val);
+          }
+        });
+
+        await nextTick();
+        $validator.validateAll();
```
martinkaintas, line 181: It seems like validation is not working correctly, using the link below I can create a network whose name is bigger than 15 characters

https://feat-create-new-network-from-a-link.wallet.z52da5wt.xyz/more/settings/networks/add?name=ysssdass12s3dasdo&url=https%3A%2F%2Ftest.com

martinkaintas (changes requested):

src/popup/pages/NetworkForm.vue
peronczyk: Nice finding. I added custom veeValidate rule for this.

PR #2085, opened 2023-05-21
[SW-916] fix: account avatar loading

ifaouibadi:


src/composables/ui.ts
```diff
@@ -2,7 +2,7 @@ import { onBeforeUnmount, onMounted, ref } from '@vue/composition-api';
 import { ROUTE_ACCOUNT } from '../popup/router/routeNames';
 
 const homeRouteName = ref(ROUTE_ACCOUNT);
-const isAppActive = ref(false);
+const isAppActive = ref(true);
```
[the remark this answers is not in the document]
pstachel: We can do it, as it will be set after entering the app - so the initial state should be active

src/popup/utils/utils.ts
```diff
@@ -727,3 +727,14 @@ export function connectFrames(sdk: ISdk | AeSdkWallet) {
...
   }
 }
+
+export function checkImageAvailability(url: string): Promise<boolean> {
+  return fetch(url)
+    .then((response) => {
+      if (response.ok) {
+        return true;
+      }
+      return false;
```
pstachel, line 737:
```suggestion
      return !!response.ok
```

src/popup/utils/utils.ts
```diff
@@ -727,3 +727,9 @@ export function connectFrames(sdk: ISdk | AeSdkWallet) {
     handleUnknownError(error);
   }
 }
+
+export function checkImageAvailability(url: string): Promise<boolean> {
```
peronczyk, line 731: I suggest to add a comment that describes why we have to use custom image existence check instead of the previous solution (problems on iOS).

src/popup/utils/utils.ts
```diff
@@ -727,3 +727,9 @@ export function connectFrames(sdk: ISdk | AeSdkWallet) {
...
+  return fetch(url)
```
peronczyk, line 732: Instead of fetching whole file I suggest to only fetch headers: https://stackoverflow.com/questions/47240975/how-to-fetch-using-fetch-api-only-headers-in-javascript-not-the-whole-content

peronczyk (changes requested):

src/composables/ui.ts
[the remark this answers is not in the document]
peronczyk: This change is bad in my opinion. When user opens up the browser and the recently opened tabs opens this will be set to true, but actually should be false as the tab may not be visible to the user. Instead of setting this to true I suggest to run the `handleVisibilityChange` at the beginning of the `initVisibilityListeners`:

```
function initVisibilityListeners() {
  handleVisibilityChange();
  ...
```

peronczyk (changes requested):

src/composables/ui.ts
```diff
@@ -12,11 +12,12 @@ export function useUi() {
     }
   }
 
-  function handleVisibilityChange(event: Event) {
+  function handleVisibilityChange(event: Partial<Event>) {
     isAppActive.value = !(event.target as Document).hidden;
   }
 
   function initVisibilityListeners() {
+    handleVisibilityChange({ target: document });
```
peronczyk, line 20: I think there is no need to pass any arguments to the `handleVisibilityChange` function. As we are operating on the `document` object it could be directly used there this way:
```
isAppActive.value = !document.hidden;
```

src/popup/utils/utils.ts
peronczyk: Your comment actually didn't describe why this function exists. I suggest to add an info describing that we need to check for the image existence because regular image error catching didn't work on iOS.

peronczyk (changes requested):

peronczyk (approved):

PR #2086, opened 2023-05-23
Hotfix: Adjust send modal qr code scanning

CedrikNikita:


peronczyk (approved):

Liubov-crypto:
LGTM

pstachel (approved):

CedrikNikita:
> For me its not working with tokens. It can scan and detect the address but it does not select the token automatically. It still displayed ae. 

Does the device you scanning with have the same network chosen as the device you are scanning from?

csiarab:
I was trying to scan a token from a different network. So scanning tokens works. 
As for the issue with the disappearing qr code i was using the wallet on chrome on a windows laptop. 
I will retest it again.

CedrikNikita:
> Also sometimes the qr code disappears and reappears after a few seconds 

This PR is not changing qr code generation, and this file in general.

csiarab:
In that case Lgtm. i can scan the qr codes.

paolomolo:
LGTM as well

PR #2090, opened 2023-05-29
[SW-886] Chore/migrate tests

martinkaintas:
Sorry for the big PR but changes are dependent on each other
🤞 hopefully this will be the last big PR of the migration 🤞


src/store/getters.js
```diff
@@ -44,18 +44,25 @@ export default {
     return accounts[activeIdx] || {}; // TODO: Return null
   },
   networks({ userNetworks }) {
+    if (!userNetworks) return {};
```
CedrikNikita, line 47: That is incorrect.

src/store/index.js
```diff
@@ -20,6 +20,11 @@ import { defaultNetwork, NODE_STATUS_CONNECTING } from '../popup/utils/constants
 
 export default new Vuex.Store({
   state: {
+    popupProps: {
```
CedrikNikita, line 23: @peronczyk will be against adding something in vuex.

tests/e2e/integration/aex2-popups.js
```diff
@@ -55,7 +55,7 @@ describe('Tests cases for AEX-2 popups', () => {
     popups.forEach((popup) => {
       cy.openAex2Popup(popup, popup === 'sign' && SCHEMA.TX_TYPE.spend)
         .get('[data-cy=deny]')
-        .click()
+        .click({ force: true })
```
CedrikNikita, line 58: Why `{ force: true }` is needed?

tests/e2e/integration/aex2-popups.js
```diff
@@ -65,7 +65,7 @@ describe('Tests cases for AEX-2 popups', () => {
     popups.forEach((popup) => {
       cy.openAex2Popup(popup, popup === 'sign' && SCHEMA.TX_TYPE.spend)
         .get('[data-cy=accept]')
-        .click()
+        .click({ force: true })
```
CedrikNikita, line 68: IDEM

jest.config.js
```diff
@@ -16,14 +16,18 @@ module.exports = {
   moduleNameMapper: {
     '^.*\\.svg\\?vue-component$': '<rootDir>/config/jest/EmptySvg.vue',
     '^.*\\.svg\\?skip-optimize$': '<rootDir>/config/jest/EmptySvg.vue',
+    '\\.(css|scss)$': 'identity-obj-proxy',
+
```
CedrikNikita, line 20:
```suggestion
```

src/popup/components/Modals/ConfirmRawSign.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <Modal
+    show
```
CedrikNikita, line 3: How was it working before without `show`?

tests/e2e/integration/auth.js
```diff
@@ -5,15 +5,13 @@ describe('Test cases for login functionality', () => {
       .should('be.visible')
 
       .shouldRedirect('/', '/account')
-      .shouldRedirect('/import-account', '/account')
-      .shouldRedirect('/more/about/terms', '/more/about/terms')
+      .shouldRedirect('#/more/about/terms', '/more/about/terms')
```
CedrikNikita, line 8: Why have we returned to the `#` infront of the the route? Are the tests run on the web?

src/store/plugins/pendingTransactionHandler.js
```diff
@@ -38,7 +38,7 @@ export default async (store) => {
     }
   };
 
-  await watchUntilTruthy(() => store.getters['sdkPlugin/sdk']);
+  await watchUntilTruthy(() => store.state.sdkPlugin.ready);
```
CedrikNikita, line 41: Why ths change is necessary?

src/store/getters.js
```diff
@@ -44,18 +44,25 @@ export default {
...
     return [
       NETWORK_MAINNET,
       NETWORK_TESTNET,
       ...userNetworks.map((network, index) => ({ index, ...network })),
     ].reduce((acc, n) => ({ ...acc, [n.name]: n }), {});
   },
-  activeNetwork({ current: { network } }, { networks }) {
+  activeNetwork(networkRef, { networks }) {
+    if (!networkRef) return;
```
CedrikNikita, line 55: Why are these changes needed?

src/manifest.json
```diff
@@ -2,6 +2,7 @@
   "name": "Superhero",
   "description": "Superhero Wallet",
   "manifest_version": 2,
+  "content_security_policy": "default-src 'self' 'unsafe-inline'; script-src 'self' 'sha256-BCCVwoZgDUk/cw+Ln2/t3z9bm8qy+sICaZP5BWQIyUg='; connect-src *; font-src * data:; img-src * data:; style-src-elem *",
```
CedrikNikita, line 5:
```
Refused to execute inline script because it violates the following 
Content Security Policy directive: "script-src 'self' 'sha256-BCCVwoZgDUk/cw+Ln2/t3z9bm8qy+sICaZP5BWQIyUg='". Either the 'unsafe-inline' keyword, a hash ('sha256-LDP+OjgDpfArTDlfhpqHPU5mcov+qbvbnr657nRMBZY='), or a nonce ('nonce-...') is required to enable inline execution
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following 
Content Security Policy directive: "script-src 'self' 'sha256-BCCVwoZgDUk/cw+Ln2/t3z9bm8qy+sICaZP5BWQIyUg='".
```

src/popup/components/Modals/ConfirmRawSign.vue
martinkaintas: `show` was introduced in `feature/migrate-to-vue3` when we removed the Vue migration build. In Vue3 when `transition` is used as root it wont trigger animations the way we were using it [[see](https://v3-migration.vuejs.org/breaking-changes/transition-as-root.html)]. 
`show` was set to true through `modals.ts/openModal()` but some modals can be accessed directly (without being opened from `openModal()`) so they need the `show` prop otherwise they don't show up

src/store/plugins/pendingTransactionHandler.js
martinkaintas: In this case `await watchUntilTruthy(() => store.getters['sdkPlugin/sdk'])` never gets resolved

tests/e2e/integration/auth.js
martinkaintas: `vue/router` in develop is initialized with `mode: IS_WEB ? 'history' : 'hash',`
after migrating to Vue3 it is initialized with `history: IS_WEB ? createWebHistory() : createWebHashHistory(),`
So if `IS_WEB` is not set we are using hash history. When running tests we are not specifying `IS_WEB` but if we do then the `#` is not required anymore

martinkaintas: Should I add `PLATFORM=web` when running e2e tests and remove the `#`?

src/popup/router/index.ts
```diff
@@ -69,7 +69,9 @@ router.beforeEach(async (to, from, next) => {
     }[POPUP_TYPE];
 
     if (name !== to.name) {
-      next({ name, params: await getPopupProps() as Dictionary });
+      const params = await getPopupProps() as Dictionary;
```
CedrikNikita, line 72: All this params related changes can be removed after router mode will be set to `history`.

src/manifest.json
martinkaintas: Are you executing `npm run build:extension` or `npm run watch`? When building for development webpack is using `eval()` which is not allowed
`npm run build:extension` should work

tests/e2e/integration/aex2-popups.js
martinkaintas: Cypress is claiming that the buttons are hidden, but they are visible. By using `{force: true }` we are telling cypress not to check for this.

src/popup/router/index.ts
```diff
@@ -69,7 +69,9 @@ router.beforeEach(async (to, from, next) => {
     }[POPUP_TYPE];
 
     if (name !== to.name) {
-      next({ name, params: await getPopupProps() as Dictionary });
```
CedrikNikita, line 72: After that the saved `popupProps` would not be removed from the state, and will be `v-bind` to every mounted component which is bad.
Previous approach stoped working becuase of that https://github.com/vuejs/router/blob/main/packages/router/CHANGELOG.md#414-2022-08-22
We have several options:
1) We can use one of the approaches the `vue-router` team suggesting.
2) We can write a composable that will load this info and use it for every modal that can be accesed in popup.
3) We can use `JSON.stringify` to `app` object and then `JSON.parse` it back in the popup.

Also it means that every place in the project where prop is not defined in the query with new vue-router it is not working anymore.

CedrikNikita (changes requested):

CedrikNikita (commented):
There are `app` inside `app` in this branch (and i guess the migration itself).
image
image

tests/e2e/integration/aex2-popups.js
CedrikNikita: I know what for the `force` option for. The question is what is covering/why is force needed.

martinkaintas: This is related to the duplicate `#app` issue. After fixing that it now works without the `force` option

src/popup/components/Modals/ConfirmRawSign.vue
```diff
@@ -57,12 +58,14 @@
 
 <script>
```
CedrikNikita, line 59:
```suggestion
<script lang="ts">
```

CedrikNikita (changes requested):
Currently both `cancel` and `confirm` buttons doesnt trigger anything in the web iframe popups.

.github/workflows/main.yml
```diff
@@ -16,6 +16,8 @@ jobs:
         with:
           path: ~/.npm
           key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
+      - name: Upgrade NPM
+        run: npm install -g npm
```
peronczyk, line 20: Why the pipeline needs to update the NPM each time? Can you add more light to this?

src/composables/popupProps.ts
```diff
@@ -0,0 +1,23 @@
...
+
+interface IPopupPropsState {
+  popupProps: Ref<IPopupProps | null>
+  // eslint-disable-next-line no-unused-vars
+  setPopupProps: (props: IPopupProps) => void
+}
+
+const popupProps = ref<IPopupProps | null>(null);
+
+export function usePopupProps(): IPopupPropsState {
```
peronczyk, line 14: I suggest to add a comment that would describe what are the popup props.

src/composables/transactionTx.ts
```diff
@@ -115,8 +115,8 @@ export function useTransactionTx({
   const direction = computed(() => innerTx.value?.function === TX_FUNCTIONS.claim
     ? TX_DIRECTION.received
     : getTxDirection.value(
-        outerTx.value?.payerId ? outerTx.value : innerTx.value,
-        externalAddress
+      outerTx.value?.payerId ? outerTx.value : innerTx.value,
+      externalAddress
```
peronczyk, line 119: In my opinion the previous indentation was more accurate.

.github/workflows/main.yml
davidyuk: Can you update node version instead? node@14 gets installed with npm@6, but node@18 with npm@9 👍 
https://nodejs.org/ro/download/releases

src/popup/locales/en.json
```diff
@@ -116,7 +116,7 @@
     },
     "payloadInfo": {
       "title": "Payload message",
-      "msg": "A payload is an optional public message composed by the sender and sent to the recipient along with a blockchain transaction.<br /><br />Payload can be utilized as additional transaction identifier, destination tag (required by some centralized exchanges when depositing funds), or just to provide any extra information in free-form text."
+      "msg": "A payload is an optional public message composed by the sender and sent to the recipient along with a blockchain transaction.{'<'}br /{'>'}{'<'}br /{'>'}Payload can be utilized as additional transaction identifier, destination tag (required by some centralized exchanges when depositing funds), or just to provide any extra information in free-form text."
```
peronczyk, line 119: Why the square brackets are wrapped now? This makes the code to look really bad. Maybe there is a different solution for this?

src/popup/pages/Popups/Connect.vue
```diff
@@ -1,12 +1,13 @@
...
   <Modal
+    show
     full-screen
     class="connect"
     data-cy="popup-aex2"
   >
     <TransactionInfo
       :custom-title="$t('pages.connectConfirm.title')"
-      :sender="{ name: appName, address: app.host, url: app.url }"
+      :sender="{ name: appName, address: app?.host, url: app?.url }"
```
peronczyk, line 10: I suggest to move the data passed to the prop to the computed property and set the return type to enforce the proper structure.

src/popup/pages/Popups/Connect.vue
```diff
@@ -119,20 +114,21 @@ export default defineComponent({
...
+        ...popupProps.value?.app,
         ...permission.value,
       });
-      props.resolve();
+      popupProps.value?.resolve();
     }
 
     function cancel() {
-      props.reject(new Error('Rejected by user'));
+      popupProps.value?.reject(new Error('Rejected by user'));
```
peronczyk, line 124: Please use `RejectedByUserError` instead.

src/popup/pages/Popups/Connect.vue
```diff
@@ -119,20 +114,21 @@ export default defineComponent({
...
     }
 
     return {
       POPUP_CONNECT_ADDRESS_PERMISSION,
       POPUP_CONNECT_TRANSACTIONS_PERMISSION,
       appName,
+      app: popupProps.value?.app,
```
peronczyk, line 131: Please avoid setting values in the return block.

src/popup/pages/Popups/MessageSign.vue
```diff
@@ -1,29 +1,34 @@
...
     data-cy="popup-aex2"
   >
     <TransactionInfo
       :custom-title="$t('pages.popupMessageSign.title')"
-      :sender="{ name: app.name, address: app.host, url: app.url }"
+      :sender="{
+        name: popupProps?.app?.name,
+        address: popupProps?.app?.host,
+        url: popupProps?.app?.url
+      }"
```
peronczyk, line 14: I suggest to move this into composable and set the correct type for the return value to enforce correct data structure.

src/popup/pages/Popups/MessageSign.vue
```diff
@@ -90,12 +89,13 @@ export default defineComponent({
     }));
 
     function cancel() {
-      props.reject(new Error('Rejected by user'));
+      popupProps.value?.reject(new Error('Rejected by user'));
```
peronczyk, line 92: Please use `RejectedByUserError` instead. Probably you didn't rebased your branch for a while - on current develop this was already changed.

src/popup/router/index.ts
```diff
@@ -69,7 +70,9 @@ router.beforeEach(async (to, from, next) => {
...
+      const params = await getPopupProps() as Dictionary;
```
peronczyk, line 73:
```suggestion
      const popupProps = await getPopupProps() as Dictionary;
```

src/store/plugins/pendingTransactionHandler.js
peronczyk: Instead I suggest to use the composable `const { getSdk } = useSdk()` and to ensure the sdk is accessible run `await getSdk()`.

src/types/index.ts
```diff
@@ -645,8 +645,8 @@ export interface IKeyPair {
 }
 
 export interface ILabel {
-  text: string | LocaleMessages | TranslateResult,
-  customPending?: string | LocaleMessages | TranslateResult
+  text: string | TranslateResult,
+  customPending?: string | TranslateResult
```
peronczyk, line 649: Please use semicolons at the end of the lines.

tests/e2e/support/commands.js
```diff
@@ -11,7 +11,7 @@ import {
 } from '../../../src/popup/utils';
 
 Cypress.Commands.add('openPopup', (onBeforeLoad, route) => {
-  cy.visit(`${route ? `#${route}` : ''}`, { onBeforeLoad });
+  cy.visit(`${route || ''}`, { onBeforeLoad });
```
peronczyk, line 14:
```suggestion
  cy.visit(route || '', { onBeforeLoad });
```

tests/unit/snapshot.spec.js
```diff
@@ -1,52 +1,73 @@
...
 import Index from '../../src/popup/pages/Index.vue';
 import About from '../../src/popup/pages/About.vue';
 import TermsOfService from '../../src/popup/pages/TermsOfService.vue';
 import PrivacyPolicy from '../../src/popup/pages/PrivacyPolicy.vue';
+import * as environment from '../../src/lib/environment';
 
-Object.assign(Vue.prototype, {
-  $t: () => 'locale-specific-text',
+const OLD_ENV = process.env;
+beforeEach(() => {
```
peronczyk, line 9: Please put empty line before this one :)

peronczyk (changes requested):

src/store/plugins/pendingTransactionHandler.js
martinkaintas: This is also not working 🥲. It never gets resolved

src/popup/locales/en.json
martinkaintas: when using `@intlify/vue-i18n-loader` html is not allowed in translation texts. When trying to build the extension we will get an error

src/store/plugins/pendingTransactionHandler.js
peronczyk: Any errors when doing `getSdk()`? It works perfectly in every other place.

martinkaintas: No errors, the promise just doesn't resolve

peronczyk (changes requested):

.github/workflows/main.yml
martinkaintas: This has been fixed. Some extra packages are required to be installed in order to use node v18 (v16 as well)

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -308,23 +305,25 @@ export default defineComponent({
     }
 
     function cancel() {
-      props.reject(new Error('Rejected by user'));
+      popupProps.value?.reject(new Error('Rejected by user'));
```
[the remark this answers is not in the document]
martinkaintas: Errors related to rebase will be fixed in a separate PR. This error type does not exist in this branch

src/popup/components/Modals/ConfirmRawSign.vue
```diff
@@ -73,35 +79,47 @@ export default {
...
+  setup() {
+    const { popupProps } = usePopupProps();
+    const { t } = useI18n();
+    const store = useStore();
+
+    const dataAsString = computed(() => {
+      if (typeof popupProps.value?.data === 'string') return popupProps.value?.data;
+      return Buffer.from(popupProps.value?.data as any).toString('hex');
+    });
+    const sender: ComputedRef<IAccountOverView> = computed(() => ({
```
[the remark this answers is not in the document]
martinkaintas: With all the refactoring I've forgotten how to write code 🫠

src/store/plugins/pendingTransactionHandler.js
peronczyk: OK, for now let's go with your approach. Soon we will remove Vuex.

peronczyk (approved):

.github/workflows/main.yml
CedrikNikita: _You have jsdom (jest) dependent on an old version of canvas that tries to build under nodejs@18. Current fix is a workaround, it's better to update dependencies._
originally mentioned by @davidyuk

martinkaintas: `@vue/cli-plugin-unit-jest` doesn't support jest@28/29.  [see](https://github.com/vuejs/vue-cli/issues/7129)

davidyuk: Can you update jsdom only? The `canvas` package is only needed to emulate it for tests

martinkaintas: Updating `canvas` fixes the issue

davidyuk: Maybe `canvas` is not built for the env you have in gh actions 🤷‍♀️

martinkaintas: Turns out `npm ci` doesn't run locally either. So the issue still remains.
Yes, canvas is trying to build it self because it doesn't a prebuilt version, but then it fails. The packages I had added solved this.
This is also an official fix [[see](https://github.com/Automattic/node-canvas#compiling)]
I think adding the packages is the best option here

CedrikNikita (approved):

PR #2091, opened 2023-05-30
fix: swiper slide arrow navigation & seed-phrase badges

martinkaintas:


src/popup/components/AccountSwiperSlide.vue
```diff
@@ -12,15 +12,15 @@
       <BtnPlain
         v-if="idx !== 0"
         class="swiper-button prev"
-        @click="$emit('slide', idx - 1)"
+        @click.prevent="$emit('slide', idx - 1)"
```
martinkaintas, line 15: `.prevent` is used here so that parent button isn't triggered as well and the user doesn't get navigated to account-details

ZGiovas (approved):
LGTM after @pstachel 's recommended changes

pstachel (approved):

PR #2093, opened 2023-05-30
fix: duplicate app

martinkaintas:
see https://v3-migration.vuejs.org/breaking-changes/mount-changes.html

make outer `#app` act in the same way as body & change inner `#app` to `.app-wrapper`

### DIFF
--- a/src/styles/global.scss
+++ b/src/styles/global.scss
@@ -15,11 +15,13 @@
 }
 
 html,
-body {
+body,
+#app {
   height: var(--height);
 }
 
-body {
+body,
+#app {
   margin: 0;
   background-color: $color-black;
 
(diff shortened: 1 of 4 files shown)


src/styles/global.scss
```diff
@@ -1,7 +1,7 @@
-@use './typography';
-@use './variables' as *;
-@use './mixins';
-@use './transitions';
+@use "./typography";
+@use "./variables" as *;
+@use "./mixins";
+@use "./transitions";
```
pstachel, line 4: Single quotes should be used here

ZGiovas (approved):
LGTM  and seems that it fixes at least 2 minor content-height related issues on mobile view.

PR #2095, opened 2023-05-31
refactor(middleware): use v2 endpoints

### LINKED ISSUE #1431: Adjust Middleware interface to V2

CedrikNikita:
fixes #1431


src/store/plugins/names.js
```diff
@@ -105,19 +105,20 @@ export default (store) => {
...
             getPendingNameClaimTransactions(address),
-            middleware.getNamesOwnedBy(address)
-              .then(({ active }) => active.map(({ info, name, hash }) => ({
-                createdAtHeight: info.activeFrom,
-                expiresAt: info.expireHeight,
-                owner: info.ownership.current,
-                pointers: info.pointers,
-                autoExtend: !!(
+            middleware.getNames({ owned_by: address })
+              .then(({ data }) => data.filter(({ active }) => active)
```
peronczyk, line 109: I suggest to move the `.filter` to the next line to make the code more readable.

src/store/plugins/names.js
```diff
@@ -105,19 +105,20 @@ export default (store) => {
...
+                .map(({ info, name, hash }) => ({
+                  createdAtHeight: info.activeFrom,
+                  expiresAt: info.expireHeight,
+                  owner: info.ownership.current,
+                  pointers: info.pointers,
+                  autoExtend: !!(
```
peronczyk, line 115: Next 2 lines should also be indented.

src/composables/middleware.ts
peronczyk, line 1: I suggest to update the `fetchFromMiddleware` to use the `/v2` path by default. Now sometimes running this function requires using the `v2/` and sometimes no.

src/composables/middleware.ts
```diff
@@ -64,21 +64,17 @@ export function useMiddleware({ store }: IDefaultComposableOptions) {
               required: true,
               type: 'integer',
             },
+            {
+              in: 'query',
+              name: 'direction',
+              default: 'backward',
+              type: 'integer',
```
peronczyk, line 71: Are you sure this should be an integer?

peronczyk (changes requested):

src/composables/middleware.ts
CedrikNikita: The `next` property including `/v2` in the route, and at this point we would need to make the logic of fetching all pages more complicated.

Liubov-crypto:
I can't see all my registered chain names in my account. On dev branch I can see them all.

src/store/plugins/names.js
```diff
@@ -105,19 +105,22 @@ export default (store) => {
...
+            middleware.getNames({ owned_by: address })
+              .then(({ data }) => data
+                .filter(({ active }) => active)
+                .map(({ info, name, hash }) => ({
+                  createdAtHeight: info.activeFrom,
+                  expiresAt: info.expireHeight,
+                  owner: info.ownership.current,
+                  pointers: info.pointers,
+                  autoExtend:
+                  !!(
```
peronczyk, line 117: I don't see a reason to move this into new line. I think the way how it was before was OK.

peronczyk (approved):

Liubov-crypto:
The issue with names has been fixed. LGTM.

Liubov-crypto:
Retested, LGTM

konradodo (approved):

PR #2096, opened 2023-05-31
fix: filters and searchbar in transactions

ZGiovas:
Refactor usage of instance.ref so it will update to latest value


src/popup/components/Modals/QrCodeReader.vue
```diff
@@ -173,7 +173,7 @@ export default {
     },
     async stopReading() {
       if (this.mobile) {
-        ['body', '#app'].forEach((s) => {
+        ['body', '#app', '.app-wrapper'].forEach((s) => {
```
konradodo, line 176: like above

martinkaintas (approved):
LGTM

konradodo (approved):

PR #2099, opened 2023-05-31
feat: utils for generate url with query params

konradodo:


src/popup/utils/utils.ts
```diff
@@ -336,6 +336,10 @@ export const calculateNameClaimFee = (name: string): BigNumber => calculateFee(
   },
 );
 
+export function urlWithParams(url: string, params: Record<string, any>) {
+  return `${url}?${new URLSearchParams(JSON.parse(JSON.stringify(params)))}`;
```
peronczyk, line 340: 1. Why parse & stringify? I don't see a reason for deep merge.
2. This piece of code assumes that the URL has no query string. In my opinion more safe would be to allow passing any URL and manipulate it. Here's a untested proposition:

```suggestion
  const urlObj = new URL(url);
  Object.keys(params).forEach((key) => urlObj.searchParams.set(key, params[key]));
  return urlObj.toString();
```

src/popup/utils/utils.ts
```diff
@@ -336,6 +336,10 @@ export const calculateNameClaimFee = (name: string): BigNumber => calculateFee(
   },
 );
 
+export function urlWithParams(url: string, params: Record<string, any>) {
```
peronczyk, line 339: Functions executes code so their names should have imperative form. Like `do...`, `create...`, `fetch...`. I suggest to rename this function to `addParamsToUrl`.

src/store/plugins/fungibleTokens.js
```diff
@@ -65,7 +66,9 @@ export default (store) => {
           try {
             if (isEmpty(availableTokens)) return;
             const tokens = await fetchAllPages(
-              () => fetchFromMiddleware(`/v2/aex9/account-balances/${address}?limit=100`),
+              () => fetchFromMiddleware(urlWithParams(`/v2/aex9/account-balances/${address}`, {
```
peronczyk, line 69: I don't think this change adds any benefit. Previous version was actually more readable.

src/popup/utils/utils.ts
peronczyk: Please skip  pt2 from my previous comment as it will throw error. It's not possible to use `URL` with relative paths. This makes me convinced that this util makes no sense for us. Whenever you want to create query string just do this:
```
const query = (new URLSearchParams({
  <the data>
})).toString();
const fullUrl = `${path}${query}`;
```
No need for util.

peronczyk (changes requested):

PR #2100, opened 2023-06-01
[SW-855, SW-1032] Remove "Transaction has been sent" modal and redirect user Latest tx widget

konradodo:


tests/e2e/integration/withdraw.js
```diff
@@ -60,18 +60,19 @@ describe('Test cases for Withdraw Page', () => {
...
-      .should('be.visible')
-      .click()
-      .get('[data-cy=spend-success]', { timeout: 10000 })
-      .should('be.visible')
-      .get('[data-cy=btn-close]')
-      .click()
-      .openTransactions()
-      .get('[data-cy=pending-txs]')
-      .should('be.visible');
+    // TODO change condition on this test
```
CedrikNikita, line 65: I think you should change the condition within this PR.

src/store/plugins/pendingTransactionHandler.js
```diff
@@ -1,36 +1,15 @@
...
-          showSpendModal();
-          break;
-        case 'spendToken':
-          store.dispatch('fungibleTokens/loadTokenBalances');
-          showSpendModal();
-          break;
-        default:
+    try {
+      if (type === 'spendToken') {
+        store.dispatch('fungibleTokens/loadTokenBalances');
```
CedrikNikita, line 12: I think we can remove this, since the balance of fungible token is not updating as fast as transaction mining, and we are currently fetching the fungible token balance every 10 seconds.

src/popup/components/OpenTransferSendModalButton.vue
```diff
@@ -3,7 +3,7 @@
     :text="isMultisig ? $t('dashboard.proposeCard.title') : $t('common.send')"
     :subtitle="subtitle"
     :icon="ArrowSendIcon"
-    :disabled="!isOnline || !isConnected || (!!pendingMultisigTransaction && isMultisig)"
+    :disabled="false && (!isOnline || !isConnected || (!!pendingMultisigTransaction && isMultisig))"
```
CedrikNikita, line 6: That's a mistake.

peronczyk, line 6: `false`? Looks like something temporary.

peronczyk (changes requested):

src/composables/latestTransactionList.ts
```diff
@@ -98,6 +121,16 @@ export function useLatestTransactionList({ store }: IDefaultComposableOptions) {
...
 
+  watch(
+    localPendingTransactionsList,
+    async (transactions) => {
+      if (transactions.length) {
+        await updateTransactionListData(transactions);
+      }
+    },
+    { deep: true, immediate: true },
+  );
```
pstachel, line 132: Do you think we should reload all of the transactions if there's one pending? What do you think about just pushing that one on top of the list?

pstachel (approved):

src/composables/latestTransactionList.ts
konradodo: Push could be resolved before fetching the list. We should have always the newest data, when we want "push" something by myself.

src/composables/latestTransactionList.ts
subhod-i, line 1: I would like to hear more about this proposal.

`LatestTransactionList` is not really a reusable code. We use this at just one component. This shouldn't be composable instead a composable which takes care of pending, fetching transactions is right.

The logic to compute the latest transaction list could be moved to the `LatestTransactionsCard.vue` file

src/composables/latestTransactionList.ts
```diff
@@ -13,7 +22,10 @@ import { createNetworkWatcher } from './composablesHelpers';
 import { useTransactionTx } from './transactionTx';
 
 const isTransactionListLoading = ref(false);
-const transactionList = ref<ITransaction[]>([]);
+const fetchedTransactions = ref<ITransaction[]>([]);
+const pendingTransactions = ref<IPendingTransaction[]>([]);
+const latestTransaction = ref<ITransaction[]>([]);
```
subhod-i, line 27:
```suggestion
const latestTransactions = ref<ITransaction[]>([]);
```

src/popup/components/LatestTransactionsCard.vue
```diff
@@ -38,22 +45,44 @@ export default defineComponent({
...
     );
+
     const showWidget = computed(() => (
       isOnline.value
-      && (isTransactionListLoading.value || transactionList.value.length)
+      && (isTransactionListLoading.value || latestTransaction.value.length)
     ));
 
+    function conditionalUpdatingState(param?: string) {
+      if (param && viewportElement.value && latestTransactionCard.value) {
```
subhod-i, line 68: A code comment on this condition would be helpful.

src/popup/components/TransferReview.vue
```diff
@@ -384,29 +387,30 @@ export default defineComponent({
...
           selectedAsset,
           note,
         });
+      } else {
+        const hash = await transfer({
+          amount,
+          recipient,
+          selectedAsset,
+        });
+        root.$router.push({ name: homeRouteName.value, query: { latestTransaction: hash } });
```
subhod-i, line 412:
```suggestion
        root.$router.push({ name: homeRouteName.value, query: { latestTxHash: hash } });
```

subhod-i (changes requested):

src/composables/latestTransactionList.ts
konradodo: This is question for @peronczyk

src/composables/latestTransactionList.ts
```diff
@@ -28,14 +40,40 @@ export function useLatestTransactionList({ store }: IDefaultComposableOptions) {
...
+        );
+      }
+    });
+
+    latestTransaction.value = uniqWith(transactions, (a, b) => (
+      a.hash === b.hash && a.transactionOwner === b.transactionOwner
+    ))
+      .sort(sortTransactionsByDateCallback)
+      .slice(0, DASHBOARD_TRANSACTION_LIMIT);
+  }
```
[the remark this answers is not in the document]
konradodo: This function not works correctly, i'd like will use my version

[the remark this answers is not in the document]
subhod-i: Sure

src/composables/latestTransactionList.ts
konradodo: i think this solution give as only the cashing data.

subhod-i: This is more to deal with the Composables definition. They cannot cache! They are meant for reusable code. @peronczyk can confirm this.

peronczyk: @subhod-i Konrad have kind of right. He used word "caching" which is not accurate, but in this place makes sense. So this composable holds **global state** of recent transactions. So whenever we go back to the dashboard this state is just displayed without the need of fetching it again every time. It's a major optimisation done some time ago.

src/composables/latestTransactionList.ts
```diff
@@ -28,14 +40,39 @@ export function useLatestTransactionList({ store }: IDefaultComposableOptions) {
...
   const tokens = computed(() => store.state.fungibleTokens.tokens);
 
+  const localPendingTransactionsList = computed(
+    (): IPendingTransaction[] => {
+      const { networkId } = store.getters.activeNetwork;
+      return (store.state.transactions as ITransactionsState).pending[networkId] || [];
+    },
+  );
+
+  function buildLatestTransaction() {
```
peronczyk, line 50: Plural:
```suggestion
  function buildLatestTransactions() {
```

src/composables/latestTransactionList.ts
```diff
@@ -28,14 +40,39 @@ export function useLatestTransactionList({ store }: IDefaultComposableOptions) {
...
+    const transactions = [...fetchedTransactions.value];
+    pendingTransactions.value.forEach((transaction) => {
+      if (!transactions.some(({ hash }) => transaction.hash === hash)) {
+        transactions.unshift(transaction as ITransaction);
```
peronczyk, line 54: `transaction` has different type and forcing TS to change it is in my opinion a bad thing. If we want to allow to put pending transactions into the collection of transactions we need to change how the `transactions` are typed.

src/composables/latestTransactionList.ts
```diff
@@ -28,14 +40,39 @@ export function useLatestTransactionList({ store }: IDefaultComposableOptions) {
...
 
+  const localPendingTransactionsList = computed(
+    (): IPendingTransaction[] => {
+      const { networkId } = store.getters.activeNetwork;
+      return (store.state.transactions as ITransactionsState).pending[networkId] || [];
+    },
+  );
+
+  function buildLatestTransaction() {
+    const transactions = [...fetchedTransactions.value];
```
peronczyk, line 51: Why we need to make a copy of the transactions here?

src/composables/latestTransactionList.ts
```diff
@@ -28,14 +40,39 @@ export function useLatestTransactionList({ store }: IDefaultComposableOptions) {
...
+    pendingTransactions.value.forEach((transaction) => {
+      if (!transactions.some(({ hash }) => transaction.hash === hash)) {
+        transactions.unshift(transaction as ITransaction);
+      } else {
+        pendingTransactions.value = pendingTransactions.value.filter(
```
peronczyk, line 56: This line is placed inside of a `forEach` loop which in my opinion is a bad pattern. I suggest to use more functional approach by utilising `map` & `filter` on higher level.

peronczyk (changes requested):

src/composables/latestTransactionList.ts
peronczyk, line 50: Maybe this function could be replaced with a composable? You already store all the transaction types in refs.

peronczyk (changes requested):

src/store/plugins/pendingTransactionHandler.js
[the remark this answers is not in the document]
CedrikNikita: The `poll` shouldn't be removed, we are checking here the status of the pending transactions. We still want to update the pending transaction status and remove it.

src/composables/latestTransactionList.ts
konradodo: I do not see a reason

src/composables/latestTransactionList.ts
subhod-i: Got it. Thanks.

subhod-i (approved):

src/composables/latestTransactionList.ts
```diff
@@ -25,17 +38,40 @@ const { onNetworkChange } = createNetworkWatcher();
...
+      .pending[nodeNetworkId.value!] || [],
+  );
+
+  function buildLatestTransactions() {
+    pendingTransactions.value = pendingTransactions.value.filter(({ hash }) => (
+      !fetchedTransactions.value.some((transaction) => transaction.hash === hash)
+    )) as IPendingTransaction[];
+
+    const transactions: ITransaction[] = [
+      ...pendingTransactions.value as ITransaction[],
```
peronczyk, line 56: `pendingTransactions` has the type of `IPendingTransaction` - I suggest to avoid forcing the type here as this is not true.

peronczyk (changes requested):

src/types/index.ts
```diff
@@ -388,17 +388,18 @@ export interface ITx {
...
+  VSN?: string;
 }
 
 export interface ITransaction {
-  blockHeight: number;
-  claim: any; // TODO find type
   hash: string;
+  pending: boolean;
+  microIndex?: number;
+  microTime?: number;
```
peronczyk, line 398: Please arrange the properties alphabetically.

src/lib/accounts/superhero.ts
```diff
@@ -13,6 +13,8 @@ import { POPUP_TYPE_MESSAGE_SIGN, POPUP_TYPE_SIGN, POPUP_TYPE_TX_SIGN } from '..
 import { showPopup } from '../../background/popupHandler';
 import type { IPopupType } from '../../types';
 
+// TODO Subhod please check it
+// @ts-ignore
```
peronczyk, line 17: Please ask Subhod directly to help with solving the problem. Leaving such a comments in the code is a bad thing because it suggests that we are not treating the code in a proper way.

PR #2104, opened 2023-06-07
[SW-944] refactor: use capacitor status-bar

ZGiovas:
Replace cordova status bar to the capacitor equivalent. 
Added scripts for android build with ionic and changed IS_CORDOVA param to IS_IONIC at all files.
Also fixed an lint issue on VeeValidate.js file


config.xml
```diff
@@ -108,8 +108,6 @@
     </platform>
     <preference name="SplashScreen" value="screen" />
     <preference name="ShowSplashScreenSpinner" value="false" />
-    <preference name="StatusBarOverlaysWebView" value="false" />
-    <preference name="StatusBarBackgroundColor" value="#141414" />
```
martinkaintas, line 112: Doesn't this need to be replaced in capacitor status-bar? How are these colors set in the new plugin?

ZGiovas: The equivalent initial configuration has been added at the initial commit of this Branch and more specifically at capacitor.config.ts folder lines 16,17

martinkaintas (approved):

package.json
```diff
@@ -13,6 +13,8 @@
     "lint": "cross-env NODE_ENV=production npm run lint:commit && npm run lint:vue && npm run lint:css",
     "build": "vue-cli-service build",
     "build:cordova": "cross-env PLATFORM=cordova vue-cli-service build",
+    "build:ionic": "cross-env PLATFORM=ionic && ionic build",
+    "build:ionic:android": "cross-env PLATFORM=ionic && ionic build && ionic capacitor add android && ionic capacitor sync",
```
pstachel, line 17: Shouldn't we consider adding a command for iOS build?

src/store/plugins/veeValidate.js
```diff
@@ -112,11 +112,9 @@ export default (store) => {
...
   defineRule('name_registered_address', (value) => checkAensName(value) && checkNameRegisteredAddress(value));
   defineRule('token_to_an_address',
-    (value, [isToken]) => (
-      !checkAensName(value)
-      || (checkAensName(value) && !isToken)
-    ),
-    { params: ['isToken'] });
+    (value, [isToken]) => !checkAensName(value) || (checkAensName(value) && !isToken), {
+      params: ['isToken'],
+    });
```
pstachel, line 117: I feel that the previous formatting of those lines was clearer

package.json
ZGiovas: We will change all ionic run/build commands after we replace all the plugins. Right now I just added these two for dev reasons.

src/store/plugins/veeValidate.js
ZGiovas: Actually that change was done by npm run lint:vue:fix because error on serve/build existed.

pstachel (approved):

PR #2107, opened 2023-06-08
[SW-901] Add popup for aepps requesting all account addreseses

subhod-i:
You need an aepp requesting all account addresses to test this. I have deployed one.
1. use the extension from this artifact
2. visit aepp https://subhod-i.github.io/ 
3. click 'Connect', First aepp will request for typical connection popup and then the second popup is where it is requesting all the account addresses. See if it is according to the design and test by toggling the new permission settings.

### DIFF
--- a/src/popup/locales/en.json
+++ b/src/popup/locales/en.json
@@ -33,6 +33,7 @@
     "send": "Send",
     "smartContract": "Smart contract",
     "swap": "Swap",
+    "title": "Superhero wallet",
     "total": "Total",
     "totalMultisig": "Total in multisig vaults",
     "tx": "Tx",
@@ -558,9 +559,14 @@
       "remaining-time": "Remaining Time",
       "ending-height": "Ending height"
     },
+    "accountListConfirm": {
+      "websiteRequestConnect": "would like to connect to your Superhero wallet.",
+      "addressesRequest": "This aepp is requesting access to view all of your account addresses.",
+      "message": "Make sure you trust the aepp provider before accepting the request."
+    },
     "connectConfirm": {
       "title": "Allow access",
-      "websiteRequestconnect": "would like to connect to your account.",
+      "websiteRequestConnect": "would like to connect to your account.",
       "websiteRequest": "This website is requesting access to view your current account address. Always make sure you trust the website you interact with. ",
       "cancelButton": "Deny",
       "confirmButton": "Accept",
@@ -1086,6 +1092,7 @@
     "permissions": {
       "login": "Don't ask for login requests",
       "message-sign": "Auto-sign messages",
+      "addressList": "Don't ask for list of all addresses",
       "daily-spending-limit": "Daily spending limit without asking",
       "transaction-sign": "Limit for spending without asking per day",
       "spent-today": "Spent today",
--- a/src/popup/pages/Popups/AccountList.vue
+++ b/src/popup/pages/Popups/AccountList.vue
@@ -2,13 +2,13 @@
   <Modal
     show
     full-screen
-    class="connect"
+    class="account-list"
     data-cy="popup-aex2"
   >
     <TransactionInfo
       :custom-title="$t('pages.connectConfirm.title')"
       :sender="sender"
-      :recipient="activeAccountExtended"
+      :recipient="recipient"
     />
 
     <div
@@ -16,7 +16,7 @@
       data-cy="aepp"
     >
       <span class="app-name">{{ sender.name }}</span>
-      ({{ sender.address }}) {{ $t('pages.connectConfirm.websiteRequestConnect') }}
+      ({{ sender.address }}) {{ $t('pages.accountListConfirm.websiteRequestConnect') }}
     </div>
 
     <div class="permissions">
@@ -24,17 +24,14 @@
         <span class="title">
           <CheckMark class="icon" /> {{ $t('common.address') }}
         </span>
-        <span class="description">
-          {{ $t('pages.connectConfirm.addressRequest') }}
-        </span>
-      </template>
-      <template v-if="access.includes(POPUP_CONNECT_TRANSACTIONS_PERMISSION)">
-        <span class="title">
-          <CheckMark class="icon" /> {{ $t('pages.connectConfirm.transactionLabel') }}
-        </span>
-        <span class="description">
-          {{ $t('pages.connectConfirm.transactionRequest') }}
-        </span>
+        <div class="description">
+          <p>
+            {{ $t('pages.accountListConfirm.addressesRequest') }}
+          </p>
+          <p class="color-warning">
+            {{ $t('pages.accountListConfirm.message') }}
+          </p>
+        </div>
       </template>
     </div>
 
@@ -63,15 +60,14 @@ import {
   onUnmounted,
 } from 'vue';
 import { useStore } from 'vuex';
-import type { IPermission } from '../../../types';
+import type { IAccountOverview, IPermission } from '../../../types';
 import { RejectedByUserError } from '../../../lib/errors';
 import {
   PERMISSION_DEFAULTS,
   POPUP_CONNECT_ADDRESS_PERMISSION,
-  POPUP_CONNECT_TRANSACTIONS_PERMISSION,
 } from '../../utils';
 import { useState } from '../../../composables/vuex';
-import { useAccounts, usePopupProps } from '../../../composables';
+import { usePopupProps } from '../../../composables';
 
 import Modal from '../../components/Modal.vue';
 import BtnMain from '../../components/buttons/BtnMain.vue';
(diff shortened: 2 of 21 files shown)


subhod-i:
https://github.com/aeternity/superhero-wallet/assets/88760043/1e92f018-293b-469f-bc7c-6017b3107060

Demo

src/popup/pages/Popups/AccountList.vue
```diff
@@ -0,0 +1,175 @@
...
+      root.$store.commit('permissions/addPermission', {
+        ...PERMISSION_DEFAULTS,
+        ...props.app,
+        ...permission.value,
+      });
+      props.resolve();
+    }
+
+    function cancel() {
+      props.reject(new Error('Rejected by user'));
```
peronczyk, line 105: I suggest to use `RejectedByUserError` taken from `errors.ts` file.

src/popup/pages/Popups/AccountList.vue
```diff
@@ -0,0 +1,175 @@
+<template>
+  <Modal
+    full-screen
+    class="connect"
```
peronczyk, line 4: The root class should be the same as the component name.

src/popup/pages/Popups/AccountList.vue
```diff
@@ -0,0 +1,175 @@
...
+
+      display: block;
+      padding-bottom: 6px;
+      color: variables.$color-white;
+      text-align: left;
+    }
+
+    .message {
+      color: variables.$color-warning;
+    }
```
peronczyk, line 172: This can be removed and instead you can use `color-warning` CSS class (see `global.scss` file).

src/popup/pages/Popups/AccountList.vue
```diff
@@ -0,0 +1,175 @@
...
+      <template v-if="access.includes(POPUP_CONNECT_ADDRESS_PERMISSION)">
+        <span class="title">
+          <CheckMark class="icon" /> {{ $t('common.address') }}
+        </span>
+        <span class="description">
+          {{ $t('pages.accountListConfirm.addressesRequest') }}
+        </span>
+        <span class="description message">
+          {{ $t('pages.accountListConfirm.message') }}
+        </span>
```
peronczyk, line 30: I suggest to replace the spans with `<p>`. This way you wan't need to set `display: block` in the CSS definitions block.

src/popup/pages/Popups/AccountList.vue
```diff
@@ -0,0 +1,175 @@
...
+      confirm,
+      cancel,
+      recipient,
+    };
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@use '../../../styles/variables';
```
peronczyk, line 121:
```suggestion
@use '../../../styles/variables' as *;
```
This way there is no need to prefix all variables with `variables.`.

peronczyk (changes requested):

src/popup/pages/Popups/AccountList.vue
subhod-i: I have tried maintaining this structure as other popup files. Maintains readability.

src/popup/pages/Popups/AccountList.vue
subhod-i: This is not working for the popups. Popups don't include the global styles in my opinion

ifaouibadi (approved):

src/store/modules/permissions.js
```diff
@@ -78,6 +78,7 @@ export default {
         'connection.open': 'address',
         'address.subscribe': 'address',
         'message.sign': 'messageSign',
+        'address.get': 'addressList',
```
davidyuk, line 81: Maybe combine it with `address.subscribe`? It adapt allowed to cal `address.get` without confirmation then it can subscribe by polling.

src/popup/pages/Popups/AccountList.vue
```diff
@@ -0,0 +1,175 @@
...
+    <div
+      class="subtitle"
+      data-cy="aepp"
+    >
+      <span class="app-name">{{ appName }}</span>
+      ({{ app.host }}) {{ $t('pages.accountListConfirm.websiteRequestconnect') }}
+    </div>
+
+    <div class="permissions">
+      <template v-if="access.includes(POPUP_CONNECT_ADDRESS_PERMISSION)">
```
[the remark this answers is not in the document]
peronczyk: OK, but still in my opinion looks bad.

src/popup/pages/Popups/AccountList.vue
peronczyk: `global.scss` is merged with all other files. Why wouldn't it be not merged?

src/popup/pages/Popups/AccountList.vue
peronczyk, line 1: I see that this file is a copy of `Connect.vue` component. I suggest to avoid code duplication. Maybe creating a reusable template for both components would be better?

src/popup/pages/Popups/AccountList.vue
peronczyk: > I have tried maintaining this structure as other popup files
You mean you just copy+pasted the code from `Connect.vue`? :) I don't think this is a good way of solving the problem of code duplication?

src/popup/pages/Popups/AccountList.vue
peronczyk: Why you resolved this comment without reasoning why you won't do anything?

src/popup/locales/en.json
```diff
@@ -568,6 +569,11 @@
       "remaining-time": "Remaining Time",
       "ending-height": "Ending height"
     },
+    "accountListConfirm": {
+      "websiteRequestconnect": "would like to connect to your Superhero wallet.",
```
peronczyk, line 573:
```suggestion
      "websiteRequestConnect": "would like to connect to your Superhero wallet.",
```

src/popup/locales/en.json
```diff
@@ -1096,6 +1102,7 @@
     "permissions": {
       "login": "Don't ask for login requests",
       "message-sign": "Auto-sign messages",
+      "address-list": "Don't ask for address list requests",
```
peronczyk, line 1105: I suggest to use camelCase here.

src/popup/pages/Popups/AccountList.vue
```diff
@@ -0,0 +1,190 @@
...
+  setup() {
+    const store = useStore();
+
+    const { popupProps, sender, setPopupProps } = usePopupProps();
+
+    const permission = useState<IPermission>('permissions', popupProps.value?.app?.host);
+
+    const appName = computed(() => permission.value?.name || popupProps.value?.app?.name);
+
+    const recipient = {
```
peronczyk, line 99: I suggest to type this constant.

src/types/index.ts
```diff
@@ -170,6 +170,7 @@ export interface IAccountOverview extends Partial<Omit<IAccount, 'address'>> {
   contractCreate?: boolean;
   aens?: boolean;
   label?: TranslateResult;
+  wallet?: string;
```
peronczyk, line 173: Can you add a description for this property? Maybe some possible values also?

peronczyk (changes requested):

CharisSiarampalis:
@subhod-i When i have this option selecting the extension cannot connect to the aepp.
is this normal?

image

src/popup/pages/Popups/AccountList.vue
subhod-i: Yes, It's a duplicate of `Connect.vue`. Imagine supporting 3 popup types and having just 2 files under `/Popups` directory. Someone reading fresh would get confused. Either the solution is to combine all 3 into one or maintain 3 files. But will not maintain 2 files where one file just represents two popup types.

src/popup/pages/Popups/AccountList.vue
subhod-i: I guess it's better to keep the file for fine control as the team is leaning on this new approach. Also comment related to the same
https://github.com/aeternity/superhero-wallet/pull/2107#discussion_r1274806869

src/popup/pages/Popups/AccountList.vue
subhod-i: We had a discussion regarding this, and were able to fix this. Thanks @pstachel

src/types/index.ts
subhod-i: Added

subhod-i:
> @subhod-i When i have this option selecting the extension cannot connect to the aepp.
is this normal?

@CharisSiarampalis The usual connection should work irrespective of this new permission

src/store/modules/permissions.js
subhod-i: `askAddresses` is the `METHOD.address` request and it is `address.get`

```
  /**
   * Ask addresses from wallet
   * @returns Addresses from wallet
   */
  async askAddresses(): Promise<Encoded.AccountAddress[]> {
    this._ensureAccountAccess();
    return this.rpcClient.request(METHODS.address, undefined);
  }
  ```

src/popup/pages/Popups/AccountList.vue
peronczyk: I don't like code duplication, but if you insist on leaving it as it is I'm OK with it. Maybe we'll refactor this in the future.

peronczyk (approved):

src/store/modules/permissions.js
davidyuk: Also you can implement `address.get` to return a single address, I'm not sure that it is defined by protocol.

davidyuk (approved):

src/store/modules/permissions.js
subhod-i: I see, @davidyuk wouldn't it conflict with the all-account address list then?

> Also you can implement address.get to return a single address
I don't see a callback method like `onAskAccounts` for single address requests. Maybe something to consider for the Aepp communication improvement epic.

PR #2108, opened 2023-06-08
refactor: replace cordova clipboard plugin

ZGiovas:
Replace cordova clipboard plugin with the capacitor built-in one.


src/popup/utils/utils.ts
```diff
@@ -597,22 +598,23 @@ export async function readValueFromClipboard(): Promise<string | undefined> {
   if (!process.env.UNFINISHED_FEATURES) {
     return undefined;
   }
-  let value = '';
+  let text = '';
 
   if (IS_CORDOVA) {
-    value = await new Promise((...args) => window.cordova!.plugins!.clipboard.paste(...args));
+    const { type, value } = await Clipboard.read();
+    if (type === 'string') { text = value; }
```
martinkaintas, line 605:
```suggestion
    if (type === 'string') { 
      text = value;
    }
```

Otherwise brackets aren't really useful ^^

martinkaintas (changes requested):

src/popup/utils/utils.ts
ZGiovas: You are right. PR Updated

martinkaintas (approved):
🤝

pstachel (approved):

PR #2109, opened 2023-06-08
[SW-945] refactor: replace cordova share plugin with ionic share

martinkaintas:
The Share API works on iOS, Android, and the Web (using the new [Web Share API](https://web.dev/web-share/))
So `navigator.share` is no longer needed (as it is used by capacitor if the user is on web)


src/store/actions.js
```diff
@@ -192,12 +193,15 @@ export default {
...
+   * @param {{text: String}} options
+   */
   async share(_, options) {
-    await (process.env.IS_CORDOVA
-      ? new Promise((resolve) => window.plugins.socialsharing.shareW3C(
-        options,
-        ({ app }) => app && resolve(),
-      ))
-      : navigator.share(options));
+    await Share.share({
```
CedrikNikita, line 201: Why are you always trying to share with the ionic plugin first?

martinkaintas: You are correct 👀

ZGiovas (approved):
LG2M

src/store/actions.js
```diff
@@ -192,12 +193,18 @@ export default {
...
+   */
   async share(_, options) {
-    await (process.env.IS_IONIC
-      ? new Promise((resolve) => window.plugins.socialsharing.shareW3C(
-        options,
-        ({ app }) => app && resolve(),
-      ))
-      : navigator.share(options));
+    // The Share API works on iOS, Android, and the Web
+    const canShare = (await Share.canShare()).value;
```
pstachel, line 202: Are there any fallback options in case we can't share?

martinkaintas: I didn't implement any fallback options as there weren't any originally. 
This implementation should work everywhere where the previous implementation worked, so if there was no need for a fallback option then, then I believe there is no need one now. 
If you have any ideas we could add them though.

PR #2114, opened 2023-06-09
[SW-852] Update design of Import AirGap account modal

konradodo:


src/popup/components/AccountSelectOptionsItem.vue
pstachel, line 1: Do you think we still need that component? Maybe the balance could be handled by the parent (as it's an only extension of AccountInfoCard)?

pstachel (approved):

src/popup/components/AccountInfoCard.vue
```diff
@@ -50,25 +49,25 @@ export default defineComponent({
...
+    address: { type: String, required: true },
+    name: { type: String, default: '' },
+    idx: { type: Number, default: 0 },
+    balance: { type: Number, default: 0 },
+    isSelected: Boolean,
+    isAirGapAccount: Boolean,
   },
   setup(props, { root }) {
     const { getAccountBalance } = useBalances({ store: root.$store });
+    const bgColorStyle = computed(() => ({ '--bg-color': getAddressColor(props.address) }));
```
peronczyk, line 61: In my opinion this is bad to return CSS/HTML code within the JS logic. I suggest to put the object directly into template.

peronczyk (approved):

PR #2116, opened 2023-06-09
[SW-958] Use one InputAmount component

peronczyk:


src/popup/components/InputAmount.vue
peronczyk, line 1: Actually the InputAmount component was removed and the `InputAmountV2` was renamed to `InputAmount`. But GIT displays this change as the modification of the `InputAmount`. So please skip this file while reviewing the PR because I did no changes to the `InputAmountV2` file.

pstachel (approved):

src/popup/components/InviteItem.vue
```diff
@@ -82,85 +98,120 @@ export default {
...
+      const sdk = await getSdk();
       try {
-        if (this.topUpAmount > 0) {
-          await this.sdk.spend(this.topUpAmount, this.address, {
-            payload: 'referral',
-            denomination: AmountFormatter.AE_AMOUNT_FORMATS.AE,
-          });
-          await this.updateBalance();
+        await sdk.spend(formModel.value.amount, address.value, {
+          payload: 'referral',
```
CedrikNikita, line 180: The `payload` is being set, and since the `formModel` doesn't have a `payload` set the `max` and `fee` is being calculated wrongly. The defined `payload` here is not doing anything except wasting money during this transfer. It is not defined in the sdk. It should be removed. Same in the `Invite` page.

CedrikNikita (changes requested):

CedrikNikita (approved):

PR #2117, opened 2023-06-09
[SW-952, SW-957] The token asset automatically changes after scanning the recipient's address

pstachel:


src/popup/components/TransferSendForm.vue
```diff
@@ -371,7 +371,11 @@ export default defineComponent({
       payload,
       token,
     }: Dictionary) {
-      formModel.value.selectedAsset = availableTokens.value[token] || aeternityCoin.value;
+      if (token) {
+        formModel.value.selectedAsset = availableTokens.value[token] || aeternityCoin.value;
+      } else if (!formModel.value.selectedAsset) {
+        formModel.value.selectedAsset = aeternityCoin.value;
+      }
```
peronczyk, line 378: Maybe I'm missing something, but this `else` block makes no sense: there is already a check for the existence of the token, and if it's falsy the AE coin is assigned.

peronczyk (changes requested):

peronczyk (approved):

PR #2124, opened 2023-06-14
[SW-943] refactor: replace cordova qr scanner

ZGiovas:


martinkaintas (approved):

src/popup/utils/utils.ts
```diff
@@ -600,7 +600,7 @@ export async function readValueFromClipboard(): Promise<string | undefined> {
   }
   let text = '';
 
-  if (IS_CORDOVA) {
+  if (IS_IONIC) {
```
peronczyk, line 603: Maybe instead of using direct framework names like "cordova" or "ionic" we could have something more generic? Like `IS_MOBILE_APP`?

src/popup/components/Modals/QrCodeReader.vue
```diff
@@ -156,9 +157,9 @@ export default {
         ? new Promise((resolve, reject) => {
           this.setQrScanner(true);
           window.plugins.webviewcolor.change('#00FFFFFF');
-
-          window.QRScanner.scan((error, text) => (!error && text ? resolve(text) : reject(error)));
-          window.QRScanner.show();
+          BarcodeScanner.hideBackground();
+          BarcodeScanner.startScan((error, text) => (!error && text ? resolve(text)
+            : reject(error)));
```
peronczyk, line 162: Try to use this notation for one liners:
```
(something)
  ? firstResult
  : secondResult;
```
So this becomes:
```suggestion
          BarcodeScanner.startScan((error, text) => (!error && text)
            ? resolve(text)
            : reject(error));
```

peronczyk (approved):

PR #2129, opened 2023-06-16
[DRAFT] Multichain architecture design POC

peronczyk:


src/lib/networks/Network.ts
```diff
@@ -0,0 +1,124 @@
...
+  ]);
+
+  function getEmptyNetworkState(): INetworkState {
+    return {
+      isReady: ref(false),
+      isConnected: ref(false),
+      isError: ref(false),
+      accounts: ref([]),
+      activeAccountIdx: ref(0),
+      activeAccount: computed(() => ({} as IAccount)),
```
pstachel, line 41: Why network is storing the information about accounts?

peronczyk: Good question :) I know that this might look like the `Network` lib is holding this states, but it's not fully true.

It is only passing the states received by the adapter to the places that requires the access. So before having any network running we have a moment where the components should have access to the state (`Network.current.state`), the state should have the reactive props in place, and the values should be "empty".

`getEmptyNetworkState` creates this empty state, which is filled in (replaced) by the `setNetworkState` at the moment of network adapter initialization.

Is this clear? If not I think we could try to update this/

PR #2132, opened 2023-06-19
[SW-987] perf: use pre-generated ACIs instead of source code

CedrikNikita:


Liubov-crypto:
LGTM

src/lib/contracts/FungibleTokenFullInterfaceACI.json
subhod-i, line 1: Only this file is sufficient, the Full ACI is not needed I believe.

src/popup/components/TransactionOverview.vue
```diff
@@ -186,10 +188,10 @@ export default defineComponent({
...
       const sdk = await getSdk();
       const { bytecode } = await sdk.getContractByteCode(innerTx.value.contractId);
-      const txParams: ITx = await sdk.compilerApi.decodeCalldataBytecode({
-        bytecode,
-        calldata,
-      });
+      const txParams: ITx = await postJson(
+        `${activeNetwork.value.compilerUrl}/decode-calldata/bytecode`,
+        { body: { bytecode, calldata } },
+      );
```
subhod-i, line 194: Nice move. Add a todo to replace this with SDK method whenever we update the SDK version that supports this method.

subhod-i (approved):

src/lib/contracts/FungibleTokenFullInterfaceACI.json
CedrikNikita: `burn_trigger_pos` is not existing in this file, but required.

subhod-i: Then you could just use the just src/lib/contracts/FungibleTokenFullACI.json file only

CedrikNikita: `burn_trigger_pos` does not exist in `FungibleTokenFullACI`, but required by `ZeitToken`.

subhod-i: So maybe the file naming is misleading.

CedrikNikita: This is how the contract named inside of the file,

PR #2136, opened 2023-06-22
[SW-989] show contract address in the details

konradodo:


subhod-i (approved):

src/popup/locales/en.json
```diff
@@ -951,6 +951,7 @@
     "transactionDetails": {
       "poolFee": "pool fee",
       "heading": "Transactions Details",
+      "contract": "Contract",
```
peronczyk, line 954: I think this is a common word and should be moved to the "common" group.

peronczyk (approved):

PR #2140, opened 2023-06-23
[SW-751] Multisig using sdk 13

CedrikNikita:


src/background/popupHandler.ts
```diff
@@ -8,6 +8,7 @@ import {
   POPUP_TYPE_RAW_SIGN,
 } from '../popup/utils';
 
+// @ts-ignore
```
CedrikNikita, line 11: I don't like this solution. I'm open for the suggestions.

src/composables/sdk13.ts
```diff
@@ -148,6 +150,31 @@ export function useSdk13({ store }: IDefaultComposableOptions) {
...
+   */
+  async function getDrySdk(): Promise<AeSdk> {
+    if (!drySdk) {
+      const nodeInstance = await createNodeInstance(activeNetwork.value.url);
+      drySdk = new AeSdk({
+        nodes: [{
+          name: activeNetwork.value.name,
+          instance: nodeInstance!,
+        }],
+        onCompiler: new CompilerHttp(activeNetwork.value.compilerUrl),
```
subhod-i, line 165: None of the dry SDK methods are calling the node or compiler. We should find a way to initialize drySDK without node and compiler

subhod-i: For the one where we are sending the confirmed tx, we can use normal SDK and not dry sdk.

src/composables/sdk13.ts
```diff
@@ -148,6 +150,31 @@ export function useSdk13({ store }: IDefaultComposableOptions) {
...
+      });
+      return drySdk;
+    }
+    const networkId = await sdk.api.getNetworkId();
+    if (activeNetwork.value.networkId !== networkId) {
+      drySdk.pool.delete(sdkCurrentNetwork.name);
+      const nodeInstance = await createNodeInstance(activeNetwork.value.url);
+      drySdk.addNode(activeNetwork.value.name, nodeInstance!, true);
```
subhod-i, line 173: This optimization is redundant. We can always assume networks are different and proceed. Both approaches make one API request to the node. but in the approach I suggested we can eliminate branching(if-statement). Eliminating branching is a great runtime optimization.

src/composables/transactionTx.ts
```diff
@@ -38,14 +39,14 @@ export function useTransactionTx({
...
 
   function setTransactionTx(newTx: ITx) {
     outerTx.value = newTx;
     innerTx.value = getInnerTransaction(newTx);
   }
 
-  function setExternalAddress(address: string) {
+  function setExternalAddress(address: Encoded.AccountAddress) {
     ownerAddress.value = address;
   }
```
subhod-i, line 51: Function name is confusing. Name is `setExternalAddress` but it is setting `ownerAddress`

src/types/index.ts
```diff
@@ -191,16 +191,16 @@ export interface IMultisigAccountResponse {
...
-  signers: string[];
+  signers: Encoded.AccountAddress[];
   hasPendingTransaction: boolean;
   pending?: boolean;
 }
 
 export interface IRawMultisigAccount {
-  multisigAccountCreationEncodedCallData?: string;
-  signedAttachTx?: string;
+  multisigAccountCreationEncodedCallData?: Encoded.ContractBytearray;
```
subhod-i, line 202: Nice

src/types/index.ts
```diff
@@ -385,7 +385,7 @@ export interface ITransaction {
   pending: boolean; // There are cases that not only the IPendingTransaction can be pending
   rawTx?: any; // TODO find type
```
subhod-i, line 386: RawTx type is available in the SDK

subhod-i (changes requested):

src/composables/multisigAccountCreate.ts
```diff
@@ -162,21 +175,24 @@ export function useMultisigAccountCreate({ store }: IDefaultComposableOptions) {
...
+    ) {
       throw Error(`Raw PayForTransaction not found for account ${accountId}, Prepare PayForTransaction first`);
     }
 
     const sdk = await getSdk();
 
     // Send transaction to the chain
-    const { txHash } = await sdk.api.postTransaction(
-      { tx: pendingMultisigCreationTxs.value[accountId].rawTx },
+    const a = await sdk.api.postTransaction(
```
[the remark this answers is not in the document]
CedrikNikita: Leftover.

src/background/popupHandler.ts
peronczyk: What kind of problem it tries to solve?

src/background/popupHandler.ts
```diff
@@ -18,6 +19,7 @@ export const showPopup = async (aepp: any, type: string, params?: any) => {
   const id = uuid();
   const { href, protocol, host } = typeof aepp === 'object' ? getAeppUrl(aepp) : new URL(aepp);
   const tabs = await browser.tabs.query({ active: true });
+  // @ts-ignore
```
peronczyk, line 22: Same here...

src/composables/multisigAccountCreate.ts
```diff
@@ -118,39 +128,42 @@ export function useMultisigAccountCreate({ store }: IDefaultComposableOptions) {
    * First step of creating the multisig account
    * Prepare multisig account creation transaction
    */
-  async function prepareVaultCreationRawTx(payerId: string, accountId: string) {
-    if (!pendingMultisigCreationTxs.value[accountId]?.signedAttachTx) {
+  async function prepareVaultCreationRawTx(
```
peronczyk, line 131: I suggest to add empty line before this one.

src/composables/multisigAccountCreate.ts
```diff
@@ -162,21 +175,24 @@ export function useMultisigAccountCreate({ store }: IDefaultComposableOptions) {
...
-    accountId: string,
+    accountId: Encoded.AccountAddress,
     confirmationsRequired: number,
-    signers: string[],
+    signers: Encoded.AccountAddress[],
   ) {
-    if (!pendingMultisigCreationTxs.value[accountId]?.signedAttachTx) {
+    if (
+      !pendingMultisigCreationTxs.value[accountId]?.signedAttachTx
+      || !pendingMultisigCreationTxs.value[accountId].rawTx
```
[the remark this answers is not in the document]
peronczyk: Follow up to @subhod-i 's code suggestion (if it's correct):
```
if (pendingMultisigCreationTxs.value[accountId]?.rawTx) {
```

src/background/popupHandler.ts
CedrikNikita: https://github.com/aeternity/superhero-wallet/actions/runs/5356614184/jobs/9716655449#step:9:104

src/composables/transactionTx.ts
CedrikNikita: I don't think it should be tackled in this PR.

src/types/index.ts
CedrikNikita: This change would require a deep refactoring in the whole project.

src/popup/pages/MultisigProposalDetails.vue
```diff
@@ -400,9 +402,9 @@ export default defineComponent({
 
       multisigTx.value = {
         ...activeMultisigTx.tx,
-        type: SCHEMA.TX_TYPE.gaMeta,
-        tag: SCHEMA.TX_TYPE.gaMeta,
-      } as ITx;
+        type: Tag[Tag.GaMetaTx],
+        tag: Tag.GaMetaTx,
+      } as any;
```
peronczyk, line 407: Maybe there is a way for not using the `as any` here?
For example the `tag` property of the `ITx` requires `string`, but the `Tag.GaMetaTx` is a number. Is this a mistake?

peronczyk (changes requested):

src/popup/pages/MultisigProposalDetails.vue
```diff
@@ -474,7 +476,18 @@ export default defineComponent({
       try {
         const { gaAccountId, txHash, nonce } = activeMultisigAccount.value;
         const rawTx = await fetchTransactionByHash(txHash as string);
-        transaction.value = await sendTx(gaAccountId, (rawTx as any).tx, nonce);
+        if (!rawTx) {
+          throw Error('failed to load a transaction');
+        }
+        transaction.value = JSON.parse(
+          JSON.stringify(
```
peronczyk, line 483: Why `JSON.parse/JSON.stringify` is used here?

src/popup/pages/MultisigProposalDetails.vue
```diff
@@ -474,7 +476,18 @@ export default defineComponent({
...
+            await sendTx(gaAccountId, rawTx.tx, nonce),
+            // TODO: use bigint in every place for transcation amount and fee
+            (key, value) => typeof value === 'bigint'
+              ? value.toString()
+              : value,
```
peronczyk, line 488: I'm not sure about this. All `BigInt`, `string` and `number` has the `toString()` method.

peronczyk (changes requested):

src/composables/sdk13.ts
subhod-i: irrelevant to what I commented. My suggestion is 
```suggestion
      drySdk.pool.delete(sdkCurrentNetwork.name);
      const nodeInstance = await createNodeInstance(activeNetwork.value.url);
      drySdk.addNode(activeNetwork.value.name, nodeInstance!, true);
```
This works, the optimization using `if` branching is not really an optimization.

src/popup/pages/MultisigProposalDetails.vue
CedrikNikita: The problem here is that ITx is perfectly describing the middleware transactions, but if we are going to use the type that is returned from `unpackTx` like in this particular case, we would need a deep refactoring of the whole project. And i suggest to split this into different task.

src/composables/sdk13.ts
CedrikNikita: Got it.

src/composables/multisigAccountCreate.ts
CedrikNikita: I didn't understand where.

src/popup/pages/MultisigProposalDetails.vue
CedrikNikita: image

src/popup/pages/MultisigProposalDetails.vue
CedrikNikita: In order to convert all the inner `bigint` values into `string`.

src/composables/sdk13.ts
davidyuk: As I see, with no network switching `sdk.api.getNetworkId` wouldn't make an extra request because https://github.com/aeternity/aepp-sdk-js/blob/75097952ee689c591973048b0c52ea1c516dcfb5/src/Node.ts#L155-L160
but `createNodeInstance` would do it for sure. So, `if` makes sense.

davidyuk: > Eliminating branching is a great runtime optimization

This makes sense only as improving readability. Don't worry about sync code optimizations unless you have warnings like `[Violation] Handler took 83ms of runtime`.

Liubov-crypto:
LGTM

src/composables/multisigAccountCreate.ts
```diff
@@ -118,39 +127,38 @@ export function useMultisigAccountCreate({ store }: IDefaultComposableOptions) {
    * First step of creating the multisig account
    * Prepare multisig account creation transaction
    */
-  async function prepareVaultCreationRawTx(payerId: string, accountId: string) {
-    if (!pendingMultisigCreationTxs.value[accountId]?.signedAttachTx) {
+  async function prepareVaultCreationRawTx(
+    payerId: Encoded.AccountAddress,
+    accountId: Encoded.AccountAddress,
+  ) {
+    if (pendingMultisigCreationTxs.value[accountId].signedAttachTx === undefined) {
```
pstachel, line 134: In this function you are referring to `pendingMultisigCreationTxs.value[accountId]` three times, what do you think of assigning it to variable before this line

```
const { signedAttachTx, rawTx } = pendingMultisigCreationTxs.value[accountId]
```

src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -276,7 +277,7 @@ export default defineComponent({
 
     function addNewSigner() {
       signers.value.push({
-        address: '',
+        address: '' as any,
```
pstachel, line 280: maybe it could be typed?

src/popup/pages/MultisigProposalDetails.vue
peronczyk: Why not `.toString()` ?

src/popup/pages/MultisigProposalDetails.vue
peronczyk: OK, so please add `TODO` comment that describes why and what needs to be refactored and then resolve my comment.

src/background/popupHandler.ts
peronczyk: If this is caused by adding `strictFunctionTypes` to the TS config file maybe we should remove it to avoid this errors?

CedrikNikita: [This is necessary to avoid type issues coming from sdk, and a part of sdk13 usage.](https://github.com/aeternity/aepp-sdk-js/blob/0c9386096701a238edcc55c09287b4487a865a95/docs/README.md?plain=1#L85)

src/lib/logger.ts
```diff
@@ -7,6 +7,9 @@ import { getState } from '../store/plugins/persistState';
 import { useModals } from '../composables';
 import { RejectedByUserError } from './errors';
 
+// @ts-ignore
```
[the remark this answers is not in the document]
CedrikNikita: It's causing an error.

src/types/index.ts
```diff
@@ -153,12 +155,11 @@ export interface IAccount {
 /**
  * Account fetched from the node with the use of `sdk.api.getAccountByPubkey`
  */
-export interface IAccountFetched {
+type accountFetched = Awaited<ReturnType<InstanceType<typeof Node>['getAccountByPubkey']>>;
```
peronczyk, line 158: Types should use PascalCase naming. Also for the future adoption of other blockchains I suggest to add "Aeternity" to the name.

src/composables/sdk13.ts
```diff
@@ -150,6 +155,30 @@ export function useSdk13({ store }: IDefaultComposableOptions) {
...
+      const nodeInstance = await createNodeInstance(activeNetwork.value.url);
+      drySdk = new AeSdk({
+        nodes: [{
+          name: activeNetwork.value.name,
+          instance: nodeInstance!,
+        }],
+      });
+      return drySdk;
+    }
+    const networkId = await drySdk.api.getNetworkId();
```
peronczyk, line 173: I'm not sure if fetching the dry SDK network ID every time we use `getDrySdk` is necessary. Maybe it's better to store the dry SDK network ID on the first initialization and then comparing it with the `nodeNetworkId` that comes from the main SDK?

src/composables/popupProps.ts
```diff
@@ -22,7 +24,7 @@ export function usePopupProps(): IPopupPropsState {
 
   const sender = computed((): IAccountOverView => ({
     name: popupProps.value?.app?.name,
-    address: popupProps.value?.app?.host,
+    address: popupProps.value?.app?.host as Encoded.AccountAddress,
```
peronczyk, line 27: App host is an AccountAddress? Looks like a mistake.

src/composables/sdk13.ts
CedrikNikita: https://github.com/aeternity/superhero-wallet/pull/2140#discussion_r1243609216

src/composables/multisigAccounts.ts
```diff
@@ -202,19 +202,19 @@ export function useMultisigAccounts({ store, pollOnce = false }: MultisigAccount
...
               currentAccount?.signers
                 ? { decodedResult: currentAccount.signers }
-                : contractInstance.methods.get_signers(),
-              contractInstance.methods.get_consensus_info(),
-              gaAccountId ? sdk.balance(gaAccountId) : 0,
-            ]);
+                : contractInstance.get_signers(),
+              contractInstance.get_consensus_info(),
+              gaAccountId ? sdk.getBalance(gaAccountId) : new Promise(() => 0),
+            ])) as any;
```
peronczyk, line 217: I think this is bad to treat all resolved promises as `any`. Maybe you can add some typings?

src/types/index.ts
```diff
@@ -9,7 +9,9 @@ import { RouteLocationRaw } from 'vue-router';
 import { TranslateResult } from 'vue-i18n';
 import BigNumber from 'bignumber.js';
 import { Store } from 'vuex';
-import { ContractMethodsBase, Encoded } from '@aeternity/aepp-sdk-13';
+import {
+  ContractMethodsBase, Encoded, Node, Tag,
+} from '@aeternity/aepp-sdk-13';
```
pstachel, line 14: Please split this line

src/composables/popupProps.ts
CedrikNikita: `IAccountOverview` type is used. Which is based on `IAccount` which is used for the accounts, where address is supposed to be `ak_...`. The question is why it was used in the first place. Maybe it can be done separately.

peronczyk (changes requested):

CharisSiarampalis:
Lgtm

src/background/popupHandler.ts
peronczyk: OK< didn't know that. Resolving.

Liubov-crypto:
Retested, LGTM.

src/composables/popupProps.ts
peronczyk: I suggest to change the type of `address` in the `IAccountOverview` to `Encoded.AccountAddress | string` for now.
And maybe add some comment to this prop in the types file that the `string` stands for URL.

In the future we should make a new interface for this `sender` stuff. What do you think?

src/composables/sdk13.ts
peronczyk: OK, I'm also surprised same as Subhod was. Weird that things under `api` property sometimes does not do fetching :O

pstachel (approved):

peronczyk (approved):

subhod-i (approved):
Please have a look in the comments.

PR #2147, opened 2023-06-30
[SW-997, SW-999] Update styling of paragraph warning messages and show on multisig proposal if ft is used

pstachel:


src/popup/components/TransferSendForm.vue
```diff
@@ -372,7 +381,12 @@ export default defineComponent({
       token,
     }: Dictionary) {
       if (token) {
-        formModel.value.selectedAsset = availableTokens.value[token] || aeternityCoin.value;
+        if (props.isMultisig && token !== AETERNITY_SYMBOL) {
+          hasMultisigTokenWarning.value = true;
+        } else {
+          formModel.value.selectedAsset = availableTokens.value[token] || aeternityCoin.value;
+          hasMultisigTokenWarning.value = false;
+        }
```
peronczyk, line 389: I suggest to set the `hasMultisigTokenWarning` as a computed property. It needs to check if the selected asset is not an AE coin and it's a multisig.

src/popup/components/TransferSendForm.vue
```diff
@@ -537,6 +551,8 @@ export default defineComponent({
       multisigVaultOwnedByManyAccounts,
       activeMultisigAccount,
       isUrlTippingEnabled,
+      hasMultisigTokenWarning,
+      INFO_BOX_TYPES,
```
peronczyk, line 555: I suggest to move the constants to the top of return object.

src/popup/components/TransferSendForm.vue
```diff
@@ -2,6 +2,11 @@
   <div class="transfer-send-form">
     <template v-if="isMultisig">
       <ModalHeader :title="$t('modals.multisigTxProposal.title')" />
+      <InfoBox
+        v-if="hasMultisigTokenWarning"
+        :type="INFO_BOX_TYPES.warning"
+        :text="$t('modals.multisigTxProposal.tokenWarning')"
+      />
```
peronczyk, line 9: Why not just disable the possibility of changing the asset if it's a multisig?

src/popup/pages/NetworkForm.vue
```diff
@@ -211,6 +212,7 @@ export default defineComponent({
       buttonDisabled,
       isEdit,
       formConfig,
+      INFO_BOX_TYPES,
```
peronczyk, line 215: I suggest to put the constants art the top of the return object.

peronczyk (changes requested):

src/popup/components/TransferSendForm.vue
pstachel: It is disabled, but if someone scans different token they can be confused why it was changed to AE

src/popup/components/TransferSendForm.vue
pstachel: It can't be done as the token is not set in the form, this is the only place where we receive the info about the scanned token

src/popup/components/TransferSendForm.vue
peronczyk: OK, makes sense.

src/popup/components/TransferSendForm.vue
peronczyk: Makes sense, thanks.

peronczyk (approved):

konradodo (approved):

PR #2154, opened 2023-07-07
fix(transaction-overview): use correct label for a contract call

CedrikNikita:
missed in https://github.com/aeternity/superhero-wallet/pull/2136

### DIFF
--- a/src/popup/components/TransactionOverview.vue
+++ b/src/popup/components/TransactionOverview.vue
@@ -116,7 +116,9 @@ export default defineComponent({
           const contract = {
             address: contractId,
             url: getExplorerPath.value(contractId),
-            label: t(`transaction.overview.${isDex.value ? 'superheroDex' : 'contract'}`),
+            label: isDex.value
+              ? t('transaction.overview.superheroDex')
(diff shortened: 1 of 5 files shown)


pstachel (approved):

konradodo (approved):

Liubov-crypto:
LGTM. This pr has been fixed existing issue with labels. 

On some screenshots, the "smart" prefix is present. I suggest fixing this in a separate pr.

onvisions:
I was asked by @CedrikNikita to provide my opinion about this issue from UX perspective. The question was should we also update the transaction type labels in such a way that "Contract call" becomes "Smart contract call". In short "No" because we are striving for having the labels text as short as possible. 
So my suggestion for consistency update will be the following:

1. Use "Smart contract" for the text under the identicon and above the smart contract address (see screenshot below);
2. Continue to use just "Contract call" (NOT "Smart contract call").
image

CharisSiarampalis:
Lgtm

Liubov-crypto:
LGTM

CharisSiarampalis:
Lgtm

PR #2158, opened 2023-07-12
[SW-1034] Fix translation dynamic keys

peronczyk:


src/popup/utils/constants.ts
```diff
@@ -65,6 +68,55 @@ export const TX_FUNCTIONS = {
...
 } as const;
 
+export const TX_FUNCTION_TYPE_MULTISIG = {
+  propose: 'propose',
+  confirm: 'confirm',
+  refuse: 'refuse',
+  revoke: 'revoke',
+} as const;
+
+export const TX_FUNCTION_TYPE_DEX: Record<'pool' | 'addLiquidity' | 'removeLiquidity' | 'swap' | 'allowance' | 'maxSpent' | 'minReceived', TxFunctionRaw[]> = {
```
konradodo, line 78: what do you think about separate this union type ?

konradodo (approved):

src/popup/utils/constants.ts
peronczyk: OK, I created separate type for this.

pstachel (approved):

src/popup/utils/constants.ts
```diff
@@ -68,6 +72,69 @@ export const TX_FUNCTIONS = {
   claim: 'claim',
 } as const;
 
+export const TX_FUNCTION_TYPE_MULTISIG = {
```
subhod-i, line 75:
```suggestion
export const MULTISIG_TX_FUNCTIONS = {
```

src/popup/utils/constants.ts
```diff
@@ -42,6 +43,9 @@ export const TX_DIRECTION = {
   received: 'received',
 } as const;
 
+/**
+ * ITx.function
+ */
 export const TX_FUNCTIONS = {
```
subhod-i, line 49:
```suggestion
export const TIPPING_TX_FUNCTIONS = {
```

src/popup/utils/constants.ts
```diff
@@ -68,6 +72,69 @@ export const TX_FUNCTIONS = {
...
 } as const;
 
+export const TX_FUNCTION_TYPE_MULTISIG = {
+  propose: 'propose',
+  confirm: 'confirm',
+  refuse: 'refuse',
+  revoke: 'revoke',
+} as const;
+
+export const TX_FUNCTION_TYPE_DEX: Record<DexFunctionType, TxFunctionRaw[]> = {
```
subhod-i, line 82:
```suggestion
export const DEX_TX_FUNCTIONS: Record<DexFunctionType, TxFunctionRaw[]> = {
```

src/popup/utils/constants.ts
```diff
@@ -68,6 +72,69 @@ export const TX_FUNCTIONS = {
...
+  ],
+  minReceived: [
+    'swap_exact_tokens_for_tokens', 'swap_exact_ae_for_tokens', 'swap_exact_tokens_for_ae',
+  ],
+};
+
+/**
+ * ITx.tag
+ */
+export const TX_TAGS_SUPPORTED = [
```
subhod-i, line 111:
```suggestion
export const SUPPORTED_TX_TAGS = [
```

src/popup/utils/constants.ts
```diff
@@ -68,6 +72,69 @@ export const TX_FUNCTIONS = {
...
+export const TX_RETURN_TYPE_OK = 'ok';
+export const TX_RETURN_TYPE_ABORT = 'abort';
+export const TX_RETURN_TYPE_REVERT = 'revert';
+
+export const TX_RETURN_TYPES = [
+  TX_RETURN_TYPE_OK,
+  TX_RETURN_TYPE_ABORT,
+  TX_RETURN_TYPE_REVERT,
+] as const;
+
```
subhod-i, line 137: IDEM. The naming of the constants could be improved.

src/types/index.ts
```diff
@@ -295,22 +296,29 @@ export interface ICurrency {
...
 
-export type TxFunction = TxFunctionRaw | TxFunctionParsed;
+export type TxFunctionMultisig = keyof typeof TX_FUNCTION_TYPE_MULTISIG;
+
+export type TxFunction = TxFunctionRaw | TxFunctionParsed | TxFunctionMultisig;
+
+/**
+ * String representation of the ITx.tag.
+ */
+export type TxType = keyof typeof Tag;
```
subhod-i, line 321: Nice.

subhod-i (changes requested):

src/popup/utils/constants.ts
peronczyk: I don't understand this suggestions. In my opinion it is nice to name things starting from more general part and finish with specific. This way when using constants it looks more grouped:
```
TX_FUNCTION_TYPE_DEX
TX_FUNCTION_TYPE_FOO
TX_FUNCTION_TYPE_BAR
```

src/popup/utils/constants.ts
peronczyk: Also here I think starting from more general part and finish with specifics is a more well laid out approach.

src/popup/utils/constants.ts
peronczyk: Yeah, I already improved the naming :D

src/types/index.ts
peronczyk: Thanks :)

src/composables/transactionTx.ts
```diff
@@ -30,6 +38,56 @@ interface UseTransactionOptions extends IDefaultComposableOptions {
...
+  NamePreclaimTx: () => tg('transaction.listType.namePreClaimTx'),
+  NameClaimTx: () => tg('transaction.listType.nameClaimTx'),
+  NameUpdateTx: () => tg('transaction.listType.nameUpdateTx'),
+  NameTransferTx: () => tg('transaction.listType.nameTransferTx'),
+  NameRevokeTx: () => tg('transaction.listType.nameRevokeTx'),
+  SpendTx: () => tg('transaction.listType.sentTx'),
+};
+
+const txFunctionTranslations: TxPropertyTranslations<TxFunctionRaw | TxFunctionMultisig> = {
+  propose: () => tg('transaction.function.propose'),
```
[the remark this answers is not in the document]
peronczyk: Makes sense - I'll move this into separate `i18nTranslations.ts` file.

src/popup/utils/constants.ts
subhod-i: It's look good this way where you defined them but not so good at where you are using it
For eg:
```js
TX_FUNCTION_TYPE_DEX.contains('tip');

vs 

DEX_TX_FUNCTIONS.contains('tip') 

```

later one is more readable.

subhod-i: Let me know what you think about this. The existing one is fine too.

src/popup/utils/constants.ts
subhod-i: Right! Sorry! The diff was showing only Tipping contract methods.

subhod-i (approved):

src/popup/utils/constants.ts
lukeromanowicz: I was asked for my point of view on that matter so here are my two cents: 
1. If there is more than one variable holding alternative content of the same type, I usually call it [SOME_NAME]_[MODIFIER] so when I put the first part of the const, the IDE suggests all the alternatives I can use without me memorizing what the right [modifier] is called. e.g. 
```
MY_TOYS_CARS = []
MY_TOYS_BLOCKS = []
```
2.  if there is only one variable without any variants then I just follow the English grammar which would be for example `MY_TOY_CARS = []`

---
I think here the first case applies so I would probably go with `TX_FUNCTION_TYPE_DEX` since there are multiple TX_FUNCTION_TYPE variants

peronczyk: I created also a poll on our chat and 4 out of 4 available people voted for the new naming convention. I'm closing this.

PR #2160, opened 2023-07-13
feat: multi purpose websocket client

ifaouibadi:


subhod-i (approved):
After from fthe comments the code changes look good to me.

src/lib/WebSocketClient.ts
```diff
@@ -0,0 +1,193 @@
...
+  WEB_SOCKET_SOURCE,
+  WEB_SOCKET_CHANNELS,
+  WEB_SOCKET_SUBSCRIBE,
+  WEB_SOCKET_UNSUBSCRIBE,
+  WEB_SOCKET_RECONNECT_TIMEOUT,
+  handleUnknownError,
+  NETWORK_MAINNET,
+} from '../popup/utils';
+
+class WebSocketClient {
```
davidyuk, line 20: I would add middleware to the name because it is aware of it

davidyuk (commented):
There is a code for the same goal in sdk repo https://github.com/aeternity/aepp-sdk-js/blob/develop/src/MiddlewareSubscriber.ts (already released)
Though it has own issues

PR #2162, opened 2023-07-13
refactor: remove unused files in preference to use sdk supported ledger account

CedrikNikita:


pstachel (approved):

src/store/modules/accounts/index.js
```diff
@@ -1,10 +1,9 @@
 /* eslint no-param-reassign: ['error', { 'ignorePropertyModificationsFor': ['state'] }] */
 
 import hdWallet from './hdWallet';
-import ledger from './ledger';
 import { ACCOUNT_HD_WALLET } from '../../../popup/utils';
 
-const modules = { hdWallet, ledger };
+const modules = { hdWallet };
```
peronczyk, line 6: I suggest to add a TODO comment to describe why this is an object and that we need to remove the concept of "modules" here.

PR #2163, opened 2023-07-13
[SW-1062] Tech enabler: Add browser extension typings

peronczyk:


src/popup/pages/TipsClaim.vue
```diff
@@ -153,7 +153,7 @@ export default defineComponent({
     onMounted(async () => {
       if (IS_EXTENSION && browser) {
         const [tab] = await browser.tabs.query({ active: true, currentWindow: true });
-        if (tab && validateTipUrl(tab.url)) {
+        if (tab && tab.url && validateTipUrl(tab.url)) {
```
pstachel, line 156:
```suggestion
        if (tab?.url && validateTipUrl(tab.url)) {
```

src/popup/router/index.ts
```diff
@@ -51,7 +51,7 @@ const unbind = router.beforeEach(async (to, from, next) => {
     (
       !RUNNING_IN_POPUP
       && to.name === ROUTE_INDEX
-      && (await browser?.storage.local.get(lastRouteKey))[lastRouteKey]
+      && ((await browser?.storage.local.get(lastRouteKey)) as any)[lastRouteKey]
```
pstachel, line 54: aren't we able to define the type here?

peronczyk: Not that easy. The typing of the storage getter defines the return value to be `string | number | true | RegExp | ArrayBuffer | StorageObject | DataView | StorageArray | StorageMap | StorageSet | undefined` and the TS engine is complaining that it's not assignable to `NavigationGuardNextCallback`. I hate this `as any` situations, but after spending some time with this I didn't come up with anything meaningful :/

src/popup/pages/TipsClaim.vue
peronczyk: Makes sense. Changed...

pstachel (approved):

CedrikNikita (approved):
Code changes look good. Didn't test it.

PR #2164, opened 2023-07-13
[SW-1056] Remove sdk v11 instance from the state

subhod-i:


src/composables/sdk13.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
pstachel, line 1: why did you disable it?

src/composables/sdk13.ts
```diff
@@ -71,20 +79,16 @@ export function useSdk13({ store }: IDefaultComposableOptions) {
...
     try {
       // TODO: remove ignore version once HTTP compiler dependency is removed
       nodeInstance = new Node(url, { ignoreVersion: true });
       nodeNetworkId.value = (await nodeInstance.getStatus()).networkId;
-      isNodeReady.value = true;
+      store.commit('setNodeStatus', NODE_STATUS_CONNECTED);
     } catch (error) {
-      isNodeError.value = true;
       nodeNetworkId.value = undefined;
+      store.commit('setNodeStatus', NODE_STATUS_ERROR);
```
pstachel, line 90: why can't we store this variable in a composable?

src/lib/wallet.js
```diff
@@ -51,30 +50,16 @@ export default async function initSdk() {
       try {
         store.commit('setNodeStatus', NODE_STATUS_CONNECTING);
 
-        let sdk;
-        let sdk13;
+        const sdk = await getSdk();
+        await getMiddleware();
```
pstachel, line 54:
```suggestion
        const [sdk] = await Promise.all[getSdk(), getMiddleware()]
```

src/popup/router/index.ts
```diff
@@ -69,7 +69,9 @@ router.beforeEach(async (to, from, next) => {
     return;
   }
 
-  if (!store.getters['sdkPlugin/sdk'] && !RUNNING_IN_POPUP) initSdk();
+  const { isSdkReady } = useSdk13({ store });
+
+  if (!isSdkReady && !RUNNING_IN_POPUP) initSdk();
```
pstachel, line 74:
```suggestion
  if (!isSdkReady && !RUNNING_IN_POPUP) {
    initSdk();
  }
```

src/store/modules/accounts/hdWallet.js
```diff
@@ -21,8 +21,10 @@ export default {
     nextAccountIdx: 1,
   },
   actions: {
-    async isAccountUsed({ rootGetters }, address) {
-      return rootGetters['sdkPlugin/sdk'].api.getAccountByPubkey(address).then(() => true, () => false);
+    async isAccountUsed({ rootGetters: { store } }, address) {
```
pstachel, line 24: rootGetters are not needed here
```suggestion
    async isAccountUsed(context, address) {
```

src/store/modules/accounts/hdWallet.js
```diff
@@ -21,8 +21,10 @@ export default {
...
+      const { getSdk } = useSdk13({ store });
```
pstachel, line 25:
```suggestion
      const { getSdk } = useSdk13({ store: context });
```

pstachel (changes requested):

src/composables/sdk13.ts
peronczyk: I agreed with Subhod that I'll do the refactor for keeping the connection status inside of the composable: https://aeternity-blockchain.atlassian.net/browse/SW-1061

pstachel: Ok, resolving my comment then

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -325,9 +326,7 @@ export default defineComponent({
             { bytecode },
           ] = await Promise.all([
             fetchJson(`${activeNetwork.value.url}/v3/contracts/${popupProps.value.tx.contractId}/code`),
-            // SDK is needed to establish the `networkId` and the dex contracts for the network
-            // TODO replace with `getSdk` after migration to SDK13
-            store.dispatch('sdkPlugin/initialize'),
+            getSdk(),
```
peronczyk, line 329: I suggest to leave the previous comment because without it using the `getSdk` has no context.

src/popup/router/index.ts
peronczyk, line 74: I think this needs to be removed as the logic of the SDK initialization is totally different than when this lines was added to the codebase.

src/store/modules/accounts/hdWallet.js
peronczyk: Totally true. I think this was a mistake as the store is not accessible from the root getters.

src/types/index.ts
```diff
@@ -359,7 +359,7 @@ export interface ITx {
   nameSalt?: string
   nonce?: number
   payerId?: string
-  payload?: string
+  payload?: Encoded.Any
```
peronczyk, line 362: Properties should end with `;`.

peronczyk (changes requested):

src/composables/sdk13.ts
subhod-i: This was by mistake. Thanks

CharisSiarampalis:
Latest transactions are not working for me

image

pstachel:
@CharisSiarampalis that should be fixed in another PR #2166

src/store/modules/accounts/hdWallet.js
subhod-i: Thanks.

Liubov-crypto:
> Latest transactions are not working for me
> 
> image
> 
> Only new transactions appear image

the same issue on dev branch. please recheck all issues there.

subhod-i:
> > Latest transactions are not working for me
> > image
> > Only new transactions appear image
> 
> the same issue on dev branch. please recheck all issues there.

This has been fixed.  Please check!

src/popup/utils/bigint.ts
pstachel, line 1: Do we need another library for handling bigints or can we use bignumber.js for that also?

src/store/plugins/pendingTransactionHandler.js
```diff
@@ -2,20 +2,21 @@ import { useSdk13 } from '../../composables';
...
+      const sdk = await getSdk();
+      await sdk.poll(hash);
       store.commit('setPendingTransactionSentByHash', { hash, network: nodeNetworkId.value });
     } catch (e) {
       store.commit('removePendingTransactionByHash', { hash, network: nodeNetworkId.value });
     }
   };
 
-  await watchUntilTruthy(() => store.state.sdkPlugin.ready);
+  await watchUntilTruthy(() => isSdkReady);
```
pstachel, line 19:
```suggestion
  await watchUntilTruthy(isSdkReady);
```
or 
```suggestion
  await watchUntilTruthy(() => isSdkReady.value);
```

src/popup/utils/bigint.ts
subhod-i: bignumber.js is for creating BigNumber instances. `json-bigint` for processing and parsing JSON responses.

src/store/plugins/pendingTransactionHandler.js
subhod-i: Thanks

peronczyk (changes requested):

src/popup/utils/bigint.ts
```diff
@@ -0,0 +1,7 @@
+import JsonBig from 'json-bigint';
+
+const jsonBig = JsonBig({ storeAsString: true });
+
+export function parseBigIntJson(response: any): any {
+  return jsonBig.parse(jsonBig.stringify(response));
```
peronczyk, line 6: Stringifying object and then parsing it again can be unpredictable. I strongly suggest not to use this approach to apply data manipulation.

peronczyk (changes requested):

src/popup/router/index.ts
```diff
@@ -69,7 +69,11 @@ router.beforeEach(async (to, from, next) => {
     return;
   }
 
-  if (!store.getters['sdkPlugin/sdk'] && !RUNNING_IN_POPUP) initSdk();
+  const { isSdkReady } = useSdk13({ store });
+
+  if (!isSdkReady.value && !RUNNING_IN_POPUP) {
+    initSdk();
+  }
```
peronczyk, line 76: I think we should remove this as the SDK is not needed on every route. Especially that the SDK instance is lazy loaded when `getSdk` method is used.

peronczyk (changes requested):

src/popup/utils/bigint.ts
subhod-i: This is how it is often done for the BigInt responses. I see this as totally valid.  If not for this then the engine in general doesn't know how to serialize/deserialize JSON objects with BigInts.

src/composables/sdk13.ts
```diff
@@ -203,6 +215,26 @@ export function useSdk13({ store }: IDefaultComposableOptions) {
...
+          if (oldValue) {
+            sdk.pool.delete(oldValue.name);
+          }
+          sdk.addNode(newValue.name, (await createNodeInstance(newValue.url))!, true);
+        } finally {
+          sdkBlocked = false;
+        }
+      },
+    );
+  }
```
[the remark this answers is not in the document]
subhod-i: I have moved the watchers to individual files, and dedicated watchers for background and popup. The approach with on-demand resetNode doesn't work as expected. 
When you are on the network page and change the network and navigate back to the dashboard, there  are multiple `getSdk` methods fired simultaneously, resulting in a `Duplicate node, node already exists error`. This could be addressed by a flag, due to the parallel execution of the components this is a poor approach to solving this problem, and expects the first `getSdk` call rest of the calls need to wait until the node has been updated. This makes the dashboard rendering really slow. Hence the source of truth remains in the watchers, I guess this also improves the readability too given we are gonna use too many nonpure variables in the `getSdk` method.

src/popup/router/index.ts
subhod-i: IDEM https://github.com/aeternity/superhero-wallet/pull/2164#discussion_r1266318401

src/background/wallet.js
```diff
@@ -71,27 +69,12 @@ export async function init() {
...
-          sdk13.pool.delete(oldNetwork.name);
-        }
-        sdk.addNode(network.name, await createNewNodeInstance(network.url), true);
-        sdk13.addNode(network.name, await createNodeInstance(network.url), true);
-      } finally {
-        initSdkRunning = false;
-      }
+      resetNode(oldNetwork, network);
     },
+    { deep: true },
```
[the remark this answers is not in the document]
subhod-i: For resetting the node, we need the old network name to remove the existing node from the pool and the new network URL, and the name for adding a new node. `deep` is also needed if the user edits just one item in the custom network form.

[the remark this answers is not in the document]
subhod-i: Otherwise, SDK would maintain the same node with a different network name than that of the application

`sdk.addNode(NAME, nodeInstance, true)`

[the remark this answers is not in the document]
CedrikNikita: So it wasn't checking this fields without `deep`? How was it working before as every fields in this object is plain ones?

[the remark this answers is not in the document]
subhod-i: Same as before but without deep. You may be right
https://github.com/aeternity/superhero-wallet/blob/b8ce17ee1afd87b56861be8796fbb9ecf37e8cf9/src/store/plugins/sdk.js#L180C28-L180C28

src/background/wallet.js
```diff
@@ -71,26 +69,10 @@ export async function init() {
   store.watch(
     (state, getters) => getters.activeNetwork,
     async (network, oldNetwork) => {
-      if (initSdkRunning || isEqual(network, oldNetwork)) {
+      if (isEqual(network, oldNetwork)) {
```
peronczyk, line 72: Why not
```
if (!isEqual(network, oldNetwork)) {
  resetNode(oldNetwork, network);
}
```

peronczyk (approved):

pstachel (approved):

PR #2135, opened 2023-06-21
Feature/migrate to vue3

martinkaintas:

PR #2166, opened 2023-07-13
[SW-482] Use SDK13 in latest transactions & fix account discovery

pstachel:


subhod-i (approved):

src/popup/components/Modals/AccountImport.vue
```diff
@@ -99,9 +102,10 @@ export default defineComponent({
       store.commit('setBackedUpSeed');
       props.resolve();
       setTimeout(async () => {
-        await getSdk();
-        store.dispatch('accounts/hdWallet/discover');
-      }, 100);
+        watchUntilTruthy(() => isSdkReady.value).then(
```
peronczyk, line 105: Why not awaiting the `getSdk` ?

peronczyk (approved):

src/popup/components/Modals/AccountImport.vue
CedrikNikita: https://github.com/aeternity/superhero-wallet/pull/2135#discussion_r1246485890

peronczyk: We'll resolve this in the future :)

pstachel: That change doesn't reflect what happened properly - `getSdk` is not working here but previously we were only checking if sdk is ready. Usage of `getSdk` here was a mistake as it wasn't working

PR #2167, opened 2023-07-14
[SW-1050] Remove usage of SDK v11 transaction schema

subhod-i:


pstachel (approved):

src/popup/utils/utils.ts
```diff
@@ -374,7 +374,7 @@ export function categorizeContractCallTxObject(transaction: ITransaction): {
   url?: string
   note?: string
 } | null {
-  if (!compareCaseInsensitive(transaction.tx.type, SCHEMA.TX_TYPE.contractCall)) {
+  if (!compareCaseInsensitive(transaction.tx.type, Tag[Tag.ContractCallTx])) {
```
peronczyk, line 377: Maybe the `tx.type` can be now compared directly to the `Tag[Tag.ContractCallTx]`. I mean do they have the same casing?

src/types/index.ts
```diff
@@ -367,8 +367,8 @@ export interface ITx {
   recipientId?: string
   senderId?: string
   selectedTokenContractId?: string
-  tag?: Tag;
-  type: keyof typeof Tag,
+  tag?: Tag
+  type: string
```
peronczyk, line 371: TS properties should end up with `;`.

peronczyk, line 371: I think this change is bad - we need to have explicit list of possible values.

peronczyk (changes requested):

CharisSiarampalis:
Lgtm

src/popup/utils/utils.ts
subhod-i: As discussed, I have added code comments for the usage of tx `type`. We need to support case sensitive contract call URLs

src/types/index.ts
subhod-i: Fixed it. Added comment too.

peronczyk (approved):

PR #2171, opened 2023-07-17
[SW-1064] Fix/transaction list composables

konradodo:


peronczyk (changes requested):

src/composables/transactionsList.ts
```diff
@@ -0,0 +1,229 @@
...
+    pending: {},
+    tipWithdrawnTransactions: [],
+  };
+
+  transactions.value = [...transactions.value, newAccountTransactionState];
+
+  return newAccountTransactionState;
+}
+
+function updateTransactionByAccount(address: Encoded.AccountAddress, transaction: ITransaction) {
```
pstachel, line 52: I'm not a fan of the names `doSthForAccount`, `doSthFromAccount`,`doSthByAccount`. What do you think of naming it `doAccountSth` - in this case `updateAccountTransactions`?

src/composables/transactionsList.ts
```diff
@@ -0,0 +1,226 @@
...
+const setTransactionsNextPage = (address: string, url: string) => {
+  const index = getTransactionStateIndexByAccountAddress(address);
+  transactions.value[index].nextPageUrl = url;
+};
+
+export const useTransactionsList = () => {
+  const store = useStore();
+  const { nodeNetworkId } = useSdk13({ store });
+  const { activeAccount } = useAccounts({ store });
+  const { getSdk } = useSdk({ store });
```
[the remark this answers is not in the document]
subhod-i: This peers needs rebasing on top of develop to review.

subhod-i (changes requested):

CedrikNikita:
The name of the commit is incorrect.
[feat: multichain account create modal](https://github.com/aeternity/superhero-wallet/pull/2171/commits/4d407acc58831b906d6fb62e00919cde1cf83e06)

src/composables/transactionList.ts
```diff
@@ -0,0 +1,272 @@
...
+    );
+
+    if (pendingTransactionForAccount.length) {
+      transactions.value[address].pending[nodeNetworkId.value!] = (
+        pendingTransactionForAccount.filter((transaction) => transaction.hash !== hash)
+      );
+    }
+  }
+
+  async function fetchTransactions(param: IFetchTransactionsParams): Promise<ITransaction[]> {
```
subhod-i, line 182: I don't think an interface `IFetchTransactionsParams` is needed here. It could be just plain function arguments.

CedrikNikita (dismissed):
I'm no longer able to see my pending transactions in the latest transaction widget.
pending

src/composables/transactionList.ts
```diff
@@ -0,0 +1,272 @@
...
+  const accountTransactionState = getTransactionStateByAccountAddress(address);
+
+  if (accountTransactionState) {
+    return accountTransactionState;
+  }
+
+  const newAccountTransactionState: ITransactionsState = {
+    loaded: [],
+    nextPageUrl: '',
+    pending: {},
```
CedrikNikita, line 38: Currently we are storing the pending transactions to if the user will close the extension (for example and open it), we will still be able to show him his pending transaction, also because transactions might not be returned by the mdw fast enough. If we will not store in this case the user won't see his transaction in the list neither as pending nor as mined and then it will appear.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,272 @@
...
+              callerId: transaction.payload.senderId,
+              type: 'ContractCallTx',
+            },
+            hash: transaction.payload.txHash,
+            incomplete: true,
+          } as ITransaction
+          : {}),
+      }));
+
+    preparedTransactions = uniqBy(preparedTransactions.reverse(), 'hash').reverse();
```
subhod-i, line 227: Reverse is being called twice

CharisSiarampalis:
The extensions behaviour is weird. When i first open it it flashes and goes through the menus.


https://github.com/aeternity/superhero-wallet/assets/42606212/ace98729-44f9-418b-b310-8f6040e87919

CharisSiarampalis:
Name registrations is not working. Its only loading up for ever.

image

src/composables/transactionList.ts
konradodo: yes, and in this case it is a correct. Data should be reverse for correct making uniq, after that we should reverse again to have proper order. This code is copy from old store action.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,272 @@
...
+  const { getAeSdk } = useAeSdk({ store });
+
+  const allTransactionsForAccount = computed(() => {
+    if (!activeAccount.value.address) {
+      return [];
+    }
+    const transactionState = getOrInitTransactionStateForAccount(activeAccount.value.address);
+    const pendingTransaction = transactionState?.pending[nodeNetworkId.value!] || [];
+
+    return [...transactionState.loaded, ...pendingTransaction];
```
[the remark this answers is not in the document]
konradodo: i don't see value on this moment. Now i think we should leave that. In another PR we can adding new feature/refactor. The big refactor should be did in few iteration.

src/composables/transactionList.ts
subhod-i: Okay.

src/composables/transactionList.ts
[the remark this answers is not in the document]
subhod-i: I see this common code block floating in multiple places.

src/types/index.ts
```diff
@@ -396,6 +396,8 @@ export interface ITransaction {
   pendingTokenTx?: boolean;
   rawTx?: any; // TODO find type
   tipUrl?: string;
+  sent?: boolean;
```
peronczyk, line 399: How the account can be `sent`? I guess this is a mistake.

src/types/index.ts
```diff
@@ -165,7 +165,7 @@ export interface IAccountFetched extends Omit<AeternityAccountFetched, 'balance'
 
 export interface IAccountOverview extends Partial<Omit<IAccount, 'address'>> {
   // TODO: use a proper type for an address since it can be a url
-  address?: Encoded.AccountAddress | string;
+  address?: Encoded.AccountAddress;
```
peronczyk, line 168: If you removed the `string` please also remove the comment above.

tests/e2e/integration/other.js
peronczyk, line 1: I think to consider this task done the tests should be actually fixed / aligned with changes.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,324 @@
...
+  if (transactionIndex >= 0) {
+    transactions.value[address].loaded[transactionIndex] = transaction;
+  }
+}
+
+const setTransactionsNextPage = (address: Encoded.AccountAddress, url: string) => {
+  transactions.value[address].nextPageUrl = url;
+};
+
+export const useTransactionList = () => {
```
peronczyk, line 68: Please convert this to regular function as this is the way ow all other composables are written.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,324 @@
...
+      if (!recent) {
+        setTransactionsNextPage(address, next);
+      }
+      return data;
+    } catch (error) {
+      return [];
+    }
+  }
+
+  async function fetchPendingTransaction(
```
peronczyk, line 135: I guess the name should use plural version of the "transaction" word.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,324 @@
...
+    address: Encoded.AccountAddress,
```
peronczyk, line 136: I think this argument can be moved to the same line as the function name - same as othe functions are defined.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,324 @@
...
+      return [];
+    }
+  }
+
+  async function fetchTipWithdrawnTransactions(address: Encoded.AccountAddress, recent: boolean) {
+    try {
+      const response = await fetchJson(
+        `${nodeNetworkId.value}/cache/events/?address=${address}&event=TipWithdrawn${recent ? '&limit=5' : ''}`,
+      );
+      if (response.message) return [];
```
peronczyk, line 162:
```suggestion
      if (response.message) {
        return [];
      }
```

src/composables/transactionList.ts
```diff
@@ -0,0 +1,324 @@
...
+      const tipWithdrawnTransactions = (response as any[]).map(({
```
peronczyk, line 163: Instead of using `as any[]` here I suggest to type the source `response` constant.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,324 @@
...
+          contractId: contract,
+          type: Tag[Tag.ContractCallTx],
+        },
+        ...t,
+        microTime: new Date(t.createdAt).getTime(),
+        blockHeight: height,
+        claim: true,
+      }));
+
+      return tipWithdrawnTransactions as ITransaction[];
```
peronczyk, line 183: Instead of using `as` here I suggest to type the `tipWithdrawnTransactions` where it's defined.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,324 @@
...
+
+  watch(transactions, (value) => {
+    setLocalStorageItem(['transactionsState'], value);
+  }, {
+    deep: true, immediate: true,
+  });
+
+  watch(nodeNetworkId, () => {
+    transactions.value = {};
+  });
```
peronczyk, line 293: What about situations where this composable is not used by any component and user changed the network?

src/composables/transactionList.ts
```diff
@@ -0,0 +1,324 @@
...
+  computed, onMounted, ref, watch,
+} from 'vue';
+import { Encoded, Tag } from '@aeternity/aepp-sdk';
+import {
+  ITransaction,
+  ITransactionsState, ITransactionStateList,
+} from '../types';
+import { useMiddleware } from './middleware';
+import {
+  AEX9_TRANSFER_EVENT, fetchJson, getLocalStorageItem, setLocalStorageItem, TX_DIRECTION,
```
peronczyk, line 13: Please split this line.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,324 @@
+import { useStore } from 'vuex';
+import { orderBy, uniqBy } from 'lodash-es';
+import {
+  computed, onMounted, ref, watch,
+} from 'vue';
+import { Encoded, Tag } from '@aeternity/aepp-sdk';
+import {
+  ITransaction,
+  ITransactionsState, ITransactionStateList,
```
peronczyk, line 9: Please split this line.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,324 @@
+import { useStore } from 'vuex';
+import { orderBy, uniqBy } from 'lodash-es';
+import {
+  computed, onMounted, ref, watch,
```
peronczyk, line 4: Please split this line.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,324 @@
...
+} from 'vue';
+import { Encoded, Tag } from '@aeternity/aepp-sdk';
+import {
```
peronczyk, line 7:
```suggestion
import type {
```

src/composables/popupProps.ts
```diff
@@ -22,7 +23,7 @@ export function usePopupProps(): IPopupPropsState {
 
   const sender = computed((): IAccountOverview => ({
     name: popupProps.value?.app?.name,
-    address: popupProps.value?.app?.host,
+    address: popupProps.value?.app?.host as Encoded.AccountAddress,
```
peronczyk, line 26: Actually this is not correct. The `host` is an URL. This is like saying that `cat` is a `chair` :D

src/composables/transactionList.ts
```diff
@@ -0,0 +1,324 @@
...
+const generateEmptyTransactionState = (): ITransactionsState => ({
+  loaded: [],
+  nextPageUrl: '',
+  pending: {},
+  tipWithdrawnTransactions: [],
+});
+
+const transactions = ref<ITransactionStateList>(
+  getLocalStorageItem<ITransactionStateList>(LOCAL_STORAGE_KEY) || {},
+);
```
peronczyk, line 32: Please use the `createStorageRef` and then `useStorageRef` composable helper instead.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,324 @@
...
+const LOCAL_STORAGE_KEY = ['transactionsState'];
+
+const isInitalized = ref(false);
+
+const generateEmptyTransactionState = (): ITransactionsState => ({
+  loaded: [],
+  nextPageUrl: '',
+  pending: {},
+  tipWithdrawnTransactions: [],
+});
```
peronczyk, line 28: Why not just creating `ref`s here? Simplest example can be found in the `ui.ts` composable.

CharisSiarampalis:
Lgtm in general

src/types/index.ts
konradodo: i added TODO for this line

src/composables/transactionList.ts
konradodo: This is the function to create new empty element for list

src/composables/transactionList.ts
konradodo: i tested it, and i don't found situation where this is not working properly

src/composables/transactionList.ts
konradodo: i added TODO for this type.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,329 @@
...
+}
+
+function updateAccountTransactions(address: Encoded.AccountAddress, transaction: ITransaction) {
+  const transactionIndex = transactions.value[address]?.loaded?.findIndex(
+    (tr) => tr.hash === transaction.hash,
+  );
+
+  if (transactionIndex >= 0) {
+    transactions.value[address].loaded[transactionIndex] = transaction;
+  }
```
subhod-i, line 67: If not already exists then maybe you can push the transaction to the array. 
This case maybe rare.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,329 @@
...
+    }
+  }
+
+  async function waitTransactionMined(address: Encoded.AccountAddress, hash: string) {
+    try {
+      const aeSdk = await getAeSdk();
+      await aeSdk.poll(hash as Encoded.TxHash);
+      setPendingTransactionSentByHash(address, hash);
+    } catch (e) {
+      store.commit('removePendingTransactionByHash', { hash, network: nodeNetworkId.value });
```
subhod-i, line 103: Maybe you have placed this code by mistake.  Pending transactions have been removed from the store as I see.

tests/e2e/integration/other.js
```diff
@@ -1,5 +1,5 @@
 import { transactions } from '../../../src/popup/utils/testsConfig';
-
+//
```
subhod-i, line 2:
```suggestion
 
```

subhod-i (changes requested):

subhod-i:
@CharisSiarampalis please test the latest transactions widget when the view is in Multisig vaults and try to propose a TX with a multisig account and see if it appears in the wallet. There is a certain test that breaks the latest transaction widget @Liubov-crypto knows it well.

CharisSiarampalis:
> @CharisSiarampalis please test the latest transactions widget when the view is in Multisig vaults and try to propose a TX with a multisig account and see if it appears in the wallet. There is a certain test that breaks the latest transaction widget @Liubov-crypto knows it well.

I can see the multisig transaction on my wallet. I am not sure about the test that breaks it.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,338 @@
...
+
+import type {
+  ITransaction,
+  ITransactionsState,
+  ITransactionStateList,
+} from '../types';
+import { useMiddleware } from './middleware';
+import {
+  AEX9_TRANSFER_EVENT,
+  fetchJson, getLocalStorageItem, setLocalStorageItem,
```
peronczyk, line 19: Please split this line.

src/types/index.ts
```diff
@@ -443,11 +440,13 @@ export type ICommonTransaction = ITransaction | IActiveMultisigTransaction
 
 export type ITransactionsState = {
   loaded: ITransaction[];
-  nextPageUrl?: string;
+  nextPageUrl: string | null;
   pending: Record<string, ITransaction[]>;
   tipWithdrawnTransactions: ITransaction[];
 }
 
+export type ITransactionStateList = Record<Encoded.AccountAddress, ITransactionsState>
```
peronczyk, line 448: I suggest to rename this to `IAccountTransactionList` in order to make it more meaningfull.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,338 @@
...
+  TX_DIRECTION,
+} from '../popup/utils';
+import { useAccounts } from './accounts';
+import JsonBig from '../lib/json-big';
+import { useAeSdk } from './aeSdk';
+
+const LOCAL_STORAGE_KEY = 'transactions_state';
+
+const isInitalized = ref(false);
```
peronczyk, line 28: Typo
I strongly suggest to install some kind of "Code Spell Checker" plugin to your IDE.

src/types/index.ts
```diff
@@ -443,11 +440,13 @@ export type ICommonTransaction = ITransaction | IActiveMultisigTransaction
 
 export type ITransactionsState = {
   loaded: ITransaction[];
-  nextPageUrl?: string;
+  nextPageUrl: string | null;
   pending: Record<string, ITransaction[]>;
```
peronczyk, line 444: Please add a comment that the `string` here refers to node network ID.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,338 @@
...
+    nextPageUrl: '',
+    pending: {},
+    tipWithdrawnTransactions: [],
+  };
+}
+function getTransactionStateByAccountAddress(address: Encoded.AccountAddress) {
+  return transactions.value[address];
+}
+
+function getOrInitTransactionStateForAccount(address: Encoded.AccountAddress): ITransactionsState {
```
peronczyk, line 44: In my opinion the name of this function is too complex. I suggest to use `getAccountTransactionsState`.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,338 @@
...
+  const accountTransactionState = getTransactionStateByAccountAddress(address);
+
+  if (accountTransactionState) {
+    return accountTransactionState;
+  }
+
+  transactions.value[address] = generateEmptyTransactionState();
```
peronczyk, line 51: I think this is a bad pattern to just add empty data to the state for the reason of returning it. Why not just:
```
return (transactions.value[address])
  ? transactions.value[address]
  : {
    loaded: [],
    nextPageUrl: '',
    pending: {},
    tipWithdrawnTransactions: [],
  };
```

src/composables/transactionList.ts
```diff
@@ -0,0 +1,338 @@
...
+    if (!isLoggedIn) {
+      return [];
+    }
+    const { pending, loaded } = getOrInitTransactionStateForAccount(activeAccount.value.address);
+
+    return [...loaded, ...(pending[nodeNetworkId.value!] || [])];
+  });
+
+  function setPendingTransactionSentByHash(address: Encoded.AccountAddress, hash: string) {
+    getOrInitTransactionStateForAccount(address);
```
peronczyk, line 91: I think this line can be removed. Just do:
```
const accountTransactionsState = getAccountTransactionsState(address);
```
And then reuse this instead of repeating `transactions.value[address]`.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,338 @@
...
+const setTransactionsNextPage = (address: Encoded.AccountAddress, url: string) => {
+  transactions.value[address].nextPageUrl = url;
+};
+
+export function useTransactionList() {
+  const store = useStore();
+  const { nodeNetworkId, getAeSdk } = useAeSdk({ store });
+  const { activeAccount, isLoggedIn } = useAccounts({ store });
+
+  const allTransactionsForAccount = computed(() => {
```
peronczyk, line 81: `forAccount` doesn't tell for which account. So I suggest to follow pattern used by the `accounts` composable and rename this to `activeAccountAllTransactions`.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,338 @@
...
+    const { pending, loaded } = getOrInitTransactionStateForAccount(activeAccount.value.address);
+
+    return [...loaded, ...(pending[nodeNetworkId.value!] || [])];
+  });
+
+  function setPendingTransactionSentByHash(address: Encoded.AccountAddress, hash: string) {
+    getOrInitTransactionStateForAccount(address);
+    const index = transactions.value[address].pending[nodeNetworkId.value!]
+      ?.findIndex((transaction) => transaction.hash === hash);
+    transactions.value[address]
```
peronczyk, line 94: I suggest to run this line only if the index is different than `-1`. It can happen when for example there was no pending transaction with this hash or the transaction state for the account does not exist.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,338 @@
...
+    if (pendingTransactionForAccount.length) {
+      transactions.value[address].pending[nodeNetworkId.value!] = (
+        pendingTransactionForAccount.filter((transaction) => transaction.hash !== hash)
+      );
+    }
+  }
+
+  async function waitTransactionMined(address: Encoded.AccountAddress, hash?: string) {
+    try {
+      if (hash) {
```
peronczyk, line 112: I suggest to move this test before `try`. This way there will be no need to use exclamation mark in line 118.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,338 @@
...
+    waitTransactionMined(address, transaction.hash);
+  }
+
+  async function fetchTransactionsFromMiddleware(
+    transactionState: ITransactionsState,
+    address: Encoded.AccountAddress,
+    recent: boolean,
+    limit: number,
+  ) {
+    const { fetchFromMiddlewareCamelCased } = useMiddleware({ store });
```
peronczyk, line 139: I suggest to move this line to the beginning of the composable.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,338 @@
...
+    limit: number,
+    recent: boolean,
+    address: Encoded.AccountAddress,
+  ): Promise<ITransaction[]> {
+    const transactionState = getOrInitTransactionStateForAccount(address);
+
+    if (transactionState.nextPageUrl === null && !recent) {
+      return [];
+    }
+    const { getMiddleware } = useMiddleware({ store });
```
peronczyk, line 221: I suggest to move this line to the beginning of the composable.

peronczyk (changes requested):

src/popup/pages/TransactionDetails.vue
```diff
@@ -269,12 +269,14 @@ import {
 } from '../utils';
 import { ROUTE_NOT_FOUND } from '../router/routeNames';
 import type { ITransaction, TxFunctionRaw, INetwork } from '../../types';
+
```
peronczyk, line 272: I this line is not necessary.

peronczyk (changes requested):

tests/e2e/support/commands.js
```diff
@@ -91,6 +99,11 @@ Cypress.Commands.add('login', (options = {}, route, isMockingExternalRequests =
   cy.openPopup(async (contentWindow) => {
     /* eslint-disable-next-line no-param-reassign */
     contentWindow.localStorage.state = JSON.stringify(await getLoginState(options));
+
+    if (options.pendingTransaction) {
+      // eslint-disable-next-line no-param-reassign
+      contentWindow.localStorage['sh-wallet_transactions_state_ae_uat'] = JSON.stringify(preparePendingTransactionToLocalStorage(options.pendingTransaction));
+    }
```
peronczyk, line 106: In my opinion this is rather bad that the test case is creating its own local storage state. I think our logic should handle the lack of pending transactions in the local storage.

Maybe you could use `addCustomPendingTransactionForAccount` method instead?

tests/e2e/integration/aex2-popups.js
```diff
@@ -56,13 +54,6 @@ describe('Tests cases for AEX-2 popups', () => {
       .get('[data-cy=sender]')
       .should('be.visible')
       .should('contain', props2.app.host);
-
-    const props3 = popupProps[POPUP_TYPE_ACCOUNT_LIST];
-    cy.openAex2Popup(POPUP_TYPE_ACCOUNT_LIST)
-      .get('[data-cy=aepp]')
-      .should('be.visible')
-      .should('contain', props3.app.name)
-      .should('contain', props3.app.host);
```
peronczyk, line 65: @subhod-i also see this.

src/store/plugins/names.js
```diff
@@ -98,10 +106,9 @@ export default (store) => {
         state: { owned, pendingAutoExtendNames },
         rootGetters: { accounts },
         commit,
-        dispatch,
       }) {
         commit('setAreNamesFetching', true);
-        const getPendingNameClaimTransactions = (address) => dispatch('fetchPendingTransactions', address, { root: true })
+        const getPendingNameClaimTransactions = (address) => fetchPendingTransactions(address)
```
peronczyk, line 111:
```suggestion
        const fetchPendingNameClaimTransactions = (address) => fetchPendingTransactions(address)
```

peronczyk (changes requested):

tests/e2e/integration/aex2-popups.js
subhod-i: @konradodo We need this test too.

src/composables/transactionList.ts
konradodo: this place was changed, i close that comment

tests/e2e/support/commands.js
konradodo: Original test worked on localStorage. What do you think @CedrikNikita @subhod-i ?

src/composables/transactionList.ts
```diff
@@ -0,0 +1,338 @@
...
+      setLocalStorageItem([LOCAL_STORAGE_KEY, oldValue!], transactions.value);
+      transactions.value = getLocalStorageItem([LOCAL_STORAGE_KEY, value!]) || {};
+    }
+  });
+
+  watch(transactions, (value) => {
+    setLocalStorageItem([LOCAL_STORAGE_KEY, nodeNetworkId.value!], value);
+  }, { deep: true });
+
+  onMounted(async () => {
```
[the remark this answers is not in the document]
konradodo: When i used IIFE then getAeSdk not worked correctly, and application crashed then.

src/popup/utils/testsConfig.ts
```diff
@@ -44,15 +43,6 @@ export const popupProps: Record<string, IPopupConfig> = {
...
   },
-  [POPUP_TYPE_ACCOUNT_LIST]: {
-    type: POPUP_TYPE_ACCOUNT_LIST,
-    app: {
-      url: 'http://localhost:5000/aepp/aepp',
-      name: 'AEPP',
-      protocol: 'http:',
-      host: 'localhost',
-    },
-  },
```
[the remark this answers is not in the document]
konradodo: this is my mistake, i rollback this change

src/composables/transactionList.ts
```diff
@@ -0,0 +1,337 @@
...
+    }
+  }
+
+  async function fetchTransactions(
+    limit: number,
+    recent: boolean,
+    address: Encoded.AccountAddress,
+  ): Promise<ITransaction[]> {
+    await getAeSdk();
+    createTransactionsState(address);
```
peronczyk, line 215: This can be removed as we don't need to be sure the state exists.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,337 @@
...
+function getAccountTransactionsState(address: Encoded.AccountAddress): ITransactionsState {
+  return transactions.value[address] || generateEmptyTransactionState();
+}
+
+function createTransactionsState(address: Encoded.AccountAddress) {
+  if (!transactions.value[address]) {
+    transactions.value[address] = generateEmptyTransactionState();
+  }
+}
+function getTransactionByHash(address: Encoded.AccountAddress, hash: string) {
```
peronczyk, line 51: Please add empty line before this function.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,337 @@
...
+    pending: {},
+    tipWithdrawnTransactions: [],
+  };
+}
+
+function getAccountTransactionsState(address: Encoded.AccountAddress): ITransactionsState {
+  return transactions.value[address] || generateEmptyTransactionState();
+}
+
+function createTransactionsState(address: Encoded.AccountAddress) {
```
peronczyk, line 46: As this function has an `if` inside I think this can be renamed to `ensureAccountTransactionStateExists`.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,337 @@
...
+        const aeSdk = await getAeSdk();
+        await aeSdk.poll(hash as Encoded.TxHash);
+        setPendingTransactionSentByHash(address, hash);
+      } catch (e) {
+        removePendingTransactionByAccount(address, hash);
+      }
+    }
+  }
+
+  function addCustomPendingTransactionForAccount(
```
peronczyk, line 123: This function doesn't add new pending transaction but replaces the pending transactions with the new one. So I think the name of this function is misleading.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,337 @@
...
+  if (!transactions.value[address]) {
+    transactions.value[address] = generateEmptyTransactionState();
+  }
+}
+function getTransactionByHash(address: Encoded.AccountAddress, hash: string) {
+  return transactions.value[address]?.loaded.find((transaction) => transaction.hash === hash);
+}
+
+function updateAccountTransactions(address: Encoded.AccountAddress, transaction: ITransaction) {
+  createTransactionsState(address);
```
peronczyk, line 56: This can be removed as the next line searches data inside of the state.

src/composables/transactionList.ts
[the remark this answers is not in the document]
peronczyk: What you mean by "not worked correctly"?

peronczyk (changes requested):

src/composables/transactionList.ts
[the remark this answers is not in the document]
konradodo: await getAeSdk(); is never resolve.

src/composables/transactionList.ts
konradodo: if this object is not exist, a lot of things below not working properly.,like:
`    
transactions.value[address].loaded = uniqBy(
      [...transactions.value[address].loaded, ...preparedTransactions],
      'hash',
    );
`
or 

`
    const oldPendingTransactionForAccount: ITransaction[] = (
      transactions.value[address]?.pending[nodeNetworkId.value!] || []
    );
`

src/composables/transactionList.ts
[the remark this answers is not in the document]
peronczyk: Probably because you use `useStore` from the `vuex` here: https://github.com/aeternity/superhero-wallet/pull/2171/files#diff-dbd73d02753926bab2746dfd799dec8c1b05c3920dfb52003d5f02bb3539521dR32.

Here's a topic on this: https://www.reddit.com/r/vuejs/comments/mki721/is_it_possible_to_access_usestore_from_another/

tests/e2e/support/commands.js
```diff
@@ -91,6 +99,11 @@ Cypress.Commands.add('login', (options = {}, route, isMockingExternalRequests =
   cy.openPopup(async (contentWindow) => {
     /* eslint-disable-next-line no-param-reassign */
     contentWindow.localStorage.state = JSON.stringify(await getLoginState(options));
+
+    if (options.pendingTransaction) {
+      // eslint-disable-next-line no-param-reassign
+      contentWindow.localStorage['sh-wallet_transactions_ae_uat'] = JSON.stringify(preparePendingTransactionToLocalStorage(options.pendingTransaction));
```
peronczyk, line 105: The key for the local storage should be created out of the constants we have in the system.

src/composables/transactionList.ts
peronczyk: Uhhh, makes sense. Resolving.

tests/e2e/support/commands.js
```diff
@@ -5,7 +5,7 @@ import { ROUTE_ACCOUNT_DETAILS_TRANSACTIONS } from '../../../src/popup/router/ro
 import { STUB_CURRENCY, testAccount } from '../../../src/popup/utils/testsConfig';
 import {
   formatDate,
-  formatTime,
+  formatTime, LOCAL_STORAGE_PREFIX, NETWORK_TESTNET, TRANSACTIONS_LOCAL_STORAGE_KEY,
```
peronczyk, line 8: Please split this line.

src/types/index.ts
```diff
@@ -443,11 +442,13 @@ export type ICommonTransaction = ITransaction | IActiveMultisigTransaction
 
 export type ITransactionsState = {
   loaded: ITransaction[];
-  nextPageUrl?: string;
+  nextPageUrl: string | null; // this refers to node network ID.
```
peronczyk, line 445: In what way the `nextPageUrl` is related to `nodeNetworkId`?

peronczyk (approved):

tests/e2e/support/commands.js
```diff
@@ -31,7 +34,15 @@ export async function getLoginState({
     balance,
     ...(name && { names: { defaults: { [`${account.address}-ae_uat`]: name } } }),
     ...(pendingTransaction
-        && { transactions: { loaded: [], pending: { ae_uat: [pendingTransaction] } } }),
+      && { transactions: { loaded: [], pending: { ae_uat: [pendingTransaction] } } }),
```
subhod-i, line 37: `ae_uat` could be imported from constants

subhod-i (commented):
Code changes LGTM. The pipeline needs to be resolved. Since this PR touches a lot of file,  I suggest to have full regression testing @CharisSiarampalis @Liubov-crypto

src/composables/transactionList.ts
```diff
@@ -0,0 +1,328 @@
...
+
+    ensureAccountTransactionStateExists(address);
+    const transactionState = transactions.value[address];
+
+    if (transactionState.nextPageUrl === null && !recent) {
+      return [];
+    }
+
+    const [
+      normal,
```
peronczyk, line 222: `normal` is a too generic name in my opinion. I suggest to rename this to `regularTransactions`.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,328 @@
...
+      pending,
```
peronczyk, line 223: I suggest to rename this to `pendingTransactions`.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,328 @@
...
+      tip,
```
peronczyk, line 224: I suggest to rename this to `tipWithdrawnTransactions`.

src/popup/components/LatestTransactionsCard.vue
```diff
@@ -34,8 +34,8 @@ import {
   watch,
   ref,
 } from 'vue';
-import { useStore } from 'vuex';
 import { useRoute } from 'vue-router';
+import { useStore } from 'vuex';
```
peronczyk, line 38: This change gives no value but increase the possibility of future conflicts. I suggest to not apply this change.

src/composables/transactionList.ts
```diff
@@ -0,0 +1,328 @@
...
+        const aeSdk = await getAeSdk();
+        await aeSdk.poll(hash as Encoded.TxHash);
+        setPendingTransactionSentByHash(address, hash);
+      } catch (e) {
+        removePendingTransactionByAccount(address, hash);
+      }
+    }
+  }
+
+  function insertCustomPendingTransactionForAccount(
```
peronczyk, line 121: Previously I commented that word `add` does not reflect what this function does (replace the pending transactions with new one). `insert` is a synonym of `add` so it's still not valid in my opinion.

src/popup/components/TransactionList.vue
```diff
@@ -57,7 +60,7 @@ import {
   pipe,
   includesCaseInsensitive,
 } from '../utils';
-import { useDispatch, useGetter, useState } from '../../composables/vuex';
+import { useGetter, useState } from '../../composables/vuex';
```
peronczyk, line 63:
```suggestion
import { useGetter, useState } from '@/composables/vuex';
```

Liubov-crypto:
@konradodo please check it: Transactions from regular accounts are visible in multisig accounts: 

https://github.com/aeternity/superhero-wallet/assets/69896204/fe226acf-474e-444d-8f48-10b3c969b506

Accounts discovery is broken in FF extension:
tot

Liubov-crypto:
The issue with Transaction has been fixed. 
I've registered a separate bug with Accounts discovery. 

I found another issue: if I quick enter the amount I will be able to propose a Tx from vault with 0 balance. There is some delay in blocking Next button. This option is blocked on develop. @konradodo please check it: 

https://github.com/aeternity/superhero-wallet/assets/69896204/7ca8262a-d445-4042-86f4-3c2ef079735e

uval1

Also I tried to claim a name, Tx was successful, but on this branch Tx is in Pending status for a long time . On dev branch it's completed. Maybe it's also worth to check: 

https://github.com/aeternity/superhero-wallet/assets/69896204/da4e5f57-8692-48c5-b712-3ba5dce27d82

Liubov-crypto:
@konradodo  I found that issue with blocking Next button is reproducible on another PRs too, so I registered a separate issue for it: if I quick enter the amount I will be able to propose a Tx from vault with 0 balance. There is some delay in blocking Next button. This option is blocked on develop.

src/composables/latestTransactionList.ts
```diff
@@ -37,29 +33,38 @@ export function useLatestTransactionList({ store }: IDefaultComposableOptions) {
...
-      ...fetchedTransactions.value,
-    ];
+  const tokens = computed(() => store.state.fungibleTokens.tokens);
 
-    latestTransactions.value = uniqWith(transactions, (a, b) => (
+  const latestTransactions = computed(() => {
+    const allTransactions = Object.entries(transactions.value)
+      .map(([
+        accountAddress,
+        { loaded, pending }]) => [...(pending[nodeNetworkId.value!] || []), ...(loaded || [])]
```
pstachel, line 47: Please fix the formatting here - if you drop a bracket a the first line it should be also dropped on the last making the params more readable
```suggestion
      .map(([
        accountAddress,
        { loaded, pending },
      ]) => [...(pending[nodeNetworkId.value!] || []), ...(loaded || [])]
```

src/composables/transactionList.ts
```diff
@@ -0,0 +1,329 @@
...
+    }
+  }
+
+  async function waitTransactionMined(address: Encoded.AccountAddress, hash?: string) {
+    if (hash) {
+      try {
+        const aeSdk = await getAeSdk();
+        await aeSdk.poll(hash as Encoded.TxHash);
+        setPendingTransactionSentByHash(address, hash);
+      } catch (e) {
```
pstachel, line 114: you are not being consistent with the error variable name - you should stick either to `e` or `error` but I think in most places we use the second one

src/composables/transactionList.ts
```diff
@@ -0,0 +1,329 @@
...
+        await sdk.api.getPendingAccountTransactionsByPubkey(address)
+      );
+      const transactionState = getAccountTransactionsState(address);
+
+      return JsonBig.parse(JsonBig.stringify(
+        fetchedPendingTransaction?.transactions || [],
+      ))
+        .filter((transaction: ITransaction) => !transactionState.pending[nodeNetworkId.value!]
+          ?.find((tx) => tx?.hash === transaction?.hash))
+        .map((transaction: ITransaction) => ({ ...transaction, pending: true }));
```
pstachel, line 164: At first glance it looks like it has wrong indentation here - consider refactoring
```suggestion
        .filter(
          (transaction: ITransaction) => !transactionState.pending[nodeNetworkId.value!]?.find(
            (tx) => tx?.hash === transaction?.hash
          )
        )
        .map((transaction: ITransaction) => ({ ...transaction, pending: true }));
```

pstachel (approved):

tests/e2e/support/commands.js
```diff
@@ -29,9 +32,17 @@ export async function getLoginState({
     backedUpSeed,
     current: { network: network || 'Testnet' },
     balance,
-    ...(name && { names: { defaults: { [`${account.address}-ae_uat`]: name } } }),
+    ...(name && { names: { defaults: { [`${account.address}-${NETWORK_ID_TESTNET}`]: name } } }),
     ...(pendingTransaction
-        && { transactions: { loaded: [], pending: { ae_uat: [pendingTransaction] } } }),
+      && { transactions: { loaded: [], pending: { [NETWORK_ID_TESTNET]: [pendingTransaction] } } }),
```
CedrikNikita, line 37: The `transactions` variable no longer exists in the current `state`.

CedrikNikita (changes requested):
- This should also be removed.
https://github.com/aeternity/superhero-wallet/blob/b68de712db3ed0a704855b79e12d62d5fe2d1a27/src/popup/components/DefaultPagesRouter.vue#L26
- Issue with pending transactions still exists
`Error: Missing required param "pathMatch"`
While clicking on pending transaction in last transactions widget, that was fetched from the node
image

src/composables/transactionList.ts
```diff
@@ -0,0 +1,329 @@
...
+      }
+    }
+  }
+
+  function upsertCustomPendingTransactionForAccount(
+    address: Encoded.AccountAddress,
+    transaction: ITransaction,
+  ) {
+    ensureAccountTransactionStateExists(address);
+    transactions.value[address].pending[nodeNetworkId.value!] = [transaction];
```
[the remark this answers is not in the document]
konradodo: https://github.com/aeternity/superhero-wallet/pull/2171#discussion_r1276159007

src/composables/latestTransactionList.ts
```diff
@@ -140,8 +113,11 @@ export function useLatestTransactionList({ store }: IDefaultComposableOptions) {
     { deep: true },
   );
 
+  watch(() => store.state.names.owned, () => {
```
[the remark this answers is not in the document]
konradodo: this watch is not need anymore, i deleted it

src/composables/middleware.ts
```diff
@@ -2,14 +2,14 @@ import { computed, ref } from 'vue';
 import camelCaseKeysDeep from 'camelcase-keys-deep';
 import { load } from 'js-yaml';
 import { camelCase } from 'lodash-es';
-import { genSwaggerClient, mapObject } from '../lib/swagger';
-import { fetchJson, watchUntilTruthy } from '../popup/utils';
+import { fetchJson, watchUntilTruthy } from '@/popup/utils';
```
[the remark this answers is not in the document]
konradodo: mistake

cypress.json
```diff
@@ -6,7 +6,7 @@
   "viewportHeight": 660,
   "numTestsKeptInMemory": 1,
   "chromeWebSecurity": false,
-  "defaultCommandTimeout": 30000,
+  "defaultCommandTimeout": 40000,
```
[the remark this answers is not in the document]
konradodo: I reverted this change. Before something lagging the app and random tests had problem with this. Now looks everything ok.

src/popup/pages/TransactionDetails.vue
```diff
@@ -404,13 +410,22 @@ export default defineComponent({
...
+      let rawTransaction = getTransactionByHash(activeAccount.value.address, props.hash);
+
       if (!rawTransaction || rawTransaction.incomplete) {
         const middleware = await getMiddleware();
         try {
           rawTransaction = await middleware.getTx(props.hash);
         } catch (e) {
-          router.push({ name: ROUTE_NOT_FOUND });
+          // This case is for pending transaction
+          await fetchAllPendingTransactions();
```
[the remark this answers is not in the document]
konradodo: This endpoints not resolve all problem. I don't have any other idea. I sugerate merge this pr, and this problem resolving in separate task
https://aeternity-blockchain.atlassian.net/jira/software/c/projects/SW/boards/2?modal=detail&selectedIssue=SW-994&assignee=62f0d1a41323922c61e350e6

CedrikNikita (approved):

PR #2176, opened 2023-07-19
Fetch confirmed transactions on load

subhod-i:
I see the latest transaction widget is not shown when there is no pending transaction or there is no balance/network change. 
The wallet doesn't show confirmed past transactions. Hence on load, fetch past transactions once.


pstachel (approved):

Liubov-crypto:
I have pending tx in multisig - widget isn't working
e
pen

CharisSiarampalis:
> I have pending tx in multisig - widget isn't working e pen

Reproduceable on dev

CharisSiarampalis:
Latest multisig transaction now works

CedrikNikita (approved):

PR #2180, opened 2023-07-20
[SW-1078] Rename aepp-sdk instances to aeSdk standard

subhod-i:
as we import multiple SDKs for multichain architecture, Naming sdk as aeSdk( as per SDK repo), and other public examples is good.

Removed the `ISDK` type too.

### DIFF
--- a/src/background/wallet.js
+++ b/src/background/wallet.js
@@ -4,24 +4,24 @@ import { CONNECTION_TYPES } from '../popup/utils/constants';
 import { removePopup, getPopup } from './popupHandler';
 import { detectConnectionType } from './utils';
 import store from './store';
-import { useSdk } from '../composables';
+import { useAeSdk } from '../composables';
 
 window.browser = require('webextension-polyfill');
 
 let connectionsQueue = [];
 
 const addAeppConnection = async (port) => {
-  const { getSdk } = useSdk({ store });
-  const sdk = await getSdk();
+  const { getAeSdk } = useAeSdk({ store });
+  const aeSdk = await getAeSdk();
   const connection = new BrowserRuntimeConnection({ port });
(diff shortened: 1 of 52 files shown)


pstachel:
Shouldn't we also consider renaming the base file for sdk from 'sdk.ts' to 'aeSdk.ts'?

subhod-i:
> Shouldn't we also consider renaming the base file for sdk from 'sdk.ts' to 'aeSdk.ts'?

Of course. I missed it.

pstachel (approved):

src/lib/shSdkWallet.ts
```diff
@@ -20,7 +20,11 @@ type ISpendOptions = Omit<Parameters<typeof spend>[2], 'onAccount' | 'onNode'>
     payload?: Encoded.Any,
   }
 
-export class ShSdkWallet extends AeSdkWallet {
+/**
+ * Class extends `AeSdkWallet` from aepp-sdk-js
+ * provides flexibility to manage the accounts the way wallet would like to handle
+ */
+export class ShAeSdk extends AeSdkWallet {
```
peronczyk, line 27: 1. I suggest to use `AeShSdkWallet` or `ShAeSdkWallet` here to maintain the base class naming strategy.
2. Good practice is to name the file the same way as the exported class name.

peronczyk (changes requested):

src/lib/shSdkWallet.ts
subhod-i: Renamed it to `AeSdkSupehero` to maintain the base class naming strategy.

AeSdkBase -> AeSdkWallet -> AeSdkSuperhero

konradodo (approved):

peronczyk (approved):

PR #2183, opened 2023-07-21
[SW-1039] [Multichain] Components architecture

peronczyk:


src/popup/protocols/aeternity/components/AccountDetails.vue
peronczyk, line 1: Please bare in mind that this component was moved to `protocols/aeternity/components/` location. So every **protocol** needs to have its own `AccountDetails` components. This allows for high flexibility on what to display on the page.

pstachel (approved):

src/popup/protocols/aeternity/components/AccountDetails.vue
martinkaintas: What happens if we support 100 protocols? Would we need to have 100 different `AccountDetails` components?

peronczyk: Yes, exactly. But still it's better than having 100x ifs in one component. This inspired by the Micro Front-ends approach (https://micro-frontends.org/) - like a micro app within the app. Maybe the overall amount of code would be bigger, but it easier to maintain this kind of structure. What do you think @martinkaintas ?

src/popup/pages/ProtocolSpecificComponent.vue
```diff
@@ -0,0 +1,46 @@
...
+  },
+  props: {
+    componentName: { type: String, required: true },
+  },
+  setup(props) {
+    const store = useStore();
+    const { activeAccount } = useAccounts({ store });
+
+    const componentToDisplay = defineAsyncComponent(
+      () => import(`../protocols/${activeAccount.value.protocol}/components/${props.componentName}.vue`),
```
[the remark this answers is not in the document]
peronczyk: I revisited my code as your point raised some concerns, that not only pages would be built, but also some other components within the `protocols` directory. I changed the way how the components are loaded. What do you think?

src/popup/protocols/aeternity/components/AccountDetails.vue
martinkaintas: I totally agree that is way it is easier to maintain. I also hadn't seen your comment about using boilerplates (base components) which greatly reduces the need for duplicate code.

martinkaintas (approved):
Looks good to me!

src/popup/pages/ProtocolSpecificComponent.vue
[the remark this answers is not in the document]
peronczyk: Right now this is the list of files that are built:
image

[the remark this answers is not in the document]
davidyuk: src/popup/protocols/aeternity/views/index.ts
looks fine to me 👍

src/popup/components/Modals/TransferReceiveBase.vue
```diff
@@ -136,31 +139,25 @@ export default defineComponent({
...
-      ? activeMultisigAccountId.value
-      : activeAccount.value.address);
-
-    const availableTokens = computed<ITokenList>(
-      () => store.state.fungibleTokens.availableTokens,
-    );
-
     function getTokenInfoQuery(account?: string): Record<string, string> {
-      if (!amount.value || amount.value <= 0) return {};
+      if (!amount.value || +amount.value <= 0) return {};
```
konradodo, line 160:
```suggestion
      if (!amount.value || +amount.value <= 0) {
        return {}
      }
```

src/popup/router/routes.ts
```diff
@@ -113,7 +115,15 @@ export const routes: WalletAppRouteConfig[] = [
...
-        component: AccountDetails,
+        /**
+         * New way of loading pages specific to the protocol.
+         * `componentName` prop is required.
+         * TODO:
+         *   1. Use the same method with all child components
+         *   2. Remove this comment
+         */
+        component: ProtocolSpecificView,
+        props: { componentName: PROTOCOL_VIEW_ACCOUNT_DETAILS },
```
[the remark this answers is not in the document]
peronczyk: After reading this document I think those are 2 different things. The document is about passing params **when navigating** (eg when using `router.push`). Commented line is the **route definition** and it is totally perfect to pass some props this way: https://router.vuejs.org/guide/essentials/passing-props.html#object-mode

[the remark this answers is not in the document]
CedrikNikita: Right, thought in future you would call this with `router.push`, that will result in the problem above.

pstachel (approved):

konradodo (approved):

src/popup/router/routes.ts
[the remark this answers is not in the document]
peronczyk: I tried to test passing some stuff to this route when using `router.push` and nothing bad happened.

src/types/protocols.ts
subhod-i, line 1: In my PR https://github.com/aeternity/superhero-wallet/pull/2193, I have introduced files and types related in the protocol directory because it would be easy for someone from the community or the team can add a new protocol to the Superhero Wallet. I guess it's better to keep the protocol related types at one place.

src/popup/utils/constants.ts
```diff
@@ -22,6 +22,22 @@ export const MOBILE_WIDTH = 480;
...
+] as const;
+
+export const PROTOCOL_VIEW_ACCOUNT_DETAILS = 'AccountDetails';
+export const PROTOCOL_VIEW_TRANSFER_RECEIVE = 'TransferReceiveModal';
+
+export const PROTOCOL_VIEWS = [
+  PROTOCOL_VIEW_ACCOUNT_DETAILS,
+  PROTOCOL_VIEW_TRANSFER_RECEIVE,
+] as const;
+
```
subhod-i, line 40: Better naming may be needed here. eg: `DISTINCT_PROTOCOL_VIEWS` to indicate these are separate views for each protocol

src/composables/accounts.ts
```diff
@@ -8,13 +8,43 @@ import type {
...
+import { testAccount } from '../popup/utils/testsConfig';
 
 export function useAccounts({ store }: IDefaultComposableOptions) {
   // TODO in the future the state of the accounts should be stored in this composable
   const activeIdx = computed((): number => store.state.accounts?.activeIdx || 0);
-  const accounts = computed((): IAccount[] => store.getters.accounts || []);
+  const accounts = computed((): IAccount[] => {
+    const baseAccounts: IAccount[] = (store.getters.accounts || []).map((acc: IAccount) => ({
+      ...acc,
+      protocol: PROTOCOL_AETERNITY,
```
subhod-i, line 27: Instead of hardcoding the protocol type in the Account object, we can derive the Protocol type from the account address. We can introduce a method `identifyAccountProtocol` in this PR or in my per as per this comment https://github.com/aeternity/superhero-wallet/pull/2193/files#r1274653420

src/types/protocols.ts
peronczyk: Yes, I've seen this and I've made this comment: https://github.com/aeternity/superhero-wallet/pull/2193/files#r1276024101 to move the typing related things into `types/` directory. It's a common pattern to store global types in one place.

src/popup/router/routes.ts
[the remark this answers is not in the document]
CedrikNikita: Was referring to situation as such.
https://github.com/aeternity/superhero-wallet/pull/2135#pullrequestreview-1498001226
https://github.com/aeternity/superhero-wallet/pull/2090#discussion_r1210213569

src/composables/accounts.ts
peronczyk: Yes, we can derive it from the address. But in my opinion it's more readable and easier to debug to have the `protocol` property. Imagine you want to `console.log` the account data - you'll immediately know which protocol the account is.

Also as this property is not a thing that would change in time it's better to set it once instead of calculating it whenever we want to establish the protocol.

subhod-i: My proposal is to calculate it once and then store it as part of `IAccount` interface.

src/types/protocols.ts
subhod-i: Got it.

subhod-i: Makes sense to move `IProtocol` to `types/` directory and keep only protocol specific types in libs.

src/popup/router/routes.ts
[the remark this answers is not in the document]
peronczyk: OK, in case there will be some issues we'll try to pass the prop in a different way.

src/types/protocols.ts
peronczyk: Protocol specific types in my opinion should be stored in the protocol directory - to encapsulate all the logic/UI/types into one place.

subhod-i: Logic could be moved under the protocol directory but in this PR the protocol directories are under `popup/`

subhod-i (approved):
The approach is nice. I do not have any comments on the routing or the view, rest everything looks good.

src/types/protocols.ts
peronczyk: OK, I moved all `protocols` dir above the `popup` dir.

CedrikNikita (changes requested):
The destination branch is incorrect.
image

src/popup/router/routes.ts
davidyuk, line 126: It may be that ProtocolSpecificView is overengineered. Have you considered doing separate components like ViewAccountDetailsCommon which render a blockchain-specific component? For example, write a function that would generate a component like ViewAccountDetailsCommon by child components.
```js
// ViewAccountDetailsCommon.js
export default generateProtocolSpecificView({
  aeternity: () => import('...'),
  bitcoin: () => import('...'),
  ...
});
```
This way would be simpler to track dependencies: if ViewAccountDetailsCommon is used somewhere, then all child components are used as well. Otherwise, you need to track the usage of ProtocolSpecificView with a specific prop value. Also, ViewAccountDetailsCommon would be a bit clearer to read.

subhod-i:
> The destination branch is incorrect. image

to `feat/multichain`

src/popup/router/routes.ts
peronczyk: Yes, this was my first thought to do it this way. But during the implementation I found out that it will be bad for the scalability. This PR covers 2 views, but there will be more (10-15). With every view you would need to create one `Common` component, one boilerplate and one for every protocol. With 6 protocols it will be 8 components. So it's 8*15 = 120 components.

Also every common component would have to have the set of imports for every protocol. So whenever you add/remove protocol you would need to edit every common component.

My goal was to make the protocol installation as easy as possible, so I wanted to encapsulate all the imports inside of the protocol directory.

davidyuk: So, you would have 15 Common components that would combine 120 specific components. Sounds clear. The only thing to propose is to define all common components in a single file. It would take less than 200loc. It doesn't seem to be bad because all these components work the same, except they combine different child components.

davidyuk (approved):

src/popup/router/routes.ts
peronczyk: It's rather subjective but I'm opting for the current variant. The `ProtocolViewsConfig` type covers the required components for each of the protocols. And the one component that we have right now implements it nicely. But I'm open for the opinion from the rest of the team.
@pstachel , @subhod-i , @CedrikNikita , @konradodo , @ifaouibadi ?

pstachel: I'm open to both solutions - I was also thinking about what @davidyuk proposed here but maintaining it in the future may be a curse for us. On the other hand it will provide us clear structure of components and one entry point for each view. I feel that @peronczyk solution is cleaner (despite the fact that I'm not so keen on using component names as props) and immediately will show us that we need to deal with the protocol specific view.
What do you think of mixing both solutions - to have dynamic imports based on the protocol to prevent maintaining the if-s or switches inside base components and to use it inside mentioned component templates?

peronczyk: @pstachel can you give some code example of your proposal? I'm not sure if I understand it correctly.

pstachel: Sure, let's do it with an example of account details. 

We would have the base component AccountDetails.vue
```
<template>
  <ProtocolSpecificView :viewComponentName="PROTOCOL_VIEW_ACCOUNT_DETAILS" />
</template>
```

pstachel: But on a second thought I'm more convinced on your proposal as we wouldn't use any views directly in the templates, so we won't see the usage of `<ProtocolSpecificView />` tag

peronczyk: To unblock further development I decided to create separate spike for this topic: https://aeternity-blockchain.atlassian.net/browse/SW-1100

PR #2184, opened 2023-07-21
[SW-988] Validate transaction before signing

CedrikNikita:
labels in `advancedDetails` were missed in #2135, used the same approach as in #2158

### DIFF
--- a/src/popup/components/Modals/ConfirmTransactionSign.vue
+++ b/src/popup/components/Modals/ConfirmTransactionSign.vue
@@ -84,7 +84,7 @@
         <DetailsItem
           v-for="key in filteredTxFields"
           :key="key"
-          :label="$t('modals.confirmTransactionSign')[key]"
+          :label="TX_FIELDS_TO_DISPLAY[key]"
           :value="popupProps?.tx?.[key]"
           :class="{ 'hash-field': isHash(key) }"
         />
@@ -128,6 +128,7 @@ import BigNumber from 'bignumber.js';
 import { getExecutionCost } from '@aeternity/aepp-sdk';
 import ContractByteArrayEncoder from '@aeternity/aepp-calldata/src/ContractByteArrayEncoder';
 
+import { tg } from '../../../store/plugins/languages';
 import { RejectedByUserError } from '../../../lib/errors';
 import {
   DEX_TRANSACTION_TAGS,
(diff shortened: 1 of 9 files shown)


pstachel (approved):

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -321,7 +346,42 @@ export default defineComponent({
       popupProps.value?.reject(new RejectedByUserError());
     }
 
-    onMounted(async () => {
+    async function verifyTransaction() {
+      if (popupProps.value?.txBase64 && popupProps.value.tx) {
```
subhod-i, line 350: Does both params needed or is one of them optional?
I see `&&` check here.

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -321,7 +346,42 @@ export default defineComponent({
...
+        try {
+          verifying.value = true;
+          const { balance, updateBalances } = useBalances({ store });
+
+          await updateBalances();
```
subhod-i, line 355: Updating the balance of only the active account is required here not all accounts.

subhod-i (commented):
@Liubov-crypto @CharisSiarampalis Please test this PR with all types of Aepps transactions like confirm connect, message signing, transaction signing, contract create, contract call.

src/popup/components/Modals/ConfirmTransactionSign.vue
CedrikNikita: Shall we maybe consider it a future improvement, since other places where we are using `updateBalances` except `balances.ts` also requiring updating of only active account balance?

CedrikNikita: Asking for a `balances` from the `balances.ts` is already applying `updateBalances` every 3 seconds. 
- Maybe you think that current concerns are worth not to use `balances` composable at all here, and just to fetch the active account balance directly, even without sdk initialization?
- Or you are suggesting to add additional flag to the `balances` composable to fetch the balance only for an active account. But will the current logic in `initPollingWatcher` be able to load run `updateBalances` in one component and `updateActiveAccountBalance` in another component in the same scope (parent component).

subhod-i: > Maybe you think that current concerns are worth not to use balances composable at all here, and just to fetch the active account balance directly, even without sdk initialization?

This sounds more reasonable.
Again, not expecting over-engineering in the name of optimization but I guess we can achieve this without using `balances` composable with periodic watchers. Just saw the code, `balance` composable is being called `onMounted` hook. Hence it may not be adding significant delay.

peronczyk: I think this makes sense to use `aeSdk.getBalance` here instead of fetching balances for all the accounts.

CedrikNikita: Updating the balance of only the active account is required in every place we are using `updateBalances` currently. So what would the general suggestion be?

subhod-i (approved):
Code changes LGTM. PR must be tested by signing different transactions from Aepps.

CharisSiarampalis:
Lgtm

PR #2193, opened 2023-07-26
[SW-1019] Implement multichain protocol interface

subhod-i:
Approach
1. add new property `protocol` for `IAccount` which holds the account chain. Added default `aeternity` to existing accounts
2. Define interface `IProtocol` with common chain operations
3. Implement `AeternityProtocol, BitcoinProtocol, DogecoinProtocol implements IProtocol`  
4. Define a new composable `sdk.ts` which is crucial and exposes the Protocol instances to the project
5. Use `getSdk` method to send coins and to get the account address balance.

### DIFF
--- a/src/composables/accounts.ts
+++ b/src/composables/accounts.ts
@@ -8,9 +8,7 @@ import type {
   INetwork,
 } from '@/types';
 import { tg } from '@/store/plugins/languages';
-import { PROTOCOL_AETERNITY, PROTOCOL_BITCOIN } from '@/constants';
 import { getAccountNameToDisplay } from '@/utils';
-import { STUB_ACCOUNT } from '@/constants/stubs';
 import { AE_FAUCET_URL } from '@/protocols/aeternity/config';
 import { buildSimplexLink } from '@/protocols/aeternity/helpers';
 import { AeScan } from '@/protocols/aeternity/libs/AeScan';
@@ -18,30 +16,7 @@ import { AeScan } from '@/protocols/aeternity/libs/AeScan';
 export function useAccounts({ store }: IDefaultComposableOptions) {
   // TODO in the future the state of the accounts should be stored in this composable
   const activeIdx = computed((): number => store.state.accounts?.activeIdx || 0);
-  const accounts = computed((): IAccount[] => {
-    const baseAccounts: IAccount[] = (store.getters.accounts || []).map((acc: IAccount) => ({
-      ...acc,
-      protocol: PROTOCOL_AETERNITY,
-    }));
-
-    /**
-     * For the purpose of multichain Vue components architecture testing
-     * we add this dummy BTC account.
-     * TODO remove when PoC is approved
-     */
-    if (baseAccounts.length) {
-      baseAccounts.push({
-        idx: baseAccounts.length,
-        address: STUB_ACCOUNT.address,
-        protocol: PROTOCOL_BITCOIN,
-        publicKey: new Uint8Array(),
-        secretKey: new Uint8Array(),
-        showed: true,
-        type: '',
-      });
-    }
-    return baseAccounts;
-  });
+  const accounts = computed((): IAccount[] => store.getters.accounts || []);
   const accountsAddressList = computed(() => accounts.value.map((acc) => acc.address));
   const activeAccount = computed((): IAccount => accounts.value[activeIdx.value] || {});
   const isLoggedIn = computed(
--- a/src/composables/balances.ts
+++ b/src/composables/balances.ts
@@ -1,5 +1,4 @@
 import { computed } from 'vue';
-import { Encoded } from '@aeternity/aepp-sdk';
 import { isEmpty, mapValues } from 'lodash-es';
 import BigNumber from 'bignumber.js';
 import type {
@@ -14,6 +13,7 @@ import {
   AE_CONTRACT_ID,
 } from '@/protocols/aeternity/config';
 import { aettosToAe } from '@/protocols/aeternity/helpers';
+import { ProtocolAdapterFactory } from '@/lib/ProtocolAdapterFactory';
 import {
   createNetworkWatcher,
   createPollingBasedOnMountedComponents,
@@ -21,7 +21,6 @@ import {
 } from './composablesHelpers';
 import { useCurrencies } from './currencies';
 import { useAccounts } from './accounts';
-import { useAeSdk } from './aeSdk';
 
 type Balances = Record<string, Balance>;
 
@@ -43,7 +42,6 @@ const { useStorageRef } = createStorageRef<Balances>({}, LOCAL_STORAGE_BALANCES_
  * to live update the values. If no components are using it the polling stops.
  */
 export function useBalances({ store }: IDefaultComposableOptions) {
-  const { getAeSdk } = useAeSdk({ store });
   const { aeternityData } = useCurrencies();
   const { activeAccount, accounts } = useAccounts({ store });
 
@@ -68,10 +66,11 @@ export function useBalances({ store }: IDefaultComposableOptions) {
   }
 
   async function updateBalances() {
-    const aeSdk = await getAeSdk();
     const balancesPromises = accounts.value.map(
-      // TODO - type address in IAccount interface
-      ({ address }) => aeSdk.getBalance(address as Encoded.AccountAddress).catch((error) => {
+      async ({
+        address,
+        protocol,
+      }) => (ProtocolAdapterFactory.getAdapter(protocol)).getBalance(address).catch((error) => {
         if (!isNotFoundError(error)) {
           handleUnknownError(error);
         }
--- a/src/lib/ProtocolAdapterFactory.ts
+++ b/src/lib/ProtocolAdapterFactory.ts
@@ -1,0 +1,34 @@
+import type {
+  Class,
+  Protocol,
+} from '@/types';
+import { BaseProtocolAdapter } from '@/protocols/BaseProtocolAdapter';
+import { AeternityAdapter } from '@/protocols/aeternity/libs/AeternityAdapter';
+import { BitcoinAdapter } from '@/protocols/bitcoin/libs/BitcoinAdapter';
+import {
+  PROTOCOL_AETERNITY,
+  PROTOCOL_BITCOIN,
+} from '@/constants';
+
+/**
+ * Encapsulates helper methods related to multiple protocol instances
+ */
+export const ProtocolAdapterFactory = (() => {
+  const protocolAdapterClassRegistry: Record<Protocol, Class<BaseProtocolAdapter> > = {
+    [PROTOCOL_AETERNITY]: AeternityAdapter,
+    [PROTOCOL_BITCOIN]: BitcoinAdapter,
+  };
+
+  const protocolAdapterRegistry: Partial<Record<Protocol, BaseProtocolAdapter>> = {};
+
+  function getAdapter(protocol: Protocol): BaseProtocolAdapter {
+    if (!protocolAdapterRegistry[protocol]) {
+      protocolAdapterRegistry[protocol] = new protocolAdapterClassRegistry[protocol]();
+    }
+    return protocolAdapterRegistry[protocol]!;
+  }
+
+  return {
+    getAdapter,
+  };
+})();
--- a/src/store/getters.js
+++ b/src/store/getters.js
@@ -4,10 +4,11 @@ import { mnemonicToSeed } from '@aeternity/bip39';
 import { Tag } from '@aeternity/aepp-sdk';
 import { toShiftedBigNumber } from '@/utils';
 import {
+  ACCOUNT_HD_WALLET,
   NETWORK_MAINNET,
   NETWORK_TESTNET,
   NODE_STATUS_CONNECTED,
-  ACCOUNT_HD_WALLET,
+  PROTOCOL_AETERNITY,
   TX_DIRECTION,
 } from '@/constants';
 import { useAeSdk } from '@/composables';
@@ -26,9 +27,12 @@ export default {
   accounts({ accounts: { list } }, getters) {
     if (!getters.wallet) return [];
     return list
-      .map(({ idx, type, ...acc }) => ({
+      .map(({
+        idx, type, protocol, ...acc
+      }) => ({
         idx,
         type,
+        protocol: protocol || PROTOCOL_AETERNITY,
         ...acc,
         ...(type === ACCOUNT_HD_WALLET ? getHdWalletAccount(getters.wallet, idx) : {}),
       }))
--- a/src/store/modules/accounts/hdWallet.js
+++ b/src/store/modules/accounts/hdWallet.js
@@ -11,6 +11,7 @@ import {
   ACCOUNT_HD_WALLET,
   MODAL_CONFIRM_RAW_SIGN,
   MODAL_CONFIRM_TRANSACTION_SIGN,
+  PROTOCOL_AETERNITY,
 } from '@/constants';
 import { getHdWalletAccount, isTxOfASupportedType } from '@/protocols/aeternity/helpers';
 
@@ -47,7 +48,12 @@ export default {
     create({ state, commit }, isRestored = false) {
       commit(
         'accounts/add',
-        { idx: state.nextAccountIdx, type: ACCOUNT_HD_WALLET, isRestored },
+        {
+          idx: state.nextAccountIdx,
+          type: ACCOUNT_HD_WALLET,
+          isRestored,
+          protocol: PROTOCOL_AETERNITY,
+        },
         { root: true },
       );
       state.nextAccountIdx += 1;
--- a/src/store/modules/accounts/index.js
+++ b/src/store/modules/accounts/index.js
@@ -1,6 +1,9 @@
 /* eslint no-param-reassign: ['error', { 'ignorePropertyModificationsFor': ['state'] }] */
 
-import { ACCOUNT_HD_WALLET } from '@/constants';
+import {
+  ACCOUNT_HD_WALLET,
+  PROTOCOL_AETERNITY,
+} from '@/constants';
 import hdWallet from './hdWallet';
 
 // TODO: modules file is an object, because previously it contained more than one module,
@@ -13,7 +16,7 @@ export default {
 
   state: {
     list: [{
-      idx: 0, showed: true, type: ACCOUNT_HD_WALLET,
+      idx: 0, showed: true, type: ACCOUNT_HD_WALLET, protocol: PROTOCOL_AETERNITY,
     }],
     activeIdx: 0,
   },
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -36,6 +36,8 @@ export * from './filter';
 export * from './forms';
 export * from './protocols';
 
+export type Class<T> = new (...args: unknown[]) => T
+
 export type Dictionary<T = any> = Record<string, T>;
 
 export type Truthy<T> = T extends false | '' | 0 | null | undefined ? never : T;
--- a/src/protocols/BaseProtocolAdapter.ts
+++ b/src/protocols/BaseProtocolAdapter.ts
@@ -1,0 +1,6 @@
+/**
+ *  Represents common attributes and behavior of a protocol
+ */
+export abstract class BaseProtocolAdapter {
+  abstract getBalance(address: string): Promise<string>;
+}
--- a/src/protocols/aeternity/libs/AeternityAdapter.ts
+++ b/src/protocols/aeternity/libs/AeternityAdapter.ts
@@ -1,0 +1,15 @@
+/* eslint-disable class-methods-use-this */
+
+import { Encoded } from '@aeternity/aepp-sdk';
+import { useStore } from 'vuex';
+import { useAeSdk } from '@/composables/aeSdk';
+import { BaseProtocolAdapter } from '@/protocols/BaseProtocolAdapter';
+
+export class AeternityAdapter extends BaseProtocolAdapter {
+  async getBalance(address: Encoded.AccountAddress): Promise<string> {
+    const store = useStore();
+    const { getAeSdk } = useAeSdk({ store });
+    const sdk = await getAeSdk();
+    return sdk.getBalance(address);
+  }
+}
--- a/src/protocols/bitcoin/libs/BitcoinAdapter.ts
+++ b/src/protocols/bitcoin/libs/BitcoinAdapter.ts
@@ -1,0 +1,11 @@
+/* eslint-disable class-methods-use-this */
+
+import { BaseProtocolAdapter } from '@/protocols/BaseProtocolAdapter';
+
+export class BitcoinAdapter extends BaseProtocolAdapter {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async getBalance(address: string): Promise<string> {
+    // TODO: Implement this with bitcoin adapter
+    throw new Error('Method not implemented.');
+  }
+}


src/types/index.ts
```diff
@@ -151,6 +151,7 @@ export interface IAccount {
   secretKey: Uint8Array
   showed: boolean
   type: string
+  protocol: string,
```
peronczyk, line 154: The lines (properties) of the interface should be ended with semi (`;`).

src/composables/sdk.ts
```diff
@@ -0,0 +1,39 @@
...
+
+let sdkRegistry: Record<string, IProtocol>;
+
+export function useSdk({ store }: IDefaultComposableOptions) {
+  const { activeAccount } = useAccounts({ store });
+
+  sdkRegistry = {
+    aeternity: new AeternityProtocol(store),
+    bitcoin: new BitcoinProtocol(),
+    dogecoin: new DogecoinProtocol(),
```
subhod-i, line 18: The `keys` could be derived from constants. hardcoded here just to get the approach reviewed.

src/store/getters.js
```diff
@@ -29,6 +29,7 @@ export default {
         type,
         ...acc,
         ...(type === ACCOUNT_HD_WALLET ? getHdWalletAccount(getters.wallet, idx) : {}),
+        protocol: 'aeternity',
```
subhod-i, line 32: Not really needed. Chain type can be derived from the account address.
Aeternity: `ae_....` 
Bitcoin: `bc1......`
Dogecoin: `D....`

src/types/index.ts
subhod-i: Fixed it.

src/composables/sdk.ts
peronczyk, line 1: I'm strongly against this kind of approach of generic getters for the SDK. Mostly because all the protocols has different features.

src/composables/sdk.ts
peronczyk, line 18: This is bad because whenever someone will run `useSdk` new instances of all the protocol classes will be created. Major performance issue.

peronczyk (changes requested):

src/composables/sdk.ts
subhod-i: The inspiration for this approach is the proposed UX for the Multi-chain, if all of the protocols have the same `Send Form`, `Balance view`, and `Fee` view then what would be special to each protocol? 
Aeternity protocol I understand and that is why I have kept the `useAeSdk` composable. The rest could be the same.

src/composables/sdk.ts
```diff
@@ -0,0 +1,39 @@
...
+import { AeternityProtocol } from '../lib/protocols/AeternityProtocol';
+import { BitcoinProtocol } from '../lib/protocols/BitcoinProtocol';
+import { DogecoinProtocol } from '../lib/protocols/DogecoinProtocol';
+
+let sdkRegistry: Record<string, IProtocol>;
+
+export function useSdk({ store }: IDefaultComposableOptions) {
+  const { activeAccount } = useAccounts({ store });
+
+  sdkRegistry = {
```
peronczyk, line 15: This needs to be typed from the start.

src/composables/sdk.ts
peronczyk: The PR draft I created doesn't have all the changes because I didn't want to go to deep to avoid doing changes that could be not accepted. But if the Vue components structure changes would be approved a lot of views, like the Send Form would need to be moved into the protocol specific directory.
Maybe I should  add this change to show the bigger picture. Give me some time please and I'll inform you when I'll finish.

ifaouibadi (approved):

src/lib/protocols/AeternityProtocol.ts
peronczyk, line 1: To

src/lib/protocols/IProtocol.ts
peronczyk, line 1: I suggest to move this file here to have one single source of types (`types/` dir): https://github.com/aeternity/superhero-wallet/pull/2183/files#diff-d305aadf0cf64545c1401728006b6b098ddb9bf0627d8bdeed26f1e414f48d8d

peronczyk (changes requested):

src/lib/protocols/IProtocol.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Transaction fee varies to control the transaction confirmation speed slow, normal and fast
+ */
+export interface SpendTransactionFee {
+  low: string;
+  average: string;
+  max: string;
+}
+
+export interface IProtocol {
```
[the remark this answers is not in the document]
davidyuk: Have you considered to use abstract base class? Though I'm not sure about advantages comparing to `interface`

src/lib/protocols/AeternityProtocol.ts
```diff
@@ -0,0 +1,39 @@
...
+    const sdk = await getAeSdk();
+    return sdk.getBalance(address);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async estimateFeeForSpendTransaction(amount: string): Promise<SpendTransactionFee> {
+    const { getAeSdk } = useAeSdk({ store: this.store });
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const sdk = await getAeSdk();
+    // TODO: calculate Fee using the SDK/Helper functions
```
davidyuk, line 32: Currently, the minimum fee would be `low`, `average`, and `max` 😃

src/store/getters.js
peronczyk: I think it's better to have the `protocol` property instead of calculating it every time we need to know the exact protocol. Here's me comment on this: https://github.com/aeternity/superhero-wallet/pull/2183/files#r1276211607

src/lib/protocols/IProtocol.ts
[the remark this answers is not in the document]
subhod-i: I don't see any common behavior implementation among the protocols. Hence no abstract class. `interface` assuming we won't be relying on middlewares for these protocols in the future because otherwise these common middleware APIs could be part of these protocol files.

src/lib/protocols/AeternityProtocol.ts
subhod-i: Sorry, I didn't get your comment.

src/composables/balances.ts
```diff
@@ -69,10 +68,9 @@ export function useBalances({ store }: IDefaultComposableOptions) {
   }
 
   async function updateBalances() {
-    const aeSdk = await getAeSdk();
     const balancesPromises = accounts.value.map(
       // TODO - type address in IAccount interface
-      ({ address }) => aeSdk.getBalance(address as Encoded.AccountAddress).catch((error) => {
+      ({ address, protocol }) => getProtocolSdk(protocol).getBalance(address).catch((error) => {
```
peronczyk, line 73: By the way the comment above can be removed as it's already typed.

subhod-i: Haha. Sure

src/composables/sdk.ts
subhod-i: Already typed https://github.com/aeternity/superhero-wallet/blob/65e74e6f22e6b843e1001d924f8a7be0693013eb/src/composables/sdk.ts#L17

src/lib/protocols/IProtocol.ts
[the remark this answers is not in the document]
subhod-i: @peronczyk renamed it to `IChainProtocol` and this file has been moved to the `types/` directory

src/store/getters.js
davidyuk: you can define a `protocol` getter based on address, so it would behave like a property 🙃

src/lib/protocols/IProtocol.ts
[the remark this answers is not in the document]
davidyuk: > I don't see any common behavior implementation among the protocols

This is why it is abstract, not a usual class.

Not sure how middleware is related to this topic.

As I know, interfaces a solution for multiple inheritance, but it is not the case here, so by default I would use an abstract class. You would get ability to use `instanceof ChainProtocol` in runtime and a simpler naming (without `I` 🙂).

src/lib/protocols/AeternityProtocol.ts
davidyuk: The fee calculated by sdk can be used for all cases

src/types/protocol.ts
peronczyk, line 1: In my PR the name of the file uses plural form ("protocols") as it refers not to one protocols, but to all of them in general. Same as the `forms` file.

src/types/protocol.ts
```diff
@@ -0,0 +1,25 @@
...
+  low: string;
+  average: string;
+  max: string;
+}
+
+export interface IChainProtocol {
+  name: string;
+  symbol: string;
+  derivationPath: string;
+
```
peronczyk, line 17: This line is not necessary.

src/types/protocol.ts
```diff
@@ -0,0 +1,25 @@
+import { PROTOCOLS_SUPPORTED } from '../popup/utils';
+
+export type Protocol = typeof PROTOCOLS_SUPPORTED[number];
+/**
```
peronczyk, line 4: Please add empty line before this one.

src/popup/utils/utils.ts
```diff
@@ -704,3 +706,12 @@ export function checkImageAvailability(url: string): Promise<boolean> {
...
     .catch(() => false);
 }
+
+export function deriveProtocolFromAddress(address: string): Protocol {
+  switch (address.charAt(0).toLowerCase()) {
+    case 'a': return PROTOCOLS_SUPPORTED[0];
+    case 'b': return PROTOCOLS_SUPPORTED[1];
+    case 'd': return PROTOCOLS_SUPPORTED[2];
+    default: throw new Error('Unsupported protocol');
+  }
```
peronczyk, line 716: In my opinion checking only the first letter of the `address` is not a good way to detect the protocol. For the AE account addresses we have the `checkAddress` utility function.

src/composables/sdk.ts
```diff
@@ -0,0 +1,49 @@
...
+  const { activeAccount } = useAccounts({ store });
+
+  if (!isSdkRegistryReady) {
+    sdkRegistry = {
+      [PROTOCOL_AETERNITY]: new AeternityProtocol(store),
+      [PROTOCOL_BITCOIN]: new BitcoinProtocol(),
+      [PROTOCOL_DOGECOIN]: new DogecoinProtocol(),
+    };
+  }
+  isSdkRegistryReady = true;
```
peronczyk, line 29: I still think this not good to initialize all classes whenever the `useSdk` composable is fired. What about situations where user opens a component, which uses this composable, but user didn't take any action that requires any of the protocol SDKs?

I think the libs should be lazy loaded only when needed.

src/composables/sdk.ts
```diff
@@ -0,0 +1,49 @@
+import type {
+  IDefaultComposableOptions,
+  Protocol,
+} from '../types';
+import { useAccounts } from './accounts';
+import { IChainProtocol } from '../types';
```
peronczyk, line 6: This can be added to already existing types imports at the beginning of the file.

peronczyk (changes requested):

src/popup/utils/utils.ts
subhod-i: This code is no longer needed hence removed.

src/composables/sdk.ts
subhod-i: Done.

src/composables/sdk.ts
subhod-i: Sure. I will add the `initSdk` method as part of `getSdk` method

src/types/protocol.ts
subhod-i: Fixed

src/types/protocol.ts
subhod-i: Got it. This file has been deleted.

src/store/getters.js
subhod-i: I have completely removed the `deriveProtocolFromAddress` method as it is a must to know the protocol upfront to derive the account from seed to use the appropriate derivation path.

src/lib/protocols/IProtocol.ts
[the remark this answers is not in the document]
subhod-i: @davidyuk I will replace the interface with the Abstract class.

src/composables/accounts.ts
```diff
@@ -9,42 +9,16 @@ import type {
...
-        idx: baseAccounts.length,
-        address: testAccount.address,
-        protocol: PROTOCOL_BITCOIN,
-        publicKey: new Uint8Array(),
-        secretKey: new Uint8Array(),
-        showed: true,
-        type: '',
-      });
-    }
-    return baseAccounts;
```
subhod-i, line 46: Removed it cause the Bitcoin adaptor is not yet implemented.

src/store/getters.js
```diff
@@ -26,9 +27,12 @@ export default {
   accounts({ accounts: { list } }, getters) {
     if (!getters.wallet) return [];
     return list
-      .map(({ idx, type, ...acc }) => ({
+      .map(({
+        idx, type, protocol, ...acc
+      }) => ({
         idx,
         type,
+        protocol: protocol || PROTOCOL_AETERNITY,
```
subhod-i, line 35: The existing accounts do not have any protocol type and the existing accounts are always Aeternity accounts.

src/store/modules/accounts/hdWallet.js
```diff
@@ -48,7 +49,12 @@ export default {
     create({ state, commit }, isRestored = false) {
       commit(
         'accounts/add',
-        { idx: state.nextAccountIdx, type: ACCOUNT_HD_WALLET, isRestored },
+        {
+          idx: state.nextAccountIdx,
+          type: ACCOUNT_HD_WALLET,
+          isRestored,
+          protocol: PROTOCOL_AETERNITY,
```
subhod-i, line 56: Ideally, protocol is a function parameter as we introduce multiple buttons to create protocol-specific accounts.

src/store/modules/accounts/index.js
```diff
@@ -13,7 +16,7 @@ export default {
 
   state: {
     list: [{
-      idx: 0, showed: true, type: ACCOUNT_HD_WALLET,
+      idx: 0, showed: true, type: ACCOUNT_HD_WALLET, protocol: PROTOCOL_AETERNITY,
```
subhod-i, line 19: The first account is always Aeternity.

src/composables/sdk.ts
```diff
@@ -0,0 +1,66 @@
...
+import { DogecoinProtocol } from '@/protocols/dogecoin/DogecoinProtocol';
+import {
+  PROTOCOL_AETERNITY,
+  PROTOCOL_BITCOIN,
+  PROTOCOL_DOGECOIN,
+} from '@/popup/utils';
+import { watchUntilTruthy } from '@/utils';
+import { useAccounts } from './accounts';
+
+const isSdkRegistryReady = ref(true);
```
peronczyk, line 18: If it's not returned by the composable it doesn't need to be reactive - use `let` instead.

src/protocols/aeternity/AeternityProtocol.ts
peronczyk, line 1: Please put this file in additional `lib` directory (`aeternity/lib`).

src/protocols/bitcoin/BitcoinProtocol.ts
peronczyk, line 1: Please put this file in additional `lib` directory.

src/protocols/dogecoin/DogecoinProtocol.ts
peronczyk, line 1: Please put this file in additional `lib` directory.

src/protocols/aeternity/AeternityProtocol.ts
```diff
@@ -0,0 +1,47 @@
...
+import { PROTOCOL_AETERNITY } from '../../popup/utils';
+
+type AeternityAddress = Encoded.AccountAddress;
+
+export class AeternityProtocol implements BaseProtocol {
+  static title = PROTOCOL_AETERNITY;
+
+  static symbol = 'AE';
+
+  static derivationPath = 'm/44h/457h'
```
peronczyk, line 15: Missing semi.

src/composables/sdk.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  function initSdk(protocol: Protocol): BaseProtocol {
+    switch (protocol) {
+      case PROTOCOL_AETERNITY:
+        return new AeternityProtocol(store);
+      case PROTOCOL_BITCOIN:
+        return new BitcoinProtocol();
+      case PROTOCOL_DOGECOIN:
+        return new DogecoinProtocol();
```
peronczyk, line 42: Instead of building large switch...case I suggest to first create a constant that would store references to all classes, eg.:
```
const protocolClassReferences: Record<Protocol, BaseProtocol> = {
  aeternity: AeternityProtocol,
  ...
};
```
and then returning it:
```
const protocolClass = protocolClassReferences[protocol];
if (protocolClass) {
  return new protocolClass(store);
} else {
  throw...
}
```
I see that not all protocol classes requires the store argument, but I think the abstract class should force not only the same internals, but also the arguments. Soon we will remove the Vuex so it won't be a problem.

src/composables/sdk.ts
```diff
@@ -0,0 +1,66 @@
...
+import {
+  PROTOCOL_AETERNITY,
+  PROTOCOL_BITCOIN,
+  PROTOCOL_DOGECOIN,
+} from '@/popup/utils';
+import { watchUntilTruthy } from '@/utils';
+import { useAccounts } from './accounts';
+
+const isSdkRegistryReady = ref(true);
+const sdkRegistry: Partial<Record<Protocol, BaseProtocol>> = {};
```
peronczyk, line 19: This is not an `sdkRegistry` but more like a `protocolClassRegistry`.

src/composables/sdk.ts
```diff
@@ -0,0 +1,66 @@
...
+      throw new Error('Protocol instance already exists in the registry');
+    }
+    sdkRegistry[protocol] = instance;
+  }
+
+  function getSdkFromRegistry(protocol: Protocol): BaseProtocol | undefined {
+    return sdkRegistry[protocol];
+  }
+
+  function initSdk(protocol: Protocol): BaseProtocol {
```
peronczyk, line 35: This is not initiating the SDK, but the protocol class so I think the name is wrong.

src/composables/sdk.ts
```diff
@@ -0,0 +1,66 @@
...
+  const { activeAccount } = useAccounts({ store });
+
+  function registerSdk(protocol: Protocol, instance: BaseProtocol) {
+    if (sdkRegistry[protocol]) {
+      throw new Error('Protocol instance already exists in the registry');
+    }
+    sdkRegistry[protocol] = instance;
+  }
+
+  function getSdkFromRegistry(protocol: Protocol): BaseProtocol | undefined {
```
peronczyk, line 31: This is not returning the SDK but the protocol class.

src/popup/components/TransferReview.vue
peronczyk, line 1: This component will be moved into `@/protocols/aeternity/components` so it won't need to use the common sdk - it will use directly the aeSdk.

src/protocols/aeternity/AeternityProtocol.ts
subhod-i: Sure

src/popup/components/ProtocolSpecificView.vue
```diff
@@ -37,6 +37,7 @@ import InfoBox from './InfoBox.vue';
 const views: Record<Protocol, ProtocolViewsConfig> = {
   aeternity: aeternityViews,
   bitcoin: bitcoinViews,
+  dogecoin: bitcoinViews, // TODO: implement dogecoin view
```
peronczyk, line 40: For now I suggest to postpone this.

src/lib/protocols/IProtocol.ts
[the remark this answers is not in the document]
subhod-i: The file has been moved here [src/protocols/BaseProtocol.ts](https://github.com/aeternity/superhero-wallet/pull/2193/files#diff-f6bde9421eb2156fa45d884fc21678748e1c4917a1138bf7d32a9182a0c9bdcd) 
@davidyuk Please have a look

src/lib/protocols/AeternityProtocol.ts
subhod-i: This diff is no longer in the PR.

src/composables/sdk.ts
subhod-i: Removed it.

src/composables/sdk.ts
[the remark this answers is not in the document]
subhod-i: Renamed it to `protocol` and  `useProtocol`

src/popup/components/TransferReview.vue
subhod-i: Rolled back the changes.

src/popup/components/ProtocolSpecificView.vue
subhod-i: Done.

src/protocols/BaseProtocol.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ *  Represents common attributes and behavior of a protocol
+ */
+export abstract class BaseProtocol {
+  static title: string;
+
+  static symbol: string;
+
+  static derivationPath: string;
```
davidyuk, line 9: I would keep this on sdk side and add a method to create an account instead.

src/protocols/bitcoin/libs/BitcoinProtocol.ts
```diff
@@ -0,0 +1,30 @@
...
+type BitcoinAddress = `bc_${string}`;
+
+export class BitcoinProtocol extends BaseProtocol {
+  static title = PROTOCOL_BITCOIN;
+
+  static symbol = '₿';
+
+  static derivationPath = 'm/44h/0h';
+
+  async spend(amount: string, recipientId: BitcoinAddress, options: any): Promise<string> {
```
davidyuk, line 19: Sender address/account would be provided in options?

src/composables/protocol.ts
```diff
@@ -0,0 +1,56 @@
...
+  PROTOCOL_BITCOIN,
+} from '@/constants';
+import { watchUntilTruthy } from '@/utils';
+import { useAccounts } from './accounts';
+
+let isProtocolRegistryReady = true;
+
+const protocolClassReferences: Record<Protocol, Class<BaseProtocol> > = {
+  [PROTOCOL_AETERNITY]: AeternityProtocol,
+  [PROTOCOL_BITCOIN]: BitcoinProtocol,
```
davidyuk, line 20: `Class<BaseProtocol>` looks too meta, maybe just
```suggestion
const protocolClassReferences: Record<Protocol, () => BaseProtocol> = {
  [PROTOCOL_AETERNITY]: () => new AeternityProtocol(),
  [PROTOCOL_BITCOIN]: () => new BitcoinProtocol(),
```

subhod-i: Nice. Thanks.

src/protocols/bitcoin/libs/BitcoinProtocol.ts
subhod-i: I am plannning to address to redinfe this function prototype when writing bitcoin adapter methods.

src/composables/protocol.ts
peronczyk: I don't think the change suggested by Denys is good for us. Storing the class instead of the callback that returns new instance gives us possibility to access the static properties of the class.

src/protocols/BaseProtocol.ts
subhod-i: I kept it here so that I can achieve this https://github.com/aeternity/superhero-wallet/pull/2219/files#diff-629f43758798f7d5446cb5e8fc37a20d33204b606127a338f28924376b805d71R22
by using existing methods from the SDK to generate private keys for Bitcoin.

src/composables/protocol.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  function getProtocolFromRegistry(protocol: Protocol): BaseProtocol | undefined {
+    return protocolInstanceRegistry[protocol];
+  }
+
+  /**
+   * get SDK of given protocol or active account protocol
+   */
+  async function getProtocolInstance(targetProtocol?: Protocol): Promise<BaseProtocol> {
```
peronczyk, line 41: Why `async`? The Base class has no constructor at all so instantiation is synchronous.

src/composables/protocol.ts
```diff
@@ -0,0 +1,56 @@
...
+import { AeternityProtocol } from '@/protocols/aeternity/libs/AeternityProtocol';
+import { BitcoinProtocol } from '@/protocols/bitcoin/libs/BitcoinProtocol';
+import {
+  PROTOCOL_AETERNITY,
+  PROTOCOL_BITCOIN,
+} from '@/constants';
+import { watchUntilTruthy } from '@/utils';
+import { useAccounts } from './accounts';
+
+let isProtocolRegistryReady = true;
```
peronczyk, line 16: Why we need this? The protocol classes are synchronous so no waiting is required.
Also what actually means "protocol registry ready"? When it's full, or at least one protocol is instantiated?

src/composables/protocol.ts
```diff
@@ -0,0 +1,56 @@
...
+};
+const protocolInstanceRegistry: Partial<Record<Protocol, BaseProtocol>> = {};
+
+export function useProtocol({ store }: IDefaultComposableOptions) {
+  const { activeAccount } = useAccounts({ store });
+
+  function registerProtocol(protocol: Protocol, instance: BaseProtocol) {
+    if (protocolInstanceRegistry[protocol]) {
+      throw new Error('Protocol instance already exists in the registry');
+    }
```
peronczyk, line 30: I think there's too much logic here. Why not just create the `getProtocolInstance` without using the `registerProtocol` method at all? Eg.:
```
function getProtocolInstance(targetProtocol?: Protocol): BaseProtocol {
  const protocol = targetProtocol || activeAccount.value.protocol;
  const registeredInstance = protocolInstanceRegistry[protocol];
  if (!registeredInstance) {
    const classReference = protocolClassReferences[protocol];
    if (classReference) {
      protocolInstanceRegistry[protocol] = new protocolClassReferences[protocol]();
      return protocolInstanceRegistry[protocol];
    } else {
      throw new Error('...');
    }
  }
  return registeredInstance;
}
```
Not tested, but in general skips the possibility of trying to register protocol that already is instantiated.

src/composables/protocol.ts
subhod-i: Yes but we https://github.com/aeternity/superhero-wallet/blob/48716742e58d2426d8b03743da7a85c4c494a502/src/composables/protocol.ts#L44

If not then when we load the app many components would call `getProtocolInstance` at once and we might end up creating multiple instances of the class. 

I can keep it very simple but it would create a class instance every time but that's fine I guess as it's stateless.
```js
 function getProtocolInstance(targetProtocol?: Protocol): BaseProtocol {
    const protocol = targetProtocol || activeAccount.value.protocol;
    return new (protocolClassReferences[protocol])();
  }
```

src/composables/protocol.ts
peronczyk, line 1: As this file doesn't hold any reactive stuff, that can be consumed by the components I don't think this file can be considered as a `composable`. I suggest to maybe convert it into some kind of singleton. See for example: `src/lib/FramesConnection.ts`.

src/composables/protocol.ts
peronczyk: Yeah, but it's totally wrong. The `watchUntilTruthy` requires reactive property. `isProtocolRegistryReady` is not watchable. The code you created works only because the class instantiation is immediate.

peronczyk: > I can keep it very simple but it would create a class instance every time
No, it won't because the instantiation is immediate and the registry is filled with the instance immediate after calling the `getProtocolInstance` :)

src/protocols/aeternity/libs/AeternityProtocol.ts
```diff
@@ -0,0 +1,30 @@
...
+import { PROTOCOL_AETERNITY } from '@/constants';
+
+type AeternityAddress = Encoded.AccountAddress;
+
+export class AeternityProtocol extends BaseProtocol {
+  static title = PROTOCOL_AETERNITY;
+
+  static symbol = 'AE';
+
+  static derivationPath = 'm/44h/457h';
```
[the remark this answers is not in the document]
peronczyk: I already mentioned this in our call that it totally makes sense. So for example accessing the coin data this way would be awesome:
```
const assetSymbol = AeterntyProtocol.symbol
```

Maybe consider using UPPERCASE for this static properties, so we would access them like constants: `AeternityProtocol.SYMBOL`. What do you think?

src/protocols/aeternity/libs/AeternityProtocol.ts
```diff
@@ -0,0 +1,30 @@
+import { Encoded } from '@aeternity/aepp-sdk';
+import { useStore } from 'vuex';
+import { useAeSdk } from '@/composables/aeSdk';
+import { BaseProtocol } from '@/protocols/BaseProtocol';
+import { PROTOCOL_AETERNITY } from '@/constants';
+
+type AeternityAddress = Encoded.AccountAddress;
+
+export class AeternityProtocol extends BaseProtocol {
+  static title = PROTOCOL_AETERNITY;
```
peronczyk, line 10: Do we really need this?
In my opinion this class should be self enclosed, so it should not reach for any global stuff. Only types.

src/composables/protocol.ts
peronczyk: Not sure how you see this as the `/protocols/index.ts`. Importing it would be little weird.
I lean more into something like: `import { ProtocolAdapter } from '@/libs';`. What do you think?
So the lib name should be forced and stored in a place, that fits the category.

src/protocols/BaseProtocol.ts
peronczyk, line 1: As we are moving in the concept od "adapters" I suggest to rename this to `BaseProtocolAdapter`.

src/protocols/aeternity/libs/AeternityProtocol.ts
peronczyk, line 1: As we are moving in the concept of "adapters" I suggest to rename this to `AeternityAdapter`.

src/protocols/bitcoin/libs/BitcoinProtocol.ts
peronczyk, line 1: As we are moving in the concept of "adapters" I suggest to rename this to `BitcoinAdapter`.

peronczyk (changes requested):

src/composables/protocol.ts
peronczyk: After spending some time with the recent reviews I think this should be a `ProtocolAdapterFactory` library.
So it could be used this way:
```
import { ProtocolAdapterFactory } from '@/libs';
...
const protocolAdapter = ProtocolAdapterFactory.getAdapter(someProtocol);
```

src/protocols/aeternity/libs/AeternityProtocol.ts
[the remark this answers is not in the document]
subhod-i: Since it's different for each Protocol, I guess we can keep it same.

src/composables/protocol.ts
subhod-i: Kept it very simple. Please have a fresh look https://github.com/aeternity/superhero-wallet/pull/2193/files#diff-b04f42206a6a2e9b4688bc42b5193f5cf351c98c98f28610bc7f842a6e07489b
@peronczyk

src/composables/protocol.ts
subhod-i: No longer needed

src/composables/protocol.ts
subhod-i: https://github.com/aeternity/superhero-wallet/pull/2193/files#diff-b04f42206a6a2e9b4688bc42b5193f5cf351c98c98f28610bc7f842a6e07489b
Please have a look here

src/composables/protocol.ts
subhod-i: Done

src/protocols/BaseProtocol.ts
davidyuk: > existing methods from the SDK to generate private keys for Bitcoin

aepp-sdk-js can't do it https://github.com/aeternity/superhero-wallet/pull/2219#discussion_r1286085523

src/composables/protocol.ts
subhod-i: Keeping it according to Bart's suggestions.

src/protocols/aeternity/libs/AeternityProtocol.ts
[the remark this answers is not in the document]
subhod-i: removed the static properties as it's no longer used in this PR.

src/protocols/BaseProtocol.ts
subhod-i: Yes. Static properties have been removed. We will use the methods from each protocol SDK to generate accounts.

src/lib/ProtocolAdapterFactory.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Encapsulates helper methods related to multiple protocol instances
+ */
+export const ProtocolAdapterFactory = (() => {
+  const protocolAdapterRegistry: Record<Protocol, Class<BaseProtocolAdapter> > = {
+    [PROTOCOL_AETERNITY]: AeternityAdapter,
+    [PROTOCOL_BITCOIN]: BitcoinAdapter,
+  };
+
+  function getAdapter(protocol: Protocol): BaseProtocolAdapter {
+    return new protocolAdapterRegistry[protocol]();
```
peronczyk, line 23: This should not return new instance each time you run `getAdapter`. The instances should be stored. So I think oversimplified this now.

peronczyk (changes requested):

src/lib/ProtocolAdapterFactory.ts
subhod-i: Fixed it with a cache as discussed.

peronczyk (approved):

Liubov-crypto:
The balance of the accounts have been fetched correctly, LGTM.

davidyuk:
There is a hidden unresolved issue about store migration: don't forget to add one if it is needed ☝️

subhod-i:
> There is a hidden unresolved issue about store migration: don't forget to add one if it is needed ☝️

You mean https://github.com/aeternity/superhero-wallet/pull/2193#discussion_r1284080140 ?

PR #2197, opened 2023-07-28
[SW-1102] Move AE Coin constants to protocol dir

peronczyk:


pstachel (approved):

src/popup/utils/utils.ts
subhod-i, line 1: There are many methods in this file specific to Aeternity.  could be moved to `protocol/aeternity/utils` directory

src/popup/utils/constants.ts
```diff
@@ -227,9 +210,6 @@ export const NODE_STATUS_CONNECTING = 'connecting';
 export const NODE_STATUS_CONNECTED = 'connected';
 export const NODE_STATUS_ERROR = 'error';
 
-export const AGGREGATOR_URL = 'https://superhero.com/';
-export const CONTACT_EMAIL = 'superherowallet@protonmail.com';
-
 export const TXS_PER_PAGE = 30;
 export const AENS_DOMAIN = '.chain';
 export const AENS_NAME_MAX_LENGTH = 63 + AENS_DOMAIN.length;
```
subhod-i, line 215: These are Aeternity-specific constants but I don't think this should be moved to the 'protocol/aeternity/` directory. Since our app in general offers more features of Aeternity, we can move constants or methods to each specific protocol level directory which are only common for all the protocols but distinct for each one of them.

src/popup/utils/utils.ts
peronczyk: Yes, I'm doing this within another PR. There will be bunch of them to split the work.

peronczyk: Here's a next PR which covers part of the utils.

src/popup/utils/constants.ts
peronczyk: The AENS is specific to Aeternity protocol and it definitely should be moved to the `protocols/aeternity/`.
Agreed on the call.

subhod-i (approved):

PR #2203, opened 2023-08-01
[Multichain] Create global utils directory and store common utils

peronczyk:


src/popup/components/InviteItem.vue
```diff
@@ -87,7 +87,8 @@ import {
 } from '@aeternity/aepp-sdk';
 import { useStore } from 'vuex';
 import { useRouter } from 'vue-router';
-import { APP_LINK_WEB, formatDate } from '../utils';
+import { formatDate } from '@/utils';
+import { APP_LINK_WEB } from '../utils';
```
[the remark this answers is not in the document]
peronczyk: Nope, the new utils are stored in the `src/` directory. The previous one is `src/popup/`. But you're right - it's confusing. I'll update the second path to use `@`.

src/protocols/aeternity/utils.ts
```diff
@@ -0,0 +1,185 @@
...
+
+export function getTransactionTipUrl(transaction: ITransaction): string {
+  return (
+    transaction.tipUrl
+    || transaction.url
+    || (
+      !transaction.pending
+      && !transaction.claim
+      && transaction.tx.log?.[0]
+      && transaction.tx?.function
```
pstachel, line 108: I think you should check the presence of `tx` in line 107 - there's no need of doing it in 108

pstachel (approved):

src/protocols/aeternity/utils.ts
peronczyk: `tx` property is required in the `ITransaction` interface. So there's no need to check for it's existance.

tests/unit/transaction-details.spec.js
```diff
@@ -113,7 +113,6 @@ function mountComponent({ hasError = false } = {}) {
       getTx: () => () => testTransaction,
       getTxType: () => () => 'provide liquidity',
       getTxSymbol: () => () => AE_SYMBOL,
-      getTransactionTipUrl: () => () => '',
```
subhod-i, line 116: Why?

src/utils/crypto.ts
subhod-i, line 1: This is not related to crypto. This is just another helper function that converts a number or string to BigNumber.
The method could be renamed to `toBigNumber` and could be moved to `./common.ts`

src/store/plugins/veeValidate.js
subhod-i, line 1: Can this be moved to `utils/validators.ts` file?

src/protocols/aeternity/helpers/index.ts
subhod-i, line 1: `helpers` directory has just one file. Instead, Would `aeternity/helpers.ts` file work?

subhod-i (changes requested):

tests/unit/transaction-details.spec.js
peronczyk: Because `getTransactionTipUrl` is not anymore present in the store so there is no need to mock it.

src/utils/crypto.ts
peronczyk: Makes sense. Done - I changed the name to `toShiftedBigNumber`.

src/store/plugins/veeValidate.js
peronczyk: What do you mean by "this"? The whole file?

src/protocols/aeternity/helpers/index.ts
peronczyk: For now it has only one file. But another file moves more files here. Like for example the transaction resolvers.

src/store/plugins/veeValidate.js
subhod-i: Yes.

src/protocols/aeternity/helpers/index.ts
subhod-i: Okay.

src/store/plugins/veeValidate.js
peronczyk: I don't think it's reasonable until we get rid of Vuex. The `veeValidate` file is imported in the `store/index` file and then passed to the `plugins` array of the store. So right now it is a store plugin.
I totally agree we need to move it from the store but not within the scope of the current PR.

subhod-i: Sure.

subhod-i (approved):

PR #2210, opened 2023-08-02
[SW-1111] [Multichain] Split Transaction Details protocol views

peronczyk:


konradodo (approved):

pstachel (approved):

CedrikNikita:
This approach will lead to the next scenario:
User will have `btc` account active. He will try to open (via send url for example) the `ae` transaction (currently we have such case handled). And the btc `transactionDetails` page will be opened.

PR #2211, opened 2023-08-02
[In-App Browser] feat: enhancement to support dapp browser card

ZGiovas:
image


martinkaintas (approved):
🤝

src/popup/pages/Dashboard.vue
```diff
@@ -57,9 +69,11 @@ import ArrowReceiveIcon from '../../icons/arrow-receive.svg?vue-component';
 import ArrowSendIcon from '../../icons/arrow-send.svg?vue-component';
 import CardIcon from '../../icons/credit-card.svg?vue-component';
 import MenuCardIcon from '../../icons/menu-card-fill.svg?vue-component';
+import GlobeIcon from '../../icons/globe-small.svg?vue-component';
 
 import buyBackground from '../../image/dashboard/buy-ae.jpg';
 import chainNameBackground from '../../image/dashboard/chain-name.jpg';
+import daeppBrowserBackground from '../../image/dashboard/daepp-browser.png';
```
peronczyk, line 76:
```suggestion
import daeppBrowserBackground from '@/image/dashboard/daepp-browser.png';
```

src/popup/pages/Dashboard.vue
```diff
@@ -57,9 +69,11 @@ import ArrowReceiveIcon from '../../icons/arrow-receive.svg?vue-component';
 import ArrowSendIcon from '../../icons/arrow-send.svg?vue-component';
 import CardIcon from '../../icons/credit-card.svg?vue-component';
 import MenuCardIcon from '../../icons/menu-card-fill.svg?vue-component';
+import GlobeIcon from '../../icons/globe-small.svg?vue-component';
```
peronczyk, line 72:
```suggestion
import GlobeIcon from '@/icons/globe-small.svg?vue-component';
```

src/popup/components/Card.vue
```diff
@@ -71,12 +71,19 @@ export default {
...
+    bgDarken: { type: Boolean, default: false },
+    bgPosition: {
+      type: String,
+      default: null,
+    },
   },
   setup(props) {
     const hiddenCards = useState('hiddenCards');
 
+    const bgImage = computed(() => props.bgDarken ? `linear-gradient(rgba(0, 0, 0, 0.5), rgba(0, 0, 0, 0.5)), url("${props.background}")` : `url("${props.background}")`);
```
peronczyk, line 83: I suggest to avoid putting CSS rules in Javascript block. I suggest to add/remove class `bg-darken` based on the `bgDarken` prop and allow this class to take care of the styling.

Next thing: why `linear-gradient`? Maybe `::before` pseudo class with opacity would be better?

src/popup/pages/Dashboard.vue
```diff
@@ -12,6 +12,18 @@
     <template #cards>
       <LatestTransactionsCard />
 
+      <DashboardCard
+        v-if="isAeAccount"
```
peronczyk, line 16: I'm not 100% sure if we should remove this card on non AE accounts. Did you consult this with Tsvetan?

peronczyk (changes requested):

src/popup/components/Card.vue
ZGiovas: I am not sure, opacity will affect also the text inside of element. We only want to darken the background

peronczyk: If you apply the opacity to pseudo element (like `::before` or `::after`) it won't affect other layers.

PR #2219, opened 2023-08-04
Create bitcoin account

subhod-i:


src/protocols/hdWallet.ts
```diff
@@ -0,0 +1,36 @@
...
+  }[protocol];
+  return derivePathFromSeed(derivationPath as any, seed);
+}
+
+export function getHdWalletAccount(wallet: IWallet, accountIdx = 0, protocol: Protocol) {
+  const { secretKey, chainCode } = derivePathFromKey(`${accountIdx}h/0h/0h`, wallet as any);
+  const generateAccountAddress = {
+    [PROTOCOL_AETERNITY]: AeternityProtocol,
+    [PROTOCOL_BITCOIN]: BitcoinProtocol,
+  }[protocol].getAddressFromKey;
```
subhod-i, line 30: Not a really great fan of this design. Suggestions are welcome.

src/popup/components/Modals/AccountCreate.vue
```diff
@@ -45,8 +53,13 @@
...
 import { defineComponent, PropType, ref } from 'vue';
 import { useStore } from 'vuex';
-import { MODAL_MULTISIG_VAULT_CREATE } from '@/constants';
+import {
+  MODAL_MULTISIG_VAULT_CREATE,
+  PROTOCOL_AETERNITY,
+  PROTOCOL_BITCOIN,
+} from '@/constants';
 import { useConnection, useModals } from '@/composables';
+import { Protocol } from '@/types';
```
peronczyk, line 62: Please put this before all local imports and use `import type` to avoid dependency cycles.

src/protocols/hdWallet.ts
```diff
@@ -0,0 +1,36 @@
...
+  Protocol,
+} from '@/types';
+import {
+  PROTOCOL_AETERNITY,
+  PROTOCOL_BITCOIN,
+} from '@/constants';
+import { AeternityProtocol } from './aeternity/libs/AeternityProtocol';
+import { BitcoinProtocol } from './bitcoin/libs/BitcoinProtocol';
+
+export function generateHdWallet(seed: Uint8Array, protocol: Protocol) {
```
peronczyk, line 17: I think this should be moved to each of the protocol adapters.

peronczyk (changes requested):

src/protocols/BaseProtocol.ts
```diff
@@ -9,4 +9,11 @@ export abstract class BaseProtocol {
   static derivationPath: string;
 
   abstract getBalance(address: string): Promise<string>;
+
+  abstract isAccountUsed(address: string): Promise<boolean>;
```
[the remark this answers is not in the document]
subhod-i: Added

src/protocols/BaseProtocol.ts
```diff
@@ -9,4 +9,11 @@ export abstract class BaseProtocol {
...
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  static getAddressFromKey(key: Uint8Array, chainCode?: Uint8Array): string {
+    throw new Error('Not implemented in the base class');
+  }
```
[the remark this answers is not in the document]
subhod-i: Yes, maintained a same style as you said.

src/protocols/hdWallet.ts
subhod-i: Exactly, That worked better. Please have a look @peronczyk

src/protocols/BaseProtocolAdapter.ts
```diff
@@ -9,4 +14,30 @@ export abstract class BaseProtocolAdapter {
...
+   * Check if the network has seen this account
+   * @param address Account address
+   */
+  abstract isAccountUsed(address: string): Promise<boolean>;
+
+    /**
+   * get address from the keys
+   * @param secretKey Secret key
+   * @param chainCode Chaincode
+   */
```
peronczyk, line 28: Indentation problem.

src/protocols/BaseProtocolAdapter.ts
```diff
@@ -9,4 +14,30 @@ export abstract class BaseProtocolAdapter {
   static derivationPath: string;
 
   abstract getBalance(address: string): Promise<string>;
+
+  /**
+   * Check if the network has seen this account
```
peronczyk, line 19: I'm not still sture what actually "seen" means. Maybe you could describe it in more clear way?

src/protocols/BaseProtocolAdapter.ts
```diff
@@ -9,4 +14,30 @@ export abstract class BaseProtocolAdapter {
...
+   * @param address Account address
+   */
+  abstract isAccountUsed(address: string): Promise<boolean>;
+
+    /**
+   * get address from the keys
+   * @param secretKey Secret key
+   * @param chainCode Chaincode
```
peronczyk, line 27: This lines gives no value in my opinion.

src/protocols/bitcoin/libs/BitcoinAdapter.ts
```diff
@@ -13,8 +23,38 @@ export class BitcoinAdapter extends BaseProtocolAdapter {
...
+    return true;
+  }
+
+  override getAddressFromKey(key: Uint8Array, chainCode: Uint8Array): string {
+    const bip32 = BIP32Factory(ecc);
+    const { publicKey } = bip32.fromPrivateKey(
+      Buffer.from(key.buffer),
+      Buffer.from(chainCode.buffer),
+    );
+    return (bitcoin.payments.p2wpkh({
```
peronczyk, line 43: No need to wrap this with round brackets.

src/store/getters.js
```diff
@@ -11,33 +10,41 @@ import {
...
 import { AE_SYMBOL } from '@/protocols/aeternity/config';
 import {
   aettosToAe,
   categorizeContractCallTxObject,
 } from '@/protocols/aeternity/helpers';
+import { ProtocolAdapterFactory } from '@/lib/ProtocolAdapterFactory';
 
 export default {
-  wallet({ mnemonic }) {
+  wallet: ({ mnemonic }) => (protocol = PROTOCOL_AETERNITY) => {
```
peronczyk, line 22: Why not to pass the `protocol` as a second property of the base function (next to the `mnemonic`)?

src/store/getters.js
```diff
@@ -11,33 +10,41 @@ import {
...
     if (!mnemonic) return null;
-    return generateHdWallet(mnemonicToSeed(mnemonic));
+    return ProtocolAdapterFactory
+      .getAdapter(protocol)
+      .generateHdWallet(mnemonicToSeed(mnemonic));
   },
   accounts({ accounts: { list } }, getters) {
-    if (!getters.wallet) return [];
+    if (!getters.wallet()) return [];
```
pstachel, line 29:
```suggestion
    if (!getters.wallet()) {
      return [];
    }
```

src/store/modules/accounts/hdWallet.js
```diff
@@ -6,17 +6,16 @@ import {
...
   MODAL_CONFIRM_RAW_SIGN,
   MODAL_CONFIRM_TRANSACTION_SIGN,
   PROTOCOL_AETERNITY,
+  PROTOCOL_BITCOIN,
 } from '@/constants';
-import {
-  getHdWalletAccount,
-} from '@/popup/utils';
 import { isTxOfASupportedType } from '@/protocols/aeternity/helpers';
+import { ProtocolAdapterFactory } from '@/lib/ProtocolAdapterFactory';
```
peronczyk, line 18: I suggest to have this unwritten rule that first we import the common things, like types, utils, libs and then the protocol specific. WHat do you think?

peronczyk (approved):

src/store/getters.js
```diff
@@ -11,33 +10,41 @@ import {
...
+        protocolIdx[protocol] = protocolIdx[protocol] !== undefined ? protocolIdx[protocol] + 1 : 0;
+        return {
+          idx,
+          type,
+          ...acc,
+          protocol,
+          ...(type === ACCOUNT_HD_WALLET
+            ? ProtocolAdapterFactory
+              .getAdapter(protocol)
+              .getHdWalletAccount(getters.wallet(protocol), protocolIdx[protocol], protocol)
```
pstachel, line 44: you pass the `protocol` a lot - maybe that could be optimised somehow?

src/store/modules/accounts/hdWallet.js
```diff
@@ -29,33 +28,55 @@ export default {
...
+          lastNotEmptyIdx = nextIdx;
+        }
+      }
+      for (let i = state.nextAccountIdx; i <= lastNotEmptyIdx; i += 1) {
+        dispatch('create', {
+          isRestored: true,
+          protocol: PROTOCOL_BITCOIN,
+        });
       }
     },
```
pstachel, line 70: You are repeating those loops for each protocol - maybe they could be extracted to protocol specific functions and used here or to some base functions as this code differs only with the protocol argument
```
for (let i = state.nextAccountIdx; i <= lastNotEmptyIdx; i += 1) {
        dispatch('create', {
          isRestored: true,
          protocol: PROTOCOL_BITCOIN,
        });

File: src/protocols/bitcoin/libs/BitcoinAdapter.ts [hunk 15]
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable class-methods-use-this */
+
+import {
+  derivePathFromSeed,
```
davidyuk, line 4: Are you sure getting the correct keys using this method? According BIP-32 master key should be generated with "Bitcoin seed" key and in aeternity we are using "ed25519 seed"
https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki
https://github.com/aeternity/aepp-sdk-js/blob/5df22dd297abebc0607710793a7234e6761570d4/src/utils/hd-wallet.ts#L47

subhod-i: Of course not. The keys are not the same, I did compare though😊. The Bitcoin Adapter would use the bitcoinjs-lib library methods along with other implementations which is pending eg: `getBalance`

src/store/modules/accounts/hdWallet.js
subhod-i: All these imports (types, utils, and libs) are generally the most edited imports. keeping them first would lead to conflicts between parallel commits. I suggest keeping them at last except the types import that could be at the first.
Thus we can minimize the merge conflicts.

src/store/getters.js
subhod-i: https://vuex.vuejs.org/guide/getters.html#method-style-access
The first parameter is the state.

src/protocols/BaseProtocolAdapter.ts
subhod-i: Changed it to `Check whether the network has encountered this account.` @peronczyk please check

src/protocols/BaseProtocolAdapter.ts
subhod-i: Removed it.

src/store/modules/accounts/hdWallet.js
subhod-i: This has been improved drastically, Please have a look here https://github.com/aeternity/superhero-wallet/pull/2219/files#diff-37085670e4cc4d0853535bb598e3c36b31ffb49af1badbebe2018c1591d51b9c

Resolving this, if you got suggestions for the new approach then please open a new comment.

@pstachel

src/store/getters.js
subhod-i: No longer needed.

src/protocols/BaseProtocolAdapter.ts
```diff
@@ -1,6 +1,28 @@
...
+   * @param address Account address
+   */
+  abstract isAccountUsed(address: string): Promise<boolean>;
+
+  /**
+   * Generate account from Mnemonic
+   * @param seed 12 word seed array buffer
+   * @param accountIndex Account Index in derivation path
+   */
+  abstract getHdWalletAccountFromMnemonic(seed: Uint8Array, accountIndex: number): IHdWalletAccount;
```
davidyuk, line 20: Uint8Array is not a mnemonic, it is seed or mnemonic seed 🤔

src/protocols/aeternity/libs/AeternityAdapter.ts
```diff
@@ -1,15 +1,51 @@
...
+    let lastNotEmptyIdx = 0;
+    // First Aeternity account is present in the state, hence index starts from 1
+    for (let i = 1; i < MAXIMUM_ACCOUNTS_TO_DISCOVER; i += 1) {
+      const account = this.getHdWalletAccountFromMnemonic(seed, i);
+      // eslint-disable-next-line no-await-in-loop
+      if (await this.isAccountUsed(account.publicKey)) {
+        lastNotEmptyIdx = i - 1;
+      }
+    }
+    return lastNotEmptyIdx;
```
davidyuk, line 49:
```suggestion
    let i;
    // First Aeternity account is present in the state, hence index starts from 1
    for (i = 1; i < MAXIMUM_ACCOUNTS_TO_DISCOVER; i += 1) {
      const account = this.getHdWalletAccountFromMnemonic(seed, i);
      // eslint-disable-next-line no-await-in-loop
      if (!(await this.isAccountUsed(account.publicKey))) {
        break;
      }
    }
    return i - 1;
```
this way it won't do extra requests if meet an unused account

src/protocols/bitcoin/libs/BitcoinAdapter.ts
```diff
@@ -1,11 +1,58 @@
 /* eslint-disable class-methods-use-this */
 
+import * as ecc from '@bitcoin-js/tiny-secp256k1-asmjs';
+import { BIP32Factory } from 'bip32';
+import * as bitcoin from 'bitcoinjs-lib';
```
davidyuk, line 5: I would import specific parts to allow tree-shaking if it supported by bitcoinjs-lib

pstachel (approved):

src/store/modules/accounts/hdWallet.js
```diff
@@ -63,7 +67,7 @@ export default {
       const { secretKey } = (options?.fromAccount)
         ? getAccountByAddress(options.fromAccount)
         : activeAccount.value;
-      return sign(data, secretKey);
+      return sign(data, new Uint8Array(Buffer.from(secretKey, 'hex')));
```
davidyuk, line 70:
```suggestion
      return sign(data, Buffer.from(secretKey, 'hex'));
```
usually Buffer extends Uint8Array, so this conversion shouldn't be necessary

src/protocols/aeternity/libs/AeternityAdapter.ts
CedrikNikita: https://github.com/aeternity/superhero-wallet/pull/2236

Liubov-crypto:
LGTM

PR #2246, opened 2023-08-21
[SW-1139] BTC receive modal - show proper currency and fiat value

pstachel:


src/popup/components/AddressFormatted.vue
```diff
@@ -21,6 +21,7 @@ export default defineComponent({
   props: {
     address: { type: String, required: true },
     columnCount: { type: Number, default: 6 },
+    splitAddress: Boolean,
```
konradodo, line 24: maybe "protocol" prop is more suitable

pstachel: It's not about protocol - I made it as temporary solution until we have proper address validation for all protocols

src/popup/locales/en.json
```diff
@@ -315,6 +315,7 @@
     },
     "receive": {
       "title": "Receive funds to public address",
+      "titleBtc": "Receive Bitcoin to public address",
```
peronczyk, line 318: I strongly suggest to avoid using separate translations for each protocols. Imagine we have 20 protocols and 50 translations spread across translations file. I suggest to use translate params.

peronczyk (approved):

konradodo (approved):

PR #2249, opened 2023-08-22
[SW-1138] Show more decimals for BTC account card

pstachel:


src/utils/formatters.ts
```diff
@@ -4,9 +4,11 @@ export function escapeSpecialChars(str: string = ''): string {
   return str.replace(/(\r\n|\n|\r|\n\r)/gm, ' ').replace(/"/g, '');
 }
 
-// TODO: Use the current language from i18n module
+export const defaultLocale = 'en-US';
```
peronczyk, line 7: I suggest to move this to constants.

src/utils/formatters.ts
```diff
@@ -18,8 +20,12 @@ export function formatDate(time: number) {
  * Eg.: 411149751.49932 -> 411 149 751,499 (for the europe)
  *      411149751.49932 -> 411.149.751,499 (for the USA)
  */
-export function formatNumber(value: number) {
-  return new Intl.NumberFormat(navigator.language).format(value);
+export function formatNumber(value: number, options: Intl.NumberFormatOptions = {}) {
+  return new Intl.NumberFormat(defaultLocale, options).format(value);
+}
+
+export function formatNumberParts(value: number, options: Intl.NumberFormatOptions = {}) {
```
peronczyk, line 27: Can you add comment that would provide quick example of the transformation?

src/styles/typography.scss
peronczyk, line 1: I don't think this is reasonable to add every possible variant of the sami bold to this file. Maybe if there is a component that displays string with different sizes just use one variant and manipulate `font-size` in the scope of the component.

src/protocols/bitcoin/libs/BitcoinAdapter.ts
```diff
@@ -40,6 +40,10 @@ export class BitcoinAdapter extends BaseProtocolAdapter {
     },
   ];
 
+  override getAmountDecimalPlaces(): number {
+    return BTC_COIN_PRECISION;
```
peronczyk, line 44: I think we need to maintain one naming convention `precision` or `decimalPlaces`.

peronczyk (changes requested):

konradodo (approved):

peronczyk (approved):

PR #2253, opened 2023-08-23
[SW-1045, SW-1160, SW-1170, 1169] Implement input validation for Bitcoin addresses.

konradodo:


package-lock.json
pstachel, line 1: so many changes just from one library?

src/protocols/bitcoin/components/TransferSendForm.vue
```diff
@@ -9,12 +9,11 @@
     class="transfer-send-form"
   >
     <template #recipient>
-      <!--      TODO - set validation rules -->
       <TransferSendRecipient
         v-model.trim="formModel.address"
         :placeholder="$t('modals.send.recipientPlaceholderBtc')"
         :errors="errors"
-        :validation-rules="{}"
+        :validation-rules="{ token_to_an_btc_address: formModel.address }"
```
pstachel, line 16: I feel like you could get rid of `an` from the validation rule

src/store/plugins/veeValidate.js
```diff
@@ -124,6 +126,9 @@ export default (store) => {
     ),
     { params: ['isToken'] });
 
+  defineRule('token_to_an_btc_address',
+    (value) => btcValidate(value));
```
pstachel, line 130: does the line need to be split?

subhod-i (approved):
LGTM. The pipeline is failing!

src/popup/locales/en.json
```diff
@@ -1156,6 +1156,7 @@
     "nameUnregistered": "This name has already been registered.",
     "notSameAs": "Sender and receiver address are the same. You will send tokens to your own account.",
     "tokenToAnAddress": "Unfortunately you can only send AEX-9 Tokens to an address",
+    "tokenToAnBtcAddress": "Invalid BTC address",
```
peronczyk, line 1159: I suggest to not use any protocol specific translations. Maybe you can use dynamic properties?

src/store/plugins/veeValidate.js
```diff
@@ -124,6 +126,9 @@ export default (store) => {
     ),
     { params: ['isToken'] });
 
+  defineRule('token_to_an_btc_address',
```
peronczyk, line 129: Why not just `address_btc`? The validation checks if the address string is a valid BTC address.

package.json
```diff
@@ -41,6 +41,7 @@
     "@zxing/library": "^0.19.1",
     "bignumber.js": "^9.0.2",
     "bip32": "^4.0.0",
+    "bitcoin-address-validation": "^2.2.3",
```
CedrikNikita, line 44: Maybe it is enough to use this regex and not to bring the external library?
`public addressValidationPattern: string = '^(?:[13]{1}[a-km-zA-HJ-NP-Z1-9]{25,34}|bc1[a-z0-9]{39,59})$'`
https://github.com/airgap-it/airgap-coin-lib/blob/912f326124ab9928bcc6b23f63cca55919a5b655/packages/bitcoin/src/v0/protocol/BitcoinProtocol.ts#L210

konradodo: you have right, regex is enough

src/popup/locales/en.json
konradodo: BTC validation check only address, in AE is checking also the .chainame. I think this should be separated.

src/protocols/bitcoin/helpers/index.ts
```diff
@@ -1,5 +1,11 @@
 import { BTC_COIN_PRECISION } from '../config';
 
+const btcRegExp = new RegExp('^(?:[13]{1}[a-km-zA-HJ-NP-Z1-9]{25,34}|bc1[a-z0-9]{39,59})$');
+
 export function satoshiToBtc(amount: number) {
   return amount / 10 ** BTC_COIN_PRECISION;
 }
+
+export function validBtcAddress(address: string) {
```
CedrikNikita, line 9: I suggest to change the function name to `isBtcAddressValid`

CedrikNikita (approved):
Code changers look reasonable. I didn't test it.

CedrikNikita (changes requested):
For me it doesn't work. I used an address account from one of the created account and set to another.

Liubov-crypto:
LGTM

CedrikNikita (approved):

PR #2255, opened 2023-08-23
[SW-1148, SW-1149] Fix account swiping & add protocol to default account name

pstachel:


src/utils/common.ts
```diff
@@ -128,12 +129,20 @@ export async function fetchJson<T = any>(
   return response.json();
 }
 
+/**
+ * Prepare human-readable name with protocol for an account.
+ * E.g.: `Æternity account 1`, `Bitcoin account 2`
+ */
+export function getDefaultAccountLabel({ protocol, idx }: { protocol?: Protocol, idx?: number }) {
+  return `${protocol || ''} ${tg('pages.account.heading')} ${(idx || 0) + 1}`;
```
peronczyk, line 137: `protocol` = "aeternity" | "bitcoin". So it will not produce "Æternity". Are you sure this is correct?

pstachel: I'm sure it is not - will fix

PR #2256, opened 2023-08-24
[SW-1137, SW-1120] Bitcoin fee and max calculation

CedrikNikita:


peronczyk (approved):

CharisSiarampalis:
@CedrikNikita Sometimes it takes two clicks for the max amount to appear.


https://github.com/aeternity/superhero-wallet/assets/42606212/fa1d2eeb-b9f5-4e23-8093-fbfc8b326164

Liubov-crypto:
LGTM

CedrikNikita:
> @CedrikNikita Sometimes it takes two clicks for the max amount to appear.

The problem is that we are fetching the fee from the endpoint and don't have any visual representation of it so the max amount is different.

PR #2242, opened 2023-08-21
Feature/in app browser

martinkaintas:

PR #2262, opened 2023-08-25
feat: make in-app browser feature mobile-only

martinkaintas:
In-app browser feature is mobile only so we want to hide it from other platforms.
**Note:** should only be merged when #2242 is tested and is ready to be merged.

### DIFF
--- a/src/popup/router/index.ts
+++ b/src/popup/router/index.ts
@@ -70,10 +70,19 @@ router.beforeEach(async (to, from, next) => {
     return;
   }
 
-  if (to.name === ROUTE_APPS_BROWSER && activeAccount.value.protocol !== PROTOCOL_AETERNITY) {
-    setActiveAccountByIdx(0);
-    next({ name: ROUTE_APPS_BROWSER });
-    return;
+  if (to.name === ROUTE_APPS_BROWSER) {
+    // In-app browser is mobile-only
+    if (!IS_CORDOVA) {
+      next({ name: ROUTE_INDEX });
+      return;
+    }
+
+    // In-app browser only works with AE accounts
+    if (activeAccount.value.protocol !== PROTOCOL_AETERNITY) {
+      setActiveAccountByIdx(0);
+      next({ name: ROUTE_APPS_BROWSER });
+      return;
+    }
   }
 
   const { isAeSdkReady } = useAeSdk({ store });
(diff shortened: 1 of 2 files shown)


ZGiovas (approved):

src/popup/router/index.ts
```diff
@@ -70,10 +70,19 @@ router.beforeEach(async (to, from, next) => {
...
-    return;
+  if (to.name === ROUTE_APPS_BROWSER) {
+    // In-app browser is mobile-only
+    if (!IS_CORDOVA) {
+      next({ name: ROUTE_INDEX });
+      return;
+    }
+
+    // In-app browser only works with AE accounts
+    if (activeAccount.value.protocol !== PROTOCOL_AETERNITY) {
```
CedrikNikita, line 81: Some similar logic was already done in https://github.com/aeternity/superhero-wallet/pull/2264, please review it and align both of your PRs to the common ground.

martinkaintas: This check is not relevant to the mobile-only PR. 
But we should update the #2242 PR once #2264 is merged

src/popup/router/index.ts
```diff
@@ -70,10 +70,19 @@ router.beforeEach(async (to, from, next) => {
     return;
   }
 
-  if (to.name === ROUTE_APPS_BROWSER && activeAccount.value.protocol !== PROTOCOL_AETERNITY) {
-    setActiveAccountByIdx(0);
-    next({ name: ROUTE_APPS_BROWSER });
-    return;
+  if (to.name === ROUTE_APPS_BROWSER) {
```
CedrikNikita, line 73: It can be done by having mobile only list of routes.
```
const router = new Router({
    mode: process.env.VUE_APP_CORDOVA ? 'hash' : 'history',
    scrollBehavior(to, from, savedPosition) {
      if (savedPosition) {
        return savedPosition;
      }
      return { x: 0, y: 0 };
    },
    routes: (await Promise.all([
      import(/* webpackChunkName: "ui-common" */ './routes/common'),
      ENV_MOBILE_DEVICE
        ? import(/* webpackChunkName: "ui-mobile" */ './routes/mobile')
        : undefined
    ])).reduce((p, module) => [...p, ...module.default], []),
  });
  ```

ZGiovas: we can add it to backlog as an improvement

CedrikNikita (approved):
OK, for MVP. But would be better to go to the 404 page than index.

PR #2266, opened 2023-08-25
[SW-1148, SW-1149, SW-1158, SW-1181, SW-1164] Fix account swiping & add protocol to default account name

pstachel:
…

### DIFF
--- a/src/composables/accounts.ts
+++ b/src/composables/accounts.ts
@@ -1,4 +1,7 @@
-import { computed } from 'vue';
+import {
+  computed,
+  ref,
+} from 'vue';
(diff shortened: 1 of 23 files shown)


konradodo (approved):

Liubov-crypto:
This pr is fixing some issues, but not all.

CedrikNikita:
> This pr is fixing some issues, but not all.
> 
> 1. Ae account details view is broken:
> 2. Btc modals are present in multisig account.

1. Fixed.
2. Suggest to fix it in different PR, because the issue is more complex.

PR #2271, opened 2023-08-25
[SW-1178] feat(bitcoin): adjust amount validation

CedrikNikita:


src/popup/locales/en.json
```diff
@@ -1153,7 +1153,7 @@
     "tokenToAnAddress": "Unfortunately you can only send AEX-9 Tokens to an address",
     "minValue": "This field must be {0} or more",
     "minValueExclusive": "This field must be more than {0}",
-    "enoughAe": "Not enough AE balance",
+    "enoughAe": "Not enough {0} balance",
```
peronczyk, line 1156: I suggest to change the key as it's not longer "AE". Also I suggest to use named translation properties.

peronczyk (approved):

Liubov-crypto:
LGTM

PR #2272, opened 2023-08-25
[SW-1163, SW-1166, SW-1168] Adjust texts on modals

pstachel:


src/popup/components/Modals/RecipientInfo.vue
peronczyk, line 1: This is bad to use conditional text displaying based on the protocol. I think this needs to be discussed with Tsvetan. He needs to come up with something more generic that could be used for all protocols.

peronczyk (approved):

konradodo (approved):

src/popup/components/Modals/RecipientInfo.vue
onvisions: Yes. It totally makes sense to have different texts in the info modal as for æternity blockchain we may have æternity address or **.chain name** entered in the recipient's field and for Bitcoin or Dogecoin we will have ONLY address.
IMO it's a good idea to provide some more clarity to the user what is a .chain name. For BTC or DOGE there is no need for the second paragraph of the text.

PR #2277, opened 2023-08-26
[SW-1188] feat: adjust connect modal wording

### LINKED ISSUE #2238: More accurate wording in aepp connection popup regarding transactions permissions
photo_2023-08-17_01-26-42

Superhero announces that a transaction will happen. Probably "may" should be instead of "will".

There may be cases when aepp doesn't need to make a transaction, just identify the user.

And there ARE transactions whenever a user uses dry-run to evaluate contract calls, but those are bypassing Superhero and the user is NOT (!) in charge of that in the sense that Superhero means and announces: there is no additional ok of the user needed.

So the announcement that Superhero shows is double misleading.

reported by @brainiacfive

CedrikNikita:
fixes #2238

### DIFF
--- a/src/popup/locales/cn.json
+++ b/src/popup/locales/cn.json
@@ -259,7 +259,7 @@
       "addressLabel": "地址",
       "transactionLabel": "交易",
       "addressRequest": "aepp请求查看你的当前地址",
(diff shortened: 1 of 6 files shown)


konradodo (approved):

Liubov-crypto:
LGTM

PR #2302, opened 2023-08-29
Use last aeternity account for signing in Aepp communication

subhod-i:


CedrikNikita (approved):
This is fixing the remain issue from the last ae account PR, **but** is a big change and should be tested carefully.

PR #2310, opened 2023-08-31
[SW-1241] Refactor: AE Node connection status as SDK composable state

peronczyk:


src/composables/aeSdk.ts
```diff
@@ -45,24 +42,28 @@ import { useModals } from './modals';
...
 let aeSdkBlocked = false;
 let storedNetworkName: string;
-const isAeSdkReady = ref(false);
+
 const nodeNetworkId = ref<string>();
+
+const isAeSdkReady = ref(false);
+const isAeNodeReady = ref<boolean>(false);
+const isAeNodeConnecting = ref<boolean>(false);
+const isAeNodeError = ref<boolean>(false);
```
konradodo, line 51: <boolean> is not needed

src/protocols/aeternity/views/TransferSendModal.vue
```diff
@@ -90,6 +95,13 @@ export default defineComponent({
         && isUrlValid(transferData.value.address)
     ));
 
+    const isPrimaryButtonDisabled = computed(() => (
+      error.value
+      || !isAeNodeReady.value
+      || !transferData.value.address
+      || !transferData.value.amount
```
konradodo, line 102:
```suggestion
      !error.value
      && isAeNodeReady.value
      && transferData.value.address
      && transferData.value.amount
```

this is only personal suggestion. I think, if we have a lot of negation, then code is more not readable.

konradodo (approved):

src/composables/aeSdk.ts
peronczyk: Why? It is used in the connection status bar.

src/protocols/aeternity/views/TransferSendModal.vue
peronczyk: It's a matter of readability. When you convert the code into sentence it makes more sense with negations: "the button is disabled if there is an error or the node is not ready, or address is empty or there is no amount". If we invert it it will take more time to understand.

src/composables/aeSdk.ts
subhod-i, line 51:
```suggestion
Enum ConnectionStatus = {
   Ready,
   Connecting,
   Error,
};
const aeNodeConnectionStatus = ref<ConnectionStatus>(ConnectionStatus.Connecting);
```

I remember having this discussion before. Why wouldn't this single variable wouldn't work?

subhod-i (commented):
LGTM

src/composables/aeSdk.ts
konradodo: "<boolean>" is  not needed

I don't know this type is missing in comment

konradodo: "< boolean >" is  not needed

I don't know this type is missing in comment

src/protocols/aeternity/views/TransferSendModal.vue
konradodo: like i said, this is individual. For me is ok

src/composables/aeSdk.ts
peronczyk: OK, makes sense. Please be more clear next time :)

peronczyk: Because maybe at the beginning the amount of variables is bigger, but with one variable you'll still need to establish things like `isError`, `isConnected` etc... And those would require having computed properties. So overall we will end up with more code.

src/protocols/aeternity/views/TransferSendModal.vue
peronczyk: OK, resolving.

subhod-i (approved):

src/popup/components/OpenTransferSendModalButton.vue
```diff
@@ -3,7 +3,7 @@
     :text="isMultisig ? $t('dashboard.proposeCard.title') : $t('common.send')"
     :subtitle="subtitle"
     :icon="ArrowSendIcon"
-    :disabled="!isOnline || !isConnected || (!!pendingMultisigTransaction && isMultisig)"
```
CedrikNikita, line 6: Why did you decide to remove this check?

PR #2311, opened 2023-08-31
[SW-1229] fix: open external links in mobile correctly

CedrikNikita:


Liubov-crypto:
Fixed, LGTM

src/popup/components/NotificationItem.vue
```diff
@@ -146,7 +147,11 @@ export default defineComponent({
     function handleClick() {
       if (props.notification.path) {
         if (typeof props.notification.path === 'string' && /^\w+:\D+/.test(props.notification.path)) {
-          window.open(props.notification.path, IS_MOBILE_DEVICE ? '_self' : '_blank');
+          if (IS_CORDOVA && window.cordova?.InAppBrowser?.open) {
```
peronczyk, line 150: This test `window.cordova?.InAppBrowser?.open` is already present in few places. I suggest to create environment variable out of it, for example: `IN_APP_BROWSER_AVAILABLE`.

peronczyk (approved):

src/popup/components/NotificationItem.vue
CedrikNikita: This run time const will calculate once in that case, and will calculate before `inAppBrowser` initialization resulting in `false`.

subhod-i (approved):

PR #2315, opened 2023-09-04
[SW-1196] feat: discover accounts for every protocol

CedrikNikita:


Liubov-crypto:
LGTM

src/store/modules/accounts/hdWallet.js
```diff
@@ -30,16 +38,44 @@ export default {
...
+        return { protocol, lastNotEmptyIdx };
+      }));
+      accountsToRecover.forEach(({ protocol, lastNotEmptyIdx }) => {
+        for (let i = protocol === PROTOCOL_AETERNITY ? 1 : 0; i <= lastNotEmptyIdx; i += 1) {
+          dispatch('create', {
+            isRestored: true,
+            protocol,
+          });
+        }
+      });
```
subhod-i, line 78: Please leave a code comment on the account discovery algorithm. The approach looks complicated. This could be simplified by breaking the single-function into multiple functions.

subhod-i: Standard algorithm for Account discovery https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account-discovery

CedrikNikita: This is exactly what we are doing (with the fact that we are using only `account`, without `addresses`).

CedrikNikita: Actually there is an explanation above `ADDRESS_GAP_LIMIT` variable, that explains the logic behind account discovery.

subhod-i: Sure if you think that describes the logic, No problem then.

subhod-i (approved):

src/store/modules/accounts/hdWallet.js
```diff
@@ -12,10 +12,18 @@ import {
...
 } from '@/constants';
 import { isTxOfASupportedType } from '@/protocols/aeternity/helpers';
 import { ProtocolAdapterFactory } from '@/lib/ProtocolAdapterFactory';
 
+/**
+ * Address gap limit is currently set to 5.
+ * If the software hits 5 unused addresses in a row,
+ * it expects there are no used addresses beyond this point and stops searching the address chain.
+*/
+const ADDRESS_GAP_LIMIT = 5;
```
peronczyk, line 25: I suggest to move this into the constants file.

src/store/modules/accounts/hdWallet.js
```diff
@@ -30,16 +38,44 @@ export default {
...
-        .getAdapter(PROTOCOL_AETERNITY) // Discover only aeternity accounts for now
-        .discoverAccounts(wallet);
-
-      for (let i = 0; i < numberOfAccounts; i += 1) {
-        dispatch('create', {
-          isRestored: true,
-          protocol: PROTOCOL_AETERNITY,
-        });
-      }
+      const accountsToRecover = await Promise.all(PROTOCOLS.map(async (protocol) => {
```
peronczyk, line 41: To be honest I don't like the fact that the logic of discovering the accounts was moved from the adapters. I know for AE and BTC the logic is the same and duplicating same code would make no sense. But maybe we could have an utility function for this purpose, which could be reused within the adapters? This would make the code more modular and make it easier to add new protocols that has different way of account discovery.

peronczyk (changes requested):

src/utils/common.ts
```diff
@@ -359,3 +360,39 @@ export function detectProtocolByOwner(network: string, address?: string) {
   }
   return null;
 }
+
+export async function defaultAccountDiscovery(
+  isAccountUsed: Function,
+  getHdWalletAccountFromMnemonicSeed: Function,
+  protocol: string,
```
peronczyk, line 367: If the protocol is not used in the logic but just returned at the end it makes no sense to use ti at all.

src/protocols/BaseProtocolAdapter.ts
```diff
@@ -65,7 +66,7 @@ export abstract class BaseProtocolAdapter {
    * @param seed 12 word seed array buffer
    * @returns total number of used accounts
    */
-  abstract discoverAccounts(seed: Uint8Array): Promise<number>;
+  abstract discoverAccounts(seed: Uint8Array): Promise<DiscoveredAccounts>;
```
peronczyk, line 69: I think it makes no sense to return the protocol ID by the `discoverAccounts` method.

src/store/modules/accounts/hdWallet.js
```diff
@@ -30,16 +31,19 @@ export default {
...
-          isRestored: true,
-          protocol: PROTOCOL_AETERNITY,
-        });
-      }
+      const accountsToRecover = await Promise.all(
+        PROTOCOLS.map(async (protocol) => (
+          await ProtocolAdapterFactory.getAdapter(protocol)
+        ).discoverAccounts(wallet)),
+      );
+      accountsToRecover.forEach(({ protocol, lastNotEmptyIdx }) => {
```
peronczyk, line 39: As I think the `accountsToRecover` don't need to hold the protocol name - only the number, so here instead of looping the `accountsToRecover` you can loop again the `PROTOCOLS` constant. I did this as a part of the PR that covers the migration of accounts from Vuex to composable: https://github.com/aeternity/superhero-wallet/pull/2334/files#diff-10c662816c7957770974da4b39e9e9a3ca2135b2fa50d7bba66c45f62a9b369cR196

src/utils/common.ts
```diff
@@ -359,3 +360,37 @@ export function detectProtocolByOwner(network: string, address?: string) {
   }
   return null;
 }
+
+export async function defaultAccountDiscovery(
+  isAccountUsed: Function,
+  getHdWalletAccountFromMnemonicSeed: Function,
```
peronczyk, line 366: Instead of `Function` I suggest to use for example: `(address: string) => Promise<boolean>`.

src/utils/common.ts
```diff
@@ -359,3 +361,40 @@ export function detectProtocolByOwner(network: string, address?: string) {
...
+          .map((index) => isAccountUsed(
+            getHdWalletAccountFromMnemonicSeed(wallet, index).address,
+          )),
+      );
+    } catch (e) {
+      break;
+    }
+  } while (!(
+    isAccountUsedArray.lastIndexOf(true) === -1
+      || isAccountUsedArray.filter((isAccountUsedValue) => !isAccountUsedValue)
```
peronczyk, line 396: Wrong indentation.

peronczyk (approved):

Liubov-crypto:
LGTM

PR #2325, opened 2023-09-06
[SW-1251] Fix adding new networks in chrome extension

peronczyk:


pstachel (approved):

davidyuk:
Is it to properly handle reactive arrays?

Liubov-crypto:
LGTM

src/composables/composablesHelpers.ts
```diff
@@ -28,31 +33,46 @@ interface ICreateStorageRefOptions<T> {
...
+  function setState(val: any) {
+    if (val) {
+      watcherDisabled = true;
+      state.value = (serializer?.read) ? serializer.read(val) : val;
+      setTimeout(() => { watcherDisabled = false; }, 0);
+    }
+  }
+
+  watch(state, (val, oldVal) => {
+    // Arrays are not compared as there is a bug which makes the new and old val always the same.
```
CedrikNikita, line 57: did you find the cause of the bug? Are you planning to resolve this bug?

peronczyk: Unfortunately no. I know that this bug was present in Vue 2 but I thought in Vue 3 it was fixed. There's this note in the Vue documentation:
image
I think this is the reason, but I'm not sure why this is happening only with the arrays here.
I talked about this with @lukeromanowicz but he also has no idea how to solve the problem.

peronczyk: > Are you planning to resolve this bug?
This is not a bug on our side but it's a Vue issue. So for now I don't know how to approach it in a different way as I did.

lukeromanowicz: well... I guess we could do some hacking with cloning an array instead of pushing to force the ref change and maybe store it in second ref to have two-step update. But it's a dirty trick :/

CedrikNikita (approved):

PR #2327, opened 2023-09-06
[SW-1208] Total is not displaying the sum of BTC accounts amounts but only the sum of AE accounts

pstachel:


ifaouibadi (approved):

src/composables/currencies.ts
```diff
@@ -53,17 +52,19 @@ const initPollingWatcher = createPollingBasedOnMountedComponents(POLLING_INTERVA
 export function useCurrencies({
   withoutPolling = false,
   store,
-  selectedProtocol = PROTOCOL_AETERNITY, // TODO - remove default value & make the protocol required
 }: UseCurrenciesOptions) {
   const { protocolsInUse, isLoggedIn } = useAccounts({ store });
-  const minTipAmount = computed(() => 0.01 / (currencyRates.value?.[selectedProtocol].usd || 1));
-  const currentCurrencyRate = computed(
-    (): number => currencyRates.value?.[selectedProtocol]?.[currentCurrencyCode.value] || 0,
+  const aeMinTipAmount = computed(
```
peronczyk, line 57: This value is used only in one place - `veeValidate` file. So I suggest to move this logic there.

src/composables/balances.ts
```diff
@@ -45,6 +47,21 @@ export function useBalances({ store }: IDefaultComposableOptions) {
       .toFixed(),
   );
 
+  const balancesCurrenciesTotal = computed(
+    () => {
+      const result = Object.entries(balances.value)
+        .reduce((total, [accountAddress, accountBalance]) => {
+          const selectedAccount = accounts.value.find((acc) => acc.address === accountAddress);
```
peronczyk, line 54: `accounts` composable has `getAccountByAddress` method for this purpose.

src/composables/balances.ts
```diff
@@ -45,6 +47,21 @@ export function useBalances({ store }: IDefaultComposableOptions) {
...
+          const currencyRate = selectedAccount?.protocol
```
peronczyk, line 55: Protocol is required for each accounts so there is no need to check if it exists.

src/composables/balances.ts
```diff
@@ -45,6 +47,21 @@ export function useBalances({ store }: IDefaultComposableOptions) {
       .toFixed(),
   );
 
+  const balancesCurrenciesTotal = computed(
+    () => {
+      const result = Object.entries(balances.value)
```
peronczyk, line 52: I feel that we are doing the reversed logic :)
So what actually we need to achieve is to sum ap all the **accounts** balances. So I suggest to loop over the accounts and try to collect the balances for them. This way you won't need to search account in every iteration.

src/composables/balances.ts
```diff
@@ -37,6 +38,7 @@ const balances = useStorageRef<Balances>({}, LOCAL_STORAGE_BALANCES_KEY, {
  */
 export function useBalances({ store }: IDefaultComposableOptions) {
   const { activeAccount, accounts } = useAccounts({ store });
+  const { getCurrentCurrencyRate } = useCurrencies({ store });
 
   const balance = computed(() => balances.value[activeAccount.value.address] || new BigNumber(0));
   const balancesTotal = computed(
```
peronczyk, line 44: I guess this can be removed now?

src/popup/components/TotalWalletAmount.vue
```diff
@@ -10,20 +10,32 @@
...
-import { useCurrencies } from '../../composables';
+import { Protocol } from '@/types';
+import { PROTOCOL_AETERNITY } from '@/constants';
+import { useCurrencies } from '@/composables';
 
 export default defineComponent({
   props: {
-    totalBalance: { type: String, required: true },
+    totalBalance: { type: String, default: '0' },
+    totalBalanceCurrency: { type: String, default: '' },
```
peronczyk, line 26: This looks rather bad. In my opinion this components should just receive the balance and display it. Why two separate props?

CedrikNikita (approved):

PR #2331, opened 2023-09-09
[SW-1217, 1216] Improve btc utxo selection alogrithm and prevent dust error

subhod-i:


ifaouibadi (approved):

src/protocols/bitcoin/libs/BitcoinAdapter.ts
```diff
@@ -222,23 +228,18 @@ export class BitcoinAdapter extends BaseProtocolAdapter {
...
-    const utxoDetails = await Promise.all(fetchHexPromises);
-
-    const amountInSatoshi = toSatoshi(amountInBtc);
-    const feeInSatoshi = toSatoshi(options.fee);
-    let totalBalance: number = 0;
-    let hasSufficientBalance = false;
+    // Sort UTXOs by ascending absolute difference from the amount to send
+    const spendableUtxos = fullUtxos.sort(
+      (a, b) => Math.abs(a.value - (amountInSatoshi + feeInSatoshi))
+              - Math.abs(b.value - amountInSatoshi + feeInSatoshi),
```
pstachel, line 236: You could remove parentheses or add them in the second line to have same style or you could extract the sum of amount and fee to common variable and then subtract it
```suggestion
      (a, b) => Math.abs(a.value - (amountInSatoshi + feeInSatoshi))
              - Math.abs(b.value - amountInSatoshi + feeInSatoshi),
```

pstachel (approved):

Liubov-crypto:
Please, do not merge it! Check my comments in Jira related task.

src/protocols/bitcoin/libs/BitcoinAdapter.ts
peronczyk, line 236: I understand that it looks nice to have the `Math.abs` aligned in one column. But aligning stuff to columns in longer term is rather bad practice - imagine everyone will be aligning stuff this way :) Also with this approach when one line changes it's indentation you need to align all others, which leads to more affected lines than it's needed.
```suggestion
      (a, b) => (
        Math.abs(a.value - (amountInSatoshi + feeInSatoshi))
        - Math.abs(b.value - amountInSatoshi + feeInSatoshi)
      ),
```

peronczyk (approved):

src/protocols/bitcoin/libs/BitcoinAdapter.ts
subhod-i: Fixed as per Bart's suggestion.

PR #2334, opened 2023-09-11
[DRAFT] Move accounts vuex state to composable

peronczyk:


src/protocols/aeternity/helpers/index.ts
```diff
@@ -124,6 +125,17 @@ export function categorizeContractCallTxObject(transaction: ITransaction): {
...
 }
 
+export function convertMultisigAccountToAccount(
+  multisigAccount: IMultisigAccount,
+): Partial<IAccount> {
+  return {
+    address: multisigAccount.gaAccountId,
+    idx: 0,
+    protocol: PROTOCOL_AETERNITY,
+    globalIdx: 0,
```
pstachel, line 135: Why will they always have idx equal to 0?

src/protocols/aeternity/libs/AeSdkSuperhero.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable class-methods-use-this */
```
pstachel, line 1: is it needed?

src/protocols/bitcoin/libs/BitcoinAdapter.ts
```diff
@@ -153,6 +153,7 @@ export class BitcoinAdapter extends BaseProtocolAdapter {
     };
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
pstachel, line 156: Isn't it used?

src/composables/accounts.ts
```diff
@@ -122,16 +149,11 @@ export function useAccounts({ store }: IDefaultComposableOptions) {
...
-    } else {
-      protocolNextAccountIdx.value[protocol]! += 1;
-    }
-  }
-
-  function setProtocolNextAccountIdx(value: number, protocol: Protocol) {
-    protocolNextAccountIdx.value[protocol] = value;
+  function getLastProtocolAccount(protocol: Protocol): IAccount | undefined {
+    const protocolAccounts = accountsGroupedByProtocol.value[protocol];
+    return protocolAccounts
```
pstachel, line 154:
```suggestion
    return (protocolAccounts)
```

src/protocols/bitcoin/libs/BitcoinAdapter.ts
peronczyk: True. Leftover that needs to be removed.

src/protocols/aeternity/libs/AeSdkSuperhero.ts
peronczyk: Unfortunately yes :/ The changes done in the `getAccounts` method causes the error as there is no `this` reference now.

src/protocols/aeternity/helpers/index.ts
peronczyk: Good point. The reason is that we don't use indexes when displaying the multisig accounts. For example here are is the card:
image
I'll add a comment about the reason why both idx and globalIdx are equal to 0.
What do you think?

src/popup/components/TransactionLabel.vue
```diff
@@ -135,6 +135,8 @@ export default defineComponent({
 
     const availableTokens = useState<ITokenList>('fungibleTokens', 'availableTokens');
 
+    const getDefaultName = store.getters['names/getDefault'] as (a?: string) => string | undefined;
```
konradodo, line 138: i think, this getter should be moved to utils function in this pr

src/popup/components/NameItem.vue
```diff
@@ -200,8 +200,10 @@ export default defineComponent({
     const error = ref(false);
     const pointerInput = ref();
 
+    const getDefaultName = store.getters['names/getDefault'] as (a?: string) => string | undefined;
```
konradodo, line 203: i think, this getter should be moved to utils function in this pr

konradodo (changes requested):

src/composables/accounts.ts
```diff
@@ -1,94 +1,122 @@
...
+const accounts = computed((): IAccount[] => {
+  if (!mnemonic.value || !accountsRaw.value.length) {
+    return [];
+  }
+
+  return accountsRaw.value
+    .map((account, index) => {
+      const hdWallet = ProtocolAdapterFactory
+        .getAdapter(account.protocol)
+        .getHdWalletAccountFromMnemonicSeed(mnemonicSeed.value, account.idx) as any;
```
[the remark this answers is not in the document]
peronczyk: Totally right. The `any` here is only to satisfy the difference in the `IAccount.address` type. Currently we use `Encoded.AccountAddress` but the HDWallet returns `string`. We will update this soon. I'll add `TODO` comment here. OK?

src/popup/components/NameItem.vue
peronczyk: This makes no sense as next week we will replace Vuex "names" module with composable. So this line is temporary.

konradodo (approved):

src/composables/accounts.ts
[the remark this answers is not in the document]
pstachel: Of course, I'm resolving the comment then

src/protocols/aeternity/helpers/index.ts
pstachel: I agree. It's always better to have an explanation

src/protocols/aeternity/libs/AeSdkSuperhero.ts
pstachel: Ok, resolving then

src/composables/accounts.ts
```diff
@@ -115,47 +169,74 @@ export function useAccounts({ store }: IDefaultComposableOptions) {
...
 
+  function setActiveAccountByProtocolAndIdx(protocol: Protocol, idx: number) {
+    const accountFound = accountsGroupedByProtocol.value[protocol]
+      ?.find((account) => account.idx === idx);
+    if (accountFound) {
+      setActiveAccountByGlobalIdx(accountFound.globalIdx);
+    }
+  }
+
+  function setMnemonic(val: string) {
```
pstachel, line 180: Maybe you could consider more meaningful name than `val`?

src/composables/accounts.ts
```diff
@@ -115,47 +169,74 @@ export function useAccounts({ store }: IDefaultComposableOptions) {
...
+    protocol,
+  }: Omit<IAccountRaw, 'type'>): number {
+    accountsRaw.value.push({
+      protocol,
+      isRestored,
+      type: ACCOUNT_HD_WALLET,
+    });
+    const lastProtocolAccount = getLastProtocolAccount(protocol);
+    const idx: number = (lastProtocolAccount) ? lastProtocolAccount.idx : 0;
+    return idx;
```
pstachel, line 206: I think that assignment is not needed here
```suggestion
    return (lastProtocolAccount) ? lastProtocolAccount.idx : 0;
```

pstachel (approved):

src/store/plugins/fungibleTokens.js
```diff
@@ -220,10 +219,11 @@ export default (store) => {
...
   );
 
-  store.watch(
-    () => aeNextAccountIdx.value,
+  // Fetch balances whenever new account is added
+  watch(
+    aeAccounts,
     async (val, oldVal) => {
-      if (val !== oldVal) {
+      if (val.length !== oldVal.length) {
```
subhod-i, line 226: Please test this with a fresh new wallet and also when restoring a wallet. `oldval` could be undefined and accessing `.length` may lead to runtime errors`

src/store/mutations.js
```diff
@@ -10,9 +10,6 @@ export default {
   setChainNames(state, payload) {
     state.chainNames = payload;
   },
-  setMnemonic(state, payload) {
-    state.mnemonic = payload;
-  },
```
subhod-i, line 15: Where did we move this?

src/store/modules/accounts/index.js
```diff
@@ -14,65 +8,16 @@ export default {
...
     },
 
-    signTransaction({ getters: { active, getModule }, dispatch }, { txBase64, options }) {
+    signTransaction({ dispatch }, { txBase64, options }) {
       if (options && options.fromAccount) {
-        return dispatch(`${getModule(active).name}/signTransactionFromAccount`, { txBase64, options });
+        return dispatch('hdWallet/signTransactionFromAccount', { txBase64, options });
       }
-      return dispatch(`${getModule(active).name}/signTransaction`, { txBase64, options });
+      return dispatch('hdWallet/signTransaction', { txBase64, options });
```
subhod-i, line 20: We may support only `hdWallet` for now but we had the provisions in place to support ledger wallet. I am not sure why this PR is dropping the support of hardware ledger accounts.

src/protocols/aeternity/libs/AeSdkSuperhero.ts
subhod-i, line 1: Since the class property `store` is not being used, Please remove the property from class and it's constructor

```
  store: Store<any>; // This one
  constructor(store: Store<any>, options: any) {
    super(options);
    this.store = store;
  }
 
  ```

**Review by subhod-i — CHANGES_REQUESTED** [review]

File: src/composables/accounts.ts (hunk 8 above)
Comment (peronczyk) [reply src/composables/accounts.ts:180]: `mnemonic` is already used. Any suggestions?

File: src/utils/common.ts [hunk 13]
```diff
@@ -364,7 +374,10 @@ export function detectProtocolByOwner(network: string, address?: string) {
   return null;
 }
 
-export async function defaultAccountDiscovery(
+/**
+ * @returns {number} between -1 and n where -1 means there are no accounts found.
+ */
+export async function getLastNotEmptyAccountIndex(
```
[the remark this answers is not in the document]
peronczyk: I kind of understand what you want to achieve, but this requires really complex TS custom typings which in my opinion are an overkill. Especially as the TypeScript itself types and comments stuff the same way as I did. For example here's a typing info for `Array.findIndex()` which returns `-1...n`:
image
You see the return value is typed as `number` but there is an additional comment about the possible values.

src/store/plugins/fungibleTokens.js
peronczyk: `aeAccunts` is based on regular `accounts` which always returns an array. If user didn't restored the accounts the array is empty.

src/protocols/aeternity/libs/AeSdkSuperhero.ts
peronczyk: I wanted to achieve this, but the method `_resolveAccount` returns the `AccountSuperhero` which also has the `store` as dependency :/ If you see how we can remove this please let me know - I would be so happy to remove it :)

src/store/modules/accounts/index.js
peronczyk: We didn't support the ledger. Please take a closer look on the PR and this line: https://github.com/aeternity/superhero-wallet/blob/develop/src/store/modules/accounts/index.js#L11

src/popup/components/Modals/AccountCreate.vue
```diff
@@ -73,10 +73,11 @@ export default defineComponent({
           break;
 
         case PROTOCOL_BITCOIN:
-          await store.dispatch('accounts/hdWallet/create', {
+          idx = addRawAccount({
             isRestored: false,
             protocol: PROTOCOL_BITCOIN,
           });
+          setActiveAccountByProtocolAndIdx(PROTOCOL_BITCOIN, idx);
```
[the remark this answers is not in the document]
peronczyk: No it's not used, but it's not possible to create `let`/`const` within the `case` block.

src/composables/accounts.ts
pstachel: I'm thinking of `phrase` or `mnemonicPhrase` to indicate what kind of value will be passed here

src/composables/composablesHelpers.ts
```diff
@@ -33,7 +33,7 @@ interface ICreateStorageRefOptions<T> {
  */
 export function useStorageRef<T = string | object | any[]>(
   initialState: T,
-  keys: StorageKeysInput,
+  keys: StorageKey,
```
subhod-i, line 36:
```suggestion
  key: StorageKey,
```

Not plural.

subhod-i (changes requested):

src/composables/composablesHelpers.ts
peronczyk: OK. Done.

src/utils/common.ts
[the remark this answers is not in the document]
subhod-i: Is it really complex?

The typescript document itself uses `number` but IMO we can do better. I am pushing this because it's being used at  crucial helper methods and it's important to indicate this for someone to use this in the future.

src/store/modules/accounts/index.js
subhod-i: Sure. It's better to have a module-based routing but that can be introduced when adding ledger wallet feature too.

src/composables/transactionTx.ts
```diff
@@ -188,7 +187,7 @@ export function useTransactionTx({
         return {
           name: getPreferredName.value(address) || '',
           address,
-        } as IAccount; // TODO establish the required return type for the function
+        } as any; // TODO establish the required return type for the function
```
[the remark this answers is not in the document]
peronczyk: `IAccount` was actually not correct. I'm changing this to `IAccountOverview`.

src/protocols/aeternity/libs/AeSdkSuperhero.ts
subhod-i: Oho, I didn't see the `_resolveAccount`. Something to tackle when we move the signing methods out of store.

src/utils/common.ts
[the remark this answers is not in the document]
peronczyk: There are 2 issues:

1. There is no native range type. There are some ways to do it (https://stackoverflow.com/questions/39494689/is-it-possible-to-restrict-number-to-a-certain-range) but it only makes sense for small ranges. When you try to make a large range this is what you get:
image

2. There is no typing for the infinity. When you try to use `typeof Infinity` TS has no knowledge how to handle it and returns `never`. Take a look:
image

So in short: there is a reason why TS engine does not provide any range typings for native JS functions. Instead it uses `number`. And I don't know any programming language that makes this easy or even possible.

src/composables/accounts.ts
peronczyk: I'll go with `newMnemonic` as agreed on private.

src/utils/common.ts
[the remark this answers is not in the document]
subhod-i: https://github.com/microsoft/TypeScript/issues/54925
Yup, the issue is still open. We can resolve this.

subhod-i (approved):
Code changes LGTM, We need to test this if the changes in the PR can read the existing storage keys and their data. @Liubov-crypto Please use the develop branch and make sure to have an account with few interactions and then use this PR to check the existing state is continued/resumed.

src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -26,8 +26,8 @@
           v-model.trim="signer.address"
           :default-text="$t('modals.createMultisigAccount.signerInputPlaceholder')"
           :label="getSignerLabel(index)"
-          item-title="address"
           :options="aeAccountsSelectOptions"
+          item-title="value"
```
CedrikNikita, line 30: what is `value` in this context?

CedrikNikita (approved):
Code changes look reasonable.

src/popup/components/Modals/MultisigVaultCreate.vue
peronczyk: This is a common concept for custom Vue select components where the items in the list are objects. So you are able to define which object property (key) is used as the title and the value. Reference: https://vuetifyjs.com/en/api/v-select/#props-item-title

PR #2348, opened 2023-09-18
[SW-1263] Remove languages from store

pstachel:


CedrikNikita (changes requested):
Language being preserved

src/composables/languages.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import {
+  FALLBACK_LOCALE,
+  i18n,
+  languages,
+} from '@/lib/i18n';
+import { useStorageRef } from '@/composables/composablesHelpers';
+
+const preferredLanguage = ((): SupportedLanguage => {
+  const code = navigator.language.split('-')[0] as SupportedLanguage;
```
konradodo, line 12:
```suggestion
  const [code] = navigator.language.split('-')
```

src/popup/App.vue
```diff
@@ -121,6 +132,44 @@ export default defineComponent({
       || routeMeta.value?.hideHeader
     ));
 
+    const NOTIFICATION_DEFAULT_SETTINGS: INotificationSetting[] = [
```
CedrikNikita, line 135: Why this const was moved?

src/lib/i18n.ts
```diff
@@ -0,0 +1,30 @@
...
+  messages: { en },
+});
+
+// @ts-ignore type coming from VueI18n is excessively deep and possibly infinite
+export const tg = i18n.global.t;
+
+export const languages: Record<SupportedLanguage, ILanguageConfig> = {
+  [SUPPORTED_LANGUAGES.en]: {
+    name: 'English',
+    getMessages: () => import(/* webpackChunkName: "locale-en" */ '../popup/locales/en.json'),
```
peronczyk, line 24: It makes no sense to make dynamic import if this is already imported in line 2.

src/popup/App.vue
peronczyk: I see the reason why you moved this - probably you didn't like the use of function `tg` in the constants. I agree this is bad. But I think we will move it when we will be working on moving the notification settings from vuex to composable. So for now I suggest to leave it as it is.

src/popup/App.vue
```diff
@@ -98,6 +108,7 @@ export default defineComponent({
     const store = useStore();
     const route = useRoute();
     const { t } = useI18n();
+    useLanguages(); // Runs the watcher for language change
```
peronczyk, line 111: But why we need a watcher?

src/popup/pages/LanguageSettings.vue
```diff
@@ -4,44 +4,40 @@
...
-          @click="switchLanguage(code)"
-        >
-          <div>
-            {{ name }}
-            <span>({{ code }})</span>
-          </div>
-        </div>
-      </RadioButton>
-    </div>
+    <RadioButton
```
peronczyk, line 7: I don't see a reason for removing the DIV that wraps the options.
In my opinion it's good to wrap lists with div, even if the div does nothing in terms of styling.

src/types/index.ts
```diff
@@ -658,3 +659,10 @@ export interface IWalletInfo {
   origin: any;
   type: any;
 }
+
+export type SupportedLanguage = keyof typeof SUPPORTED_LANGUAGES;
+
+export interface ILanguageConfig {
+  name: string;
+  getMessages: () => Promise<any>;
+}
```
peronczyk, line 668: As far as I see this interface is used only in one file so it's not a global thing. I suggest to define it in the file that uses it.

peronczyk (changes requested):

src/popup/App.vue
pstachel: I moved it because I was rewriting and moving the i18n component so I felt that it was reasonable to move it

src/composables/languages.ts
peronczyk, line 1: I see problem w having separate files for this composable and library. These 2 files are syncing changes between each other which makes no sense because the `i18n` instance returns nice computed property `i18n.global.locale`. So in practice what we need is one file that creates the `i18n` instance, returns:
1. current locale, eg.: `const activeLocale = computed(() => i18n.global.locale);`
2. method that allows to switch language.

This piece of code is neither composable nor lib. It's more like a plugin, so maybe the best location would be the `popup/plugins/`? What do you think?

src/popup/plugins/i18n.ts
```diff
@@ -0,0 +1,69 @@
...
+};
+
+function getPreferredLanguage(): SupportedLanguage {
+  const [code] = navigator.language.split('-') as [SupportedLanguage];
+  return languages[code] ? code : FALLBACK_LOCALE;
+}
+
+export const activeLanguage = computed(
+  () => (i18n.global.locale as any).value || getPreferredLanguage(),
+);
```
peronczyk, line 48: Is there actual need for using `getPreferredLanguage` as a fallback? I think if we set up the `locale` and `fallbackLocale` to `en` when using the `createI18n` the value should always be OK.

Also I'm not 100% sure, but maybe it would be better to do it this way:
```suggestion
export const activeLanguage = i18n.global.locale as SupportedLanguage;
```

src/popup/plugins/i18n.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export async function switchLanguage(languageCode: SupportedLanguage) {
+  if (languageCode === (i18n.global.locale as any).value) {
```
peronczyk, line 51: In my opinion this test looks off - to make it readable I think we need to compare the `languageCode` with `activeLanguage`.

src/popup/plugins/i18n.ts
```diff
@@ -0,0 +1,69 @@
...
+  [SUPPORTED_LANGUAGES.cn]: {
+    name: '中国人',
+    getMessages: () => import(/* webpackChunkName: "locale-cn" */ '../locales/cn.json'),
+  },
+};
+
+function getPreferredLanguage(): SupportedLanguage {
+  const [code] = navigator.language.split('-') as [SupportedLanguage];
+  return languages[code] ? code : FALLBACK_LOCALE;
+}
```
peronczyk, line 44: As the `navigator.language.split` is set by the browser only once (it doesn't change) I think there is no need to create a function for this. I prefer the way how @CedrikNikita did it previously:
```
const preferredLocale = (() => {
  const code = navigator.language.split('-')[0];
  return languages[code] ? code : fallbackLocale;
})();
```

src/popup/plugins/i18n.ts
```diff
@@ -0,0 +1,69 @@
...
+    return;
+  }
+  if (!(i18n.global.availableLocales as string[]).includes(languageCode)) {
+    const messages = (await languages[languageCode].getMessages()).default;
+    i18n.global.setLocaleMessage(languageCode, messages);
+  }
+  (i18n.global.locale as any).value = languageCode;
+
+  // LocalStorage is accessed directly in this file to avoid circular dependencies
+  localStorage.setItem(ACTIVE_LANGUAGE, JSON.stringify(languageCode));
```
peronczyk, line 61: This is bad, we should not use the `localStorage` directly. Instead I suggest to use the `WalletStorage` lib. It is also causing the dependency cycle?

src/popup/plugins/i18n.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+(() => {
+  const storedLanguage = localStorage.getItem(ACTIVE_LANGUAGE);
```
peronczyk, line 65: IDEM

src/composables/languages.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (!(i18n.global.availableLocales as string[]).includes(languageCode)) {
+      const messages = (await languages[languageCode].getMessages()).default;
+      i18n.global.setLocaleMessage(languageCode, messages);
+    }
+    (i18n.global.locale as any).value = languageCode;
+
+    WalletStorage.set(ACTIVE_LANGUAGE_KEY, languageCode);
+  }
+
+  async function fetchStoredLanguage() {
```
peronczyk, line 27: This method is not fetching anything :)
```suggestion
  async function restoreLanguage() {
```

peronczyk (approved):

CedrikNikita (approved):
Code changes look reasonable.

Liubov-crypto:
cus

Liubov-crypto:
> feat-remove-languages-from-store.wallet.z52da5wt.xyz

Reproducible on develop too, will create a separate issue.

PR #2353, opened 2023-09-20
[SW-1266] Reset the wallet path if user is idle for 10 min

pstachel:


konradodo (approved):

src/composables/composable-helpers/mountedBasedPolling.ts
peronczyk, line 1: I suggest to rename this file based on the same pattern as composables. So if composable function is `useFoo` the file is `foo`. In this case as the function name is `createPollingBasedOnMountedComponents` I sggest to name the file `pollingBasedOnMountedComponents`.

src/composables/composable-helpers/storageRef.ts
```diff
@@ -1,16 +1,7 @@
...
-  onBeforeUnmount,
-  onMounted,
-  Ref,
-  ref,
-  watch,
-} from 'vue';
-import { isEqual } from 'lodash-es';
+import { Ref, ref, watch } from 'vue';
 import type { StorageKeysInput } from '@/types';
+import { isEqual } from 'lodash-es';
```
peronczyk, line 3: Please put this above local imports.

peronczyk (changes requested):

src/popup/router/index.ts
```diff
@@ -130,12 +127,10 @@ router.beforeEach(async (to, from, next) => {
...
 router.afterEach(async (to) => {
-  if (RUNNING_IN_POPUP) return;
-  if (to.meta?.notPersist) {
-    await browser?.storage.local.remove(lastRouteKey);
-  } else {
-    await browser?.storage.local.set({ [lastRouteKey]: to.path });
+  if (RUNNING_IN_POPUP) {
+    return;
   }
+  setLastPath(to.meta?.notPersist ? '' : to.path);
```
peronczyk, line 133:
```suggestion
  if (RUNNING_IN_POPUP) {
    setLastPath(to.meta?.notPersist ? '' : to.path);
  }
```

But to be honest I would remove the test because I don't see any downside of seving the last used path to local storage on every environment. The matter is what we do with this value.

src/composables/ui.ts
```diff
@@ -1,10 +1,35 @@
...
+    setLocalStorageItem([LAST_PATH_KEY], path);
+  }
+
+  if (
+    IS_EXTENSION
+    && extensionExpirationTime
+    && dayjs().isAfter(extensionExpirationTime)
+  ) {
+    setLastPath('/');
+  }
```
peronczyk, line 31: This block of code should be placed right before the `return`.

peronczyk (changes requested):

src/popup/router/index.ts
CedrikNikita: Not awaiting for a `await browser.storage.local.get` and `await browser?.storage.local.set/remove` methods brought us an optimization of ~200ms on popup opening.
https://github.com/aeternity/superhero-wallet/pull/2077#issue-1709559390

peronczyk: @CedrikNikita yes, your change brought optimisation because there was `await` there. In Paweł's implementation there is no await.

src/popup/router/index.ts
```diff
@@ -61,14 +63,18 @@ RouteQueryActionsController.init(router);
...
   await watchUntilTruthy(() => store.state.isRestored);
-  next(
-    (
-      !RUNNING_IN_POPUP
-      && to.name === ROUTE_INDEX
-      && ((await browser?.storage.local.get(lastRouteKey)) as any)[lastRouteKey]
-    )
-    || undefined,
-  );
+  const storedRoute = (await browser?.storage.local.get(LAST_ROUTE_KEY) as any)[LAST_ROUTE_KEY];
```
peronczyk, line 66: Please use `WalletStorage` here.

src/popup/router/index.ts
```diff
@@ -137,11 +143,13 @@ router.beforeEach(async (to, from, next) => {
 });
 
 router.afterEach(async (to) => {
-  if (RUNNING_IN_POPUP) return;
+  if (RUNNING_IN_POPUP) {
+    return;
+  }
   if (to.meta?.notPersist) {
-    await browser?.storage.local.remove(lastRouteKey);
+    await browser?.storage.local.remove(LAST_ROUTE_KEY);
```
peronczyk, line 150: This won't work as the wallet storage is prefixing all the keys.

src/popup/router/index.ts
```diff
@@ -61,14 +64,19 @@ RouteQueryActionsController.init(router);
...
   await watchUntilTruthy(() => store.state.isRestored);
-  next(
-    (
-      !RUNNING_IN_POPUP
-      && to.name === ROUTE_INDEX
-      && ((await browser?.storage.local.get(lastRouteKey)) as any)[lastRouteKey]
-    )
-    || undefined,
-  );
+  const lastRoute = await WalletStorage.get(LAST_ROUTE_KEY);
```
CedrikNikita, line 67: The point of `RUNNING_IN_POPUP` check was not to `await` this.

src/popup/router/index.ts
```diff
@@ -61,14 +64,20 @@ RouteQueryActionsController.init(router);
...
-      !RUNNING_IN_POPUP
-      && to.name === ROUTE_INDEX
-      && ((await browser?.storage.local.get(lastRouteKey)) as any)[lastRouteKey]
-    )
-    || undefined,
-  );
+
+  // This check is run to avoid unnecessary fetching from WalletStorage
+  if (RUNNING_IN_POPUP || to.name !== ROUTE_INDEX) {
+    next(undefined);
```
CedrikNikita, line 70: In this case it will still await the `lastRoute` afterwards.

peronczyk: I reorganised code that @pstachel did by creating separate lib. Can you please take a look again?

peronczyk (approved):

CedrikNikita (approved):
Code changes look reasonable.

PR #2364, opened 2023-09-28
[SW-1297] Move names from Vuex to composable

pstachel:


src/types/index.ts
```diff
@@ -483,12 +483,14 @@ export interface ITopHeader {
...
 
+export type ChainName = `${string}.chain`;
+
 export interface IName {
   autoExtend: boolean;
   createdAtHeight: number;
   expiresAt: number;
   hash: string;
-  name: string;
+  name: ChainName
```
peronczyk, line 493:
```suggestion
  name: ChainName;
```

src/types/index.ts
```diff
@@ -522,7 +525,9 @@ export interface IMiddleware {
   getName: GenericApiMethod;
   getNamePointees: GenericApiMethod;
   getNamePointers: GenericApiMethod;
-  getNames: (address: string) => Promise<any>;
+  getNames: (
+    { owned_by, state, limit }: { owned_by: Encoded.AccountAddress, state: string, limit: number }
```
peronczyk, line 529:
```suggestion
    options: { owned_by: Encoded.AccountAddress; state: string; limit: number }
```

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+interface IAuctionEntryParams {
+  name: ChainName;
+  expiration: number;
+  bids: IAuctionBid[];
+}
+
+const ownedNames = useStorageRef<IName[]>([], 'owned-names');
+const pendingAutoExtendNames = ref<ChainName[]>([]);
+const areNamesFetching = ref(false);
+const defaults = ref<Record<string, ChainName>>({});
```
peronczyk, line 57: To make i more readable can you please rename this constant to something like `defaultNamesRegistry`?
Also it would be nice if you could add a comment about what the `string` stands for.

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+const preferred = ref<Record<string, string>>({});
```
peronczyk, line 58: To make it more readable can you please rename this constant to something like `preferredNamesRegistry`?
Also it would be nice if you could add a comment about what the `string` stands for. And a general comment what `preferred` means.

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+      name,
+      address,
+      type = UPDATE_POINTER_ACTION.update,
+    }: IUpdatePointerParams,
+  ) {
+    const aeSdk = await getAeSdk();
+    const nameEntry = await aeSdk.aensQuery(name);
+    try {
+      if (type === UPDATE_POINTER_ACTION.extend) {
+        // @ts-ignore
```
peronczyk, line 180: Why?

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+        openDefaultModal({
+          msg: isInsufficientBalanceError(error)
+            ? tg('modals.insufficient-balance.msg')
+            : error.message,
+        });
+      }
+    }
+  }
+
+  function setDefault({ address, name }: IAddressNamePair) {
```
peronczyk, line 201: I suggest to rename this to `setDefaultName`.

src/protocols/aeternity/composables/aeNames.ts
peronczyk: I see that this `string` is a `<NodeNetworkID>-<address>`. Maybe we could make the object multidimentional instead?
So the structure would look as follows:
```
{
  'mainnet': {
    'ak_abcdfg...': 'somename.chain',
  }
}
```

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+    if (name) {
+      defaults.value[`${address}-${nodeNetworkId.value}`] = name;
+    } else {
+      delete defaults.value[`${address}-${nodeNetworkId.value}`];
+    }
+  }
+
+  function setDefaultNames() {
```
peronczyk, line 209: As this is not a setter I suggest to follow same pattern as other libs/composables and rename this to `updateDefaultNames`.

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+          name,
+          hash,
+        }))),
+      ])),
+    ).then((arr) => arr.flat(2)) as IName[];
+
+    areNamesFetching.value = true;
+  }
+
+  async function updatePointer(
```
peronczyk, line 169:
```suggestion
  async function updateNamePointer(
```

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+    setPreferred(address);
+
+    return preferred.value[`${address}-${nodeNetworkId.value}`] || '';
+  }
+
+  function getAuction(name: string): IAuction {
+    return auctions.value[name] || null;
+  }
+
+  function getHighestBid(name: string): IAuctionBid {
```
peronczyk, line 123:
```suggestion
  function getNameAuctionHighestBid(name: string): IAuctionBid {
```

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+    if (getLastActiveProtocolAccount(PROTOCOL_AETERNITY)?.address === address) {
+      return getDefaultName(address);
+    }
+
+    setPreferred(address);
+
+    return preferred.value[`${address}-${nodeNetworkId.value}`] || '';
+  }
+
+  function getAuction(name: string): IAuction {
```
peronczyk, line 119:
```suggestion
  function getNameAuction(name: string): IAuction {
```

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  function getDefaultName(address?: Encoded.AccountAddress) {
+    if (!address) {
+      return '';
+    }
+    return defaults.value ? defaults.value[`${address}-${nodeNetworkId.value}`] : '';
+  }
+
+  async function setPreferred(address: Encoded.AccountAddress) {
```
peronczyk, line 96:
```suggestion
  async function setPreferredName(address: Encoded.AccountAddress) {
```

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+    getMiddlewareRef,
+    fetchFromMiddlewareCamelCased,
+  } = useMiddleware();
+
+  const {
+    aeAccounts,
+    getLastActiveProtocolAccount,
+  } = useAccounts();
+
+  function getName(name: string): IName | undefined {
```
peronczyk, line 85:
```suggestion
  function getOwnedName(name: string): IName | undefined {
```

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+
+  function getAuction(name: string): IAuction {
+    return auctions.value[name] || null;
+  }
+
+  function getHighestBid(name: string): IAuctionBid {
+    return getAuction(name)?.bids.reduce((a, b) => (a.nameFee.isGreaterThan(b.nameFee) ? a : b));
+  }
+
+  async function fetchOwnedNames() {
```
peronczyk, line 127: Functions with names starting with `fetch...` should return fetched result. Here the function is updating the state so the name should start with `update...`.

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+    areNamesFetching.value = true;
+
+    function fetchPendingNameClaimTransactions(address: Encoded.AccountAddress) {
```
peronczyk, line 130: I suggest to move this function out of the scope of the `fetchOwnedNames` function.

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+    } else {
+      delete defaults.value[`${address}-${nodeNetworkId.value}`];
+    }
+  }
+
+  function setDefaultNames() {
+    aeAccounts.value.map(async ({ address }) => {
+      const response = await fetchJson(
+        `${aeActiveNetworkSettings.value.backendUrl}/profile/${address}`,
+      ).catch(() => {});
```
peronczyk, line 213:
```suggestion
      ).catch(() => ({}));
```

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+  function setAutoExtend(name: ChainName) {
+    const selectedName = ownedNames.value.find((n) => n.name === name);
+    if (selectedName) {
+      selectedName.autoExtend = !selectedName.autoExtend;
+    }
+  }
+
+  function setAuctionEntry(
+    { name, expiration, bids }: IAuctionEntryParams,
+  ) {
```
peronczyk, line 228: I don't think it's reasonable to split this line.

src/popup/components/NameItem.vue
```diff
@@ -140,23 +140,38 @@ import {
...
+import {
+  useAccounts,
+  useAeSdk,
+  useModals,
+  useTopHeaderData,
+} from '@/composables';
 import { checkAddressOrChannel } from '@/protocols/aeternity/helpers';
+import { useAeNames } from '@/protocols/aeternity/composables/aeNames';
 
+import { useAeNetworkSettings } from '@/protocols/aeternity/composables';
```
peronczyk, line 174: Please add empty line after this one and remove empty line before.

src/popup/components/NameItem.vue
```diff
@@ -193,16 +208,23 @@ export default defineComponent({
     const { activeAccount } = useAccounts();
     const { t } = useI18n();
     const { topBlockHeight } = useTopHeaderData({ store });
+    const {
+      setAutoExtend,
+      updatePointer,
+      getName,
+      getDefaultName,
+      setDefault,
```
peronczyk, line 216: Please order this list alphabetically :)

src/popup/components/NameItem.vue
```diff
@@ -193,16 +208,23 @@ export default defineComponent({
...
     const expand = ref(false);
     const newPointer = ref<string>('');
     const showInput = ref(false);
     const error = ref(false);
     const pointerInput = ref();
 
-    const getDefaultName = store.getters['names/getDefault'] as (a?: string) => string | undefined;
-
-    const nameEntry = computed<IName | null>(() => store.getters['names/get'](props.name));
+    const nameEntry = computed((): IName | undefined => getName(props.name));
```
peronczyk, line 227: As this computed simply returns the function I suggest to remove the typings as it is inherited from `getName`.

src/popup/components/NameItem.vue
```diff
@@ -252,33 +274,44 @@ export default defineComponent({
       showInput.value = false;
     }
 
-    async function setDefault() {
-      await store.dispatch('names/setDefault', {
-        address: activeAccount.value.address,
-        name: props.name,
+    async function handleSetDefault() {
```
peronczyk, line 277: I think the body of this function should be moved to `aeNames` composable - similar as it was before done in the Vuex store.

src/popup/components/TransactionOverview.vue
```diff
@@ -35,8 +35,9 @@ import {
   useMiddleware,
   useTransactionTx,
 } from '@/composables';
-import { useGetter } from '@/composables/vuex';
 import { useAeNetworkSettings } from '@/protocols/aeternity/composables';
+import { useAeNames } from '@/protocols/aeternity/composables/aeNames';
+
```
peronczyk, line 40: This line can be removed.

src/popup/pages/Names/AuctionList.vue
```diff
@@ -125,7 +132,21 @@ export default defineComponent({
...
+      activeAuctions.value = (
+        await fetchAllPages(
+          () => middleware.getNamesAuctions({ by: 'expiration', direction: 'forward', limit: 100 }),
+          fetchFromMiddlewareCamelCased,
+        )
+      ).map(({ name, info }) => ({
+        name,
+        expiration: info.auctionEnd,
+        lastBid: info.lastBid.tx,
+      }));
```
peronczyk, line 148: I think this logic should be stored in the `aeNames` composable.

src/popup/components/NameItem.vue
```diff
@@ -252,33 +274,44 @@ export default defineComponent({
...
+        body: {
+          preferredChainName: name,
+        },
       });
+
+      const respondChallenge = await fetchRespondChallenge(response);
+
+      await postJson(url, { body: respondChallenge });
+
+      setDefault({ address, name });
```
konradodo, line 292: what about try catch block?

src/popup/pages/Names/Claim.vue
```diff
@@ -189,18 +192,17 @@ export default defineComponent({
...
           if (AE_AENS_NAME_AUCTION_MAX_LENGTH < fullName.value.length) {
-            store.dispatch('names/updatePointer', {
+            await updatePointer({
               name: fullName.value,
               address: activeAccount.value.address,
             });
           }
-        } catch (e: any) {
-          openDefaultModal({ msg: e.message });
+        } catch (error: any) {
```
konradodo, line 202: can you typing this error?

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+
+  function getName(name: string): IName | undefined {
+    return ownedNames.value.find((ownedName) => ownedName.name === name);
+  }
+
+  function getDefaultName(address?: Encoded.AccountAddress) {
+    if (!address) {
+      return '';
+    }
+    return defaults.value ? defaults.value[`${address}-${nodeNetworkId.value}`] : '';
```
konradodo, line 93:
```suggestion
    return address && defaults.value ? defaults.value[`${address}-${nodeNetworkId.value}`] : '';
```

```suggestion
   if (!address || !defaults.value) {
     return ''
   }
    return  defaults.value[`${address}-${nodeNetworkId.value}`];
```

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+          pointers: info.pointers,
+          autoExtend: (
+            ownedNames.value.find((n) => n.name === name)?.autoExtend
+            || pendingAutoExtendNames.value?.includes(name)
+          ),
+          name,
+          hash,
+        }))),
+      ])),
+    ).then((arr) => arr.flat(2)) as IName[];
```
konradodo, line 164: please not mix together async await and promise

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+      if (type === UPDATE_POINTER_ACTION.extend) {
+        // @ts-ignore
+        await nameEntry.extendTtl();
+      } else if (type === UPDATE_POINTER_ACTION.update) {
+        await aeSdk.aensUpdate(name, { account_pubkey: address }, { extendPointers: true });
+      }
+      openDefaultModal({
+        msg: tg('pages.names.pointer-added', { type }),
+      });
+    } catch (error: any) {
```
konradodo, line 188: why any?

pstachel: Did we type any of the errors?

peronczyk: Generally it is not possible to type the `error` when catching it. Using `any` is a common thing.

src/popup/components/NameItem.vue
pstachel: Setting the default value for single account address occurs only here so I'd prefer to have it moved here

src/protocols/aeternity/composables/aeNames.ts
CedrikNikita: https://github.com/aeternity/aepp-sdk-js/blob/de987107ecab1e5c2b8d78e7bcbcf446fc36cfa1/src/aens.ts#L243

src/protocols/aeternity/composables/aeNames.ts
peronczyk: After some digging and explanation I see that the name `preferred` is misleading as this state holds the names of the accounts that are not local. So I think more suitable name would be for example `external` (`externalNamesRegistry`).

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+
+    const key = `${address}-${nodeNetworkId.value}`;
+    if (preferredChainName) {
+      preferred.value[key] = preferredChainName;
+    } else {
+      delete preferred.value[key];
+    }
+  }
+
+  function getPreferredName(address: Encoded.AccountAddress) {
```
peronczyk, line 109: Instead of having 2 separate functions `getDefaultName` and `getPreferredName` I suggest to merge them into one `getName` which would check if provided address is a local address (`accounts` composable / `isLocalAccountAddress`) and returning owned name. In other case try to return external address name.

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,290 @@
...
+    if (getLastActiveProtocolAccount(PROTOCOL_AETERNITY)?.address === address) {
+      return getDefaultName(address);
+    }
+
+    setPreferred(address);
+
+    return preferred.value[`${address}-${nodeNetworkId.value}`] || '';
```
peronczyk, line 116: I feel this makes no sense. Let's say we are trying to display name for an address that is not stored. Function `setPreferred` is fired and the name for the address is updated. But the `setPreferred` is an asynchronous function, so returned value would be `''`.

This is why I think instead of returning raw value we need to return `computed` property so the name reference on the page will update when the `setPreferred` function finish it's work.

konradodo (approved):

src/store/plugins/names.js
```diff
@@ -1,273 +0,0 @@
-import { watch } from 'vue';
```
CedrikNikita, line 1: This file has not been deleted.

src/constants/common.ts
```diff
@@ -414,3 +414,8 @@ export const BROWSER_ACTIONS = {
   share: 'share',
   favourite: 'favourite',
 } as const;
+
+export const UPDATE_POINTER_ACTION = {
```
CedrikNikita, line 418: Actually this const is not common to every blockchain, it is only for aeternity.

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,315 @@
...
+    fetchFromMiddlewareCamelCased,
+  } = useMiddleware();
+
+  const {
+    aeAccounts,
+    isLocalAccountAddress,
+    getLastActiveProtocolAccount,
+  } = useAccounts();
+
+  function getOwnedName(name: string): IName | undefined {
```
peronczyk, line 93: If the `getName` function returns string I suggest to rename this method to `getOwnedNameData` or `getOwnedNameDetails` as it returns an object.

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    if (preferredChainName) {
+      externalNamesRegistry.value[nodeNetworkId.value!][address] = preferredChainName;
+    } else {
+      delete externalNamesRegistry.value[nodeNetworkId.value!][address];
+    }
+  }
+
+  function getName(address?: Encoded.AccountAddress) {
```
peronczyk, line 113: I suggest to type the returned value here.
And maybe add a comment why we return `computed`.

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -0,0 +1,315 @@
...
+      return;
+    }
+    if (!defaultNamesRegistry.value[nodeNetworkId.value]) {
+      defaultNamesRegistry.value[nodeNetworkId.value] = {};
+    }
+
+    if (name) {
+      defaultNamesRegistry.value[nodeNetworkId.value][address] = name;
+    } else {
+      delete defaultNamesRegistry.value[`${address}-${nodeNetworkId.value}`];
```
peronczyk, line 236: The key here is wrong.

src/protocols/aeternity/views/TransferReceiveModal.vue
```diff
@@ -49,7 +49,7 @@ export default defineComponent({
       : activeAccount.value.address);
 
     const activeAccountName = computed(
-      () => props.isMultisig ? undefined : getDefaultName(activeAccount.value.address),
+      () => props.isMultisig ? undefined : getName(activeAccount.value.address).value,
```
peronczyk, line 52:
```suggestion
const activeAccountName = props.isMultisig ? undefined : getName(activeAccount.value.address);
```
What do you think?

src/popup/components/TransactionOverview.vue
peronczyk, line 1: I suggest to test of there is actually need to add `.value` to `getName()`. I think that it will be unwrapped automatically as the whole object is a composable.

src/popup/components/AccountInfo.vue
```diff
@@ -86,9 +87,9 @@ export default defineComponent({
   },
   setup(props) {
     const store = useStore();
-    const getDefaultName = store.getters['names/getDefault'] as (a?: string) => string;
+    const { getName } = useAeNames({ store });
 
-    const name = computed(() => getDefaultName(props.account.address));
+    const name = computed(() => getName(props.account.address!).value);
```
peronczyk, line 92: Probably there is no need to add `value` as it will be automatically unwrapped by parent computed.

peronczyk (approved):

PR #2366, opened 2023-09-29
[SW-1340] fix: be able to open a tip deeplink on mobile

CedrikNikita:


src/composables/deepLinkApi.ts
```diff
@@ -26,8 +28,11 @@ export function useDeepLinkApi({ router }: UseDeepLinkApiOptions) {
   /**
    * Function needed to support legacy tipping from superhero.com
    */
-  function checkIfOpenTransferSendModal() {
-    if (checkIfSuperheroCallbackUrl(route.query)) {
+  async function checkIfOpenTransferSendModal(currentRoute: Route) {
+    if (
+      currentRoute.path.slice(1) === ROUTE_ACCOUNT
```
pstachel, line 33: Should we open the transfer modal only on the account page?

pstachel, line 33: Wouldn't `currentRoute.name` resolve the need of using slice?

pstachel (approved):

src/composables/deepLinkApi.ts
CedrikNikita: `name` is `undefined` in this case.

CedrikNikita: This check is used only on the `/account` page. And the legacy code we are trying to support is aiming only for this page.

Liubov-crypto:
LGTM

martinkaintas (approved):

peronczyk (approved):

PR #2369, opened 2023-10-03
[SW-1347] Move permissions vuex state to composable

peronczyk:


subhod-i (changes requested):
Existing permissions are not being considered.

peronczyk:
@subhod-i 
> Existing permissions are not being considered.

Thank you for finding this out. The issue is fixed now. The reason was the lack of state synchronisation between the extension and the background. So background didn't see the permissions added within the extension app.

src/composables/permissions.ts
```diff
@@ -0,0 +1,138 @@
...
+    const limitLeft = transactionSignLimitLeft - +aettosToAe(+amount + +fee + +nameFee);
+
+    if (!transactionSignLimit) {
+      return false;
+    }
+    if (
+      !transactionSignFirstAskedOn
+      || (
+        new Date().getTime() - new Date(transactionSignFirstAskedOn).getTime()
+        >= 24 * 60 * 60 * 1000
```
pstachel, line 67: What do you think of using dayjs for date and time comparison?

src/composables/permissions.ts
```diff
@@ -0,0 +1,138 @@
...
+    if (checkPermissions(host, METHODS.address)) {
+      return true;
+    }
+    try {
+      await connectionPopupCallback();
+      return true;
+    } catch {
+      return false;
+    }
+  }
```
pstachel, line 127: Those two methods are exactly the same - maybe you could create base function for them and call it with different parameter?

pstachel (approved):

src/composables/permissions.ts
peronczyk: After doing some research I think there is a chance we could remove the `dayjs` lib as it gives no real benefit. We do only a little of date transformations and comparisons and having a 3rd party lib for this makes no sense for me.

pstachel: I think the most important usage of it is when we display relativeTime in NotificationList & TransactionList - that's when it was introduced.

src/composables/aeSdk.ts
```diff
@@ -124,49 +125,39 @@ export function useAeSdk({ store }: IDefaultComposableOptions) {
...
+            app: {
+              name: url.hostname,
+              icons: [],
+              protocol: url.protocol,
+              host: url.host,
+              url: url.href,
+            },
+          });
+
+        if (!(await requestAllAddressesForHost(url.host, connectionPopupCallback))) {
```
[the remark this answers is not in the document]
peronczyk: Good point. I moved this logic to the permissions composable.

src/composables/permissions.ts
peronczyk: Yes, I refactored it. Please take a look again.

src/constants/common.ts
```diff
@@ -346,15 +346,13 @@ export const POPUP_TYPE_ACCOUNT_LIST = 'account-list';
...
-export const POPUP_TYPE_TX_SIGN = METHODS.sign;
 
 export const POPUP_TYPES = [
   POPUP_TYPE_CONNECT,
   POPUP_TYPE_ACCOUNT_LIST,
   POPUP_TYPE_SIGN,
   POPUP_TYPE_MESSAGE_SIGN,
   POPUP_TYPE_RAW_SIGN,
-  POPUP_TYPE_TX_SIGN,
 ] as const;
```
subhod-i, line 356: Why was `POPUP_TYPE_TX_SIGN` remvoed? Is there no modal route to this?

src/lib/accounts/AccountSuperhero.ts
```diff
@@ -36,11 +36,11 @@ export class AccountSuperhero extends AccountBase {
 
   signTransaction(txBase64: Encoded.Transaction, options: any): Promise<Encoded.Transaction> {
     if (IS_MOBILE_APP && options.aeppOrigin) {
-      return this.fgPermissionCheckAndSign(POPUP_TYPE_SIGN, txBase64, options, options.aeppOrigin);
+      return this.fgPermissionCheckAndSign(METHODS.sign, txBase64, options, options.aeppOrigin);
     }
     if (IS_EXTENSION_BACKGROUND) {
       return this.bgPermissionCheckAndSign(
-        POPUP_TYPE_TX_SIGN,
+        METHODS.sign,
```
subhod-i, line 43: `POPUP_TYPE_SIGN` and `POPUP_TYPE_TX_SIGN` values are different, You can use `METHODS.sign` at both places but need to migration permission values of the key('sign' -> 'transaction.sign')

src/migrations/003-permissions-vuex-to-composable.ts
```diff
@@ -0,0 +1,16 @@
...
+import { collectVuexState } from './migrationHelpers';
+
+const migration: Migration = async (restoredValue: PermissionRegistry) => {
+  if (!restoredValue || !Object.keys(restoredValue).length) {
+    const permissions = (await collectVuexState())?.permissions;
+    try {
+      if (permissions && Object.keys(permissions).length) {
+        return permissions;
+      }
+    } catch (error) { /* noop */ }
```
subhod-i, line 11: Please introduce `handleUnknownError` in the catch block.

src/store/modules/accounts/hdWallet.js
```diff
@@ -35,21 +36,22 @@ export default {
...
         await openModal(MODAL_CONFIRM_RAW_SIGN, { data: txBase64, app });
         return;
       }
       const txObject = unpackTx(txBase64);
 
-      const checkTransactionSignPermission = await dispatch('permissions/checkTransactionSignPermission', {
-        ...txObject,
-        host: app?.host || null,
-      }, { root: true });
+      const permitted = checkTransactionSignPermission(
```
subhod-i, line 49:
```suggestion
      const hasPermission = checkTransactionSignPermission(
```
or `canAccess`

subhod-i (changes requested):

src/constants/common.ts
peronczyk: Yes, we had the discussion on chat that there is no such modal.

src/migrations/003-permissions-vuex-to-composable.ts
peronczyk: It makes no sense to do anything as this is a migration. If user has the previous state corrupted we need to catch the error to avoid breaking the app but we do nothing with this fact.

src/lib/accounts/AccountSuperhero.ts
peronczyk: Yes, that's true - the values are different. And this actually was wrong. Let's break it to small pieces: The `POPUP_TYPE_TX_SIGN` was equal to `METHODS.sign` which is equal to `"transaction.sign"` string. The `POPUP_TYPE_SIGN` was equal to `"sign"` string.
In our previous code (and the new one) there was no logic for method with value of `"sign"` - this method: https://github.com/aeternity/superhero-wallet/blob/develop/src/store/modules/permissions.js#L75 was always returning `undefined` in this case.

src/migrations/003-permissions-vuex-to-composable.ts
subhod-i: Any try catch block doing nothing at the catch is risky. It's hard to debug in the future.

src/background/popupHandler.ts
```diff
@@ -12,7 +12,7 @@ const popups: Dictionary = {};
 
 export const getAeppUrl = (v: any) => new URL(v.connection.port.sender.url);
 
-export const showPopup = async (aepp: any, type: string, params?: any) => {
+export const openPopup = async (type: PopupType, aepp: any, params?: any) => {
```
peronczyk, line 15: I changed the name of this action to match the `openModal`. So now we have:
```
openPopup
openModal
```

src/migrations/003-permissions-vuex-to-composable.ts
peronczyk: OK, I'll use `handleUnknownError`.

src/background/popupHandler.ts
subhod-i: Nice.

src/store/modules/accounts/hdWallet.js
```diff
@@ -35,21 +36,25 @@ export default {
...
       }
       const txObject = unpackTx(txBase64);
 
-      const checkTransactionSignPermission = await dispatch('permissions/checkTransactionSignPermission', {
-        ...txObject,
-        host: app?.host || null,
-      }, { root: true });
+      const hasPermission = checkTransactionSignLimit(
+        app?.host || null,
+        txObject, // TODO verify if this is correct
```
subhod-i, line 54: TODO shall be verified before merging this.

src/composables/permissions.ts
```diff
@@ -0,0 +1,183 @@
...
+    popup: POPUP_TYPE_CONNECT,
+  },
+  [METHODS.address]: {
+    modal: MODAL_CONFIRM_ACCOUNT_LIST,
+    popup: POPUP_TYPE_ACCOUNT_LIST,
+  },
+  [METHODS.signMessage]: {
+    modal: MODAL_MESSAGE_SIGN,
+    popup: POPUP_TYPE_MESSAGE_SIGN,
+  },
```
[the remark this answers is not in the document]
peronczyk: Yes, because adding `signTransaction` here would require another big refactor. It is because if you try to use `hdWallet/signTransaction` action it also fires the `confirmTxSigning` action which opens the **modal** window that asks for the permission. I will deal with this in next task. For now I left `TODO` comment there: https://github.com/aeternity/superhero-wallet/pull/2369/files#diff-37085670e4cc4d0853535bb598e3c36b31ffb49af1badbebe2018c1591d51b9cR41

src/store/modules/accounts/hdWallet.js
peronczyk: You are the the author of this code :)
I think this is wrong, but I'm afraid to touch it as it actually works. What do you think about creating separate task for this?

subhod-i: For now, I agree to keep it as it is. We can add a related task to the board. 

I am repeating this @Liubov-crypto this PR must be tested for all types of transaction signing. Full regression testing.

src/composables/permissions.ts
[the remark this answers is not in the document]
subhod-i: Okay, Commented.

src/composables/permissions.ts
```diff
@@ -0,0 +1,183 @@
...
+    const limitLeft = transactionSignLimitLeft - +aettosToAe(+amount + +fee + +nameFee);
+
+    if (!transactionSignLimit) {
+      return false;
+    }
+    if (
+      !transactionSignFirstAskedOn
+      || (
+        new Date().getTime() - new Date(transactionSignFirstAskedOn).getTime()
+        >= 24 * 60 * 60 * 1000
```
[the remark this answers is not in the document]
peronczyk: Make sense, done.

subhod-i (approved):
LGTM.

src/composables/permissions.ts
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * Composable used to store, compare or ask for permissions to do any DAPP related action,
+ * like connecting the wallet, signing messages and transactions.
+ */
+export function usePermissions() {
+  const { openModal } = useModals();
+
+  function updateTransactionSignSpent(host: string, value: number) {
+    permissions.value[host] = {
```
CedrikNikita, line 65: Why not just `permissions.value[host].transactionSignSpent = value;`?

src/composables/permissions.ts
```diff
@@ -0,0 +1,183 @@
+import { METHODS } from '@aeternity/aepp-sdk';
```
CedrikNikita, line 1: Maybe we can have a new line after that, same as we have in other files for external and internal dependencies.

CedrikNikita (approved):
Code changes look decent. Agree with @subhod-i that it requires a proper testing.

src/lib/accounts/AccountSuperhero.ts
```diff
@@ -34,84 +31,47 @@ export class AccountSuperhero extends AccountBase {
...
-            app: {
-              name: app?.hostname,
-              icons: [],
-              protocol: app?.protocol,
-              host: app?.host,
-              url: app?.href,
+      if (IS_EXTENSION_BACKGROUND && aeppOrigin) {
+        const txObject = unpackTx(txBase64);
+        if (
+          (checkPermission(new URL(aeppOrigin).host, METHODS.sign, txObject as any))
```
martinkaintas, line 46: Why `as any` and not `ITransactionSignPermissionOptions`

peronczyk:
@CedrikNikita , @subhod-i - I requested regression testing for all stuff related to signing: https://github.com/superhero-com/superhero-wallet/issues/2419#issuecomment-1774711274

src/lib/accounts/AccountSuperhero.ts
peronczyk: This was a temporary thing. I did some research and the `txObject` has `amount` property which fits the requirement of the `checkPermission` method. So I changed the `ITransactionSignPermissionOptions` type to allow to pass any object that has at least one of the required properties: `amount`, `fee` or `nameFee`. Now there is no need to use `as` here.

src/composables/permissions.ts
```diff
@@ -0,0 +1,180 @@
...
+    return true;
+  }
+
+  function checkPermission(
+    host: string,
+    method: METHODS,
+    transactionSignOptions: ITransactionSignPermissionOptions = {},
+  ): boolean {
+    const permission = permissions.value[host];
+    console.log('checkPermission', { method, transactionSignOptions, permission });
```
martinkaintas, line 118:
```suggestion
    
```

martinkaintas (approved):

src/composables/permissions.ts
peronczyk: Removed :)

PR #2373, opened 2023-10-05
[SW-1347] Move invites vuex state to composable

peronczyk:


src/composables/invites.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (!isMax) {
+      await aeSdk.spend(
+        amount,
+        recipientId,
+        // @ts-ignore
+        { denomination: AE_AMOUNT_FORMATS.AE },
+      );
+    } else {
+      await aeSdk.transferFunds(
+        1,
```
pstachel, line 63: Maybe it's worth to comment this line why are we sending `1`

pstachel (approved):

src/composables/invites.ts
peronczyk: @CedrikNikita can you provide some info about this?

CedrikNikita: First argument is a percentage from 0 to 1, where 0 is 0% and 1 is 100%.

subhod-i (approved):
LGTM.

src/composables/invites.ts
```diff
@@ -0,0 +1,93 @@
+import { STORAGE_KEYS } from '@/constants';
+import {
+  AE_AMOUNT_FORMATS,
+  AeSdk,
+  Encoded,
+  Node,
+} from '@aeternity/aepp-sdk';
```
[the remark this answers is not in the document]
peronczyk: Makes sense. Fixed.

PR #2375, opened 2023-10-06
[SW-1348] fix: font rendering issues

martinkaintas:


peronczyk (approved):

src/styles/global.scss
```diff
@@ -9,7 +9,6 @@
 /* Basic CSS for apps built with Ionic */
 @import '@ionic/vue/css/normalize.css';
 @import '@ionic/vue/css/structure.css';
-@import '@ionic/vue/css/typography.css';
```
martinkaintas, line 12: Removed because it is only causing problems and is not required by Ionic

PR #2376, opened 2023-10-06
[SW-1349] fix: svg animatios on ff extension

martinkaintas:


peronczyk (approved):

src/manifest.json
```diff
@@ -2,7 +2,7 @@
   "name": "Superhero",
   "description": "Superhero Wallet",
   "manifest_version": 2,
-  "content_security_policy": "default-src 'self'; script-src 'self' 'sha256-BCCVwoZgDUk/cw+Ln2/t3z9bm8qy+sICaZP5BWQIyUg='; connect-src *; font-src * data:; img-src * data:; style-src-elem * 'unsafe-inline' ",
+  "content_security_policy": "default-src 'self' 'unsafe-inline'; script-src 'self' 'sha256-BCCVwoZgDUk/cw+Ln2/t3z9bm8qy+sICaZP5BWQIyUg='; connect-src *; font-src * data:; img-src * data:; style-src-elem * 'unsafe-inline' ",
```
CedrikNikita, line 5: Is it actually a good idea to add `unsafe inline` as a default for every CSP directive?

martinkaintas: Correct, I will update it to be more precise

PR #2381, opened 2023-10-09
fix: icons on index page

martinkaintas:


subhod-i (approved):

src/popup/components/PlatformIcon.vue
```diff
@@ -34,6 +34,7 @@ export default defineComponent({
 
 .platform-icon {
   height: 44px;
+  width: 40px;
```
peronczyk, line 37: The issue is not caused by wrong sizing but with the fact that the element is affect by flex shrinking. I suggest to replace the change with `flex-shrink: 0`.

peronczyk (changes requested):

src/popup/components/PlatformIcon.vue
martinkaintas: Without `width: 40px;` it still doesn't look exactly like product

peronczyk (approved):

PR #2385, opened 2023-10-11
[SW-1367] feat: lock screen orientation in portrait

martinkaintas:


CedrikNikita:
As we are not supporting the landscape mode should we remove all the landscape related styles and icons/pictures, and reduce the app size?

martinkaintas:
> As we are not supporting the landscape mode should we remove all the landscape related styles and icons/pictures, and reduce the app size?

Sure, though generating the assets with the command will generate the landscape assets again, but they will be ignored by git. 
Also fixed some icons that were off

Liubov-crypto:
LGTM

peronczyk (approved):

PR #2387, opened 2023-10-11
Support Bitcoin transfers using the wallet URLs

subhod-i:


Liubov-crypto:
LGTM, tested in the testnet.

src/lib/RouteQueryActionsController.ts
```diff
@@ -38,8 +44,15 @@ export const RouteQueryActionsController = (() => {
...
      */
-    transferSend: () => {
+    transferSend: (_, query) => {
       const { openModal } = useModals();
+      const { setActiveAccountByProtocol } = useAccounts();
+
+      const token = query[TOKEN_PROP];
+
+      setActiveAccountByProtocol(token === PROTOCOL_BITCOIN
+        ? PROTOCOL_BITCOIN : PROTOCOL_AETERNITY);
```
peronczyk, line 54: This will break if we add new protocols to the app. I suggest to create come kind of map that would connect the token value prop with the protocols. Or even better set the token name in each of the adapters an loop over them here to find protocol.

src/protocols/bitcoin/components/TransferSendForm.vue
```diff
@@ -269,6 +272,19 @@ export default defineComponent({
       { deep: true },
     );
 
+    watch(
+      activeAccount,
```
peronczyk, line 276: Is it possible to change the account when the transfer send modal is open? If not I think it makes no sense to watch activeAccount here.

peronczyk (changes requested):

src/lib/RouteQueryActionsController.ts
subhod-i: Sure.

src/lib/ProtocolAdapterFactory.ts
```diff
@@ -30,8 +30,13 @@ export const ProtocolAdapterFactory = (() => {
     return protocolAdapterRegistry[protocol]!;
   }
 
+  function getAllAdapters(): ProtocolRecord<BaseProtocolAdapter> {
+    return protocolAdapterRegistry;
+  }
```
peronczyk, line 35: I don't think it's reasonable to create this method. If you want to loop the protocols and find something I suggest for example:
```
PROTOCOLS.find((protocol) => {
  const adapter = ProtocolAdapterFactory.getAdapter(protocol);
  return adapter.<UseSomeAdapterMethodHere>() == <ComparedValue>;
});
```

src/lib/RouteQueryActionsController.ts
```diff
@@ -38,8 +45,26 @@ export const RouteQueryActionsController = (() => {
...
+    transferSend: (_, query) => {
       const { openModal } = useModals();
+      const { setActiveAccountByProtocol } = useAccounts();
+
+      /**
+       * Determine the active protocol for the current transfer
+       */
+      const token = query[TOKEN_PROP];
+      const protocol = findKey(ProtocolAdapterFactory.getAllAdapters(),
+        (adapter) => adapter?.getUrlTokenKey() === token) as Protocol;
```
peronczyk, line 57: See my comment here: https://github.com/superhero-com/superhero-wallet/pull/2387/files#r1358123568

src/lib/ProtocolAdapterFactory.ts
subhod-i: You mean to define a method in this file to get one property of each protocol?
For example, `getAllProtocolSymbols` ?

src/lib/RouteQueryActionsController.ts
```diff
@@ -38,8 +45,29 @@ export const RouteQueryActionsController = (() => {
...
      * Take action after opening the link copied in the transfer receive modal.
      */
-    transferSend: () => {
+    transferSend: (_, query) => {
       const { openModal } = useModals();
+      const { setActiveAccountByProtocol } = useAccounts();
+
+      /**
+       * Determine the active protocol for the current transfer
+       */
```
peronczyk, line 54: I suggest to use regular inline comment ( `// ...`) for such a tiny lines. DOCBLOC comments in my opinion are reserved for things like functions, methods and object properties.

peronczyk (approved):

src/lib/ProtocolAdapterFactory.ts
subhod-i: Fixed

PR #2408, opened 2023-10-18
[SW-1374] fix: weird account details page appearance

### LINKED ISSUE #2412: Weird page appearance
https://aeternity-blockchain.atlassian.net/browse/SW-1374

**Describe the bug**

Maybe it's related to internet connection, when I'm switching between networks, but I'm not sure 100%.

It looks as if information from the previous page gets stuck on the current one.

**Expected result**

Steps to reproduce
1. Open account details page ae or btc acc

Environment selector
- iOS

Image

martinkaintas:
closes #2412


src/popup/router/index.ts
```diff
@@ -68,6 +70,17 @@ router.beforeEach(async (to, from, next) => {
...
+  // Bitcoin accounts can only visit default (transaction-list) view in account-details page
+  if (
+    activeAccount.value.protocol === PROTOCOL_BITCOIN
+    && to.name !== ROUTE_ACCOUNT_DETAILS
+  ) {
+    if (to.matched.some((record) => record.path === `/${ROUTE_ACCOUNT_DETAILS}/`)) {
+      next({ name: ROUTE_ACCOUNT_DETAILS });
+      return;
+    }
+  }
```
peronczyk, line 82: I think it's bad to have the protocol related logic in a global file. I can imagine that more protocols we have more bloated this file will get.

src/popup/router/routes.ts
```diff
@@ -146,7 +150,11 @@ export const routes: WalletAppRouteConfig[] = [
           },
           {
             path: 'names',
-            component: AccountDetailsNames,
+            component: ProtocolSpecificView,
+            props: {
+              viewComponentName: PROTOCOL_VIEW_ACCOUNT_DETAILS_NAMES,
+              redirectIfNull: ROUTE_ACCOUNT_DETAILS,
```
peronczyk, line 156: In my opinion this should be passed as the meta instead of prop because the meta can be typed. So in case of future refactoring it will be easier to do it by reading what TS is complaining about.

I know that I've made the initial change that was passing the `viewComponentName` value as a prop. But I consider this as a mistake that should be fixed in near future.

src/popup/components/ProtocolSpecificView.vue
```diff
@@ -58,7 +63,8 @@ export default defineComponent({
     },
   },
   setup(props) {
-    const { params, meta } = useRoute();
+    const { params, meta }: { params: RouteParams, meta: WalletRouteMeta } = useRoute();
```
peronczyk, line 66: What do you think about this approach?
```suggestion
    const route = useRoute();
    const routeMeta = route.meta as WalletRouteMeta;
    const routeParams = route.params;
```

peronczyk (approved):

src/protocols/bitcoin/views/index.ts
```diff
@@ -9,8 +9,10 @@ const protocolViews: ProtocolViewsConfig = {
     /* webpackChunkName: "btc-account-details-transactions" */
     './AccountDetailsTransactions.vue'
   ),
+  AccountDetailsTokens: null,
+  AccountDetailsNames: null,
```
CedrikNikita, line 13: So from now on on adding new blockchains we would need to set `AccountDetailsTokens` and `AccountDetailsNames` to null if it is not supported?

CedrikNikita (approved):

PR #2410, opened 2023-10-18
feat: replace default animations

### LINKED ISSUE #2420: Transition between dashboard and account details looks weird
The dashboard goes to the left and while the transition is happening it is visible that the account cards jumps to the top.

martinkaintas:
closes #2420 
Replace default animations when opening/closing `account-details` and when switching tabs

### DIFF
--- a/src/popup/animations/fade.ts
+++ b/src/popup/animations/fade.ts
@@ -1,0 +1,20 @@
+import { createAnimation } from '@ionic/vue';
+
+const ANIMATION_DURATION = 100;
+
+export const fadeAnimation = (_: Element, opts: { enteringEl: Element, leavingEl: Element }) => {
+  const enteringAnimation = createAnimation()
+    .addElement(opts.enteringEl)
+    .fromTo('opacity', 0, 1)
+    .duration(ANIMATION_DURATION);
+
+  const leavingAnimation = createAnimation()
+    .addElement(opts.leavingEl)
+    .fromTo('opacity', 1, 0);
+
+  const animation = createAnimation()
+    .addAnimation(enteringAnimation)
+    .addAnimation(leavingAnimation);
+
+  return animation;
+};
--- a/src/popup/components/Header.vue
+++ b/src/popup/components/Header.vue
@@ -1,5 +1,8 @@
 <template>
-  <IonHeader class="header ion-no-border">
+  <IonHeader
+    class="header ion-no-border"
+    :class="{ 'hidden': !showHeader }"
+  >
     <IonToolbar class="toolbar">
       <div
         class="wrapper"
@@ -73,10 +76,15 @@ import {
   useBackButton,
   useIonRouter,
 } from '@ionic/vue';
-import { computed, defineComponent } from 'vue';
+import {
+  computed,
+  defineComponent,
+  ref,
+  watch,
+} from 'vue';
 import { useRoute } from 'vue-router';
 import { useI18n } from 'vue-i18n';
-import { IS_MOBILE_APP, UNFINISHED_FEATURES } from '@/constants';
+import { IS_MOBILE_APP, PAGE_TRANSITION_DURATION, UNFINISHED_FEATURES } from '@/constants';
 import type { WalletRouteMeta } from '@/types';
 import {
   ROUTE_ACCOUNT,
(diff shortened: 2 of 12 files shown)


onvisions:
It seems that the issue that existed before is now fixed. Approved on my side to be merged.
@martinkaintas However I'd like to see 2 minor improvements (it can be done later if it requires too much time or efforts or if this will block the release). 
1. I think the Account details modal can have additionally this background blur effect that Send/Receive/Add account etc. modals have applied to. 
2. The time of animation of tabs' content can be just slightly reduced. Let's say if the value is 1 sec let's make it 0.7 (700 miliseconds). We should experiment with the exact value.

martinkaintas:
> It seems that the issue that existed before is now fixed. Approved on my side to be merged. @martinkaintas However I'd like to see 2 minor improvements (it can be done later if it requires too much time or efforts or if this will block the release).
> 
> 1. I think the Account details modal can have additionally this background blur effect that Send/Receive/Add account etc. modals have applied to.
> 2. The time of animation of tabs' content can be just slightly reduced. Let's say if the value is 1 sec let's make it 0.7 (700 miliseconds). We should experiment with the exact value.

I reduced the animation in the tabs from 150ms to 100ms
Regarding 1. I dont see the use for this as account-details covers the whole page, and blur is an expensive animation

src/popup/components/Header.vue
```diff
@@ -187,6 +200,21 @@ export default defineComponent({
...
+          // @ts-ignore
+          headerEl.value?.$el.classList.remove('hidden');
+        } else {
+          // wait for the page transition to finish
+          await new Promise((resolve) => setTimeout(resolve, 150));
+          // @ts-ignore
+          headerEl.value?.$el.classList.add('hidden');
+        }
+      },
+    );
```
peronczyk, line 216: Manipulating classes with vanilla JS is a bad practice within Vue component. Maybe you coudl add another bool ref and change it's value based on the `show` property. And then change the `hidden` class within the template block. What do you think?

src/popup/components/Header.vue
```diff
@@ -215,6 +244,10 @@ export default defineComponent({
   z-index: variables.$z-index-header;
   height: var(--header-height);
 
+  &.hidden {
+    z-index: -1;
+  }
```
peronczyk, line 249: Are you sure using only the `z-index` is OK? Looks rather off in my opinion. Maybe adding opacity/visibility change would be good?

src/popup/animations/fade.ts
peronczyk, line 1: I suggest to add `index` file that would collect all animations and return them so we could import them in a following fashion:
```
import { fadeAnimation, popAnimation } from '@/popup/animations/`;
```

onvisions:
@martinkaintas 
1. Cool for the decrease of time. Now it should be ok.
2. The point of adding blur is to add more smoothness to the animation itself and to make it consistent with our other similar animations for modals thus adding value in terms of UX. Currently there is a moment when you first see the background content "jumping" to the top while the modal is appearing from the bottom. If we add blur it will be visible for just a second but still the transition will look same style as the others. 
Not sure what you mean by "blur is expensive" but f you refer to consuming resources we use it anyway at several places.

src/popup/animations/pop.ts
peronczyk, line 1: I suggest to choose one approach: 1 file = 1 animation or 1 file = multiple animations. Right now this file has different structure than the fade.

src/popup/components/Header.vue
```diff
@@ -187,6 +202,20 @@ export default defineComponent({
...
 
+    watch(
+      () => props.show,
+      async (value) => {
+        if (value) {
+          showHeader.value = true;
+        } else {
+          // wait for the page transition to finish
+          await new Promise((resolve) => setTimeout(resolve, 150));
+          showHeader.value = false;
```
peronczyk, line 213: Maybe just:
```suggestion
          setTimeout(() => {
            showHeader.value = false;
          }, 150));
```

Also maybe save this `150` magic number to some kind of constant to describe what it actually is.

src/popup/components/Header.vue
```diff
@@ -105,14 +114,20 @@ export default defineComponent({
...
+  },
+  setup(props) {
     const route = useRoute();
     const ionRouter = useIonRouter();
     const { t } = useI18n();
 
     const { homeRouteName } = useUi();
     const { isLoggedIn } = useAccounts();
 
+    const headerEl = ref<Element>();
```
peronczyk, line 128: This probably can be removed.

peronczyk (approved):

PR #2455, opened 2023-10-19
Fix: Correct way for popups checking if user is logged in

peronczyk:


src/popup/router/index.ts
```diff
@@ -58,8 +58,14 @@ RouteQueryActionsController.init(router);
 RouteLastUsedRoutes.init(router);
 
 router.beforeEach(async (to, from, next) => {
+  // Move the route resolving to the next thread cycle in order to allow the app
+  // to access the stored data (localStorage/browserStorage) first.
+  await new Promise((resolve) => setTimeout(resolve, 0));
```
martinkaintas, line 63: does it make sense to use `nextTick()` from vue here?

martinkaintas (approved):

src/popup/router/index.ts
CedrikNikita: This looks like a workaround to me, then an actual good decision.

src/composables/accounts.ts
```diff
@@ -26,6 +26,8 @@ import { useStorageRef } from './storageRef';
 
 let isInitialized = false;
 
+const isAccountsRestored = ref(false);
```
peronczyk, line 29: Do you think this name is OK? Or maybe `areAccountsRestored`?

CedrikNikita: `areAccountsRestored` is better.

PR #2469, opened 2023-10-24
fix: bitcoin transactions not loading

martinkaintas:


Liubov-crypto:
Fixed, LGTM

CedrikNikita:
Can we also have a comment why we are using only `didEnter`?  In order to not make this mistake again.

CedrikNikita (approved):

PR #2473, opened 2023-10-25
fix: sort pending multisig transaction correctly

### LINKED ISSUE #2450: Sorting pending tx in multisig is on bottom, should be on top
https://aeternity-blockchain.atlassian.net/browse/SW-1373

Describe the bug

Sorting pending tx in multisig is on bottom, should be on top. Reproducible on the release branch. 

Expected result

Steps to reproduce
1. Initiate multisig tx
2. Check the Tx list 

Image
Image

Environment selector

MacOS
iOS
Web wallet

CedrikNikita:
closes #2450

### DIFF
--- a/src/composables/multisigTransactions.ts
+++ b/src/composables/multisigTransactions.ts
@@ -88,7 +88,7 @@ export function useMultisigTransactions({ store }: IDefaultComposableOptions) {
         hash: txHash,
         tx: rawTx ? unpackTx(rawTx.tx) as any : undefined,
         isMultisigTransaction: true,
-        microTime: rawTx ? new Date(rawTx.createdAt).getTime() : undefined,
+        microTime: rawTx ? Date.now() : undefined,
       };
     }
     return null;


src/composables/multisigTransactions.ts
```diff
@@ -88,7 +88,7 @@ export function useMultisigTransactions({ store }: IDefaultComposableOptions) {
         hash: txHash,
         tx: rawTx ? unpackTx(rawTx.tx) as any : undefined,
         isMultisigTransaction: true,
-        microTime: rawTx ? new Date(rawTx.createdAt).getTime() : undefined,
```
CedrikNikita, line 91: The `createdAt` property actually contains the time when the vault was created, not the proposal.

Liubov-crypto:
LGTM

src/composables/multisigTransactions.ts
martinkaintas: So we are setting the `microTime` to be `Date.now` in order to have the pending multisig transactions on top?

CedrikNikita: Yes, this is what we exactly doing on adding a regular pending transaction.

peronczyk (approved):

martinkaintas (approved):

PR #2474, opened 2023-10-25
fix: redirect if no multisig

### LINKED ISSUE #2477: If account has multisig vaults in testnet, but not on mainnet you would be able to receive such an issue: header in account-details page will disappear
Image

martinkaintas:
closes #2477

### DIFF
--- a/src/composables/multisigAccounts.ts
+++ b/src/composables/multisigAccounts.ts
@@ -28,6 +28,7 @@ import { useAeNetworkSettings } from '@/protocols/aeternity/composables';
 import { createPollingBasedOnMountedComponents } from './composablesHelpers';
 import { useAeSdk } from './aeSdk';
 import { useAccounts } from './accounts';
+import { createNetworkWatcher } from './networks';
 
 const POLLING_INTERVAL = 7000;
 
@@ -36,6 +37,7 @@ const LOCAL_STORAGE_MULTISIG_PENDING_KEY = 'multisig-pending';
 
 export interface MultisigAccountsOptions extends IDefaultComposableOptions {
   pollOnce?: boolean;
+  pollingDisabled?: boolean;
 }
 
 function storeMultisigAccounts(
@@ -62,8 +64,13 @@ const activeMultisigNetworkId = ref('');
 const isAdditionalInfoNeeded = ref(false);
 
 const initPollingWatcher = createPollingBasedOnMountedComponents(POLLING_INTERVAL);
+const { onNetworkChange } = createNetworkWatcher();
 
-export function useMultisigAccounts({ store, pollOnce = false }: MultisigAccountsOptions) {
+export function useMultisigAccounts({
+  store,
+  pollOnce = false,
+  pollingDisabled = false,
+}: MultisigAccountsOptions) {
   const { aeActiveNetworkPredefinedSettings } = useAeNetworkSettings();
   const { nodeNetworkId, getAeSdk } = useAeSdk({ store });
   const { aeAccounts } = useAccounts();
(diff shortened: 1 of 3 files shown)


Liubov-crypto:
The issue with header in account-details has been fixed. But I can see for some reason Tx related to my testnet multisig vault  in the mainnet, however I haven't created a vault in the mainnet. 
@martinkaintas could you check it please? 
err

https://github.com/superhero-com/superhero-wallet/assets/69896204/a45cbc7a-fc5f-49c0-96a7-f29b38746514

Liubov-crypto:
The issue has been fixed, LGTM

src/composables/multisigAccounts.ts
```diff
@@ -61,10 +61,11 @@ const activeMultisigAccountId = ref<Encoded.AccountAddress>();
 const activeMultisigNetworkId = ref('');
 const isAdditionalInfoNeeded = ref(false);
 
+const { aeActiveNetworkPredefinedSettings } = useAeNetworkSettings();
```
peronczyk, line 64: The child composables should be initiated inside of the parent composable - not outside it's scope.

martinkaintas: When being initiated inside of the parent composable it is not being updated correctly

Liubov-crypto:
I have retested and now it's broken again. I can see Txs related to multisig from the testnet in the mainnet.
@martinkaintas please check it:

https://github.com/superhero-com/superhero-wallet/assets/69896204/a1a49de3-77bf-435c-8571-ae8299da971b

src/composables/multisigAccounts.ts
```diff
@@ -309,12 +311,15 @@ export function useMultisigAccounts({ store, pollOnce = false }: MultisigAccount
...
     return allMultisigAccounts.value.find((acc) => acc.contractId === contractId);
   }
+
   if (pollOnce && !getStoredMultisigAccounts(nodeNetworkId.value!).length) {
     updateMultisigAccounts();
   } else if (!pollOnce) {
     initPollingWatcher(() => updateMultisigAccounts());
   }
 
+  onNetworkChange(updateMultisigAccounts);
```
peronczyk, line 321: Little suggestion for future. Generally I think it's a bad pattern to pass the function definition as callback. Why? Consider the fact that the `onNetworkChange` is passing some values to the callback. And in the future someone will update the `updateMultisigAccounts` so it has an boolean argument that decides about something. As the callback value is truthy it will affect the called function in unpredictable way. So in general whenever it's possible I suggest to use arrow functions.

src/popup/App.vue
```diff
@@ -218,6 +222,15 @@ export default defineComponent({
...
     });
 
+    // Redirect to account page if no multisig accounts are present
+    // only if current page is a multisig page
+    watch(multisigAccounts,
+      (value) => {
+        if (!value?.length && route.meta.isMultisig) {
+          router.replace({ name: ROUTE_ACCOUNT });
+        }
+      });
```
peronczyk, line 232: Something is messed up with the indentation here. As always I suggest to take one approach: all arguments in separate lines or in one line. Never mixed.

src/popup/App.vue
```diff
@@ -218,6 +222,15 @@ export default defineComponent({
       }
     });
 
+    // Redirect to account page if no multisig accounts are present
+    // only if current page is a multisig page
+    watch(multisigAccounts,
+      (value) => {
+        if (!value?.length && route.meta.isMultisig) {
```
peronczyk, line 229:
```suggestion
        if (!value?.length && (route.meta as WalletRouteMeta).isMultisig) {
```

It will help with some future refactors in the future.

src/popup/App.vue
```diff
@@ -132,6 +135,7 @@ export default defineComponent({
     const { loadCoinsData } = useCurrencies({ withoutPolling: true });
     const { restoreLanguage } = useLanguages();
     const { restore: restoreTransferSendForm } = useTransferSendHandler();
+    const { multisigAccounts } = useMultisigAccounts({ store });
```
peronczyk, line 138: Running this composable here will force the mutisig accounts to be fetched everywhere in the app. The goal of `createPollingBasedOnMountedComponents` was to poll data only when it's needed. So for example the multisg accounts are not required in regular account details, all settings pages, signing modals etc...
So maybe you can add `disablePolling` setting to the composable?

src/composables/multisigAccounts.ts
```diff
@@ -62,8 +64,13 @@ const activeMultisigNetworkId = ref('');
 const isAdditionalInfoNeeded = ref(false);
 
 const initPollingWatcher = createPollingBasedOnMountedComponents(POLLING_INTERVAL);
+const { onNetworkChange } = createNetworkWatcher();
 
-export function useMultisigAccounts({ store, pollOnce = false }: MultisigAccountsOptions) {
+export function useMultisigAccounts({
+  store,
+  pollOnce = false,
+  disablePolling = false,
```
peronczyk, line 72: I suggest to rename this to `pollingDisabled`.

peronczyk (approved):

Liubov-crypto:
Fixed, LGTM.

PR #2476, opened 2023-10-26
fix: router next function is called twice

martinkaintas:
image


src/lib/RouteLastUsedRoutes.ts
```diff
@@ -27,8 +27,10 @@ export const RouteLastUsedRoutes = (() => {
         if (path && time && dayjs().isBefore(dayjs(time).add(10, 'minutes'))) {
           next(path);
         }
+      } else {
+        next();
```
peronczyk, line 31: This is actually not solving every possibility as there are 2 `if`s. Instead of your change I suggest to make 2 changes:
1. Move the `unbind()` from the end of the function body to the beginning.
2. To prefix all `next` functions with `return` (`return next(...)`).

peronczyk (approved):

PR #2485, opened 2023-11-02
Refactor: Move veeValidate Vuex module to separate plugin

peronczyk:


src/popup/plugins/veeValidate.ts
```diff
@@ -0,0 +1,242 @@
...
+  (value: string) => isAddressValid(value) || isAensNameValid(value) || tg('validation.address'),
+);
+
+defineRule(
+  'account_address',
+  (value: string) => isAddressValid(value, Encoding.AccountAddress) || tg('validation.invalidAddress'),
+);
+
+defineRule(
+  'name',
```
martinkaintas, line 42: I suggest that we rename this to `aensName` since we are adding Ethereum support which has its own ENS names

martinkaintas (approved):

src/popup/plugins/veeValidate.ts
peronczyk: Totally make sense. I update the rule name and few others to group them.

CedrikNikita (approved):
Code changes look reasonable.

PR #2487, opened 2023-11-02
Ethereum transfer receive & send views

martinkaintas:
_fee calculation:_ 
- great article on how to calculate fee, [see](https://www.blocknative.com/blog/eip-1559-fees)
- ethereum docs on calculating fees, [see](https://ethereum.org/en/developers/docs/gas/#block-size)
- `Base Fee` is 100% automatic and readable directly from the network.
- Right now we are setting `Max Priority Fee` hardcoded for all settings (slow, medium, fast) as `Max Priority Fee` is normaly set by the user, so I believe we could also have a `custom fee` field where the user can set his own `Max Priority Fee` for slower or faster transactions
- **`Fee` =  `Base Fee` + `Max Priority Fee`**
- **`Max Fee` = (2 * `Base Fee`) + `Max Priority Fee`**

### DIFF
--- a/src/popup/components/Modals/TransferReceiveBase.vue
+++ b/src/popup/components/Modals/TransferReceiveBase.vue
@@ -41,7 +41,7 @@
           <Scrollable class="address-scrollable-area">
             <AddressFormatted
               :address="accountAddressToDisplay"
-              :split-address="protocol === PROTOCOL_BITCOIN && !amount"
+              :split-address="splitAddress"
             />
           </Scrollable>
         </CopyText>
@@ -111,6 +111,7 @@ import type {
 import {
   IS_MOBILE_DEVICE,
   PROTOCOL_BITCOIN,
+  PROTOCOL_ETHEREUM,
 } from '@/constants';
 import { RouteQueryActionsController } from '@/lib/RouteQueryActionsController';
 import { useAccounts, useCopy } from '@/composables';
@@ -131,6 +132,9 @@ import CopyText from '../CopyText.vue';
 import AccountItem from '../AccountItem.vue';
 
 import ShareIcon from '../../../icons/share.svg?vue-component';
+
+// TODO - validate all addresses instead of 'splitAddress' prop, see AddressFormatted.vue
+const SPLIT_PROTOCOL_ADDRESS = [PROTOCOL_BITCOIN, PROTOCOL_ETHEREUM];
 
 export default defineComponent({
   name: 'TransferReceiveBase',
@@ -180,6 +184,10 @@ export default defineComponent({
         ? RouteQueryActionsController.createUrl('/account', 'transferSend', getTokenInfoQuery(address))
         : '';
     }
+
+    const splitAddress = computed(
+      () => SPLIT_PROTOCOL_ADDRESS.includes(props.protocol) && !amount.value,
+    );
 
     const accountAddressToCopy = computed(
       () => (amount.value && +amount.value > 0)
@@ -239,10 +247,12 @@ export default defineComponent({
 
     return {
       PROTOCOL_BITCOIN,
+      PROTOCOL_ETHEREUM,
       IS_MOBILE_DEVICE,
       ShareIcon,
       amount,
       selectedAsset,
+      splitAddress,
       share,
       handleAssetChange,
       copyAddress,
--- a/src/popup/components/ProtocolIcon.vue
+++ b/src/popup/components/ProtocolIcon.vue
@@ -9,6 +9,7 @@
 
 <script lang="ts">
 import {
+  Component,
   computed,
   defineComponent,
   PropType,
@@ -16,9 +17,12 @@ import {
 import {
   PROTOCOL_AETERNITY,
   PROTOCOL_BITCOIN,
+  PROTOCOL_ETHEREUM,
 } from '@/constants';
+import type { Protocol } from '@/types';
 import AeternityIcon from '@/icons/coin/aeternity.svg?vue-component';
 import BitcoinIcon from '@/icons/coin/bitcoin.svg?vue-component';
+import EthereumIcon from '@/icons/coin/ethereum.svg?vue-component';
 
 const SIZES = ['sm', 'rg'] as const;
 
@@ -27,7 +31,7 @@ export type AllowedProtocolIconSize = typeof SIZES[number];
 export default defineComponent({
   props: {
     protocol: {
-      type: String,
+      type: String as PropType<Protocol>,
       required: true,
     },
     iconSize: {
@@ -37,19 +41,13 @@ export default defineComponent({
     },
   },
   setup(props) {
-    const selectedIcon = computed(() => {
-      switch (props.protocol) {
-        case PROTOCOL_AETERNITY: {
-          return AeternityIcon;
-        }
-        case PROTOCOL_BITCOIN: {
-          return BitcoinIcon;
-        }
-        default: {
-          return null;
-        }
-      }
-    });
+    const iconsMap: Record<Protocol, Component> = {
+      [PROTOCOL_AETERNITY]: AeternityIcon,
+      [PROTOCOL_BITCOIN]: BitcoinIcon,
+      [PROTOCOL_ETHEREUM]: EthereumIcon,
+    };
+
+    const selectedIcon = computed((): Component => iconsMap[props.protocol]);
 
     return {
       selectedIcon,
--- a/src/popup/locales/en.json
+++ b/src/popup/locales/en.json
@@ -337,6 +337,7 @@
       "recipientPlaceholder": "Enter public address or .chain name",
       "recipientPlaceholderUrl": "Enter public address, .chain name or URL",
       "recipientPlaceholderProtocol": "Enter {name} public address",
+      "recipientPlaceholderENS": "or ENS name",
       "transactionSpeed": "Transaction speed",
       "transactionWillBeCompleted": "Transaction will be completed in ≈{time}.",
       "payload": "Payload"
@@ -1195,7 +1196,7 @@
     "nameUnregistered": "This name has already been registered.",
     "notSameAs": "Sender and receiver address are the same. You will send {0} to your own account.",
     "tokenToAnAddress": "Unfortunately you can only send AEX-9 Tokens to an address",
-    "addressBtc": "Invalid BTC address",
+    "addressGeneric": "Invalid {protocol} address",
     "minValue": "This field must be {0} or more",
     "minValueExclusive": "This field must be more than {0}",
     "enoughCoin": "Not enough {0} balance",
--- a/src/protocols/ethereum/components/TransferReview.vue
+++ b/src/protocols/ethereum/components/TransferReview.vue
@@ -1,11 +1,11 @@
 <template>
   <TransferReviewBase
-    :base-token-symbol="BTC_SYMBOL"
+    :base-token-symbol="ETH_SYMBOL"
     :transfer-data="transferData"
     :loading="loading"
-    show-fiat
-    :protocol="PROTOCOL_BITCOIN"
+    :protocol="PROTOCOL_ETHEREUM"
     class="transfer-review"
+    show-fiat
   >
     <template #total>
       <DetailsItem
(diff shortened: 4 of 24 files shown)


src/popup/components/ProtocolIcon.vue
```diff
@@ -45,6 +47,9 @@ export default defineComponent({
         case PROTOCOL_BITCOIN: {
           return BitcoinIcon;
         }
+        case PROTOCOL_ETHEREUM: {
```
peronczyk, line 50: I suggest to update the way how the list of icons are built to enforce adding the icon if new protocol is added. For example:
```
const iconsMap: Record<Protocol, Component> = {
   [PROTOCOL_AETERNITY]: AeternityIcon,
   ...
}

const selectedIcon = computed((): Component | undefined => iconsMap[props.protocol])
```

This way the TS engine will inform us about adding the icon when new protocol is added.

src/popup/locales/en.json
```diff
@@ -337,6 +337,7 @@
       "recipientPlaceholder": "Enter public address or .chain name",
       "recipientPlaceholderUrl": "Enter public address, .chain name or URL",
       "recipientPlaceholderProtocol": "Enter {name} public address",
+      "recipientPlaceholderEth": "Enter Ethereum public address or ENS name",
```
peronczyk, line 340: I strongly suggest to not create any protocol related translations. In my opinion we need to be open for new protocols and make the translations in a generic way.

src/popup/locales/en.json
```diff
@@ -1196,6 +1197,7 @@
     "notSameAs": "Sender and receiver address are the same. You will send {0} to your own account.",
     "tokenToAnAddress": "Unfortunately you can only send AEX-9 Tokens to an address",
     "addressBtc": "Invalid BTC address",
+    "addressEth": "Invalid ETH address",
```
peronczyk, line 1200: I strongly suggest to not create any protocol related translations.
I know that there is already `addressBtc` here, but I think we need to refactor this and pass the coin as property.

src/protocols/ethereum/components/TransferReview.vue
```diff
@@ -0,0 +1,136 @@
...
+    <template #total>
+      <DetailsItem
+        :label="$t('common.total')"
+        class="details-item"
+      >
+        <template #value>
+          <TokenAmount
+            :amount="+transferData.total"
+            :symbol="ETH_SYMBOL"
+            high-precision
```
peronczyk, line 19: Please put the bool props at the end of the prop list.

src/protocols/ethereum/components/TransferReview.vue
```diff
@@ -0,0 +1,136 @@
...
+import { useRouter } from 'vue-router';
+import { useI18n } from 'vue-i18n';
+import { useModals, useUi } from '@/composables';
+import type { TransferFormModel } from '@/types';
+import { PROTOCOL_ETHEREUM } from '@/constants';
+
+import TransferReviewBase from '@/popup/components/TransferSend/TransferReviewBase.vue';
+import DetailsItem from '@/popup/components/DetailsItem.vue';
+import TokenAmount from '@/popup/components/TokenAmount.vue';
+import { ETH_SYMBOL } from '@/protocols/ethereum/config';
```
peronczyk, line 44: Please move this line before the component imports.

src/protocols/ethereum/components/TransferReview.vue
```diff
@@ -0,0 +1,136 @@
...
+      });
+    }
+
+    // TODO: replace that
+    // This is happening because of inner TemplateRenderer issue on parsing the blocksteam response
+    function HtmlEncode(content: string) {
+      const textAreaDiv = document.createElement('textarea');
+      textAreaDiv.textContent = content;
+      return textAreaDiv.innerHTML;
+    }
```
peronczyk, line 81: The same function is used in Bitcoin TransferReview component. I suggest to move it to utilities file. Also I suggest to add a comment that describes what this function actually do.
@CedrikNikita maybe you can suggest something?

src/protocols/ethereum/components/TransferSendForm.vue
```diff
@@ -0,0 +1,314 @@
...
+          >
+            {{ $t('common.max') }}
+          </BtnPlain>
+        </template>
+      </TransferSendAmount>
+    </template>
+
+    <template
+      #extra
+    >
```
peronczyk, line 51: In my opinion there is no need to split this line.

src/protocols/ethereum/components/TransferSendForm.vue
```diff
@@ -0,0 +1,314 @@
...
+        }"
+        @asset-selected="handleAssetChange"
+      >
+        <template #label-after>
+          <BtnPlain
+            class="max-button"
+            :class="{ chosen: formModel?.amount?.toString() === max.toString() }"
+            @click="setMaxAmount"
+          >
+            {{ $t('common.max') }}
```
peronczyk, line 43: I suggest to pass the text labels as the `:text` property instead of slot.

src/protocols/ethereum/components/TransferSendForm.vue
```diff
@@ -0,0 +1,314 @@
...
+        </template>
+      </TransferSendAmount>
+    </template>
+
+    <template
+      #extra
+    >
+      <div
+        v-show="activeNetwork.type !== NETWORK_TYPE_TESTNET"
+      >
```
peronczyk, line 54: This element does nothing. I suggest to put the `v-show` on the `DetailsItem` element.

src/protocols/ethereum/components/TransferSendForm.vue
```diff
@@ -0,0 +1,314 @@
...
+      submit,
+      setMaxAmount,
+    };
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@use '../../../styles/variables';
+@use '../../../styles/typography';
```
peronczyk, line 291:
```suggestion
@use '@/styles/variables';
@use '@/styles/typography';
```

src/protocols/ethereum/components/TransferSendForm.vue
```diff
@@ -0,0 +1,314 @@
...
+
+.transfer-send-form {
+  .max-button {
```
peronczyk, line 294: This is the 4th place in the app that the same styles are used. I suggest to create separate button component that would use the `BtnPlain` as it's base.

src/types/index.ts
```diff
@@ -145,8 +146,9 @@ export interface ITokenBalanceResponse {
  */
 export interface IToken {
   contractId: Encoded.ContractAddress
-    | typeof AE_CONTRACT_ID
-    | typeof BTC_CONTRACT_ID
+  | typeof AE_CONTRACT_ID
+  | typeof BTC_CONTRACT_ID
+  | typeof ETH_CONTRACT_ID;
```
peronczyk, line 151: Wrong indentation.

src/protocols/ethereum/helpers/index.ts
```diff
@@ -0,0 +1,26 @@
...
+export function isEthAddressValid(address: string) {
+  return isAddress(address);
+}
+
+/**
+ * Return the base fee of the latest block in Ether
+ */
+export async function getEthBaseFee() {
+  // TODO: use our own node
+  const web3Eth = new Web3Eth('https://eth.llamarpc.com');
```
peronczyk, line 15: The node URL should be set in the network settings in my opinion.

src/protocols/ethereum/helpers/index.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  const latestBlock = await web3Eth.getBlock('latest', false);
+  const baseFee = latestBlock.baseFeePerGas;
+  return fromWei(baseFee!, 'ether');
+}
+/**
```
peronczyk, line 21: Please add empty line before this one.

src/protocols/ethereum/views/TransferReceiveModal.vue
```diff
@@ -0,0 +1,37 @@
+<template>
+  <TransferReceiveBase
+    v-bind="$attrs"
```
peronczyk, line 3: I strongly suggest to avoid passing all attrs to child components. During my 1,5 year of work here I know that this approach was causing so many issues :(
It's unpredictable and hard to maintain.

src/protocols/ethereum/views/TransferSendModal.vue
```diff
@@ -0,0 +1,133 @@
+<!-- eslint-disable vue/no-v-model-argument -->
+<template>
+  <TransferSendBase
+    v-bind="$attrs"
```
peronczyk, line 4: I strongly suggest to avoid passing all attrs to child components. During my 1,5 year of work here I know that this approach was causing so many issues :(
It's unpredictable and hard to maintain.

src/protocols/ethereum/views/TransferSendModal.vue
```diff
@@ -0,0 +1,133 @@
...
+  ResolveCallback,
+  TransferFormModel,
+  TransferSendStep,
+} from '@/types';
+import {
+  useBalances,
+  useCurrencies,
+} from '@/composables';
+
+import TransferSendBase from '@/popup/components/Modals/TransferSendBase.vue';
```
peronczyk, line 43: Please group the component imports at the end of the imports list.

src/protocols/ethereum/views/TransferSendModal.vue
```diff
@@ -0,0 +1,133 @@
...
+import {
+  PROTOCOL_ETHEREUM,
+  PROTOCOL_VIEW_TRANSFER_SEND,
+  TRANSFER_SEND_STEPS,
+} from '@/constants';
```
peronczyk, line 48: I suggest to move the constants right after the types import. In general I think the order of the imports should go from less specific to more specific:
```
1. External deps
2. Types
3. Constants
4. Utils
5. Libs
6. Composables
7. Protocol related stuff
8. Components
9. Protocol related components
```

peronczyk (changes requested):

src/popup/locales/en.json
martinkaintas: With the current design reference we have to create atleast one protocol specific translation for "or ENS name"

src/protocols/ethereum/components/TransferReview.vue
martinkaintas: This can be removed from the ethereum component as we won't use blockstream

src/protocols/ethereum/helpers/index.ts
martinkaintas: Yes, once the adapter is completed we can change this dummy code

src/popup/components/Modals/TransferReceiveBase.vue
```diff
@@ -132,6 +133,8 @@ import AccountItem from '../AccountItem.vue';
 
 import ShareIcon from '../../../icons/share.svg?vue-component';
 
+const SPLIT_PROTOCOL_ADDRESS = [PROTOCOL_BITCOIN, PROTOCOL_ETHEREUM];
```
peronczyk, line 136: I suggest to add some comment on why those protocols have their addresses split.

src/popup/components/ProtocolIcon.vue
```diff
@@ -9,16 +9,20 @@
...
   computed,
   defineComponent,
   PropType,
 } from 'vue';
 import {
   PROTOCOL_AETERNITY,
   PROTOCOL_BITCOIN,
+  PROTOCOL_ETHEREUM,
 } from '@/constants';
+import { Protocol } from '@/types';
```
peronczyk, line 22:
```suggestion
import type { Protocol } from '@/types';
```

src/protocols/ethereum/helpers/index.ts
peronczyk: Yeah, but then the architecture and the source of this url will be different and I'm afraid that having just an utility function would not be possible - we would need to refactor it havilly. This is why I'm suggesting to mock this url in the place where it will be stored in the future to have the architecture done in a proper way now.

src/protocols/ethereum/composables/ethBaseFee.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  /**
+   * Return the base fee of the latest block in Ether
+   */
+  async function getBaseFee() {
+    const { nodeUrl } = ethActiveNetworkSettings.value;
+    const web3Eth = new Web3Eth(nodeUrl);
+    const latestBlock = await web3Eth.getBlock('latest', false);
+    const baseFee = latestBlock.baseFeePerGas;
+    return fromWei(baseFee!, 'ether');
```
peronczyk, line 16: The `baseFee` constant is used only once so I think it could be passed directly to `fromWei` function.

```suggestion
    return fromWei(latestBlock.baseFeePerGas, 'ether');
```

peronczyk (approved):

PR #2506, opened 2023-11-09
Feature/eth adapter

martinkaintas:
- [x] Fetch balance
- [x] Create account
- [x] Spend from account 
- [x] Recover account


src/types/index.ts
```diff
@@ -110,9 +110,9 @@ export interface IWallet {
 }
 
 export interface IHdWalletAccount {
-  publicKey: string;
-  secretKey: string;
-  address: string,
+  publicKey: string | Uint8Array;
+  secretKey: string | Uint8Array;
```
peronczyk, line 114: I'm not sure if this is correct. We have two kinds of objects that holds the `publicKey` and `secretKey`. This is one of them and it's related to the data that is actually stored in the `LocalStorage`. So it can't be anything different than `string`.

The `Uint8Array` values are assigned to `IKeyPar`. I suggest to not mix this here.

src/store/plugins/veeValidate.js
```diff
@@ -54,6 +55,7 @@ configure({
       max_redeem: ({ rule }) => tg('validation.maxRedeem', [rule.params[0]]),
       max_value_vault: ({ rule }) => tg('validation.maxValueVault', [rule.params[0]]),
       max_len: ({ rule }) => tg('validation.maxLength', [rule.params[0]]),
+      hex: () => tg('validation.hexValue'),
```
peronczyk, line 58: Can you please return the translated message directly in the `defineRule` scope? Please see the new structure of `veeValidate` file that will be soon merged: https://github.com/superhero-com/superhero-wallet/pull/2485/files#diff-012d1e7b17f11659baed376635edd5f2a3482db91cba378b29c82839f37e52fb
The idea is to have the rule and the message in one place.

src/store/plugins/veeValidate.js
```diff
@@ -31,6 +31,7 @@ defineRule('max_value', (value, [arg]) => value && BigNumber(value).isLessThanOr
 defineRule('max_redeem', (value, [arg]) => value && BigNumber(value).isLessThanOrEqualTo(arg));
 defineRule('max_value_vault', (value, [arg]) => BigNumber(value).isLessThanOrEqualTo(arg));
 defineRule('max_len', (value, [maxLength]) => value && value.length <= maxLength);
+defineRule('hex', (value) => value && value.toString().startsWith('0x') && value.length >= 3);
```
peronczyk, line 34:
```
(value) => value && ...
```
If you write the rule this way it makes the field required by default. Actually I suggest to allow to use this rule for not required field:
```
(value) => !value || ...
```

Also I don't think this logic is actually checking if the value is hexadecimal. I suggest to change the logic or the rule name.

src/types/index.ts
```diff
@@ -688,6 +688,8 @@ export type TransferSendStepExtended = TransferSendStep | typeof TRANSFER_SEND_S
 
 export interface TransferFormModel extends IFormModel {
   fee?: BigNumber
+  maxPriorityFeePerGas?: string
+  maxFeePerGas?: string
```
peronczyk, line 692:
```suggestion
  maxPriorityFeePerGas?: string;
  maxFeePerGas?: string;
```

:)

src/types/networks.ts
```diff
@@ -50,6 +50,10 @@ export interface IAdapterNetworkSetting<T = string> {
    */
   testId?: string;
   required?: boolean;
+  /**
+   * Can be used to override or add field validation rules
+   */
+  validationRules?: object;
```
peronczyk, line 56: I rather suggest to avoid setting `object` as a type because actually most of stuff in JS is an object :) In most cases `Dictionary` is the best fit in that cases. Or some kind of `Record`.

peronczyk (changes requested):

src/types/index.ts
martinkaintas: The Bitcoin adapter was casting `publicKey` & `sercetKey` `as any`. So they are saved as Buffers. 
If that is the case with `LocalStorage` then perhaps we should save them all as `strings` and not cast them

peronczyk: So maybe the adapters should convert the keys to string as it will eventually happen when saving to/from the `LocalStorage`?

src/types/networks.ts
```diff
@@ -50,6 +50,10 @@ export interface IAdapterNetworkSetting<T = string> {
    */
   testId?: string;
   required?: boolean;
+  /**
+   * Can be used to override or add field validation rules
```
peronczyk, line 54: I suggest to add a comment that the list of possible keys are available in the `veeValidate` rule definition file.

src/protocols/ethereum/types.ts
```diff
@@ -0,0 +1,8 @@
+import { INetworkProtocolSettings } from '@/types';
+
+/**
+ * Settings specific to this protocol.
+ */
+export type EthNetworkProtocolSettings = 'chainId'
```
peronczyk, line 6:
```suggestion
export type EthNetworkProtocolSettings = 'chainId';
```

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -154,7 +191,44 @@ export class EthereumAdapter extends BaseProtocolAdapter {
...
 
-  override spend() {
-    return {} as any; // TODO
+  override async spend(
+    amount: number,
+    recipient: string,
+    options: {
+      fromAccount: IAccount,
+      maxPriorityFeePerGas: string,
+      maxFeePerGas: string,
```
peronczyk, line 200:
```suggestion
      fromAccount: IAccount;
      maxPriorityFeePerGas: string;
      maxFeePerGas: string;
```

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -138,8 +139,9 @@ export class EthereumAdapter extends BaseProtocolAdapter {
     return isAddress(address);
   }
 
-  override async isAccountUsed(): Promise<boolean> {
-    return false; // TODO
+  override async isAccountUsed(address: string): Promise<boolean> {
+    const balance = parseFloat(await this.fetchBalance(address));
```
CedrikNikita, line 143: We are recovering the accounts that was in used as well (with balance 0, but with history).

CedrikNikita (changes requested):

peronczyk (approved):

CedrikNikita (commented):
Recover option is taking around 30 seconds.

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -112,24 +135,51 @@ export class EthereumAdapter extends BaseProtocolAdapter {
...
   }
 
-  override getHdWalletAccountFromMnemonicSeed(): IHdWalletAccount {
-    return {} as any; // TODO
+  override getHdWalletAccountFromMnemonicSeed(
+    seed: Uint8Array,
+    accountIndex: number,
+  ): IHdWalletAccount {
+    const hdNodeWallet = this.bip32.fromSeed(Buffer.from(seed));
+    const path = `m/44'/60'/0'/0/${accountIndex}`;
```
CedrikNikita, line 165: This derivation path is incorrect, you are deriving an address, and not a new account.

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -112,24 +135,51 @@ export class EthereumAdapter extends BaseProtocolAdapter {
...
-  override async isAccountUsed() {
-    return false; // TODO
+  override async isAccountUsed(address: string): Promise<boolean> {
+    const balance = parseFloat(await this.fetchBalance(address));
+    if (balance > 0) {
+      return true;
+    }
+    // in case of 0 balance, check if account has transaction history
+    const txCount = await this.getTransactionCount(address);
+    return txCount > 0;
```
CedrikNikita, line 157: We can use one request with all the needed info (balance and tx count in one place).
https://www.quicknode.com/docs/ethereum/eth_getAccount

CedrikNikita (changes requested):

src/protocols/ethereum/libs/EthereumAdapter.ts
martinkaintas: eth_getAccount is not part of the standard interface for Ethereum and many nodes don't have this function
https://ethereum.github.io/execution-apis/api-documentation/

But we can make both requests at the same time to save time

martinkaintas: I also changed the node which we are using for the mainnet, and recovery now only takes ~6 seconds (for all protocols)

CedrikNikita:
Clicking on ethereum address in sendModal will not lead to an explorer, instead will lead to `404` page.

martinkaintas:
> Clicking on ethereum address in sendModal will not lead to an explorer, instead will lead to `404` page.

Yes, explorers are not part of this PR, we agreed to split the adapter issue into more PRs

PR #2507, opened 2023-11-10
Refactor: Finalise replacing Vuex with composables

peronczyk:


src/composables/aeSdk.ts
```diff
@@ -204,6 +204,22 @@ export function useAeSdk() {
     );
   }
 
+  async function resetConnectedDapps() {
```
peronczyk, line 207: This code was copied 1:1 from the `wallet.js` file without changes. But I have a feeling that something is wrong here. The `_clients` property is typed on the SDK side as `Map<string, RpcClientsInfo>`. So the `aepp` is the `RpcClientsInfo` and this interface has no `info` property.

@davidyuk @CedrikNikita

davidyuk: `info` was removed in 12.0.0 because it is not needed on sdk side

> [RpcClient: doesn't contain aepp info anymore](https://github.com/aeternity/aepp-sdk-js/tree/7126ad02166747a75c8462b13386774cfc003124/docs/guides/migration/12.0.0.md#rpcclient-doesnt-contain-aepp-info-anymore)
Get aepp info in onConnection callback, and store somehow to use later.

https://github.com/aeternity/aepp-sdk-js/commit/010ebbb963ac8058ed74bff1fd01ad2e243eb3ab

Soooo, this code probably is not working

src/composables/aeSdk.ts
```diff
@@ -204,6 +210,17 @@ export function useAeSdk() {
...
   }
 
+  async function disconnectDapps() {
+    if (IN_FRAME) {
+      const aeSdkLocal = await getAeSdk();
+      aeSdkLocal._clients.forEach((aepp) => {
+        if (aepp.rpc.connection.isConnected()) {
+          aepp.rpc.connection.disconnect();
+        }
+      });
```
peronczyk, line 220: @davidyuk do you think this is a good way of disconnecting dapps?

davidyuk: The way it is currently, it would just unbind event listeners without notifying the other peer
https://github.com/aeternity/aepp-sdk-js/blob/315718fbc34ab92df8b9fc04bf186d00f2af7151/src/aepp-wallet-communication/connection/BrowserWindowMessage.ts#L88-L97

```suggestion
      aeSdkLocal._clients.forEach((aepp, aeppId) => {
        // if aepp didn't disconnected before by itself
        if (aepp.status === RPC_STATUS.CONNECTED) {
          // notify aepp that wallet disconnects
          aepp.rpc.notify(METHODS.closeConnection, null);
        }
        // internally it is closes connection and removes aepp from `_clients`
        this.aeSdk.removeRpcClient(aeppId);
      });
```
Based on https://github.com/aeternity/aepp-sdk-js/blob/315718fbc34ab92df8b9fc04bf186d00f2af7151/examples/browser/wallet-iframe/src/App.vue#L117-L126
Also, ensure that is works the same as https://docs.aeternity.com/aepp-sdk-js/v13.2.2/examples/browser/wallet-iframe/, compatible with https://docs.aeternity.com/aepp-sdk-js/v13.2.2/examples/browser/aepp/

peronczyk: Thank you very much. I updated the code with you suggestions.

dvdbak:
LTGM

PR #2510, opened 2023-11-10
Refactor: Enforce consistent interface member delimiter style

peronczyk:
Semicolons required for the line ending in the TS interfaces.


martinkaintas (approved):
`src/types/filter.ts` is missing some semicolons

PR #2515, opened 2023-11-13
Refactor: Remove global AE Middleware initialization

peronczyk:


src/composables/balances.ts
```diff
@@ -80,12 +82,16 @@ export function useBalances() {
...
 
-  onNetworkChange(() => {
-    updateBalances();
-  });
-
   initPollingWatcher(() => updateBalances());
 
+  if (!initialized) {
+    onNetworkChange(() => {
+      console.log('onNetworkChange BALA');
```
CedrikNikita, line 89:
```suggestion
```

src/composables/balances.ts
```diff
@@ -80,12 +82,15 @@ export function useBalances() {
     );
   }
 
-  onNetworkChange(() => {
-    updateBalances();
-  });
-
   initPollingWatcher(() => updateBalances());
 
+  if (!initialized) {
```
CedrikNikita, line 87: Were we initiating several `updateBalances` on `onNetworkChange` if the balances we used in several places on particular page previously?

peronczyk: No, because previous logic was working in a different way. Every time the composable was fired the logic behind `onNetworkChange` was comparing the previously stored network with the actual and if the values was different the callback was fired. New logic is based on hooks - you add callback to registry and the networks composable is responsible for firing the callbacks.

Thanks to this the composable is actually reacting to network change when it happens instead only when the composable is requested by component.

New logic requires to add the hook callback only once - this is why we needed to have the `initialized` variable here.

src/utils/common.ts
```diff
@@ -110,6 +110,26 @@ export function errorHasValidationKey(error: any, expectedKey: string): boolean
   return error.validation?.some(({ key }: any) => expectedKey === key);
 }
 
+/**
+ * Create list of callbacks
+ */
+export function createCallbackStorage<T extends(...args: any) => any>() {
```
martinkaintas, line 116: I think the name is a bit misleading, I would suggest `createCallbackList` or `createCallbackRegistry` or `createCallbackArray`

peronczyk: Totally agree, I overused the word "storage". Changed to `Registry`.

CedrikNikita (approved):
Code changes look reasonable.

martinkaintas (approved):

peronczyk:
I'm closing this PR as there is another one that uses this code to fix one more issue: https://github.com/superhero-com/superhero-wallet/pull/2598

PR #2522, opened 2023-11-15
feat: eth send max amount fee calculation

martinkaintas:


martinkaintas:
Closed in favor of #2901

PR #2523, opened 2023-11-15
feat: fetch transactions when account balance changes

martinkaintas:


src/composables/transactionList.ts
```diff
@@ -274,22 +278,25 @@ export function useTransactionList() {
...
-    await Promise.all(accounts.value.map((account) => fetchTransactions(account.address, true)));
-  }
-
   /**
    * To avoid unnecessary data transfers instead of constant polling
-   * we are fetching the transactions only if the total balance of the accounts changes.
+   * we are fetching the transactions only if the balance of an account changes.
    */
   watch(
-    accountsTotalBalance,
```
martinkaintas, line 286: `accountsTotalBalance` doesn't always change when we do small transactions or if the transaction is between accounts in the same wallet (fees are usually not enough to change the total value of the wallet)
Also, this way we are only fetching transactions for the accounts that need to be updated instead of all accounts

peronczyk (approved):

Liubov-crypto:
LGTM

PR #2525, opened 2023-11-15
feat: eth dont wait for tx confirmation when sending

martinkaintas:
Do not wait for response when sending transaction
Instead calculate hash beforehand


Liubov-crypto:
Reviewing and sending a transaction is faster than on the ETH feature branch. Sent transactions appear with some delay (1-2 min) in the wallet than on the ETH branch. I have to refresh page to see them:

https://github.com/superhero-com/superhero-wallet/assets/69896204/e5d09a54-858d-4bfd-a1d3-9bcc4b384c61

In general LGTM.

CedrikNikita (approved):

PR #2527, opened 2023-11-15
fix: eth send very small amounts

martinkaintas:


src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -242,7 +242,7 @@ export class EthereumAdapter extends BaseProtocolAdapter {
     const web3Eth = new Web3Eth(nodeUrl);
     const nonce = await this.getTransactionCount(options.fromAccount.address!);
 
-    const hexAmount = bigIntToHex(BigInt(toWei(amount, 'ether')));
+    const hexAmount = bigIntToHex(BigInt(toWei(amount.toFixed(ETH_COIN_PRECISION), 'ether')));
```
CedrikNikita, line 245: Is `toWei` unable to handle E notation `1e-18`?

martinkaintas: Yes, it throws an error that it cannot convert it

CedrikNikita (approved):
Code changes look reasonable.

Liubov-crypto:
I can send 1e-20 ETH from my wallet. Transaction was successful: https://sepolia.etherscan.io/tx/0x219bcfe5bcc7f5e88b160152977fd031bc1892ef0f4d51edee4e6d497d47d367 
The wallet is extremely slow for me and I can see many errors in console: 
errrr

Also when I'm opening Send ETH modal I'm getting this error:

PR #2533, opened 2023-11-17
refactor: remove protocol specific logic from tokens component

### LINKED ISSUE #2517: Refactor Tokens.vue component to remove/simplify protocol specific logic

martinkaintas:
closes #2517


src/protocols/BaseProtocolAdapter.ts
```diff
@@ -25,7 +27,10 @@ export abstract class BaseProtocolAdapter {
 
   abstract getAmountPrecision(args?: any): number;
 
-  abstract getCoinSymbol(getShort?: boolean): string;
+  /**
+   * selector='protocol' by default
+   */
+  abstract getCoinSymbol(selector?: ProtocolSymbolTypes): string;
```
peronczyk, line 33: Maybe we could have 3 different methods for each symbol type? Or even store it as a class property instead of getter as this is a thing that doesn't change.

src/protocols/BaseProtocolAdapter.ts
```diff
@@ -19,14 +20,18 @@ export abstract class BaseProtocolAdapter {
 
   abstract protocolName: string;
 
+  abstract coinSymbol: string;
+
+  abstract coinSymbolShort: string;
```
peronczyk, line 25:
```suggestion
  abstract coinName: string;

  abstract coinSymbol: string;
```

Maybe just coin name and coin symbol?

src/composables/fungibleTokens.ts
```diff
@@ -50,6 +50,10 @@ const { onNetworkChange } = createNetworkWatcher();
 const availableTokensPooling = createPollingBasedOnMountedComponents(60000);
 const tokenBalancesPooling = createPollingBasedOnMountedComponents(10000);
 
+/**
+ * AE network AEX-9 fungible tokens
+ * TODO With the addition of ERC-20 tokens we should rename this to useAex9Tokens
```
peronczyk, line 55: I don't full;y understand this comment. I think this composable should take care of all tokens - no matter if ERC-20 or Aex-9.

src/popup/components/ProtocolIcon.vue
```diff
@@ -17,12 +17,14 @@ import {
 import { PROTOCOLS } from '@/constants';
 import type { Protocol } from '@/types';
 import AeternityIcon from '@/icons/coin/aeternity.svg?vue-component';
+import AeternityIconSecondary from '@/icons/coin/aeternity-secondary.svg?vue-component';
```
peronczyk, line 20: Now we have 2 icons. Maybe one is used for the blockchain and one for the coin? If this is true I suggest to rename the files accordingly.

src/popup/components/ProtocolIcon.vue
```diff
@@ -35,10 +37,21 @@ export default defineComponent({
       default: 'rg',
       validator: (val: AllowedProtocolIconSize) => SIZES.includes(val),
     },
+    /**
+     * Not all icons have secondary version
+     * Secondary version is used for example when displaying token icon
```
peronczyk, line 42: You mean "not all **protocols**..."?

src/popup/components/ProtocolIcon.vue
```diff
@@ -35,10 +37,21 @@ export default defineComponent({
...
+     */
+    iconVariant: {
+      type: String as PropType<AllowedProtocolIconVariations>,
+      default: 'default',
+    },
```
peronczyk, line 47: Maybe instead use bool prop so the default variant would be the "coin" logo and if the bool prop would be passed the protocol "logo" would be used.

src/popup/components/Tokens.vue
```diff
@@ -108,29 +105,22 @@ export default defineComponent({
...
-        name = props.fullSymbol ? ETH_COIN_NAME : ETH_SYMBOL;
+      if (isTokenCoin) {
+        name = props.fullSymbol ? adapter.coinName : adapter.protocolSymbol;
       }
 
       return {
         ...token,
         name,
-        image,
+        isCoin: isTokenCoin,
```
peronczyk, line 123: Why not just use `isTokenCoin` instead of mapping it to different name?

src/popup/components/TransactionListItem.vue
```diff
@@ -108,10 +109,13 @@ export default defineComponent({
     let timerInterval: NodeJS.Timer;
     const transactionDate = ref();
 
-    const currentTransaction = computed(
-      () => (props.multisigTransaction || props.transaction),
+    const currentTransaction = computed<ITransaction>(
```
peronczyk, line 112: I strongly suggest to type the callback return type instead of just passing the type as the generic argument.

src/popup/components/TransactionListItem.vue
```diff
@@ -108,10 +109,13 @@ export default defineComponent({
...
+      () => (props.multisigTransaction as ITransaction || props.transaction),
     );
 
+    // temp if protocol undefined assume it is aeternity
+    const transactionProtocol = computed(() => props.transaction?.protocol ?? PROTOCOLS.aeternity);
```
peronczyk, line 117: Maybe instead of this just hide the elements if the `props.transaction` is falsy with the `v-if`?

src/utils/common.ts
```diff
@@ -421,3 +423,11 @@ export function checkIfSuperheroCallbackUrl(query: LocationQuery) {
...
   );
 }
+
+export function isCoin(contractId: IToken['contractId']): boolean {
+  return PROTOCOL_LIST.some(
+    (protocol) => ProtocolAdapterFactory
+      .getAdapter(protocol)
+      .getDefaultAssetContractId() === contractId,
+  );
+}
```
peronczyk, line 433: Maybe the `getDefaultCoin` method could return the flag that would determine if it's coin?

src/types/index.ts
```diff
@@ -170,6 +170,7 @@ export interface ITokenResolved extends Partial<IToken> {
   isAe?: boolean;
   isPool?: boolean;
   isReceived?: boolean;
+  isCoin?: boolean;
```
peronczyk, line 173: Maybe the `IAsset` could have a property `assetType` with possible values `'token' | 'coin'`?

src/popup/components/TransactionListItem.vue
martinkaintas: This is supposed to only check for `props.trsansaction.protocol` and uses aeternity if it is missing.
All transactions have a protocol assigned to them, but sometimes aeternity transactions do not have a protocol value

src/utils/common.ts
martinkaintas: That would mean calling the `getDefaultCoin` method with an extra `contractId` argument
We would also be calling `getDefaultCoin` without wanting to get the default coin. I think it makes sense to keep this functionality separate

src/types/index.ts
```diff
@@ -170,6 +170,7 @@ export interface ITokenResolved extends Partial<IToken> {
   isAe?: boolean;
   isPool?: boolean;
   isReceived?: boolean;
+  assetType?: 'coin' | 'token';
```
peronczyk, line 173: I suggest to create constants for those 2 values.

PR #2539, opened 2023-11-20
feat: do not treat coins as tokens

### LINKED ISSUE #2529: Refactor coins to not be treated as AE tokens

martinkaintas:
based on #2533 
closes #2529


src/composables/fungibleTokens.ts
```diff
@@ -49,7 +48,10 @@ const tokenBalances = useStorageRef<Record<string, IToken[]>>(
 const { onNetworkChange } = createNetworkWatcher();
 const availableTokensPooling = createPollingBasedOnMountedComponents(60000);
 const tokenBalancesPooling = createPollingBasedOnMountedComponents(10000);
-
+/**
```
peronczyk, line 51: Please add empty line before the comments.

src/composables/transactionTokens.ts
```diff
@@ -63,24 +63,26 @@ export function useTransactionTokens({
       }));
     }
 
+    const isReceived = direction === TX_DIRECTION.received;
+    const protocolAdapter = ProtocolAdapterFactory
```
peronczyk, line 67: I suggest to rename this to `adapter` as we use only one type of adapters within the system and the `adapter` constant is already more common.

src/protocols/bitcoin/views/TransactionDetails.vue
```diff
@@ -72,21 +72,28 @@ export default defineComponent({
 
     const transactionOwner = route.params.transactionOwner as string;
 
+    const bitcoinAdapter = ProtocolAdapterFactory.getAdapter(PROTOCOL_BITCOIN);
```
peronczyk, line 75: I suggest to rename this to `adapter` as we use only one type of adapters within the system and the `adapter` constant is already more common.

src/protocols/bitcoin/views/TransactionDetails.vue
```diff
@@ -72,21 +72,28 @@ export default defineComponent({
...
 
     const transactionFee = computed((): number => transaction.value?.tx?.fee ?? 0);
 
     const totalAmount = computed((): number => transaction.value?.tx
-      ? getTxAmountTotal(transaction.value, transaction.value.tx.senderId !== transactionOwner)
+      ? bitcoinAdapter
+        .getTxAmountTotal(transaction.value, transaction.value.tx.senderId !== transactionOwner)
       : 0);
 
+    const amount = computed(
```
peronczyk, line 90: This is a thing that I already changed within another PR. For protocols different than the AE the amount should not be calculated again and again. The `transaction.tx.amount` is the real amount.

src/types/index.ts
```diff
@@ -141,6 +141,7 @@ export interface ITokenBalanceResponse {
 
 /**
  * Fungible tokens that are available in currently used network.
+ * Including the native coin.
```
peronczyk, line 144: This is actually not true... kind of. Because we have the `ICoin` which initially was describing the coins. But almost since I joined the team I really didn't like this as the `ICoin` is the `IToken` with coin gecko data :/

 I'm in an constant process of replacing the `ICoin` and `IToken` with only one type `IAsset` which for now is equal to `ICoin | IToken`. In the future there will be only `IAsset`, and when this happens the line you added will make sense. For now it actually messes the concept of the difference between `ICoin` and `IToken`.

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -254,4 +256,13 @@ export class EthereumAdapter extends BaseProtocolAdapter {
...
   }
+
+  override getTxAmountTotal(
+    transaction: ITransaction,
+    isReceived: boolean = false,
+  ) {
+    return new BigNumber(transaction.tx?.amount || 0)
+      .plus(isReceived ? 0 : transaction.tx?.fee || 0)
+      .toNumber();
+  }
```
peronczyk, line 267: I don't think every protocol should have it's own way of calculating the total amount. We need to have one global way of doing this based on the fact that the `ITransaction` has the same values for every protocol. So for example the amount stored within the `ITransaction` is the real amount that don't need any transformations. The transformations should happen in `normalize...` transaction functions.

Same applies to other adapters.

peronczyk (changes requested):

martinkaintas:
Closed in favor of #2533

PR #2542, opened 2023-11-21
refactor: protocols const to an enum

martinkaintas:
I find it sad that we are using TypeScript yet most of our files are red because our `protocol` props are not (and cannot be) typed correctly with our current setup. This is because we are using an array of constsants `PROTOCOLS`. I think it very much makes sense to use a `Protocols` enum in this case as it simplifies things and makes them more straightforward.

### DIFF
--- a/src/constants/common.ts
+++ b/src/constants/common.ts
@@ -8,15 +8,13 @@ export const MOBILE_WIDTH = 480;
 
 export const LOCAL_STORAGE_PREFIX = 'sh-wallet';
 
-export const PROTOCOL_AETERNITY = 'aeternity';
-export const PROTOCOL_BITCOIN = 'bitcoin';
-export const PROTOCOL_ETHEREUM = 'ethereum';
+export const PROTOCOLS = {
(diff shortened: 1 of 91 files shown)


src/constants/common.ts
```diff
@@ -8,15 +8,13 @@ export const MOBILE_WIDTH = 480;
...
-  PROTOCOL_BITCOIN,
-  PROTOCOL_ETHEREUM,
-] as const;
+export const PROTOCOLS = {
+  aeternity: 'aeternity',
+  bitcoin: 'bitcoin',
+  ethereum: 'ethereum',
+} as const;
+
+export const PROTOCOLS_ARRAY = Object.values(PROTOCOLS);
```
peronczyk, line 17: I suggest to rename this to `PROTOCOL_LIST`.

peronczyk (approved):

Liubov-crypto:
In general LGTM. 

I got some issues in testnet with names (was not able to make it default because of bad middleware work) and with multisig creation, but the same is on dev branch.

PR #2548, opened 2023-11-27
cordova state migrator

ifaouibadi:


src/popup/pages/Index.vue
```diff
@@ -144,6 +151,18 @@ export default defineComponent({
...
 
+    /**
+     * TMP: for IOS Migration
+     */
+    onMounted(async () => {
+      await watchUntilTruthy(mnemonic);
+      if (mnemonic.value && IS_IOS) {
+        setActiveAccountByGlobalIdx(0);
+        await discoverAccounts();
+        router.push(loginTargetLocation.value);
```
peronczyk, line 162: I think we should move user only if the discovery process actually created at least one account.

martinkaintas: What if the user has created this account, not used it yet, but has send the address somewhere and is expecting funds?

peronczyk (approved):

Issue #2541, opened 2023-11-21
aeSdk is being initialized on mainnet instead of chosen network


CedrikNikita:
![networkId](https://github.com/superhero-com/superhero-wallet/assets/9008074/1e9ed65a-deea-4bdf-8177-a278919fbda3)

https://github.com/superhero-com/superhero-wallet/blob/86f3b25d0f514731227983de1c2210b0f96d8e79/src/composables/aeSdk.ts#L107
The network is not being restored here yet.
`nodeUrl` here is https://mainnet.aeternity.io/

PR #2549, opened 2023-11-27
fix: get correct dryAeSdk after changing network

### LINKED ISSUE #2479: Can't create multisig in the testnet
**Describe the bug**
Can't create multisig in the testnet on dev branch.

**To Reproduce**
1. Click on Create multisig vault in the testnet
2. Choose signer 1 and signer 2
3. Click on Confirm

 - OS: MacOS 12.7
 - Browser: FF, Chrome
 

**Additional context**
Reproducible in testnet on dev branch. 

Image
Image
Image
Image

martinkaintas:
closes #2479

### DIFF
--- a/src/composables/aeSdk.ts
+++ b/src/composables/aeSdk.ts
@@ -55,7 +55,6 @@ const isAeNodeError = ref(false);
 const aeppInfo: Record<string, AeppInfoData> = {};
 
 let dryAeSdk: AeSdk;
-let dryAeSdkCurrentNodeNetworkId: string;
 
 export function useAeSdk() {
   const {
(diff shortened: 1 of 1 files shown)


peronczyk (changes requested):

peronczyk (approved):

CedrikNikita:
Both the issue you are trying to fix and #2541 is based on fact that `networkName` is not being restored yet. I don't think that this exact place needed to be fixed. And `await` is not an async action if the `networkId` was already asked.

Liubov-crypto:
In general it's working. 
But sometimes I was not able to create vault especially in FF extension because of bad data fetching - I saw balances from mainnet in the testnet which is not correct. 

ff ext
v3err



https://github.com/superhero-com/superhero-wallet/assets/69896204/ac5eb48a-411d-4f49-bcb2-63140a00d0c7

martinkaintas:
Turns out the problem was not fixed in #2561. Reopening this PR since it fixes the issue.
**Steps to reproduce the issue:** 
- Reset wallet
- Change to testnet
- Try to create  multisig wallet

**This currently doesn't work on develop but should be fixed with this PR**

Liubov-crypto:
Fixed, LGTM

PR #2561, opened 2023-11-30
Fix: AE ADK Init with correct network

### LINKED ISSUE #2557: Fix restoring the networks when reopening the extension popup
The isse is well described here:
[Fixes #2541](https://github.com/superhero-com/superhero-wallet/issues/2541)

Easy way to test the issue is to monitor the balances. If we open the extension on the mainnet we should see the balance equal to X. If we switch the network to testnet the balance should update to Y. Then if we close the extension and reopen it the balance will be displaying X (mainnet value) but the network indicator would display "Testnet".

This task fixes this.

peronczyk:
Fixes #2557

### DIFF
--- a/src/composables/aeSdk.ts
+++ b/src/composables/aeSdk.ts
@@ -58,7 +58,11 @@ let dryAeSdk: AeSdk;
 let dryAeSdkCurrentNodeNetworkId: string;
 
 export function useAeSdk() {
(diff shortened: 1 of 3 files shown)


CedrikNikita (approved):

Liubov-crypto:
LGTM

PR #2575, opened 2023-12-12
Feature/migrate extension to manifest v3

### LINKED ISSUE #2482: Revisit Extension Implementation - support manifest 3
https://aeternity-blockchain.atlassian.net/browse/SW-1174

**Description**

extension: support manifest 3

apply to everything that is needed to push our app on top when you search superhero (extension example https://github.com/aeternity/superhero-wallet/issues/1966 )

martinkaintas:
closes #2482


CedrikNikita (approved):
Code changes look reasonable. Did a small testing, definitely requires a more testing to be merged.

Liubov-crypto:
During my testing I've found next: 
*On Chrome extension:* 

*Important:*
1. I tried to do some Swap/ pool txs but they are failed
swap err
2.  Retip a post on superhero.com was not working on chrome browser. I got an error. 

*Already present on our backlog:*

3. Tx Alighning in multisig tx list ( I think I reported it before, UI issue)
alighning

4. Claim back gift cards works with delay (I think this is not an issue just delay in UI (?)
5. Coin details page becomes broken  after minimise the extension (I've already registered such issue)
6. Names not applicable on testnet (this is also known issue, depends on middleware)
7. Names from mainnet are visible in testnet  (I opened a bug for it alredy)


On FF:
*Important:*
1. I’m seeing tx from previous wallet in my current wallet (I was reloged and cleared cache). It's strange:

https://github.com/superhero-com/superhero-wallet/assets/69896204/977f5204-c909-4ba8-a16c-c065cafdc029

2. ae on btc tx list page:

ae in btc tx page

*Already present on our backlog:*
3. Header is disappearing  (I opened a bug for it)
4. Btc transaction from testnet was visible on mainnet (but disappeared for a while). I think I reported this issue already. 

5. Name dissapeared after I set it default. I was observing this bug few minutes, then it was working correctly...

https://github.com/superhero-com/superhero-wallet/assets/69896204/2317a118-8b65-4498-9afa-0e4b0f9f75a4

6. coin details page becomes broken  after minimise the extension, the same in chrome browser (I've already registered such issue)

7. Wrong header in multisig tx details 

https://github.com/superhero-com/superhero-wallet/assets/69896204/f3b17f06-a70b-499f-b958-3eb541bc24da

8. Tx Alighning in multisig tx ( the same in chrome extension)
9. Names from main net are visible in testnet (known issue , depends on middleware)

martinkaintas:
https://github.com/superhero-com/superhero-wallet/pull/2575/commits/09f4540be001d919570b7bb559d137f6ba45065e brings back the complete `localStorage` clear on wallet reset which was part of the `Vuex reset` before it got removed [here](https://github.com/superhero-com/superhero-wallet/commit/33b3962ff64e4b70d7283eb127b2c749635344a8#diff-9f066e54303c69f2e1622766e9b2658d09e92ee960f9d04edb81717c178ca7d5).
I think resetting the wallet should still clear the `localStorage` and it also is the cause of the new FF bugs reported by @Liubov-crypto

Liubov-crypto:
> [09f4540](https://github.com/superhero-com/superhero-wallet/commit/09f4540be001d919570b7bb559d137f6ba45065e) brings back the complete `localStorage` clear on wallet reset which was part of the `Vuex reset` before it got removed [here](https://github.com/superhero-com/superhero-wallet/commit/33b3962ff64e4b70d7283eb127b2c749635344a8#diff-9f066e54303c69f2e1622766e9b2658d09e92ee960f9d04edb81717c178ca7d5). I think resetting the wallet should still clear the `localStorage` and it also is the cause of the new FF bugs reported by @Liubov-crypto

2 first issues in FF browser has been fixed, LGTM.

Liubov-crypto:
2 first issues in Chrome browser has been fixed, LGTM.

src/background/bgPopupHandler.ts
```diff
@@ -4,17 +4,24 @@ import type {
...
 
-const popups: Dictionary<IPopupConfig> = {};
+/**
+ * We need to duplicate these constants here because
+ * importing from constants.ts causes error in background script
+ * due to undefined window object
+ */
+const PLATFORM = process.env.PLATFORM as 'web' | 'extension' | 'ionic';
+const RUNNING_IN_TESTS = !!process.env.RUNNING_IN_TESTS;
+const IS_EXTENSION = PLATFORM === 'extension' && !RUNNING_IN_TESTS;
```
peronczyk, line 20: If the background lacks of `window` object I suggest to use `?.` to pick any window property. Eg.: `window.name.startsWith('popup-')` => `window?.name?.startsWith('popup-')`

src/background/redirectRule.ts
```diff
@@ -0,0 +1,25 @@
+function updateDynamicRules() {
```
peronczyk, line 1: Please add some comments about the purpose of this code block.

src/background/redirectRule.ts
```diff
@@ -0,0 +1,25 @@
...
+          redirect: {
+            regexSubstitution: `${extUrl}#/\\1`,
+          },
+        },
+      },
+    ],
+  });
+}
+
+export default updateDynamicRules;
```
peronczyk, line 25: I suggest to avoid default exports for any utils.

src/constants/environment.ts
```diff
@@ -44,6 +44,8 @@ export const IS_EXTENSION = PLATFORM === 'extension' && !RUNNING_IN_TESTS;
 
 export const IS_EXTENSION_BACKGROUND = IS_EXTENSION && window.location.href.endsWith('_generated_background_page.html');
 
+export const IS_OFFSCREEN_TAB = window.location.pathname === '/offscreen.html';
```
peronczyk, line 47: Can you please add add a comment that would describe what `offscreen` is.

src/lib/WalletStorage.ts
```diff
@@ -58,6 +82,10 @@ function createLocalStorageInterface(): IWalletStorage {
  * Exposes methods that allows to manipulate or watch the device storage.
  * Allows also to synchronize the state between the app and the extension background.
  */
-export const WalletStorage: IWalletStorage = (!RUNNING_IN_TESTS && IS_EXTENSION && browser)
+export const WalletStorage: IWalletStorage = (
+  !RUNNING_IN_TESTS
+  && !IS_OFFSCREEN_TAB
+  && IS_EXTENSION_BACKGROUND
+  && browser)
```
peronczyk, line 89: If the condition is multiline please put the ending bracket in new line and add indentation to the statements.

src/lib/WalletStorage.ts
```diff
@@ -50,6 +55,25 @@ function createLocalStorageInterface(): IWalletStorage {
...
+     * called in the app. Only in the offscreen tab.
+     */
+    watch: (keys, callback) => {
+      if (!IS_OFFSCREEN_TAB) {
+        return;
+      }
+      window.addEventListener('storage', (event) => {
+        if (
+          composeStorageKeys(keys) === event.key
+          || composeStorageKeys(keys).includes(event.key!)
```
peronczyk, line 69: I don't think it's reasonable to check if strings are equal or part of it is equal. The second test is enough.

src/migrations/migrationHelpers.ts
```diff
@@ -12,7 +13,7 @@ export const collectVuexState = (() => {
     if (window?.browser) {
       if (isCollecting.value) {
         await watchUntilTruthy(isCollecting);
-      } else if (!vuexState) {
+      } else if (!vuexState && !IS_OFFSCREEN_TAB) {
```
peronczyk, line 16: Are you sure the offscreen should not share the state with the extension?

src/popup/components/Modals/ResetWalletModal.vue
```diff
@@ -74,7 +74,11 @@ export default defineComponent({
       resetNetworks();
       resetUiSettings();
       disconnectDapps();
-      router.push({ name: ROUTE_INDEX });
+
+      await browser.storage.local.clear();
+      localStorage.clear();
```
peronczyk, line 79: I suggest to create `WalletStorage` method for this.

src/popup/offscreen.html
peronczyk, line 1: Maybe this file should be located in `src/offscreen` same as the "background" as this is a separate "app"?

vue.config.js
```diff
@@ -37,6 +38,12 @@ module.exports = {
         title: 'cameraPermissions',
         filename: 'CameraRequestPermission.html',
       },
+      offscreen: {
+        template: 'src/popup/offscreen.html',
+        entry: 'src/popup/offscreen.ts',
```
peronczyk, line 43: As already mentioned: I feel those files should be located in `src/offscreen` directory. What do you think?

src/background/offscreen/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+function getCleanParams(params: PopupMessageData['params']) {
+  const cleanedParams = { ...params };
+  delete cleanedParams.params?.onCompiler;
+  delete cleanedParams.params?.onNode;
+  return cleanedParams;
+}
+
+/**
+ *
```
peronczyk, line 28: This line can be removed.

src/background/offscreen/utils.ts
```diff
@@ -0,0 +1,52 @@
+import { CONNECTION_TYPES, IS_FIREFOX } from '@/constants';
+import type { Runtime } from 'webextension-polyfill';
+import { openPopup, removePopup, getPopup } from '../bgPopupHandler';
+import { PopupMessageData } from '..';
+
+export const detectConnectionType = (port: Runtime.Port) => {
+  const extensionProtocol = IS_FIREFOX ? 'moz-extension' : 'chrome-extension';
+  const [senderUrl] = port?.sender?.url?.split('?')!;
+  const isExtensionSender = senderUrl.startsWith(`${extensionProtocol}://${browser.runtime.id}/index.html`)
+    || IS_FIREFOX;
```
peronczyk, line 10:
```suggestion
  const isExtensionSender = (
    senderUrl.startsWith(`${extensionProtocol}://${browser.runtime.id}/index.html`)
    || IS_FIREFOX
  );
```

peronczyk (changes requested):

src/background/bgPopupHandler.ts
martinkaintas: Unfortunately this is not possible. `window` is can't even be present at the background script.
If it is we get the following error, even with something like `window?.name?.startsWith('popup-')`

image

src/migrations/migrationHelpers.ts
martinkaintas: Offscreen uses `localStorage` instead of `window.browser.storage.local`
and IIRC syncing happens in `WalletStorage.ts`

src/background/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  target?: 'background' | 'offscreen';
+  method?: 'openPopup' | 'removePopup' | 'getPopup';
+  type?: PopupActionType;
+  uuid?: string;
+  params?: any;
+  payload?: any;
+};
+
+/**
+ *   ? see https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage#sending_an_asynchronous_response_using_sendresponse
```
peronczyk, line 48:
```suggestion
 * @see https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage#sending_an_asynchronous_response_using_sendresponse
```

src/background/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  if (existingContexts.length > 0) {
+    return;
+  }
+
+  // create offscreen document
+  if (creating) {
+    await creating;
+  } else {
+    // @ts-expect-error
```
peronczyk, line 25: Why it may throw error? If you are sure please add some comment.

src/background/index.ts
```diff
@@ -0,0 +1,76 @@
...
+import { PopupActionType } from '@/types';
+import { openPopup, removePopup, getPopup } from './bgPopupHandler';
+import updateDynamicRules from './redirectRule';
+
+let creating: Promise<void> | null; // A global promise to avoid concurrency issues
+async function setupOffscreenDocument(path: string) {
+  // Check all windows controlled by the service worker to see if one
+  // of them is the offscreen document with the given path
+  const offscreenUrl = browser.runtime.getURL(path);
+  // @ts-expect-error
```
peronczyk, line 11: Why it may throw error? If you are sure please add some comment.

src/background/index.ts
```diff
@@ -0,0 +1,76 @@
+import '@/lib/initPolyfills';
+import { PopupActionType } from '@/types';
+import { openPopup, removePopup, getPopup } from './bgPopupHandler';
+import updateDynamicRules from './redirectRule';
+
+let creating: Promise<void> | null; // A global promise to avoid concurrency issues
```
peronczyk, line 6: If the function `setupOffscreenDocument` is fired only once right after its definition how this is possible that the `creating` value could have different values?

src/background/index.ts
```diff
@@ -0,0 +1,76 @@
+import '@/lib/initPolyfills';
+import { PopupActionType } from '@/types';
+import { openPopup, removePopup, getPopup } from './bgPopupHandler';
+import updateDynamicRules from './redirectRule';
```
peronczyk, line 4: As fart as I see thif function is used only in this file, so maybe it would be good idea to just move it here?

src/background/index.ts
```diff
@@ -0,0 +1,76 @@
...
+        sendResponse(popupConfig);
+      });
+      return true;
+    } if (msg.method === 'removePopup') {
+      sendResponse(removePopup(msg.params.id));
+      return false;
+    } if (msg.method === 'getPopup') {
+      sendResponse(getPopup(msg.params.id));
+      return false;
+    }
```
peronczyk, line 64: Maybe `switch` block would be better here.

peronczyk:
Sorry for global comment (I hate those) but I feel that this PR adds another layer of complexity to functionality that is already hard to read.

As far as I remember you were complaining that the background connection is hard to read and has tangled logic. I agree with that. But now it's even more complex. I feel that there is no structure or easy to follow flow of actions. I know this is subjective, but maybe you can take a step back and take a look if some architectural tweaks can be done. If not just ignore this comment as I don't want to hold back the work.

src/background/index.ts
martinkaintas: It's also used in `offscreen.ts`

src/background/index.ts
martinkaintas: Correct

peronczyk (approved):

PR #2580, opened 2023-12-14
fix: dex missing compiler issue through in-app browser

martinkaintas:


CedrikNikita (changes requested):
Previously our goal was to completely remove our dependency on http compiler, resolving this issue (I'm not sure what exactly is the problem) like that is a step back in my opinion.

martinkaintas:
> Previously our goal was to completely remove our dependency on http compiler, resolving this issue (I'm not sure what exactly is the problem) like that is a step back in my opinion.

I completely agree with you, the issue is that I suppose the DEX is relying on the compiler for some actions.
If you try to reproduce the [linked issue](https://github.com/superhero-com/superhero-wallet/issues/2571) you'll see that we get an error that the compiler is not ready.

peronczyk (approved):

martinkaintas:
CompilerHttp has been deprecated. Same functionalities available using the JS lib https://github.com/aeternity/aepp-calldata-js
This issue should probably be fixed in the DEX

PR #2583, opened 2023-12-14
fix: Aeternity token names fallback

peronczyk:


martinkaintas (approved):
Changes seem reasonable

PR #2584, opened 2023-12-15
feat: add support of erc20 tokens into eth adapter

CedrikNikita:


src/protocols/BaseProtocolAdapter.ts
```diff
@@ -96,6 +97,19 @@ export abstract class BaseProtocolAdapter {
 
   abstract fetchAccountTokenBalances(address: string): Promise<ITokenBalance[]>;
 
+  abstract transferToken(
+    amount: number,
+    recipient: string,
+    contractId: string,
+    options: {
+      fromAccount: IAccount;
```
peronczyk, line 105: Maybe instead of `IAccount` it would be better to use `string` and pass the address?

src/protocols/aeternity/libs/AeternityAdapter.ts
```diff
@@ -223,6 +223,17 @@ export class AeternityAdapter extends BaseProtocolAdapter {
...
 
+  override async fetchTokenInfo(): Promise<IToken> {
+    // TODO if needed
+    // @ts-expect-error
+    return undefined;
+  }
+
+  // @ts-expect-error
+  override async transferToken(): Promise<{ hash: string }> {
+    //
```
peronczyk, line 234: I suggest to return something here.

src/protocols/bitcoin/libs/BitcoinAdapter.ts
```diff
@@ -206,6 +207,17 @@ export class BitcoinAdapter extends BaseProtocolAdapter {
     return [];
   }
 
+  override async fetchTokenInfo(): Promise<IToken> {
+    // TODO if needed
+    // @ts-expect-error
```
peronczyk, line 212: I think the base protocol adapter should allow to return undefined because we know that some protocols has no tokens (BTC).

src/protocols/BaseProtocolAdapter.ts
```diff
@@ -96,6 +97,19 @@ export abstract class BaseProtocolAdapter {
...
+      maxPriorityFeePerGas: string;
+      maxFeePerGas: string;
+    },
+  ): Promise<{ hash: string }>;
```
peronczyk, line 109: The type `{ hash: string }` is already used in some places. I suggest to create separate transfer response type and reuse it.

src/protocols/aeternity/libs/AeternityAdapter.ts
```diff
@@ -223,6 +223,17 @@ export class AeternityAdapter extends BaseProtocolAdapter {
     // TODO
   }
 
+  override async fetchTokenInfo(): Promise<IToken> {
+    // TODO if needed
+    // @ts-expect-error
+    return undefined;
+  }
+
+  // @ts-expect-error
```
peronczyk, line 232: I think the base adapter should be written in a way that would not force us to use such hacks.

src/protocols/ethereum/config.ts
```diff
@@ -10,6 +10,46 @@ export const ETH_COIN_PRECISION = 18; // Amount of decimals
 export const ETH_COINGECKO_COIN_ID = 'ethereum';
 export const ETH_GAS_LIMIT = 21000;
 
+export const ERC20_ABI = [
```
peronczyk, line 13:
```suggestion
export const ERC20_ABI: AbiItem[] = [
```

The `AbiItem` type is probably available in the 'web3-utils' package.

src/protocols/BaseProtocolAdapter.ts
CedrikNikita: Are you suggesting to get the `secretKey` in the function itself using the 'address'?

peronczyk: This is what we are doing here: https://github.com/superhero-com/superhero-wallet/blob/develop/src/protocols/aeternity/libs/AeAccountHdWallet.ts#L21
I think we need to be consistent on the type passed here. If you think passing whole account is more reasonable I think the `AeAccountHdWallet` needs to be updated also.

src/protocols/bitcoin/libs/BitcoinAdapter.ts
```diff
@@ -206,6 +208,15 @@ export class BitcoinAdapter extends BaseProtocolAdapter {
     return [];
   }
 
+  override async fetchTokenInfo(): Promise<IToken | undefined> {
+    return undefined;
+  }
+
+  override async transferToken() {
+    // has no tokens
+    return { hash: '' };
```
peronczyk, line 217: Maybe in this case we should return `null` instead of an object?

src/protocols/ethereum/views/TransferSendModal.vue
```diff
@@ -12,7 +12,7 @@
         ref="currentRenderedComponent"
         v-model:transferData="transferData"
         @success="currentStepConfig.onSuccess"
-        @error="(val: any) => error = val"
+        @error="(val) => error = val"
```
peronczyk, line 15: I don't think this is a reasonable change. When you remove the `any` my IDE is marking the file as bugged (filename is red). Also the line is underlined:
image

src/types/index.ts
```diff
@@ -803,3 +803,7 @@ export type IonAnimationBuilder = (
   baseEl: Element,
   opts: { enteringEl: Element; leavingEl: Element }
 ) => Animation;
+
+export interface ITransferResponse{
```
peronczyk, line 807:
```suggestion
export interface ITransferResponse {
```

src/protocols/ethereum/views/TransferSendModal.vue
CedrikNikita: The situation is opposite for me, i'm using different plugin. I can change it back ofc.

src/protocols/BaseProtocolAdapter.ts
CedrikNikita: Same approach is chosen in `spend` method here. https://github.com/superhero-com/superhero-wallet/blob/feat/eth-support/src/protocols/ethereum/libs/EthereumAdapter.ts#L233
in order to get the `secretKey` like that
https://github.com/superhero-com/superhero-wallet/blob/feat/eth-support/src/protocols/ethereum/libs/EthereumAdapter.ts#L262
Do you want to reconsider it and for me to apply some changes on both of them?

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -191,11 +194,124 @@ export class EthereumAdapter extends BaseProtocolAdapter {
   }
 
   override async fetchAvailableTokens(): Promise<IToken[]> {
```
peronczyk, line 196: I suggest to catch errors in the body of this function or eventually in the place where it's executed.

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -191,11 +194,124 @@ export class EthereumAdapter extends BaseProtocolAdapter {
...
+      contractId: token.address,
+      decimals: token.decimals || undefined,
+      image: token.image ? `https://ethplorer.io/${token.image}` : undefined,
+      name: token.name,
+      protocol: PROTOCOLS.ethereum,
+      symbol: token.symbol,
+    })) : [];
+  }
+
+  override async fetchAccountTokenBalances(address: string): Promise<ITokenBalance[]> {
```
peronczyk, line 213: I suggest to catch errors in the body of this function or eventually in the place where it's executed.

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -191,11 +194,124 @@ export class EthereumAdapter extends BaseProtocolAdapter {
...
+    }: EthRawToken) => ({
+      address,
+      amount: rawBalance,
+      contractId,
+      convertedBalance: +toShiftedBigNumber(rawBalance, -decimals).toFixed(2),
+      protocol: PROTOCOLS.ethereum,
+    })) || [];
+  }
+
+  override async fetchTokenInfo(contractId: string): Promise<IToken> {
```
peronczyk, line 231: I suggest to catch errors in the body of this function or eventually in the place where it's executed.

src/protocols/aeternity/libs/AeternityAdapter.ts
```diff
@@ -223,6 +224,15 @@ export class AeternityAdapter extends BaseProtocolAdapter {
...
   }
 
+  override async fetchTokenInfo(): Promise<IToken | undefined> {
+    // TODO if needed
+    return undefined;
+  }
+
+  override async transferToken() {
+    return undefined;
+  }
```
peronczyk, line 234: I suggest to move the `transferToken` logic from `fungibleTokens` composable here as this logic strictly refers to AE.

peronczyk (changes requested):

src/protocols/BaseProtocolAdapter.ts
peronczyk: Whatever we choose I think we need to be consistent. In my opinion passing the address and then picking the account based on the address is more safe.

PR #2585, opened 2023-12-18
feat: ERC-20 token details

peronczyk:


src/popup/pages/Assets/AssetDetailsInfo.vue
```diff
@@ -105,25 +105,12 @@
             {{ formatCurrency(tokenData?.marketCapChange24h) }}
           </template>
         </DetailsRow>
+
         <DetailsRow
           v-if="!isAe && UNFINISHED_FEATURES"
           :label="$t('pages.token-details.feeDaily')"
         />
 
-        <DetailsRow
```
peronczyk, line 113: Removal of this code block was agreed with Paolo and Tsvetan on the chat.

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -191,7 +192,27 @@ export class EthereumAdapter extends BaseProtocolAdapter {
...
+      // TODO Replace with our own node API
+      const tokens = (await fetchJson('https://api.ethplorer.io/getTop?apiKey=freekey'))?.tokens || [];
+
+      return tokens
+        .filter(({ address, name, symbol }: any) => (address && name && symbol))
+        // Remove the ETH coin from the results because as we are adding it manually
+        .filter(({ address }: any) => address !== ETH_CONTRACT_ID_EXTERNAL)
+        .map((token: any): IToken => ({
+          contractId: token.address,
+          decimals: 0, // TODO api does not provide this value so we need to come up how to get this
```
peronczyk, line 206: @CedrikNikita we discussed this. Did you find any solution for this?

src/composables/networks.ts
```diff
@@ -150,18 +150,23 @@ export function useNetworks() {
  * user changes the network.
  */
 export function createNetworkWatcher() {
-  ensureDefaultNetworksExists();
+  // Delay the networks fill in to allow the protocol adapters to be created first
```
CedrikNikita, line 153: Feel like a workaround rather than actual solution. And the source of the failing test (IMO) is behind the source of this changes.

src/protocols/ethereum/libs/EthereumAdapter.ts
peronczyk: OK, I'll updated this in this PR: https://github.com/superhero-com/superhero-wallet/pull/2594/files#diff-a9ba9cbedd19c9f66d564d2e89912890209b98f0a7ef19187877d2587300e476R162

CedrikNikita (approved):

src/protocols/ethereum/config.ts
```diff
@@ -26,29 +27,21 @@ export const ETH_NETWORK_DEFAULT_SETTINGS: IDefaultNetworkTypeData<IEthNetworkSe
...
+> = {
   [NETWORK_TYPE_MAINNET]: {
     explorerUrl: 'https://etherscan.io',
     middlewareUrl: 'https://api.etherscan.io/api', // TODO replace temp values
+    tokenMiddlewareUrl: 'https://api.ethplorer.io/',
   },
   [NETWORK_TYPE_TESTNET]: {
     explorerUrl: 'https://sepolia.etherscan.io',
     middlewareUrl: 'https://api-sepolia.etherscan.io/api', // TODO replace temp values
+    tokenMiddlewareUrl: 'https://sepolia-api.ethplorer.io/',
```
CedrikNikita, line 41: I think we should be consistent with the way we are adding such urls, both urls above have no `/` symbol in the end of it,

src/composables/networks.ts
peronczyk: I created separate task for this: https://github.com/superhero-com/superhero-wallet/issues/2595

src/protocols/ethereum/config.ts
peronczyk: You're right. Updated.

PR #2598, opened 2023-12-27
fix: Avoid multiple composables watcher and delaying network watchers

peronczyk:


src/composables/networks.ts
```diff
@@ -142,31 +152,6 @@ export function useNetworks() {
...
-  // Delay the networks fill in to allow the protocol adapters to be created first
-  setTimeout(() => {
-    ensureDefaultNetworksExists();
-  }, 0);
-
-  let storedNetworkName: string;
-
-  return {
-    onNetworkChange: (callback: (newNetwork: INetwork, oldNetwork: INetwork) => void) => {
-      setTimeout(() => {
```
peronczyk, line 162: This `setTimeout` was one of the reasons of this PR. Now we don't need any 0s delays.

martinkaintas (approved):

Liubov-crypto:
My assets are displaying with some delay (14 sec compared with 5 sec on dev branch), and I found some error in console: 

asset

Also I can see Link token from testnet in mainnet (the same on dev branch): 

key

CedrikNikita:
> My assets are displaying with some delay (14 sec compared with 5 sec on dev branch), and I found some error in console:

For me displaying of my assets are mostly equal between this branch and `develop` or `feat/eth-support`.

Liubov-crypto:
> > My assets are displaying with some delay (14 sec compared with 5 sec on dev branch), and I found some error in console:
> 
> For me displaying of my assets are mostly equal between this branch and `develop` or `feat/eth-support`.

it could be, i have 25 assets in my wallet so probably there is a reason for the delay - the huge list.

PR #2604, opened 2024-01-05
fix: token info disappearing from token details page

### LINKED ISSUE #2574: The info about token is disappearing from Token details page
**Describe the bug**
The info about token is disappearing from Token details page when I minimise and restore extension. 

**To Reproduce**
1. Go to Token details page 
2. Open token details tab and minimise the extension
3. Open extension again

**Expected behavior**
Minimising shouldn't affect the extension info. 

**Screenshots**

 - OS: MacOS
 - Browser FF
 - Version 119

**Additional context**
Reproducible on extension on dev branch. 

https://github.com/superhero-com/superhero-wallet/assets/69896204/0f612a46-4d68-4d60-9f97-7c0114368213
Image

martinkaintas:
closes #2574

### DIFF
--- a/src/composables/coinTokensProps.ts
+++ b/src/composables/coinTokensProps.ts
@@ -1,12 +1,21 @@
-import { computed, ref } from 'vue';
+import { computed } from 'vue';
 import type { TokenProps } from '@/types';
+import { STORAGE_KEYS } from '@/constants';
+import { useStorageRef } from './storageRef';
 
 type ITokenProps = Partial<TokenProps>;
 
 type ITokenDetail = Omit<TokenProps, 'isMultisig'>
 type ITokenTransaction = Pick<TokenProps, 'contractId' | 'isMultisig'>
 
-const tokenProps = ref<ITokenProps | null>(null);
+/**
+ * Save last used token props in local storage
+ * in order to have data available if user refreshes the page
+ */
+const tokenProps = useStorageRef<ITokenProps | null>(
+  null,
+  STORAGE_KEYS.tokenProps,
+);
 
 /**
  *  Data coming from a coin that will be passed to the tab components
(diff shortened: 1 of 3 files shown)


src/composables/coinTokensProps.ts
```diff
@@ -28,6 +40,12 @@ export function useTokenProps() {
     isMultisig: tokenProps.value?.isMultisig ?? false,
   }));
 
+  watch(route, () => {
+    if (route.params.id !== tokenProps.value?.contractId) {
+      tokenProps.value = null;
+    }
+  }, { immediate: true });
```
peronczyk, line 47: I noticed that we have performance issues with watcher initiated multiple time every time the composable is fired. Please bare in mind that every time `useTokenProps` is fired the watcher is created once again. This is why I suggest to create `let initialized` before the composable and run watchers only once. Here's an example: https://github.com/superhero-com/superhero-wallet/pull/2598/files#diff-1698ea5d81d976b7229ebb41ba6b715373125bdc644776ba7774792ffbefd7e9R256

peronczyk: But to be honest I think this composable should not have watcher at all. The component that loads it should take care of resetting the state.

Liubov-crypto:
In general it has been fixed. 
But for the some reason it doesn't show the full info about AE coin on testnet, and if I'm switching to mainnet the AE coin details page will be also not fully displayed with all info about the coin:

https://github.com/superhero-com/superhero-wallet/assets/69896204/617b39d4-8a5f-4b71-bd6b-eca14c303ca0

From what I see in my wallet that the wallet is trying to connect first to mainnet even if I'm on testnet. Maybe this is the reason of bug. 


https://github.com/superhero-com/superhero-wallet/assets/69896204/4f669131-ffc3-45c3-bfd2-cbb95604c5e4

CedrikNikita (approved):

martinkaintas:
This PR is obsolete as the issue has been fixed in `feat/eth-support`

PR #2607, opened 2024-01-09
refactor(ethereum-adapter): do not start sending process if different protocol is chosen

CedrikNikita:


martinkaintas (approved):
LGTM but I would suggest to create a helper function for checking and requiring the account to be of a specific protocol. This could be useful in other places too

PR #2613, opened 2024-01-10
fix: owned names should only be updated for current network

### LINKED ISSUE #2599: Names from the testnet are visible in the mainnet and vise versa
**Describe the bug**
Names from the testnet are visible in the mainnet and vise versa

**To Reproduce**
1. Go to Names tab and set name by default
2. Observe that the names from the testnet will be visible on the mainnet

**Expected behavior**
Name should belong to network where it was registered. 

**Screenshots**

 - OS: MacOS
 - Browser FF, Chrome
 - Version  121

**Additional context**
Reproducible on the dev branch. 

https://github.com/superhero-com/superhero-wallet/assets/69896204/fc93ed89-51fa-48a3-827c-5a66c501ca19

martinkaintas:
closes #2599 

`onIonViewWillLeave` was only being called when user switched from `My names` to other sub-tabs `Auctions` & `Register name`.
When user closed the `account-details` page it was not being called, leaving an unterminated interval.

`onUnmount` will be called when user closes the `account-details` page, which means that while user is on that page, names will update every 10 seconds. But when he leaves this page names will stop updating.

### DIFF
--- a/src/popup/pages/Names/NamesList.vue
+++ b/src/popup/pages/Names/NamesList.vue
@@ -25,8 +25,8 @@
 </template>
 
 <script lang="ts">
-import { IonPage, IonContent, onIonViewWillLeave } from '@ionic/vue';
-import { computed, defineComponent } from 'vue';
+import { IonPage, IonContent } from '@ionic/vue';
+import { computed, defineComponent, onUnmounted } from 'vue';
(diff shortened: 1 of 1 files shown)


Liubov-crypto:
Fixed, LGTM. 

I will add to backlog other issues with names.

PR #2618, opened 2024-01-12
fix: transaction list is displaying partly

### LINKED ISSUE #2563: Header is missing after sending Tx within the wallet
**Describe the bug**
if user send tokens from the 1st account to the 2nd one inside the wallet, and then open the transaction details, the header disappears

**To Reproduce**
1. Send ae coin from 1st acc to 2nd acc
2. Check tx details
3. Go back

**Expected behavior**

**Screenshots**

 - OS: MacOS
 - Browser FF
 - Version 119

**Additional context**
reproducible on the dev branch. 

https://github.com/superhero-com/superhero-wallet/assets/69896204/92562a07-54ac-449d-b11e-5cc0a7ae9357

### LINKED ISSUE #2572: Screen with transaction list and assets list is displaying partly
**Describe the bug**
Screen with transaction list and assets list is displaying partly, not all information is visible

**To Reproduce**
1. Go to transaction list and assets list 
2. See error

**Expected behavior**
Should be fully visible. 

**Screenshots**

**Smartphone (please complete the following information):**
 - Device:  iPhone13
 - OS:  iOS 17.1.2

**Additional context**

https://github.com/superhero-com/superhero-wallet/assets/69896204/faf04428-c8a7-465b-9ac9-1484d6e510df

martinkaintas:
closes #2572 
closes #2563

As much as I hate using this `ionicLifecycleStatus` prop, it is the best way to achieve what we're trying to do. Which is to calculate the tabs-height after the page transition has played. `onIonViewDidEnter` fires at exactly that moment but isn't directly available to the `AccountDetailsBase` component thus we pass it as a prop.
Using the timeout didn't always work correctly because depending on the user's machine the animation can take longer or even not play at all (e.g. animations are disabled on FF or user visits the page directly)

### DIFF
--- a/src/popup/components/AccountDetailsBase.vue
+++ b/src/popup/components/AccountDetailsBase.vue
@@ -56,7 +56,7 @@
 
       <div
         class="tabs-content"
-        :style="{ height: routerHeight || '350px' }"
+        :style="{ height: routerHeight || `${INITIAL_TABS_HEIGHT}px`}"
       >
         <!-- We are disabling animations on FF because of a bug that causes flickering
           see: https://github.com/ionic-team/ionic-framework/issues/26620 -->
@@ -75,7 +75,6 @@ import { StatusBar } from '@capacitor/status-bar';
 import {
   computed,
   defineComponent,
-  nextTick,
   onBeforeUnmount,
   onMounted,
   ref,
@@ -98,6 +97,8 @@ import AccountInfo from '@/popup/components/AccountInfo.vue';
 import BtnClose from '@/popup/components/buttons/BtnClose.vue';
 import TransactionAndTokenFilter from '@/popup/components/TransactionAndTokenFilter.vue';
 import { popOutAnimation, fadeAnimation } from '@/popup/animations';
+
+const INITIAL_TABS_HEIGHT = 330;
 
 export default defineComponent({
   name: 'AccountDetailsBase',
@@ -125,6 +126,7 @@ export default defineComponent({
 
     const routerHeight = ref<string>();
     const headerEl = ref<HTMLDivElement>();
+    const resizeObserver = ref<ResizeObserver>();
 
     const balanceNumeric = computed(() => balance.value.toNumber());
     const routeName = computed(() => route.name);
@@ -145,12 +147,8 @@ export default defineComponent({
      * Tabs change height when filters are shown/hidden
      */
     function observeTabsHeight() {
-      const resizeObserver = new ResizeObserver(() => {
-        nextTick(() => {
-          calculateRouterHeight();
-        });
-      });
-      resizeObserver.observe(headerEl.value!);
+      resizeObserver.value = new ResizeObserver(calculateRouterHeight);
+      resizeObserver.value.observe(headerEl.value!);
     }
 
     watch(
@@ -161,18 +159,21 @@ export default defineComponent({
     );
 
     onMounted(() => {
+      observeTabsHeight();
+      // The timeout ensures that the height is calculated correctly in some edge cases
+      setTimeout(() => {
+        calculateRouterHeight();
+      }, 150);
+
       if (IS_MOBILE_APP) {
         StatusBar.setBackgroundColor({
           color: '#191919',
         });
       }
-      setTimeout(() => {
-        observeTabsHeight();
-        calculateRouterHeight();
-      }, 250);
     });
 
     onBeforeUnmount(() => {
+      resizeObserver.value?.disconnect();
       if (IS_MOBILE_APP) {
         StatusBar.setBackgroundColor({
           color: '#141414',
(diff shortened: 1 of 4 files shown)


Liubov-crypto:
In general the transaction list view has been fixed. 
But for the some reason I can't see all my assets ( I see only 4 from 25: https://aescan.io/accounts/ak_2UGL6qArnbWx9fEkJyszssMkCjpu3VeLFBPA6MXYV3APfDSz1k?type=tokens). And DEX related transactions are displaying as Token A/ Token B, there is no "true" token name.

assets
tok

Liubov-crypto:
The issue with assets list is also present on dev, on dev eth branch and even on product
I think there is an issue with mainnet , I got 400 error

src/popup/components/AccountDetailsBase.vue
```diff
@@ -157,16 +160,22 @@ export default defineComponent({
       },
     );
 
+    watch(
+      () => props.ionicLifecycleStatus,
+      (value) => {
+        if (value === 'didEnter') {
```
peronczyk, line 166: What if the value is different? Maybe we need to unobserve?

src/protocols/aeternity/views/AccountDetails.vue
```diff
@@ -58,6 +61,7 @@ import CreditCardIcon from '@/icons/credit-card.svg?vue-component';
 import SwapIcon from '@/icons/swap.svg?vue-component';
 import FaucetIcon from '@/icons/faucet.svg?vue-component';
 import GlobeSmallIcon from '@/icons/globe-small.svg?vue-component';
+import { IonicLifecycleStatus } from '@/types';
```
peronczyk, line 64: Please put this line above local imports.

src/popup/components/AccountDetailsBase.vue
martinkaintas: `ResizeObservers` are supposed to disconnect when the element is removed. But for clarity we can call `disconnect()`

Liubov-crypto:
https://github.com/superhero-com/superhero-wallet/issues/2563 - has been fixed. 

https://github.com/superhero-com/superhero-wallet/issues/2572 - it seems to be fixed too. But if I'm opening console and then close it I will see my list after only partly. Is this view could be caused by console opening? 


https://github.com/superhero-com/superhero-wallet/assets/69896204/449adb3b-cb2f-4a49-8d46-23039eb8417e

p.s. multisig accounts are not visible, I think this is related to old URLs.

martinkaintas:
It is related to the opening of the console because it interferes with the way we calculate the height :/
Is this something we want to fix?

Liubov-crypto:
> It is related to the opening of the console because it interferes with the way we calculate the height :/ Is this something we want to fix?

I don't think that someone from regular users will open console to check there errors :) 

if other reviewers are ok with this, I'm also ok :)

CedrikNikita:
Currently in the #2659, the `IonPage` is added in `AccountDetailsTransactionsBase`, to solve the same problem on hooks is not accessible. Can this be fixed the same way? What in general do you think of that approach?

martinkaintas:
> Currently in the #2659, the `IonPage` is added in `AccountDetailsTransactionsBase`, to solve the same problem on hooks is not accessible. Can this be fixed the same way? What in general do you think of that approach?

In the mentioned PR [this line](https://github.com/superhero-com/superhero-wallet/pull/2659/files#diff-1ed85de4c375d07db0f12a185bde23caf8979423a0add30a4624b2bd3bf9fdcaL151) was also changed so that `AccountDetailsTranactionsBase` is directly mapped with the router. Which was done to satisfy this constraint by [Ionic](https://ionicframework.com/docs/vue/lifecycle):

> These lifecycles are only called on components directly mapped by a router

Removing the protocol specific `AccountDetailsTransactions` files was possible because they were just using `AccountDetailsTranactionsBase` without any other additions.

In the case of `AccountDetails` we cannot move all the logic and the different UIs that we want to the Base component

src/protocols/aeternity/views/AccountDetails.vue
```diff
@@ -1,7 +1,10 @@
 <template>
   <IonPage>
     <IonContent class="account-ion-content">
-      <AccountDetailsBase class="account-details">
+      <AccountDetailsBase
+        class="account-details"
+        :ionic-lifecycle-status="ionicLifecycleStatus"
```
peronczyk, line 6: Instead of passing the prop deeper and deeper just use `v-if` in one of the parents and display/hide the element to force it to render again when the ionic lifecycle changes?

peronczyk (changes requested):

peronczyk (approved):

Liubov-crypto:
LGTM in general. 

But I got again issue with console opening: 

https://github.com/superhero-com/superhero-wallet/assets/69896204/e0031d42-d655-4910-98e5-3ed58aed556c

PR #2619, opened 2024-01-12
refactor: unify icon sizes

### LINKED ISSUE #2578: Unify protocol and token icon sizes
Currently the icons displayed on the account details asset list are not equal. The protocol icon used as the coin icon has different size (smaller) than the token sizes:

Image

The icon sizes should be the same and match the figma designs: **18px**.

martinkaintas:
closes #2578 

_Based on `eth-support` because there were some changes with icons but could be rebased on develop as well if needed._ 
I think having a global standard for icon sizes will help cross-component compatibility (e.g. `ProtocolIcon.vue` inside `Tokens.vue`)
One shortcoming of this approach is that some components do not have `rg` as the default size (which semantically makes sense) but I think that the clarity of the global standard is more important

### DIFF
--- a/src/constants/common.ts
+++ b/src/constants/common.ts
@@ -382,6 +382,19 @@ export const ALLOWED_ICON_STATUSES = [
   'warning',
 ] as const;
 
+/**
+ * Pixel size is defined in css variables in global.scss
+ */
+export const ICON_SIZES = {
+  xs: 'xs', // 12px
+  sm: 'sm', // 16px
+  rg: 'rg', // 18px
+  md: 'md', // 20px
+  lg: 'lg', // 24px
+  xl: 'xl', // 28px
+  xxl: 'xxl', // 30px
+} as const;
+
 export const TRANSACTIONS_LOCAL_STORAGE_KEY = 'transactions';
 export const TRANSFER_SEND_DATA_LOCAL_STORAGE_KEY = 'transfer-send-data';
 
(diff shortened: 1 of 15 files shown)


src/constants/common.ts
```diff
@@ -382,6 +382,25 @@ export const ALLOWED_ICON_STATUSES = [
...
 ] as const;
 
+/**
+ * xsm: 12px |
+ * sm: 16px |
+ * rg: 18px |
+ * md: 20px |
+ * lg: 24px |
+ * xl: 28px |
+ * xxl: 30px |
```
peronczyk, line 392: Why not making this comments in the particular lines?
For example:
```
  xsm: 'xsm', // 12px
```

I think it will be more readable.

martinkaintas: But it wont show as a comment when you hover over `ICON_SIZES`. That was what I was aiming for

peronczyk: So maybe do this in a following way so when you hover the actual property it will show you the value:
```
ICON_SIZES = {
  /** 12px */
  xs: 'xs',
  ...
}
```

martinkaintas: I wanted the developer to be able to see all the values and select the right value right away, if that doesn't provide any value lets go with what you're suggesting

peronczyk: I understand your point of view. But this object does not set the px values. It's only a name map. The CSS actually sets the px size.

PR #2627, opened 2024-01-16
fix: seed-phrase verification alert alignment

### LINKED ISSUE #2626: Seed phrase verification frame is shifted
https://aeternity-blockchain.atlassian.net/browse/SW-995

Describe the bug

Seed phrase verification frame is shifted

Expected result

Should match Figma proto

Steps to reproduce
1. Try to verify Seed phrase
2. Compare to Figma proto: https://www.figma.com/file/3oGLWzSH0oJljo4RETZtur/Superhero-Wallet-UI-(%E2%9C%94%EF%B8%8FUpdated)?type=design&node-id=15395-317043&t=LjVX5dRMYbZJIu4V-0Connect your Figma account 

Environment selector
- Extension
- Firefox
- MacOS
- Web wallet

martinkaintas:
closes #2626 

Size of the `phraser` element can change if words in seed-phrase are more & bigger. With this change no matter how the big the element is the notification element will always match it pixel perfect

### DIFF
--- a/src/popup/components/SeedPhraseNotification.vue
+++ b/src/popup/components/SeedPhraseNotification.vue
@@ -1,5 +1,8 @@
 <template>
-  <div :class="['seed-phrase-notification', { error: hasError }]">
+  <div
+    :class="['seed-phrase-notification', { error: hasError }]"
+    :style="{ bottom: `${positionBottom}px`, minHeight: `${phraserElHeight}px` }"
+  >
     <div class="icon-wrapper">
       <AlertIcon v-if="hasError" />
       <CheckCircleIcon v-else />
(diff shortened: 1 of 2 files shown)


Liubov-crypto:
Fixed, LGTM

CedrikNikita (commented):
I think the fact that user is redirected after he failed to verify a seed phrase, is a mistake. The solution for misaligning looks bulky.

martinkaintas:
> I think the fact that user is redirected after he failed to verify a seed phrase, is a mistake. The solution for misaligning looks bulky.

It is a mistake, and it is fixed in another PR. 
This solution should be pixel perfect in any given case.

PR #2629, opened 2024-01-16
feat: seed-phrase verification should not redirect

### LINKED ISSUE #2628: After verifying seed-phrase, user should not be redirected, instead "Go home" button should appear

martinkaintas:
closes #2628

### DIFF
--- a/src/popup/locales/en.json
+++ b/src/popup/locales/en.json
@@ -51,7 +51,8 @@
       "medium": "Medium",
       "slow": "Slow"
     },
-    "actionMayTakeFewMoments": "This may take a few moments."
+    "actionMayTakeFewMoments": "This may take a few moments.",
+    "backToHome": "Proceed to dashboard"
   },
   "connectionStatus": {
     "offline": "You are offline. Go online to get real time data.",
--- a/src/popup/pages/SeedPhraseVerifySettings.vue
+++ b/src/popup/pages/SeedPhraseVerifySettings.vue
@@ -57,12 +57,17 @@
         </div>
         <FixedScreenFooter>
           <BtnMain
+            v-if="hasError || !showNotification"
             class="verify-button"
-            :disabled="!selectedWordIds || selectedWordIds.length !== mnemonicShuffled.length"
+            :disabled="isVerifyButtonDisabled"
+            :text="$t('pages.seedPhrase.verify')"
             @click="verifyLastStep"
-          >
-            {{ $t('pages.seedPhrase.verify') }}
-          </BtnMain>
+          />
+          <BtnMain
+            v-else
+            :text="$t('common.backToHome')"
+            :to="{name: ROUTE_ACCOUNT}"
+          />
           <SeedPhraseNotification
             v-if="showNotification"
             :has-error="hasError"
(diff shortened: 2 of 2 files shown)


src/popup/locales/en.json
```diff
@@ -866,7 +866,8 @@
       "backUpYourSeedPhrase": "Back up your seed phrase {0} on a piece of paper or file that no one but you will ever see. Make multiple copies. Make sure you keep your seed phrase in a safe place!",
       "inCorrectOrder": "in the correct order of 12 words",
       "toBeSureYouGotItRight": "To be sure you got it right, you should {0} clicking on the button below!",
-      "verifyYourSeedPhrase": "verify your seed phrase"
+      "verifyYourSeedPhrase": "verify your seed phrase",
+      "goHome": "Go home"
```
peronczyk, line 870: Can you check this with Tsvetan? As far as I know we have different naming for such button - like on 404 page.
Also I think this should be moved to common phrases.

src/popup/pages/SeedPhraseVerifySettings.vue
```diff
@@ -54,12 +54,17 @@
...
+            :text="$t('pages.seedPhrase.verify')"
             @click="verifyLastStep"
-          >
-            {{ $t('pages.seedPhrase.verify') }}
-          </BtnMain>
+          />
+          <BtnMain
+            v-else
+            :text="$t('pages.seedPhrase.goHome')"
+            @click="goHome"
```
peronczyk, line 66: Buttons has the `to` property which makes them links which are more user friendly and easier to debug.

peronczyk (changes requested):

Liubov-crypto:
LGTM

src/popup/pages/SeedPhraseVerifySettings.vue
```diff
@@ -57,12 +57,17 @@
...
+            :text="$t('pages.seedPhrase.verify')"
             @click="verifyLastStep"
-          >
-            {{ $t('pages.seedPhrase.verify') }}
-          </BtnMain>
+          />
+          <BtnMain
+            v-else
+            :text="$t('common.backToHome')"
+            :to="{name: ROUTE_ACCOUNT, replace: true}"
```
peronczyk, line 69: Why `replace: true` is used here?

src/popup/pages/SeedPhraseVerifySettings.vue
```diff
@@ -117,22 +120,26 @@ export default defineComponent({
     const examplePhrase = ref([t('pages.seedPhrase.first'), t('pages.seedPhrase.second'), '...']);
 
     const mnemonicShuffled = computed((): string[] => shuffle(mnemonic.value.split(' ')));
+    const isVerifyButtonDisabled = computed(() => (
+      !selectedWordIds
```
peronczyk, line 124: This is always true as this is an object with `value` property.

peronczyk (approved):

src/popup/pages/SeedPhraseVerifySettings.vue
martinkaintas: The reason `replace: true` is used, is because otherwise Ionic will cache the page.
This poses a security risk since if the user verifies the seed-phrase and then goes to the dashboard, if he leaves his device unattended someone could just go back and see the seed-phrase.

But on second thought, someone could see the seed-phrase anyway from the settings. I will remove the `replace: true`

PR #2634, opened 2024-01-17
chore: add missing data-cy ids

martinkaintas:
This PR adds missing data-cy ids requested by @dvdbak

### DIFF
--- a/src/popup/components/AccountDetailsBase.vue
+++ b/src/popup/components/AccountDetailsBase.vue
@@ -1,5 +1,8 @@
 <template>
-  <div class="account-details">
+  <div
+    class="account-details"
+    :data-account-address="activeAccount.address"
+  >
     <div class="account-info-wrapper">
       <slot
         v-if="$slots['account-info']"
--- a/src/popup/components/Modals/MultisigVaultCreate.vue
+++ b/src/popup/components/Modals/MultisigVaultCreate.vue
@@ -20,6 +20,7 @@
       <div
         v-for="(signer, index) in signers"
         :key="`signer${index}`"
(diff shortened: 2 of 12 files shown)


src/popup/components/AccountDetailsBase.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <div class="account-details">
+  <div
+    class="account-details"
+    :data-account-address="activeAccount.address"
```
CedrikNikita, line 4: Why isn't it a `data-cy`?

martinkaintas: It could be, but given that it is not an id I think it's better to give it a more descriptive name

CedrikNikita (approved):

src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -20,6 +20,7 @@
       <div
         v-for="(signer, index) in signers"
         :key="`signer${index}`"
+        :data-cy="`signer-address-${index}`"
```
peronczyk, line 23: I strongly recommend to not use dynamic test IDs.

peronczyk (changes requested):

dvdbak (approved):

src/popup/components/Modals/MultisigVaultCreate.vue
martinkaintas: I changed it to `:data-signer-idx` to make it more clear

peronczyk (approved):

PR #2641, opened 2024-01-19
fix: tip-in transaction details

### LINKED ISSUE #2438: TIP IN transactions present different data in transaction details after refreshing
https://aeternity-blockchain.atlassian.net/browse/SW-558

*Describe the bug*
Presented data changes after page refresh - transaction object fetched after refresh is lacking `amount` field

*Expected result*
Data should be the same after refreshing

*Steps to reproduce*
1. Go to TIP IN transaction from the transaction list
2. Refresh page

Image
Image

martinkaintas:
closes #2438

### DIFF
--- a/src/popup/components/TransactionTagList.vue
+++ b/src/popup/components/TransactionTagList.vue
@@ -110,7 +110,13 @@ export default defineComponent({
             ? t('transaction.dexType.pool')
             : t('common.swap'),
         );
-      } else if (props.transaction.claim) {
+      } else if (
+        props.transaction.claim
+        || (
+          [tippingV1, tippingV2].includes(innerTx.value.contractId)
+          && innerTx.value.function === TX_FUNCTIONS.claim
+        )
+      ) {
         arr.push(
           t('pages.token-details.tip'),
           t('transaction.spendType.in'),
--- a/src/protocols/aeternity/views/TransactionDetails.vue
+++ b/src/protocols/aeternity/views/TransactionDetails.vue
@@ -164,6 +164,7 @@ import {
   onMounted,
   watch,
 } from 'vue';
+import { merge } from 'lodash-es';
 import { useRoute, useRouter } from 'vue-router';
 import { Encoded, Tag } from '@aeternity/aepp-sdk';
 import { IonContent, IonPage } from '@ionic/vue';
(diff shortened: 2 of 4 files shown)


src/popup/components/TransactionTagList.vue
```diff
@@ -110,7 +110,10 @@ export default defineComponent({
             ? t('transaction.dexType.pool')
             : t('common.swap'),
         );
-      } else if (props.transaction.claim) {
+      } else if (
+        props.transaction.claim
+        || innerTx.value.function === TX_FUNCTIONS.claim
```
CedrikNikita, line 115: Unfortunately, any contract can have a method named claim. This method is not completely correct.

martinkaintas: Correct, I've fixed this by also checking for the contract address

src/protocols/aeternity/views/TransactionDetails.vue
```diff
@@ -319,10 +320,11 @@ export default defineComponent({
     onMounted(async () => {
       let rawTransaction = getTransactionByHash(activeAccount.value.address, hash);
 
-      if (!rawTransaction || rawTransaction.incomplete) {
+      // Claim transactions have missing data that needs to be fetched from the middleware
+      if (!rawTransaction || rawTransaction.incomplete || rawTransaction.claim) {
         const middleware = await getMiddleware();
         try {
-          rawTransaction = await middleware.getTx(hash);
+          rawTransaction = merge(rawTransaction, await middleware.getTx(hash));
```
CedrikNikita, line 327: Why this `merge` is needed?

src/protocols/aeternity/views/TransactionDetails.vue
```diff
@@ -319,10 +320,11 @@ export default defineComponent({
     onMounted(async () => {
       let rawTransaction = getTransactionByHash(activeAccount.value.address, hash);
 
-      if (!rawTransaction || rawTransaction.incomplete) {
+      // Claim transactions have missing data that needs to be fetched from the middleware
+      if (!rawTransaction || rawTransaction.incomplete || rawTransaction.claim) {
```
CedrikNikita, line 324: Do we need to fetch `claim` transaction from the middleware, if we already fetched it from backend?

Liubov-crypto:
Fixed, LGTM

src/protocols/aeternity/views/TransactionDetails.vue
martinkaintas: Yes, the backend returns faulty and incomplete data

src/protocols/aeternity/views/TransactionDetails.vue
martinkaintas: Because in the case of `claim` transactions we want to merge data coming from the backend & the middleware so that we always have the same (correct and complete) transaction.

src/protocols/aeternity/views/TransactionDetails.vue
CedrikNikita: I see, incomplete transaction:
 - doesn't have `Gas price`
 - doesn't have `Gas used`
 - have a wrong timestamp

CedrikNikita (approved):

PR #2678, opened 2024-01-31
fix: eth account explorer links

martinkaintas:
Explorer links were wrong because for Aeternity/AeScan we can determine what endpoint to use and we were using `prepareUrlForHash` everywhere, which doesn't work for Ethereum and Bitcoin

### DIFF
--- a/src/protocols/aeternity/libs/AeScan.ts
+++ b/src/protocols/aeternity/libs/AeScan.ts
@@ -37,7 +37,7 @@ export class AeScan extends ProtocolExplorer {
     return (endpoint) ? `${this.explorerUrl}/${endpoint}/${hash}` : undefined;
   }
 
-  override prepareUrlForAccount(address: string) {
-    return `${this.explorerUrl}/accounts/${address}`;
+  override prepareUrlForAccount(addressOrName: string) {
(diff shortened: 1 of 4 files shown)


src/protocols/aeternity/libs/AeScan.ts
```diff
@@ -37,7 +37,7 @@ export class AeScan extends ProtocolExplorer {
     return (endpoint) ? `${this.explorerUrl}/${endpoint}/${hash}` : undefined;
   }
 
-  override prepareUrlForAccount(address: string) {
-    return `${this.explorerUrl}/accounts/${address}`;
+  override prepareUrlForAccount(addressOrName: string) {
+    return this.prepareUrlForHash(addressOrName) || `${this.explorerUrl}/accounts/${addressOrName}`;
```
CedrikNikita, line 41: When do we call this function with a name?

martinkaintas: In some cases `recipientId` ([here](https://github.com/superhero-com/superhero-wallet/pull/2678/files#diff-7555767fde85bbc840175b6d9e1eaa7e9459580fe4cc884750809f322c9986e2R203))  is a name address (`nm_123`)

image

CedrikNikita (approved):

Liubov-crypto:
Tested on testnet, LGTM

PR #2690, opened 2024-02-06
feat: support signing messages encoded as hex

### LINKED ISSUE #2688: Support signing messages encoded as hex string
**Is your feature request related to a problem?**
When invoking a deeplink to sign a message, the message is expected to be encoded as `utf8`. Example deeplink:

```
https://wallet.superhero.com/sign-message?message=<message>&x-success=<success-url>&x-cancel=<cancel-url>
```

In our use-case, the message is in bytes, usually represented as a hex-string. In the `messageToHash` method in the aepp-sdk, which is used by `signMessage`, the buffer is created by using the `utf8` encoding. See [here](https://github.com/aeternity/aepp-sdk-js/blob/428ebe63feac93c000d3303c12fb7b043ff04c8b/src/utils/crypto.ts#L165).

We have considered simply converting the bytes to `utf8`, so the correct message would be signed, but it looks like the decoding will not result in the same output as the input was if the bytes are not valid utf8.

**Describe the solution you'd like**
An additional URL parameter that would allow for signing of bytes represented as a hex-string. Eg.

```
https://wallet.superhero.com/sign-message?encoding=hex&message=0x1234
```

**Describe alternatives you've considered**
As described, we considered using the `utf8` encoding, but that does not seem to work.

We cannot change how the verification of the signature is done because it needs to be in a certain format for it to be accepted.

martinkaintas:
Closes #2688

### DIFF
--- a/src/popup/pages/SignMessage.vue
+++ b/src/popup/pages/SignMessage.vue
@@ -18,11 +18,14 @@ export default defineComponent({
 
       try {
         const aeSdk = await getAeSdk();
-        const message = route.query.message?.toString();
+        const rawMessage = route.query.message?.toString();
+        const isHexEncodedMessage = !!rawMessage && route.query.encoding?.toString() === 'hex';
+        const message = isHexEncodedMessage ? Buffer.from(rawMessage, 'hex') : rawMessage;
+        const displayMessage = message?.toString();
         const { host, href } = callbackOrigin.value || {} as any;
 
         await openModal(MODAL_MESSAGE_SIGN, {
-          message,
+          message: displayMessage,
           app: {
             host,
             name: host,


src/popup/pages/SignMessage.vue
```diff
@@ -19,10 +20,12 @@ export default defineComponent({
       try {
         const aeSdk = await getAeSdk();
         const message = route.query.message?.toString();
+        const isHexEncodedMessage = message && route.query.encoding?.toString() === 'hex';
+        const displayMessage = isHexEncodedMessage ? hexToUtf8(message) : message;
```
davidyuk, line 24:
```suggestion
        const displayMessage = isHexEncodedMessage ? Buffer.from(message, 'hex').toString() : message;
```

src/popup/pages/SignMessage.vue
```diff
@@ -19,10 +20,12 @@ export default defineComponent({
...
         const { host, href } = callbackOrigin.value || {} as any;
 
         await openModal(MODAL_MESSAGE_SIGN, {
-          message,
+          message: displayMessage,
```
davidyuk, line 28: So, it changes only the message shown to the user? I think Andreas meant to sign string decoded from hex. Like,
```
https://wallet.superhero.com/sign-message?encoding=hex&message=41
```
should return the same as
```
https://wallet.superhero.com/sign-message?message=A
```
As I see, currently the first deeplink will sign exactly "41", not "A".

davidyuk (commented):
It needs to be tested somehow 🙏 

In general, I'm not sure that there is a good reason to extend this api. See my reply in issue

AndreasGassmann:
> I think Andreas meant to sign string decoded from hex.

Yes, that's correct. The current PR only changes the format of how the message is displayed. But as @davidyuk points out, we would need to be able to sign the bytes decoded from hex.

So basically here (https://github.com/aeternity/aepp-sdk-js/blob/428ebe63feac93c000d3303c12fb7b043ff04c8b/src/utils/crypto.ts#L164-L171), instead of 

```
const msg = Buffer.from(message, 'utf8');
```

it should be

```
const msg = Buffer.from(message, 'hex');
```

This would interpret the message in the form that we need.

For us it doesn't really matter which deeplink we use for that. If it's a security concern, maybe a new deeplink like `/signHexMessage` could be introduced with a different UI and an additional warning.

src/popup/pages/SignMessage.vue
martinkaintas: I've applied this change
@AndreasGassmann  please note that this means that values need to be cleaned from the `0x` prefix now

AndreasGassmann:
Thanks @martinkaintas for the latest update. I think that should work. I will test it ASAP.

The only assumption we are making is that 
```
const buf = Buffer.from(message, 'hex')
buf.equals(Buffer.from(buf.toString(), 'utf8'))
```

I think this is a reasonable assumption, I'm just a little confused why I found online that this only works if it's a valid `utf-8` string, otherwise it can lead to issues. But I'm couldn't find an explanation. So we'll do some testing, and if we don't find an issue I think we can go with this.

Thanks for taking this on so quickly!

src/popup/pages/SignMessage.vue
davidyuk: `0x` prefix is not needed because it is already stated in `encoding=hex`, also we can't detect `hex` by the presence of `0x` prefix because aepp may want to sign such a message as it is.

AndreasGassmann:
Sorry for the delayed response, but we now have everything set up and can confirm that this change does not work. The assumption I posted above is not true. Converting bytes to `utf8` and back will not always result in the same data.

Take a look at this example:

```javascript
const hexMessage = '2800d4697066733a2f2f516d574b505a726e3561584164326d414d5548575a7442614271577143665768374b4561427a61726a4a4239444500017fee360000000000c085b5b18d0100004074ecb18d01000080ee360000000000102700000000000000000000000000000000000000010010a5d4e80000000000000000000000010000000000000000000000000000000000440100002100000001000000dfea37d4892a3588124f357ad36eaf545905de18c510c02604bb23354fd2bded05bc65ff5fa2a3778b0bcf0c091ce81f4efb93777ac39b7e7c74e7ef453ad630';
const buf = Buffer.from(hexMessage, 'hex');
const convertedBuf = Buffer.from(buf.toString(), 'utf8');
console.log('is equal:', buf.equals(convertedBuf), buf.length, convertedBuf.length);
```

This will log

```
is equal false 222 290
```

I think the only way would be to actually decode the data as hex and then sign the resulting buffer (eg. the variable `buf` in my example). Would it be possible to add that?

@davidyuk referring to your input that it is not possible to add something like "sign data <...> for website.com". if you have a format in mind that could make the UX better, but still keep the underlying data in hex encoding, that's something we could consider (eg. adding another prefix that would be the domain name and timestamp, but then encode it as hex before signing).

martinkaintas:
Now it should work as expected.
If the message is hex encoded we will sign the `Buffer.from(rawMessage, 'hex')` and calling `toString()` only for the display message, which will look weird but I don't think there's something we can do about this except what @davidyuk suggested

AndreasGassmann:
@martinkaintas Perfect, the last implementation works! From our side, you could deploy this to prod. Thanks a lot!

src/popup/pages/SignMessage.vue
```diff
@@ -18,11 +18,14 @@ export default defineComponent({
 
       try {
         const aeSdk = await getAeSdk();
-        const message = route.query.message?.toString();
+        const rawMessage = route.query.message?.toString();
+        const isHexEncodedMessage = !!rawMessage && route.query.encoding?.toString() === 'hex';
+        const message = isHexEncodedMessage ? Buffer.from(rawMessage, 'hex') : rawMessage;
+        const displayMessage = message?.toString();
```
peronczyk, line 24: Tiny suggestion:
```suggestion
        const message = (isHexEncodedMessage ? Buffer.from(rawMessage, 'hex') : rawMessage).toString();
```
This way you can still pass `message` to `openModal`.

peronczyk (approved):

PR #2705, opened 2024-02-14
fix: eth coin precision

martinkaintas:


src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -136,7 +136,7 @@ export class EthereumAdapter extends BaseProtocolAdapter {
       protocol: PROTOCOLS.ethereum,
       contractId: this.coinContractId,
       symbol: this.coinSymbol,
-      decimals: this.getAmountPrecision(),
+      decimals: this.coinPrecision,
```
CedrikNikita, line 139: What is the `coinPrecision`?

CedrikNikita (approved):

PR #2708, opened 2024-02-15
fix(transaction-list): show allow transaction correctly

### LINKED ISSUE #2696: Allow token transaction amount is shown without decimal shifting
**Describe the bug**
Allow token transaction amount is shown without decimal shifting.

**To Reproduce**
1. Go to Tx list 
2. Choose Allow ae token Tx
3. See error

**Expected behavior**
Should be shown properly

**Screenshots**

 - OS: MacOS
 - Browser FF
 - Version 122

**Additional context**

Image

CedrikNikita:
fixes #2696

### DIFF
--- a/src/composables/transactionTokens.ts
+++ b/src/composables/transactionTokens.ts
@@ -61,7 +61,7 @@ export function useTransactionTokens({
     const isReceived = direction === TX_DIRECTION.received;
     const adapter = ProtocolAdapterFactory.getAdapter(protocol);
     const amount = (isAllowance)
-      ? toShiftedBigNumber(innerTx.value?.fee || 0, -adapter.getAmountPrecision())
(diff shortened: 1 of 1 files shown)


src/composables/transactionTokens.ts
```diff
@@ -61,7 +61,7 @@ export function useTransactionTokens({
     const isReceived = direction === TX_DIRECTION.received;
     const adapter = ProtocolAdapterFactory.getAdapter(protocol);
     const amount = (isAllowance)
-      ? toShiftedBigNumber(innerTx.value?.fee || 0, -adapter.getAmountPrecision())
+      ? toShiftedBigNumber(innerTx.value?.fee || 0, -adapter.coinPrecision)
```
martinkaintas, line 64: I think that 18 decimals is way to much for a displayed amount, is that really what we want? 
Or do we format this amount again in a later stage before showing it?

Liubov-crypto:
LGTM

PR #2711, opened 2024-02-15
fix: show fiat value for eth spend transactions

### LINKED ISSUE #2710: Show fiat amount for ETH spend transactions on transaction list
image

martinkaintas:
closes #2710

### DIFF
--- a/src/composables/transactionTokens.ts
+++ b/src/composables/transactionTokens.ts
@@ -5,8 +5,8 @@ import type {
   ITransaction,
   TxFunctionParsed,
 } from '@/types';
-import { PROTOCOLS, TX_DIRECTION } from '@/constants';
-import { toShiftedBigNumber } from '@/utils';
+import { PROTOCOLS, TX_DIRECTION, ASSET_TYPES } from '@/constants';
+import { toShiftedBigNumber, isCoin } from '@/utils';
 import { ProtocolAdapterFactory } from '@/lib/ProtocolAdapterFactory';
 import { AE_SYMBOL } from '@/protocols/aeternity/config';
 import {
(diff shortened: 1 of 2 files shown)


src/composables/transactionTokens.ts
```diff
@@ -66,12 +66,18 @@ export function useTransactionTokens({
 
     const symbol = isAllowance ? AE_SYMBOL : getTxAssetSymbol(transaction);
     const token = protocolTokens[transaction.tx.contractId];
+    const isAe = isAllowance || (symbol === AE_SYMBOL && !isTransactionAex9(transaction));
```
peronczyk, line 69: I strongly suggest to avoid any checks for specific protocol. Maybe the adapter could have a method `isTransactionCoin` and the logic for protocol specific check could be moved there?

src/popup/components/TransactionListItem.vue
```diff
@@ -154,19 +154,19 @@ export default defineComponent({
     });
 
     const fiatAmount = computed(() => {
-      const aeToken = tokens.value?.find((t) => t?.isAe);
-      if (!aeToken || isErrorTransaction.value || isDexPool.value) {
+      const protocolCoin = tokens.value?.find(({ assetType }) => assetType === ASSET_TYPES.coin);
```
peronczyk, line 157: This check looks off for me. `tokens` should hold the list of tokens. Not coins. So the test `assetType === ASSET_TYPES.coin` for me is a workaround or broken logic.

peronczyk (changes requested):

src/composables/transactionTokens.ts
martinkaintas: `isAe` seems to be heavily used. Even if the check is moved to the adapter we still need `isAe` in order to add it to the resolved token

martinkaintas: I created a task for refactoring this #2722

src/popup/components/TransactionListItem.vue
martinkaintas: `tokens` should probably be renamed to `assets` since it seems that it was always holding all kinds of assets and not just tokens

peronczyk: Good point: we need a task that would refactor the `useTransactionTokens` to `useTransactionAssets`. Can you create one?

Liubov-crypto:
LGTM. 

One question: it shows also Link transactions as ETH for a short time period, then it updating and show proper token. Is it probably due to the reason that fee is paid in ETH? 

eth


link

martinkaintas:
> One question: it shows also Link transactions as ETH for a short time period, then it updating and show proper token. Is it probably due to the reason that fee is paid in ETH?

The reason seems to be the delay we have when fetching the tokens, at first we don't have tokens so it cannot recognize the tokens until we load the available tokens

PR #2717, opened 2024-02-16
fix: do not call cordova mnemonic migration on extension

### LINKED ISSUE #2713: Background page synchronization bug
**On develop**
It seems that background page unaware of the current mnemonic (or maybe the whole state). This results in popup being unauthorized, and trying to authorize itself. 
extension Synchonization bug

martinkaintas:
closes #2713 
Even though the actual migration was not being called because of the inner if check, it was still causing issues with the extension

### DIFF
--- a/src/composables/accounts.ts
+++ b/src/composables/accounts.ts
@@ -14,7 +14,8 @@ import {
   PROTOCOL_AETERNITY,
   PROTOCOLS,
   STORAGE_KEYS,
-  RUNNING_IN_TESTS,
+  IS_IOS,
+  IS_MOBILE_APP,
 } from '@/constants';
 import {
   prepareAccountSelectOptions,
@@ -39,11 +40,9 @@ const mnemonic = useStorageRef<string>(
   STORAGE_KEYS.mnemonic,
   {
     backgroundSync: true,
-    migrations: RUNNING_IN_TESTS ? [
+    migrations: [
+      ...((IS_IOS && IS_MOBILE_APP) ? [migrateMnemonicCordovaToIonic] : []),
       migrateMnemonicVuexToComposable,
-    ] : [
-      migrateMnemonicVuexToComposable,
-      migrateMnemonicCordovaToIonic,
     ],
   },
 );


CedrikNikita (approved):

src/composables/accounts.ts
```diff
@@ -39,7 +40,7 @@ const mnemonic = useStorageRef<string>(
   STORAGE_KEYS.mnemonic,
   {
     backgroundSync: true,
-    migrations: RUNNING_IN_TESTS ? [
+    migrations: RUNNING_IN_TESTS || IS_EXTENSION ? [
```
peronczyk, line 43: It's OK, but what do you think about making the 2 arrays as 1 (less quirky way)? I mean instead choosing between 2 arrays that has some duplicated items just:

```
    migrations: [
      migrateMnemonicVuexToComposable,
      (!RUNNING_IN_TESTS && !IS_EXTENSION) && migrateMnemonicCordovaToIonic,
    ].filter(excludeFalsy),
```

peronczyk (approved):

src/composables/accounts.ts
```diff
@@ -39,11 +40,9 @@ const mnemonic = useStorageRef<string>(
   STORAGE_KEYS.mnemonic,
   {
     backgroundSync: true,
-    migrations: RUNNING_IN_TESTS ? [
-      migrateMnemonicVuexToComposable,
-    ] : [
+    migrations: [
+      ...(!(RUNNING_IN_TESTS || IS_EXTENSION) ? [migrateMnemonicCordovaToIonic] : []),
```
CedrikNikita, line 44: I think it is suppose to be run only on previously Cordova based platforms.

CedrikNikita (changes requested):
This code is not working for an reverse iframe scenario.

martinkaintas:
> It's working in Chrome but doesn't work in FF:

You need to allow the extension to run on the DEX

CedrikNikita (approved):

PR #2718, opened 2024-02-16
fix: multisig transaction list is displaying partly

martinkaintas:
missed in #2618 👀


CedrikNikita (approved):
Code changes look reasonable.

peronczyk (approved):

PR #2730, opened 2024-02-20
fix: back to top should not add height to page if not scrolled

### LINKED ISSUE #2727: Back to top is inside every modal and making it bigger
image

martinkaintas:
closes #2727


Liubov-crypto:
LGTM

PR #2732, opened 2024-02-20
Refactor/allow optional methods in base protocol adapter

### LINKED ISSUE #2721: Stop using placeholder methods and set methods as not required on the BaseProtocolAdapter

martinkaintas:
closes #2721

### DIFF
--- a/src/protocols/BaseProtocolAdapter.ts
+++ b/src/protocols/BaseProtocolAdapter.ts
@@ -108,26 +108,27 @@ export abstract class BaseProtocolAdapter {
     options: Record<string, any>,
   ): Promise<any>;
 
-  abstract fetchAvailableTokens(): Promise<IToken[] | null>;
+  fetchAvailableTokens?(): Promise<IToken[] | null>;
 
-  abstract fetchAccountTokenBalances(address: AccountAddress): Promise<ITokenBalance[] | null>;
+  fetchAccountTokenBalances?(address: AccountAddress): Promise<ITokenBalance[] | null>;
 
-  abstract transferToken(
+  transferToken?(
     amount: number,
     recipient: string,
     contractId: AssetContractId,
     options: Record<string, any>,
   ): Promise<ITransferResponse | undefined>;
 
-  abstract fetchTokenInfo(contractId: AssetContractId): Promise<IToken | undefined>;
+  fetchTokenInfo?(contractId: AssetContractId): Promise<IToken | undefined>;
 
   abstract fetchBalance(address: AccountAddress): Promise<string>;
 
-  abstract fetchTransactionByHash(hash: string, transactionOwner?: AccountAddress): Promise<any>;
+  abstract fetchTransactionByHash(
+    hash: string,
+    transactionOwner?: AccountAddress,
+  ): Promise<ITransaction>;
 
-  abstract fetchPendingTransactions(
-    address: AccountAddress,
-  ): Promise<ITransaction[]>;
+  fetchPendingTransactions?(address: AccountAddress): Promise<ITransaction[]>;
 
   /**
    * Fetches all asset transactions (Both Coin AND Token) for an account.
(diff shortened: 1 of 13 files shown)


src/protocols/BaseProtocolAdapter.ts
```diff
@@ -102,32 +102,30 @@ export abstract class BaseProtocolAdapter {
    */
   abstract discoverLastUsedAccountIndex(seed: Uint8Array): Promise<number>;
 
-  abstract constructAndSignTx(
+  constructAndSignTx?(
```
peronczyk, line 105: I think this should be required for the adapters and we should align the AE adapter to use this method. What do you think @martinkaintas @CedrikNikita ?

src/protocols/BaseProtocolAdapter.ts
```diff
@@ -102,32 +102,30 @@ export abstract class BaseProtocolAdapter {
...
     options: Record<string, any>,
   ): Promise<ITransferResponse | undefined>;
 
-  abstract fetchTokenInfo(contractId: AssetContractId): Promise<IToken | undefined>;
+  fetchTokenInfo?(contractId: AssetContractId): Promise<IToken | undefined>;
 
   abstract fetchBalance(address: AccountAddress): Promise<string>;
 
-  abstract fetchTransactionByHash(hash: string, transactionOwner?: AccountAddress): Promise<any>;
+  fetchTransactionByHash?(hash: string, transactionOwner?: AccountAddress): Promise<any>;
```
peronczyk, line 126: I think this also should be required.

src/protocols/BaseProtocolAdapter.ts
martinkaintas: This is only used by the Bitcoin adapter

src/protocols/BaseProtocolAdapter.ts
```diff
@@ -156,5 +154,5 @@ export abstract class BaseProtocolAdapter {
     options: Record<string, any>,
   ): Promise<ITransferResponse>;
 
-  abstract waitTransactionMined(hash: string): Promise<any>;
+  waitTransactionMined?(hash: string): Promise<any>;
```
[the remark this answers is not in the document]
CedrikNikita: It seems like a valuable function regarding our own logic of "custom"/temp pending transactions.

src/protocols/BaseProtocolAdapter.ts
CedrikNikita: Since we had a Dogecoin integration planned in future, that can be required (Dogecoin is a fork of Bitcoin).

martinkaintas: It is only being used by the BTC adapter to calculate byte size of the transaction. I don't think we need this to be required

peronczyk: > This is only used by the Bitcoin adapter

Yes, but maybe we should have the signing done on the adapter level? At least as a goal forced by this file.

martinkaintas: I don't see a clear refactoring path for this method in the `AeternityAdapter` 
I suggest that if we really want this it should be done in a separate PR

martinkaintas: new task: #2773

peronczyk (approved):

Liubov-crypto:
In general LGTM. 

I have a question: 

why pending BTC tx is appearing in Tx list with delay?  it has been showing in latest tx list immediately, but I see it in BTC account tx list with delay in 2-3 min. 


https://github.com/superhero-com/superhero-wallet/assets/69896204/26e50261-51bf-4868-9ee6-d200f23162cb

and in recipients wallet it shows this Tx as pending in BTC acc Tx list, however it's received 11 min ago...


https://github.com/superhero-com/superhero-wallet/assets/69896204/c00fbb4e-0c0d-41d3-9f67-a4adaf1def60

PR #2704, opened 2024-02-13
Update dependencies

CedrikNikita:
fixes `15` vulnerabilities

PR #2737, opened 2024-02-22
refactor: config commitlint more precisely

CedrikNikita:
missed in #2704


martinkaintas (approved):
Given that the branch name is outdated, LGTM

PR #2742, opened 2024-02-22
Fix several multisig issues

CedrikNikita:
fixes issues from eth-support branch

> 4. /multisig/coins/aeternity/details page is broken:
> 
> coin det

> 2. BTC/ETH transactions , info are visible in multisig if the last active account in the wallet was BTC/ETH:
> 
>  2024-02-22.1.30.16.mov 
> e1

also fixes
```
TypeError: Cannot read properties of undefined (reading 'contractId')
```

### DIFF
--- a/src/composables/accountAssetsList.ts
+++ b/src/composables/accountAssetsList.ts
@@ -7,6 +7,7 @@ import type {
   IAsset,
   ICoin,
 } from '@/types';
+import { PROTOCOLS } from '@/constants';
 import { ProtocolAdapterFactory } from '@/lib/ProtocolAdapterFactory';
 import { useCurrencies } from './currencies';
 import { useFungibleTokens } from './fungibleTokens';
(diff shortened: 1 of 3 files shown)


martinkaintas (approved):

Liubov-crypto:
Still reproducible for me but in a different way: 

GET
https://api.ethplorer.io/getAddressInfo/0x6e503e467AE5A50A135AECBf03e93E97D7B25461?apiKey=freekey
[HTTP/2 429  30ms]


https://github.com/superhero-com/superhero-wallet/assets/69896204/7ef8313b-689e-4998-abd0-e8945e8cef5c


exp



https://github.com/superhero-com/superhero-wallet/assets/69896204/0d64962b-75c6-4005-9220-bc72e5776cbc

Liubov-crypto:
BTC transactions  are visible in multisig if the last active account in the wallet was BTC:


https://github.com/superhero-com/superhero-wallet/assets/69896204/b5101a4a-6356-407e-bb62-4d56fbad1e26

btc

multi

src/composables/accountAssetsList.ts
```diff
@@ -50,12 +51,18 @@ export function useAccountAssetsList({
     return currentBalance || new BigNumber(0);
   });
 
+  const accountProtocol = computed(
+    () => isMultisig ? PROTOCOLS.aeternity : activeAccount.value.protocol,
+  );
```
peronczyk, line 56: I suggest to create a computed property that would return `activeAccount` or `multisigActiveAccount` based on the `isMultisig` property. Then take the `protocol` from this computed property instead.

This would allow to update lines 59-61 with this computed property.

peronczyk (changes requested):

src/composables/accountAssetsList.ts
CedrikNikita: `IMultisigAccount` type doesn't have `procotol` inside

peronczyk: My bad. Skip my comment on this.

peronczyk (approved):

PR #2745, opened 2024-02-23
fix: eth transaction list displaying partly

martinkaintas:
missed in #2618  as it was done on develop branch

### DIFF
--- a/src/protocols/ethereum/views/AccountDetails.vue
+++ b/src/protocols/ethereum/views/AccountDetails.vue
@@ -1,7 +1,10 @@
 <template>
   <IonPage>
     <IonContent class="account-ion-content">
-      <AccountDetailsBase class="account-details">
+      <AccountDetailsBase
+        v-if="pageDidEnter"
+        class="account-details"
+      >
         <template #navigation>
           <AccountDetailsNavigation
             :route-names="[
@@ -31,6 +34,9 @@ export default defineComponent({
     IonPage,
     IonContent,
     AccountDetailsNavigation,
+  },
+  props: {
+    pageDidEnter: Boolean,
   },
   setup() {
     return {
--- a/src/popup/components/ProtocolSpecificView.vue
+++ b/src/popup/components/ProtocolSpecificView.vue
@@ -2,7 +2,8 @@
   <Component
     :is="componentToDisplay"
     v-if="viewComponentName"
-    :ionic-lifecycle-status="ionicLifecycleStatus"
+    :page-did-enter="pageDidEnter"
+    :page-will-enter="pageWillEnter"
   />
   <div v-else>
     <InfoBox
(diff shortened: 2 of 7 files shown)


src/protocols/ethereum/views/AccountDetails.vue
```diff
@@ -16,13 +19,19 @@
...
+  ref,
+  watch,
+} from 'vue';
 import { IonContent, IonPage } from '@ionic/vue';
 import { PROTOCOL_VIEW_ACCOUNT_DETAILS } from '@/constants';
 import { ROUTE_ACCOUNT_DETAILS, ROUTE_ACCOUNT_DETAILS_ASSETS } from '@/popup/router/routeNames';
 
 import AccountDetailsBase from '@/popup/components/AccountDetailsBase.vue';
 import AccountDetailsNavigation from '@/popup/components/AccountDetailsNavigation.vue';
+import { IonicLifecycleStatus } from '@/types';
```
peronczyk, line 34: Please put this line before the constants import.

martinkaintas: Damn auto-imports :S

src/protocols/ethereum/views/AccountDetails.vue
```diff
@@ -32,10 +41,24 @@ export default defineComponent({
     IonContent,
     AccountDetailsNavigation,
   },
-  setup() {
+  props: {
+    ionicLifecycleStatus: { type: String as PropType<IonicLifecycleStatus>, default: null },
```
peronczyk, line 45: What do you think about replacing this `ionicLifecycleStatus` with a bool prop? Is this possible? I see we have 4 possible values for this, but isn't 2 of them truthy and 2 falsy?

Or maybe having 2 separate bool props: `pageDidEnter` and `pageWillEnter`.

The benefit of boolean prop is that we can avoid importing and using the `IonicLifecycleStatus` type. Also we could stop using the `'didEnter'` string values. Instead the prop could be used directly in the template to display or hide the elements.

peronczyk (changes requested):

src/protocols/ethereum/views/AccountDetails.vue
```diff
@@ -16,7 +19,9 @@
 </template>
 
 <script lang="ts">
-import { defineComponent } from 'vue';
+import {
+  defineComponent,
+} from 'vue';
```
peronczyk, line 24: No need for this change.

Liubov-crypto:
Can't get my ETH Tx list, probably ethereum testnet issue: 

уер

Liubov-crypto:
Finally I managed to get it, in general LGTM 

But ETH transactions are visible in LINK TXs list and vice versa:
tok det

err1

PR #2748, opened 2024-02-26
fix: redirect to 404 when tx cant be fetched

martinkaintas:
fixes:
> 8. After sending ERC-20 tokens Txs I clicked on Tx to see Tx details, but the page hang out and I find these issues in console:

from #2736 

If we reach the API limit and can't fetch details for a tx we're redirected to the 404 page. 
This wasn't working properly because we didn't have a specific 404 page path and were trying to display the 404 page at `/`
Now we redirect unknown pages to `/page-not-found`
Also disables the loader before redirecting

### DIFF
--- a/src/popup/router/routes.ts
+++ b/src/popup/router/routes.ts
@@ -752,8 +752,8 @@ export const routes: WalletAppRouteConfig[] = [
     },
   },
   {
-    path: '/notifications',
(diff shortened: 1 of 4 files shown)


Liubov-crypto:
Fixed, LGTM

CedrikNikita (approved):

PR #2761, opened 2024-02-27
ci: split test and build workflows

CedrikNikita:
The goal of this PR is to test and build PRs faster.


martinkaintas (approved):
Finally 🙏

PR #2744, opened 2024-02-23
fix: remove swap button from non ae assets

martinkaintas:
closes #2739

PR #2762, opened 2024-02-27
fix(multisig): show correct fiat value for assets

CedrikNikita:
missed in #2742, because of #2744


martinkaintas (approved):

peronczyk (approved):

Liubov-crypto:
The issue with value has been fixed, LGTM

PR #2772, opened 2024-02-29
fix: do not prohibit sending to same address

CedrikNikita:


src/utils/transferSend.ts
```diff
@@ -1,24 +1,20 @@
...
 import { tg } from '@/popup/plugins/i18n';
-import { BTC_SYMBOL } from '@/protocols/bitcoin/config';
 
 // TODO: in the future we might rely on the error codes instead
-const nonAlphabeticCharsRegExp = /[^\p{L}]/gu;
-
 const WARNING_RULES_WORDING = [
-  tg('validation.addressNotSameAs').replace(nonAlphabeticCharsRegExp, ''),
-  tg('validation.maxValueVault').replace(nonAlphabeticCharsRegExp, ''),
+  tg('validation.addressNotSameAs', ['//']),
```
CedrikNikita, line 6: We may not introduce any particular splitter and use ' ' instead, or the splitter may be even more specific, but in general this is a middle option.

Liubov-crypto:
LGTM

PR #2776, opened 2024-02-29
Fix: Latest multisig transaction updating on account change

peronczyk:


src/popup/App.vue
```diff
@@ -127,11 +128,13 @@ export default defineComponent({
     const { modalsOpen } = useModals();
     const { isLoggedIn } = useAccounts();
     const { addWalletNotification } = useNotifications();
-    const { loadCoinsData } = useCurrencies({ withoutPolling: true });
+    const { loadCoinsData } = useCurrencies({ pollingDisabled: true });
     const { restoreLanguage } = useLanguages();
     const { restore: restoreTransferSendForm } = useTransferSendHandler();
     const { multisigAccounts } = useMultisigAccounts({ pollingDisabled: true });
 
+    usePendingMultisigTransaction(); // Ensure the persistency of the watchers
```
CedrikNikita, line 136: Is this actually the only way to actually remain watching? Because this composable is used within the `PendingMultisigTransactionCard` already, and it is presented in `/account`.

Liubov-crypto:
LGTM

src/popup/App.vue
peronczyk: This is the only solution I came up. The main reason is that the `watch` function behaves differently based on the composable's scope
1. If the composable is fired by any **JS/TS file** (including another composable) the watcher is persistent. So for every JS/TS file new instance of the watcher is created. If we don't block running multiple watchers the impact on app's performance is quite big.
2. If the composable is fired within a Vue component the watcher is bind to the component so if the component is unmounted the watcher is also destroyed. 

These two possibilities makes the app slightly unpredictable because depending on what kind of file fires the composable for the first time we can't be sure what will happen with the watcher. For most of the composables the 1st option is applicable so the `if (!composableInitialized)` makes 100% sense. The `pendingMultisigTransaction` is fired for the 1st time by the Vue component in the case you described @CedrikNikita so it's removed when we leave the page :)

The line we are discussing right now makes sure that the composable is first run in scope that would not remove the watcher.

If I will remove the `if (!composableInitialized)` from the composable the watcher is fired multiple times and causes multiple API calls.

If you think there is a better solution please share.

CedrikNikita: This only means that there could be more places like that. Shall we now move all composables with global watchers (that can be initialized on vue scope) to this file?

src/composables/pendingMultisigTransaction.ts
```diff
@@ -183,28 +183,43 @@ export function usePendingMultisigTransaction() {
...
-    watch(
-      activeMultisigAccount,
-      (newValue, oldValue) => {
-        if (!isEqual(newValue, oldValue)) {
-          assignPendingMultisigTx();
-
-          if (!activeMultisigAccount.value?.txHash && !latestMultisigAccountTransaction.value) {
-            fetchLatestMultisigAccountTransaction();
+    // Create persistent effect scope to avoid watcher being disposed
+    effectScope(true).run(() => {
```
peronczyk, line 197: Can you take a look on this solution @CedrikNikita and @martinkaintas ?
This is using this concept: https://github.com/vuejs/rfcs/blob/master/active-rfcs/0041-reactivity-effect-scope.md

The downside of this solution is that we need to initiate watcher dependencies within this custom scope.

src/popup/App.vue
peronczyk: I think this would be rather bad to just initiate all composables with watchers in the App.vue file.
I came up with different approach to the problem. Please see this comment: https://github.com/superhero-com/superhero-wallet/pull/2776/files#r1514116021 and the code related to it.

src/composables/pendingMultisigTransaction.ts
martinkaintas: LGTM, I would also like to see if we can refactor our shared composables to use `effectScope` like [this example](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0041-reactivity-effect-scope.md#example-a-shared-composable)

PR #2777, opened 2024-03-01
fix: validate several props and emits correctly

CedrikNikita:
fixes emit and prop validation issues
```
logger.ts:50 Invalid prop: custom validator check failed for prop "iconSize". at <AssetIcon>
at <Tokens>

[Vue warn]: Component emitted event "asset-selected" but it is neither declared in the emits option nor as an "onAsset-selected" prop.
```
Also fixes this issue on custom network creation
```
veeValidate.ts:114 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'toString')
```

### DIFF
--- a/src/popup/components/AssetIcon.vue
+++ b/src/popup/components/AssetIcon.vue
@@ -26,7 +26,7 @@ import EthereumIcon from '@/icons/coin/ethereum.svg?vue-component';
 import LexonTokenIcon from '@/icons/tokens/ct_xtk8rSz9suPb6D6VLquyfVji25FcnFRDjn3dnn5mmvHsPiESt.svg?vue-component';
 import ChainlinkTokenIcon from '@/icons/tokens/0x779877a7b0d9e8603169ddbd7836e478b4624789.svg?vue-component';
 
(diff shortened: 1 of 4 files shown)


src/popup/components/AssetIcon.vue
```diff
@@ -26,7 +26,7 @@ import EthereumIcon from '@/icons/coin/ethereum.svg?vue-component';
 import LexonTokenIcon from '@/icons/tokens/ct_xtk8rSz9suPb6D6VLquyfVji25FcnFRDjn3dnn5mmvHsPiESt.svg?vue-component';
 import ChainlinkTokenIcon from '@/icons/tokens/0x779877a7b0d9e8603169ddbd7836e478b4624789.svg?vue-component';
 
-const SIZES = [ICON_SIZES.rg, ICON_SIZES.md, ICON_SIZES.lg] as const;
+const SIZES = [ICON_SIZES.sm, ICON_SIZES.rg, ICON_SIZES.lg, ICON_SIZES.xxl] as const;
```
martinkaintas, line 29: Default icon size is `md`. I suggest to either change that, or keep `ICON_SIZED.md` in the array

martinkaintas (approved):

PR #2782, opened 2024-03-03
fix: use correct nonce to dry-run contract calls

davidyuk:
This PR is supported by the Æternity Foundation


Liubov-crypto:
LGTM

CedrikNikita (approved):

PR #2786, opened 2024-03-04
refactor(ethereum): avoid showing wrong token info

CedrikNikita:
https://github.com/EverexIO/Ethplorer/wiki/Ethplorer-API#freekey-limits


src/protocols/ethereum/libs/EthplorerService.ts
```diff
@@ -14,21 +14,38 @@ const TOO_MANY_REQUESTS_ERROR_CODE = 429;
 /**
  * @link https://github.com/EverexIO/Ethplorer/wiki/Ethplorer-API
  */
+
+let lastCallTime: number;
+
```
peronczyk, line 19: This should be placed before the comment for the service class.

peronczyk (changes requested):

peronczyk (approved):

PR #2797, opened 2024-03-05
fix(aeternity): show pending and tip withdrawn transactions

### LINKED ISSUE #2796: Pending and Tipwithdrawn transactions are no longer present in transaction list

CedrikNikita:
fixes #2796

### DIFF
--- a/src/composables/transactionList.ts
+++ b/src/composables/transactionList.ts
@@ -106,6 +106,8 @@ export function useTransactionList({
       const {
         paginationParams,
         regularTransactions,
+        pendingTransactions,
+        tipWithdrawnTransactions,
       } = await fetchTransactions(state.value.nextPagePaginationParams);
 
       if (accountAddress !== state.value.accountAddress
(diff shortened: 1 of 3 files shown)


Liubov-crypto:
In general LGTM, I can see my pending Tx in Txs list.

Liubov-crypto:
LGTM

src/protocols/aeternity/libs/AeternityAdapter.ts
```diff
@@ -304,6 +304,8 @@ export class AeternityAdapter extends BaseProtocolAdapter {
         ...transaction,
         pending: true,
         protocol: PROTOCOLS.aeternity,
+        microTime: Date.now(),
```
martinkaintas, line 307: Why do we need to add `microTime` to pending transactions?

CedrikNikita: This was our previous approach to show that it was done some time ago.

src/composables/transactionList.ts
```diff
@@ -156,13 +171,20 @@ export function useTransactionList({
           return;
         }
 
+        const filteredLoadedTransactions = (state.value.transactionsLoaded as any)
+          .filter(({ claim = false, pending = false }) => !claim && !pending);
```
peronczyk, line 175: There is no need to set default values for claim and pending.

peronczyk (approved):

src/protocols/aeternity/libs/AeternityAdapter.ts
CedrikNikita: But it seems that `microTime` for the custom pending transaction is not being set :(. And the whole logic is not working.

peronczyk: So maybe this needs another fix?

CedrikNikita: Will do it within this PR.

onvisions: 1. We should display the timestamp in the Latest transactions widget on the wallet Dashboard as we currently do because this way the user can know approximately how much time the transaction is in a pending state:
Pending-3

2. We should not display the timestamp in the Transactions tab list under Account details because we have limited UI space there and because the user can access the single transaction details screen to see the exact timestamp.

3. Add the timestamp for pending transaction on the Transactions details screen in case we can get it (the transaction is initiated by the user by the currently active wallet account): 
Pending-5


@CedrikNikita what do you think?

CedrikNikita: I suggest to move last part to another ticket and apply current fix.

PR #2799, opened 2024-03-05
fix: show correct details for pending ethereum transactions

### LINKED ISSUE #2789: ERC-20 Tx details are displaying wrong info
**Describe the bug**
ERC-20 Tx details are displaying wrong info

**To Reproduce**
1. Send ERC-20 
2. Click on Tx details when Tx is in pending status
3. See error

**Expected behavior**

**Screenshots**

 - OS: MacOS
 - Browser FF
 - Version 122

**Additional context**
Reproducible on the current dev branch. 

Image

It should be: 

Image

martinkaintas:
fixed #2789 

Also removes timestamp and block number from pending transactions 
(`getBlock` will return a block even if `blockHash` is missing)

### DIFF
--- a/src/protocols/ethereum/views/TransactionDetails.vue
+++ b/src/protocols/ethereum/views/TransactionDetails.vue
@@ -69,7 +69,7 @@ import { IonContent, IonPage } from '@ionic/vue';
 
 import type { ITokenResolved, ITransaction } from '@/types';
 import { TX_DIRECTION, PROTOCOLS } from '@/constants';
-import { useFungibleTokens, useUi } from '@/composables';
+import { useFungibleTokens, useLatestTransactionList, useUi } from '@/composables';
 import { ProtocolAdapterFactory } from '@/lib/ProtocolAdapterFactory';
 
 import { ETH_SYMBOL } from '@/protocols/ethereum/config';
@@ -95,6 +95,7 @@ export default defineComponent({
     const route = useRoute();
     const { setLoaderVisible } = useUi();
     const { getTxAmountTotal, getTxAssetSymbol } = useFungibleTokens();
+    const { accountsTransactionsPending } = useLatestTransactionList();
 
     const hash = route.params.hash as string;
     const transactionOwner = route.params.transactionOwner as string;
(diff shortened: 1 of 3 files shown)


src/protocols/ethereum/views/TransactionDetails.vue
```diff
@@ -136,6 +137,18 @@ export default defineComponent({
     );
 
     onMounted(async () => {
+      // If the transaction is a token transfer and it is still pending use the pending transaction
+      // because API wont be synced yet
+      // pendingTransaction will have the most accurate data at this point
+      if (accountsTransactionsPending.value[transactionOwner]) {
+        const pendingTransaction: ITransaction = accountsTransactionsPending.value[transactionOwner]
+          .find((tx) => tx.hash === hash) as ITransaction;
+        if (pendingTransaction && pendingTransaction.tx.contractId !== adapter.coinContractId) {
```
peronczyk, line 146:
```suggestion
        if (pendingTransaction?.tx?.contractId !== adapter.coinContractId) {
```

src/protocols/ethereum/views/TransactionDetails.vue
```diff
@@ -136,6 +137,18 @@ export default defineComponent({
...
+          transaction.value = pendingTransaction;
+          return;
+        }
+      }
```
peronczyk, line 150: In my opinion this block of code is rather superfluous and I think using `return` within the `onMounted` callback is a bad practice. Consider this approach:

```suggestion
      const pendingTransaction = (accountsTransactionsPending.value[transactionOwner] || []).find((tx) => tx.hash === hash);
      if (pendingTransaction) {
        transaction.value = pendingTransaction;
      } else {
        // The logic of fetching the transaction data
```

peronczyk (changes requested):

src/protocols/ethereum/views/TransactionDetails.vue
martinkaintas: `if (accountsTransactionsPending.value[transactionOwner])`
This statement is used to check if the pending transactions composable is initialized. Without it, if user refreshes on a tx details page he might get stuck loading.

martinkaintas: Also pending transactions do not have correct fee information, thus if the transaction isn't a contract call I think it makes sense to fetch the correct data from the node

peronczyk (approved):

Liubov-crypto:
Fixed, LGTM

PR #2808, opened 2024-03-06
fix: cache loaded transactions if state did not change

### LINKED ISSUE #2798: sh-wallet_transactions-loaded are always cleared

martinkaintas:
closes #2798

### DIFF
--- a/src/composables/transactionList.ts
+++ b/src/composables/transactionList.ts
@@ -71,15 +71,21 @@ export function useTransactionList({
   const { activeNetwork } = useNetworks();
   const { accountsTransactionsPending } = useLatestTransactionList();
 
-  state.value = {
-    accountAddress,
-    assetContractId,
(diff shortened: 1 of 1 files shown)


Liubov-crypto:
In general LGTM. 

But I found if ERC-20 transaction is in pending  status it could be shown in ETH Tx list (maybe because fee is paid in ETH and Tx is not fully fetched) 


https://github.com/superhero-com/superhero-wallet/assets/69896204/9810d0ef-e297-4e8d-ab4a-4720c57d5e65

Also I can't see my old Tx in the Tx list, Nikita has mentioned it already: 


https://github.com/superhero-com/superhero-wallet/assets/69896204/4450eadc-8450-4388-b4d8-da5cd6f3078d

CedrikNikita (approved):

PR #2809, opened 2024-03-06
fix: load old transactions

### LINKED ISSUE #2810: If the selected transaction type does not populate the transaction list, the user will not be able to load old transactions
filteredDexTransactions
fungibleTokenTransactions

CedrikNikita:
fixes #2810

### DIFF
--- a/src/popup/components/TransactionList.vue
+++ b/src/popup/components/TransactionList.vue
@@ -51,6 +51,7 @@
 import {
   computed,
   defineComponent,
+  nextTick,
   onMounted,
   onUnmounted,
   PropType,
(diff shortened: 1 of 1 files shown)


Liubov-crypto:
LGTM

src/popup/components/TransactionList.vue
```diff
@@ -203,7 +205,11 @@ export default defineComponent({
       }
     }
 
-    function checkLoadMore() {
+    async function checkLoadMore() {
+      // sleep is required because viewportElement scrollHeight
+      // is not being changed fast enough, after filter is applied
+      await sleep(50);
```
peronczyk, line 211: Maybe using `await nextTick()` would solve the issue and better fit into the Vue/DOM environment?

peronczyk (approved):

PR #2819, opened 2024-03-07
feat: extract custom tokens from token balances

### LINKED ISSUE #2792: User's ERC-20 tokens that are not in the top 50, will not appear in the assets
image
image
image

martinkaintas:
closes #2792

### DIFF
--- a/src/composables/fungibleTokens.ts
+++ b/src/composables/fungibleTokens.ts
@@ -1,6 +1,6 @@
 /* eslint-disable no-param-reassign */
 
-import { watch } from 'vue';
+import { computed, watch } from 'vue';
 import BigNumber from 'bignumber.js';
 import { Encoding } from '@aeternity/aepp-sdk';
 import { toShiftedBigNumber } from '@/utils';
@@ -22,6 +22,7 @@ import { ProtocolAdapterFactory } from '@/lib/ProtocolAdapterFactory';
 import FungibleTokenFullInterfaceACI from '@/protocols/aeternity/aci/FungibleTokenFullInterfaceACI.json';
 import { aettosToAe, categorizeContractCallTxObject } from '@/protocols/aeternity/helpers';
 
+import { uniqBy } from 'lodash-es';
 import { useAccounts } from './accounts';
 import { useAeSdk } from './aeSdk';
 import { useTippingContracts } from './tippingContracts';
@@ -33,9 +34,10 @@ let composableInitialized = false;
 
 /**
  * List of all fungible tokens available on user's protocols.
+ * Does not include custom tokens extracted from user's token balances.
  * As this list is quite big (hundreds of items) it requires processing optimizations.
  */
-const tokensAvailable = useStorageRef<ProtocolRecord<AssetList>>(
+const defaultTokensAvailable = useStorageRef<ProtocolRecord<AssetList>>(
   {},
   STORAGE_KEYS.fungibleTokenList,
 );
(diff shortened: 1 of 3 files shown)


src/composables/fungibleTokens.ts
```diff
@@ -33,13 +34,23 @@ let composableInitialized = false;
...
 );
 
+/**
+ * List of custom fungible tokens available on user's protocols.
+ * Includes tokens from the user's account that are not on the main list.
+ */
+const customTokensAvailable = useStorageRef<ProtocolRecord<AssetList>>(
+  {},
+  STORAGE_KEYS.customFungibleTokenList,
+);
```
peronczyk, line 52: Maybe instead of storing this tokens there could be a computed property that takes the token balances and return list of tokens?
This way we wouldn't need any function that sets the state and there will be no need to store the data in the local storage.

src/composables/fungibleTokens.ts
```diff
@@ -74,6 +88,31 @@ export function useFungibleTokens() {
     return tokenBalances.value.filter((token) => token.address === address) || [];
   }
 
+  /**
+   * Extract unique tokens from the `tokenBalances` list and store them
+   * in the `customTokensAvailable` list.
+   */
+  function setCustomTokens() {
+    const uniqueTokens: IToken[] = uniqBy(tokenBalances.value, 'contractId')
```
peronczyk, line 96: Is it possible that the balance would return a list with duplicated tokens?

src/composables/fungibleTokens.ts
```diff
@@ -74,6 +88,31 @@ export function useFungibleTokens() {
...
+      .map((tokenBalance) => ({
+        contractId: tokenBalance.contractId,
+        protocol: tokenBalance.protocol,
+        name: tokenBalance?.name!,
+        symbol: tokenBalance?.symbol!,
+      }));
+
+    customTokensAvailable.value = uniqueTokens.reduce((customTokens, token) => {
```
peronczyk, line 104: Maybe instead of reduce you could use `objectFrom` ?

martinkaintas: I don't think it makes sense to use `Object.fromEntries` here as `uniqueTokens ` and `customTokensAvailable` have different structures.
Did you mean something else with `objectFrom`?

src/composables/fungibleTokens.ts
```diff
@@ -48,8 +50,36 @@ const tokenBalances = useStorageRef<ITokenBalance[]>(
...
+    return customTokens;
+  }, {} as typeof customTokensAvailable.value);
+});
+
 function getProtocolAvailableTokens(protocol: Protocol): AssetList {
-  return tokensAvailable.value[protocol] || {} as AssetList;
+  return {
+    ...customTokensAvailable.value[protocol],
+    ...tokensAvailable.value[protocol],
+  };
```
peronczyk, line 82: Instead of merging these two arrays each time the function is fired I suggest to:
1. Rename `tokensAvailable` to `defaultTokensAvailable`
2. Rename `customTokensAvailable` to `tokensAvailable` and merge the default ones with the   ones coming from the balances.

src/composables/fungibleTokens.ts
```diff
@@ -48,6 +50,33 @@ const tokenBalances = useStorageRef<ITokenBalance[]>(
...
+      symbol: tokenBalance?.symbol!,
+    }));
+
+  const customTokensAvailable = uniqueTokens.reduce((customTokens, token) => {
+    const { contractId, protocol } = token;
+    if (!customTokens[protocol]) {
+      customTokens[protocol] = {} as AssetList;
+    }
+    if (!defaultTokensAvailable?.value?.[protocol]?.[contractId]) {
+        customTokens[protocol]![contractId] = token;
```
peronczyk, line 72: Wrong indentation.

src/composables/fungibleTokens.ts
```diff
@@ -48,6 +50,33 @@ const tokenBalances = useStorageRef<ITokenBalance[]>(
...
+    }
+    return customTokens;
+  }, {} as typeof tokensAvailable.value);
+
+  return merge({}, customTokensAvailable, defaultTokensAvailable.value);
```
peronczyk, line 77: Why not `{ ...a, ...b }` ?

src/composables/fungibleTokens.ts
```diff
@@ -48,6 +50,33 @@ const tokenBalances = useStorageRef<ITokenBalance[]>(
   STORAGE_KEYS.fungibleTokenBalances,
 );
 
+/**
+ * List of custom fungible tokens available on user's protocols.
```
peronczyk, line 54: I suggest to update the comment as this is not a list of "custom" but "all" tokens.

peronczyk (approved):

Liubov-crypto:
LGTM

CedrikNikita:
For me tokens appearing slow in the Assets. I would assume that this is because of the slow Ethereum middleware, but amount of tokens is shown as `1 Token`.

Tokens appearing slow

Liubov-crypto:
This is quite fast for me, probably because I wait for the Tx list to load and then proceed to check my assets:

https://github.com/superhero-com/superhero-wallet/assets/69896204/d8f6c5e5-5570-4025-a5e0-fe778fc7c780

CedrikNikita (approved):

PR #2840, opened 2024-03-15
fix: ae network settings composable custom effect scope

peronczyk:


martinkaintas (approved):

src/composables/composablesHelpers.ts
```diff
@@ -1,11 +1,27 @@
...
 } from 'vue';
 import { useConnection } from './connection';
 import { useUi } from './ui';
 
+/**
+ * Creates a custom effect scope for a composable to avoid disposing watchers
+ * and computed properties when a Vue component is unmounted. The effect scope
+ * is only created the first time you run the composable and is reused from that point on.
+ */
+export function createCustomScopedComposable<T>(composableBody: () => T) {
```
CedrikNikita, line 15: I think we need to check other places where this new way of creating composables should be applied.

peronczyk: Sure, but this process would be really time consuming because it's like testing every feature from ground up. I suggest to have this PR as a new feature and gradually fix related issues with it.

CedrikNikita (approved):

PR #2844, opened 2024-03-18
fix: hide `contractId` for WAE tokens on TransactionSign modals

peronczyk:


martinkaintas (approved):
Code changes seem reasonable

PR #2870, opened 2024-03-21
Feat/secure login

### LINKED ISSUE #2824: PIN/Biometric Login Screens
- User shall be able to enable PIN/Biometric Login when creating/restoring a wallet.
- User shall be to enable/disable PIN/Biometric Login from Settings Page.
- A lock screen shall be prompted to the user according to X time of app inactivity.
- User shall be set the time that is required for locking application during inactivity app.

### LINKED ISSUE #2830: Secure Login settings page

martinkaintas:
closes #2824
closes #2830


src/popup/App.vue
```diff
@@ -229,6 +240,37 @@ export default defineComponent({
...
+    watch(
+      isAppActive,
+      async (isActive, wasActive) => {
+        if (!IS_MOBILE_APP) {
+          return;
+        }
+
+        // App went to background
+        if (wasActive && !isActive) {
+          lastTimeAppWasActive.value = new Date();
```
CedrikNikita, line 256: We can store `Date.now()`, and not call `getTime()` every time.

CedrikNikita (approved):

src/icons/lock.svg
peronczyk, line 1: This icon is not pixel perfect which makes the icon "blurry". @onvisions can you update the icon?

src/popup/router/routes.ts
```diff
@@ -442,6 +447,33 @@ export const routes: WalletAppRouteConfig[] = [
...
   },
+  {
+    path: '/more/settings/secure-login',
+    component: SecureLoginSettings,
+    name: ROUTE_SECURE_LOGIN_SETTINGS,
+    meta: {
+      title: 'secureLogin',
+      showHeaderNavigation: true,
+    },
+    beforeEnter: (to, from, next) => {
```
peronczyk, line 458: I think user should be able to open the secure login settings even if used device does not support it.
I strongly suggest to use composables in the route definitions.

peronczyk (changes requested):

src/popup/pages/Settings.vue
```diff
@@ -46,10 +52,19 @@
...
+  useUi,
+} from '@/composables';
+import {
+  ROUTE_NETWORK_SETTINGS,
+  ROUTE_PERMISSIONS_SETTINGS,
+  ROUTE_SECURE_LOGIN_SETTINGS,
+} from '@/popup/router/routeNames';
 
 import PanelItem from '@/popup/components/PanelItem.vue';
+import { IS_MOBILE_APP } from '@/constants';
```
peronczyk, line 67: Please move the constants import before composables import.

src/popup/pages/SecureLoginSettings.vue
```diff
@@ -0,0 +1,146 @@
...
+    padding-inline: var(--screen-padding-x);
+    margin-bottom: 28px;
+  }
+
+  .text-description {
+    padding: 16px 0;
+  }
+
+  .options-wrapper {
+    border-top: 1px solid rgba(255, 255, 255, 0.15);
```
peronczyk, line 113:
```suggestion
    border-top: 1px solid rgba($color-white, 0.15);
```

src/popup/pages/SecureLoginSettings.vue
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  .options-info {
+    @include mixins.flex(flex-start, flex-start, column);
```
peronczyk, line 117: I strongly suggest to avoid using this mixin as it gives no real value but makes the code less readable.

src/popup/pages/SecureLoginSettings.vue
```diff
@@ -0,0 +1,146 @@
...
+      setSecureLoginEnabled,
+      setSecureLoginTimeout,
+      AUTHENTICATION_TIMEOUTS,
+    };
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@use '../../styles/variables';
```
peronczyk, line 98:
```suggestion
@use '../../styles/variables' as *;
```

src/popup/pages/SecureLoginSettings.vue
peronczyk: Also I sugest to take a look on global styles applied to `hr` element (global.scss). This line is used in `NetworkForm.vue`.

src/popup/pages/SecureLoginSettings.vue
```diff
@@ -0,0 +1,146 @@
...
+      <div class="secure-login-settings">
+        <div class="top-wrapper">
+          <p class="text-description">
+            {{ $t('pages.secureLogin.description') }}
+          </p>
+
+          <SwitchButton
+            :label="$t('pages.secureLogin.enableSecureLogin')"
+            :model-value="isSecureLoginEnabled"
+            @update:modelValue="setSecureLoginEnabled"
```
peronczyk, line 13: Maybe it would be possible to use:
```suggestion
            v-model="isSecureLoginEnabled"
```

src/popup/components/Modals/SecureLogin.vue
```diff
@@ -0,0 +1,127 @@
...
+      margin-bottom: 4px;
+    }
+
+    .text {
+      @extend %face-sans-16-regular;
+      line-height: 24px;
+      color: rgba($color-white, 0.85);
+    }
+  }
+
```
peronczyk, line 125: This line can be removed :)

src/popup/components/Modals/SecureLogin.vue
```diff
@@ -0,0 +1,127 @@
...
+    from-bottom
+    full-screen
+    no-padding
+    centered
+    transparent
+    @close="handleClose"
+    @click="initAuthenticate"
+  >
+    <div
+      v-if="!isAuthenticated"
```
peronczyk, line 13: Consider displaying something in case user is Authenticated.

src/popup/components/Modals/EnableSecureLogin.vue
```diff
@@ -0,0 +1,141 @@
...
+    margin-top: 44px;
+    opacity: 0.5;
+  }
+
+  .info {
+    .title {
+      color: $color-white;
+      padding:16px 0;
+
+      @extend %face-sans-18-medium;
```
peronczyk, line 120: Please put the extends at the beginning of style definitions.

src/popup/components/Modals/EnableSecureLogin.vue
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    .text {
+      color: rgba($color-white, 0.85);
+      line-height: 20px;
+      gap: 10px;
+
+      @extend %face-sans-14-light;
+
+      @include mixins.flex(center, center, column);
```
peronczyk, line 130: I strongly advise against using this mixin as it gives more harm than real value.

src/popup/components/Modals/EnableSecureLogin.vue
```diff
@@ -0,0 +1,141 @@
...
+
+      @extend %face-sans-18-medium;
+    }
+
+    .text {
+      color: rgba($color-white, 0.85);
+      line-height: 20px;
+      gap: 10px;
+
+      @extend %face-sans-14-light;
```
peronczyk, line 128: Please put this at the beginning of style definitions.

src/popup/components/Modals/EnableSecureLogin.vue
```diff
@@ -0,0 +1,141 @@
...
+
+  .icon-wrapper {
+    margin-top: 44px;
+    opacity: 0.5;
+  }
+
+  .info {
+    .title {
+      color: $color-white;
+      padding:16px 0;
```
peronczyk, line 118:
```suggestion
      padding: 16px 0;
```

src/popup/components/Modal.vue
```diff
@@ -258,6 +260,16 @@ export default defineComponent({
     }
   }
 
+  &.transparent {
+    backdrop-filter: blur(10px);
```
peronczyk, line 264: This rule is already applied with `blur-bg` class.

src/popup/components/Modal.vue
```diff
@@ -258,6 +260,16 @@ export default defineComponent({
...
+    background-color: rgba(variables.$color-black, 0.8);
```
peronczyk, line 265: Is this really necessary as the default value is the same but the opacity is `0.7` instead of `0.8`?

Liubov-crypto:
LGTM in general. Leaved 1 question to @onvisions

src/popup/components/IconBoxed.vue
```diff
@@ -12,9 +16,23 @@
...
+] as const;
+
+export type IconBoxedVariant = typeof ICON_BOXED_VARIANT[number];
+
 export default defineComponent({
   props: {
     icon: { type: Object, default: null },
+    smallIcon: Boolean,
+    variant: {
+      type: String,
```
peronczyk, line 32:
```suggestion
      type: String as PropType<IconBoxedVariant>,
```

src/popup/components/IconBoxed.vue
```diff
@@ -1,9 +1,13 @@
 <template>
-  <div class="icon-boxed">
+  <div
+    class="icon-boxed"
+    :class="{ [variant]: variant }"
```
peronczyk, line 4: Consider this change:
```suggestion
    :class="[ variant ]"
```

peronczyk (changes requested):

src/composables/biometricAuth.ts
```diff
@@ -0,0 +1,83 @@
...
+      openModal(MODAL_ENABLE_SECURE_LOGIN);
+    }
+  }
+
+  (async () => {
+    if (composableInitialized) {
+      return;
+    }
+    await checkIsAvailable();
+    composableInitialized = true;
```
peronczyk, line 71: Setting this after awaiting is in my opinion wrong, because multiple components could run this composable in parallel and all `checkIsAvailable` method would be fired multiple times.

src/composables/biometricAuth.ts
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * Prompts the user to authenticate using biometric authentication.
+   * Returns a promise that resolves when the user is authenticated
+   * or if biometric authentication is not available.
+   */
+  async function authenticate(): Promise<void> {
+    return new Promise((resolve, reject) => {
+      if (!isAvailable.value || !isSecureLoginEnabled.value) {
+        resolve();
+        return;
```
peronczyk, line 30:
```suggestion
        return resolve();
```

src/composables/ui.ts
```diff
@@ -31,6 +36,10 @@ const hiddenCards = useStorageRef<string[]>(
     ],
   },
 );
+const secureLogin = useStorageRef<ISecureLoginSettings>(
+  {},
+  STORAGE_KEYS.secureLogin,
+);
```
peronczyk, line 42: In order to decrease the amount of future migrations and amount of separate storage entries consider moving the secure login settings to other settings.
I know that "other" is not the best name for these settings. Maybe in the future we could use something like `primarySettings` or `commonSettings`.

src/popup/App.vue
```diff
@@ -134,9 +138,16 @@ export default defineComponent({
...
+    const {
+      checkIsAvailable,
+      isAuthenticated,
+      openAuthModal,
+      deauthenticate,
+    } = useBiometricAuth();
 
     const innerElement = ref<HTMLDivElement>();
     const delayedShowHeader = ref(false);
+    const lastTimeAppWasActive = ref<Date>();
```
peronczyk, line 150: I suggest to move this to `ui` composable.

src/popup/App.vue
```diff
@@ -229,6 +240,37 @@ export default defineComponent({
...
+        if (isActive && !wasActive) {
+          if (isSecureLoginEnabled.value && isAuthenticated.value && lastTimeAppWasActive.value) {
+            await checkIsAvailable(); // Check if biometric auth is still available
+            const now = new Date();
+            const elapsedTime = now.getTime() - lastTimeAppWasActive.value.getTime();
+            if (elapsedTime > secureLoginTimeout.value) {
+              deauthenticate();
+              await openAuthModal();
+            }
+          }
```
peronczyk, line 269: In my opinion this logic should be moved to `useBiometricAuth` composable.

src/composables/biometricAuth.ts
```diff
@@ -0,0 +1,83 @@
...
+import { tg as t } from '@/popup/plugins/i18n';
+import { IS_MOBILE_APP, MODAL_ENABLE_SECURE_LOGIN, MODAL_SECURE_LOGIN } from '@/constants';
+import { useUi } from './ui';
+import { useModals } from './modals';
+
+let composableInitialized = false;
+const isAvailable = ref(false);
+const isAuthenticated = ref(false);
+
+export function useBiometricAuth() {
```
peronczyk, line 13: What do you think about making it more generic by not focusing only on the biometric aspect? This way we could extend this composable with different methods of athenticating users.
```suggestion
export function useAuth() {
```

src/popup/router/index.ts
```diff
@@ -76,6 +75,10 @@ router.beforeEach(async (to, from, next) => {
     return;
   }
 
+  if (isSecureLoginEnabled.value && !isAuthenticated.value) {
+    await openAuthModal();
+  }
```
peronczyk, line 80: As this test is repeated few times in the PR consider creating some kind of method that would do the check and eventually open the modal: `checkUserAuth()`

peronczyk (changes requested):

src/popup/router/routes.ts
martinkaintas: @onvisions What do you think?

src/popup/pages/SecureLoginSettings.vue
martinkaintas: Since it is a computed value we cannot

src/popup/components/Modals/SecureLogin.vue
martinkaintas: If user is Authenticated this modal should disappear

src/popup/components/Modal.vue
martinkaintas: According to the design it's 0.8🤷‍♂️ 
@onvisions What do you think?

src/icons/lock.svg
onvisions: Updated resources were provided. However blurring issue might be related to a visual defect causing all modals to look blurred within Chrome browser extension. I mentioned this several times and I consider it important to investigate and fix. Not for this release though.

src/popup/components/Modal.vue
onvisions: We can leave the default value then. Let's make it 0.7.

src/popup/router/routes.ts
onvisions: I think the user should not even see "Secure login" tab in Settings of the _browser extension_ because currently we do not support secure login for the extension, we do only for the apps. 
It will be a different story if we introduce the feature with password or PIN for the browser extension too. 

But currently it doesn't make sense to me: As a user I see the option "Secure login" in my browser extension Settings and I click because I'd like to have it enabled. I go to Secure login page and everything there is disabled. So now I am starting to wonder why... Is there a problem with my browser, OS. It's a waste of time and creates consfusion.

Another approach will be to display message that Secure login is currently supported only for the wallet mobile apps and not for the web version or browser extensions. However I think all this is an unnecessary complication of the feature and we'd better think of adding some secure authentication support for the web and extension users. @smaroudasunicorn

src/composables/auth.ts
```diff
@@ -0,0 +1,117 @@
...
+        }).then(() => {
+          isAuthenticated.value = true;
+        }));
+      } catch (error) {
+        reject(error);
+      }
+    });
+  }
+
+  function deauthenticate() {
```
peronczyk, line 55: Although there is a `deauthenticate` word, but it relates one specific thing: Wi-Fi. I suggest to use something different.

After doing some research I found out that most common actions on the `authenticated` state is the `login` / `logout`, `signIn` / `signOut`. For example take a look here: https://auth0.com/docs/quickstart/spa/vanillajs/01-login#evaluate-the-authentication-state

src/composables/auth.ts
```diff
@@ -0,0 +1,117 @@
...
+              await openAuthModal();
+            }
+          }
+        },
+      );
+    });
+  })();
+
+  return {
+    isAvailable: readonly(isAvailable),
```
peronczyk, line 109: I think this should not be available from the outside as we are not always sure if we already set the value.

src/composables/auth.ts
```diff
@@ -0,0 +1,117 @@
...
+const isAvailable = ref(false);
+const isAuthenticated = ref(false);
+
+export function useAuth() {
+  const { isSecureLoginEnabled } = useUi();
+  const { openModal } = useModals();
+
+  async function checkIsAvailable() {
+    isAvailable.value = (await BiometricAuth.checkBiometry()).isAvailable;
+  }
```
peronczyk, line 24: I suggest to use similar approach as `getAeSdk` so we set the value only on first call and then return it.

src/composables/auth.ts
```diff
@@ -0,0 +1,117 @@
...
+  effectScope,
+} from 'vue';
+
+import { tg as t } from '@/popup/plugins/i18n';
+import { IS_MOBILE_APP, MODAL_ENABLE_SECURE_LOGIN, MODAL_SECURE_LOGIN } from '@/constants';
+import { useUi } from './ui';
+import { useModals } from './modals';
+
+let composableInitialized = false;
+const isAvailable = ref(false);
```
peronczyk, line 15: I think this could be a regular `let` variable instead of reactive reference.

src/composables/auth.ts
```diff
@@ -0,0 +1,117 @@
...
+    isAvailable.value = (await BiometricAuth.checkBiometry()).isAvailable;
+  }
+
+  /**
+   * Prompts the user to authenticate using biometric authentication.
+   * Returns a promise that resolves when the user is authenticated
+   * or if biometric authentication is not available.
+   */
+  async function authenticate(): Promise<void> {
+    return new Promise((resolve, reject) => {
```
peronczyk, line 32: I don't see big reason for wrapping the logic with additional `Promise`.

peronczyk (changes requested):

src/popup/pages/Settings.vue
```diff
@@ -6,6 +6,12 @@
           :to="{ name: 'settings-seed-phrase' }"
           :title="$t('pages.index.seedPhrase')"
         />
+        <PanelItem
+          v-if="IS_MOBILE_APP"
+          :to="{ name: ROUTE_SECURE_LOGIN_SETTINGS }"
+          :info="isSecureLoginEnabled ? 'On' : 'Off'"
```
peronczyk, line 12: Maybe those two vallues could be taken from translations?

src/composables/auth.ts
```diff
@@ -0,0 +1,117 @@
...
+        reject(error);
+      }
+    });
+  }
+
+  function deauthenticate() {
+    isAuthenticated.value = false;
+  }
+
+  async function openAuthModal() {
```
peronczyk, line 59: In my opinion the function name should reflect the modal name. So if modal name is "Foo" the method should be "OpenFooModal". So if I see the method name somewhere I can immediately know which modal file I can open to dig into the code.

src/composables/auth.ts
```diff
@@ -0,0 +1,117 @@
...
+    await checkIsAvailable(); // Check if biometric auth is still available
+    if (isAvailable.value && isSecureLoginEnabled.value && !isAuthenticated.value) {
+      await openModal(MODAL_SECURE_LOGIN);
+    }
+  }
+
+  async function openEnableAuthModal() {
```
peronczyk, line 66: IDEM

src/composables/auth.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  (async () => {
+    if (composableInitialized) {
+      return;
+    }
+    composableInitialized = true;
+    await checkIsAvailable();
+
+    // Create persistent effect scope to avoid watcher being disposed
+    effectScope(true).run(() => {
```
peronczyk, line 80: I suggest to use `createCustomScopedComposable` instead.

peronczyk (changes requested):

src/popup/router/routes.ts
peronczyk: I think I did wrong explanation so let me rephrase.

What I was suggesting is to not write any router level code that would block user from manually entering the Secure Login settings page as it makes sense because we are removing the button that allows tthe access to this page.

If user enters the page address of this settings page in the address bar he should access the page with an information that this feature is not available to him. Right now the `SecureLoginSettings` is not testing if user is actually able to use this feature which is bad in my opinion.

Also I forgot to add **no** to this sentence:
> I strongly suggest to use composables in the route definitions.

So it should be: I strongly suggest to **NOT** use composables in the route definitions.

src/popup/router/routes.ts
```diff
@@ -4,7 +4,10 @@ import {
   PROTOCOL_VIEW_TRANSACTION_DETAILS,
   PROTOCOL_VIEW_ACCOUNT_DETAILS_ASSETS,
   PROTOCOL_VIEW_ACCOUNT_DETAILS_NAMES,
+  UNFINISHED_FEATURES,
+  IS_MOBILE_APP,
 } from '@/constants';
+import { useAuth } from '@/composables';
```
peronczyk, line 10: I strongly suggest to NOT use composables in the route definitions file :)

src/icons/lock.svg
peronczyk, line 1: I think the fill color should be taken from our CSS color variables. This will allow us to change color themes in the future.

src/popup/components/IconBoxed.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <div class="icon-boxed">
+  <div
+    class="icon-boxed"
+    :class="[variant]"
```
peronczyk, line 4: To make it easier to debug I suggest to:
```suggestion
    :class="[`variant-${variant}`]"
```

src/composables/auth.ts
```diff
@@ -0,0 +1,120 @@
...
+   * Prompts the user to authenticate using biometric authentication.
+   * Returns a promise that resolves when the user is authenticated
+   * or if biometric authentication is not available.
+   */
+  async function authenticate(): Promise<void> {
+    if (!isAvailable || !isSecureLoginEnabled.value) {
+      return Promise.resolve();
+    }
+    try {
+      return (BiometricAuth.authenticate({
```
peronczyk, line 52: I think the rounded brackets around returned value can be emoved.

src/popup/router/routes.ts
onvisions: In this case I'll add the Secure login screen mockup for the case when the page is not accessible from Settings but still accessible directly via URL.I'm syncing it with Martin for the implementation.

src/popup/components/Modals/SecureLogin.vue
peronczyk: OK, so why adding `v-if` for this case if other cases won't happen?

peronczyk (changes requested):

src/popup/router/routes.ts
peronczyk: I wouldn't do anything complex. We have our `InfoBox` component so we need only some kind of text to be put there.

peronczyk (approved):

Liubov-crypto:
LGTM

PR #2875, opened 2024-03-21
feat: jwt signature requests

thepiwo:
@CedrikNikita @davidyuk let me know what you think

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   },
   "dependencies": {
     "@aeternity/aepp-calldata": "^1.5.1",
-    "@aeternity/aepp-sdk": "^13.2.2",
+    "@aeternity/aepp-sdk": "github:aeternity/aepp-sdk-js#feat/jwt-signature",
     "@aeternity/bip39": "^0.1.0",
     "@aeternity/hd-wallet": "^0.2.0",
     "@aeternity/json-bigint": "^0.3.1",
--- a/src/popup/pages/SignMessage.vue
+++ b/src/popup/pages/SignMessage.vue
@@ -39,19 +39,36 @@ export default defineComponent({
         const message = isHexEncodedMessage ? Buffer.from(rawMessage, 'hex') : rawMessage;
         const displayMessage = message?.toString();
         const { host, href } = callbackOrigin.value || {} as any;
+        const jwt = route.query.jwt?.toString() === 'true';
 
-        await openModal(MODAL_MESSAGE_SIGN, {
-          message: displayMessage,
-          app: {
-            host,
-            name: host,
-            url: href,
-          },
-        });
+        if (jwt) {
+          const messageObject = JSON.parse(message as string);
 
-        const signature = await aeSdk.signMessage(message as string);
-        const signatureHex = Buffer.from(signature).toString('hex');
-        openCallbackOrGoHome(true, { signature: signatureHex });
+          await openModal(MODAL_MESSAGE_SIGN, {
+            message: JSON.stringify(messageObject, null, 2),
+            app: {
+              host,
+              name: host,
+              url: href,
+            },
+          });
+
+          const signature = await aeSdk.signMessageJWT(messageObject);
+          openCallbackOrGoHome(true, { signature });
+        } else {
+          await openModal(MODAL_MESSAGE_SIGN, {
+            message: displayMessage,
+            app: {
+              host,
+              name: host,
+              url: href,
+            },
+          });
+
+          const signature = await aeSdk.signMessage(message as string);
+          const signatureHex = Buffer.from(signature).toString('hex');
+          openCallbackOrGoHome(true, { signature: signatureHex });
+        }
       } catch (error: any) {
         openCallbackOrGoHome(false);
 
--- a/src/popup/pages/SignMessage.vue
+++ b/src/popup/pages/SignMessage.vue
@@ -54,8 +54,8 @@ export default defineComponent({
             },
           });
 
-          const signature = await aeSdk.signMessageJWT(messageObject);
-          openCallbackOrGoHome(true, { signature, address: aeSdk.address });
+          const signedPayload = await aeSdk.signJWT(messageObject);
+          openCallbackOrGoHome(true, { 'signed-payload': signedPayload, address: aeSdk.address });
         } else {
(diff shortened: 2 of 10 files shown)


src/popup/pages/SignMessage.vue
```diff
@@ -39,7 +39,36 @@ export default defineComponent({
...
+              host,
+              name: host,
+              url: href,
+            },
+          });
+
+          const signature = await aeSdk.signMessage(message as string);
+          const signatureHex = Buffer.from(signature).toString('hex');
+          openCallbackOrGoHome(true, { signature: signatureHex });
+        }
```
CedrikNikita, line 71:
```suggestion
        } else {
```
This is actually a copy of the code underneath.

package.json
```diff
@@ -32,7 +32,7 @@
   },
   "dependencies": {
     "@aeternity/aepp-calldata": "^1.5.1",
-    "@aeternity/aepp-sdk": "^13.2.2",
+    "@aeternity/aepp-sdk": "github:aeternity/aepp-sdk-js#feat/jwt-signature",
```
CedrikNikita, line 35: We would definitely need a proper release from the sdk, in order to include it in our release.

src/protocols/aeternity/libs/AeAccountHdWallet.ts
```diff
@@ -92,6 +93,37 @@ export class AeAccountHdWallet extends AccountBase {
...
   }
 
+  override async signMessageJWT(
+    message: object,
+    options: Parameters<AccountBase['signMessageJWT']>[1] & InternalOptions,
+  ): Promise<string> {
+    if (options?.aeppOrigin) {
+      const { checkOrAskPermission } = usePermissions();
+      const permissionGranted = await checkOrAskPermission(
+        METHODS.signMessageJWT,
```
CedrikNikita, line 103: Maybe we would need to come up with the new permission for this new method.

src/popup/pages/SignMessage.vue
```diff
@@ -39,7 +39,36 @@ export default defineComponent({
         const message = isHexEncodedMessage ? Buffer.from(rawMessage, 'hex') : rawMessage;
         const displayMessage = message?.toString();
         const { host, href } = callbackOrigin.value || {} as any;
+        const jwt = route.query.jwt?.toString() === 'true';
 
+        if (jwt) {
+          const messageObject = JSON.parse(message as string);
+
+          await openModal(MODAL_MESSAGE_SIGN, {
```
CedrikNikita, line 47: We would need either more info on existing modal or even new modal to be more user friendly for this action.

src/popup/pages/SignMessage.vue
thepiwo: yes it has been moved to the else case where the if is `if(jwt)`

src/popup/pages/SignMessage.vue
thepiwo: agree

src/protocols/aeternity/libs/AeAccountHdWallet.ts
thepiwo: agree

src/popup/pages/SignMessage.vue
CedrikNikita: It wasn't moved, yet. There are currently
```
if () {
  action a;
} else {
  action b;
}
action b;
```

thepiwo: oh right, just noticed, thanks

src/protocols/aeternity/libs/AeAccountHdWallet.ts
thepiwo: isn't `METHODS.signMessageJWT` already its own permission?

CedrikNikita: No, the check will just return `false` every time we check, so the user would need to confirm every time.

src/popup/pages/SignMessage.vue
thepiwo: adjusted

src/protocols/aeternity/libs/AeAccountHdWallet.ts
thepiwo: can you then help me understand what is needed to be done?

CedrikNikita: As an MVP i would go with asking user every time. It's just something for the future improvements.

thepiwo: I mean technically, I'm not very familiar with the wallet implementation, could you help me with that.

src/protocols/aeternity/libs/AeAccountHdWallet.ts
```diff
@@ -92,6 +93,36 @@ export class AeAccountHdWallet extends AccountBase {
     );
   }
 
+  override async signJWT(
+    message: object,
+    options: Parameters<AccountBase['signJWT']>[1] & InternalOptions,
+  ): Promise<`${string}.${string}`> {
```
peronczyk, line 99: Maybe we could name the `${string}.${string}` type and save it somewhere in `types/index.tx` file so this code would be easier to read?

src/protocols/aeternity/libs/AeAccountHdWallet.ts
```diff
@@ -92,6 +93,36 @@ export class AeAccountHdWallet extends AccountBase {
...
+      if (!permissionGranted) {
+        throw new RpcRejectedByUserError('Rejected by user');
+      }
+    }
+    const { getLastActiveProtocolAccount, getAccountByAddress } = useAccounts();
+    const account = (options?.fromAccount)
+      ? getAccountByAddress(options.fromAccount)
+      : getLastActiveProtocolAccount(PROTOCOLS.aeternity);
+
+    if (account && account.secretKey && account.protocol === PROTOCOLS.aeternity) {
```
peronczyk, line 116:
```suggestion
    if (account?.secretKey && account?.protocol === PROTOCOLS.aeternity) {
```

thepiwo:
replaced by https://github.com/superhero-com/superhero-wallet/pull/2952

PR #2895, opened 2024-03-25
fix: verify account contract call displayed as ae coin

peronczyk:


src/popup/components/TransactionLabel.vue
```diff
@@ -165,6 +165,8 @@ export default defineComponent({
         text = t('transaction.listType.tipReceived');
       } else if (isTip.value) {
         text = t('transaction.listType.tipSent');
+      } else if (innerTx.value.function === TX_FUNCTIONS.verifyAccount) {
```
CedrikNikita, line 168: There are potential problems with this approach. A contract made by fraudsters can name its method `verify_account`, and we will have it marked `Verify Account` (the user may be led to believe that the wallet is aware of the contract context/"verified").

peronczyk: I removed the verify account additional labels and created separate task for this: https://github.com/superhero-com/superhero-wallet/issues/2897

src/protocols/aeternity/config.ts
```diff
@@ -127,6 +127,8 @@ export const TX_FUNCTIONS = {
   pendingTipToken: 'pending_tip_token',
   payForGaAttach: 'pay_for_ga_attach',
   gaMetaSpend: 'ga_meta_spend',
+  /** For example Connecting to superhero.chat */
+  verifyAccount: 'verify_account',
```
CedrikNikita, line 131: Is it used in the code?

peronczyk: No, but it might help in the future as this is one of the possible values.

src/composables/transactionData.ts
```diff
@@ -180,30 +180,44 @@ export function useTransactionData({
...
+              decimals,
+              ...otherAssetData
+            }) => ({
+              amount: +toShiftedBigNumber(amount!, -decimals!),
+              ...otherAssetData,
+            }));
+        }
+
+        // Convert all unresolvable transactions that are not transfers as a coin.
+        if (functionName !== TX_FUNCTIONS.transfer) {
```
CedrikNikita, line 213: This is valid only if the `contractId` in the transaction is from known `protocolTokens`.

CedrikNikita (approved):

src/composables/transactionData.ts
```diff
@@ -180,30 +180,45 @@ export function useTransactionData({
...
+              ...otherAssetData
+            }) => ({
+              amount: +toShiftedBigNumber(amount!, -decimals!),
+              ...otherAssetData,
+            }));
+        }
+
+        // Coin conversion of all unresolvable transactions that are not transfers
+        // and uses known token.
+        if (functionName !== TX_FUNCTIONS.transfer && protocolTokens[innerTx.value.contractId]) {
```
CedrikNikita, line 214: Correct equation would be `if(!protocolTokens[innerTx.value.contractId])` since for the tip, retip, allowing we are also showing AE Coin.

CedrikNikita (changes requested):

CedrikNikita (approved):

Liubov-crypto:
it looks like an issue, FT have a fiat value. in product they don't 
https://github.com/superhero-com/superhero-wallet/assets/69896204/ed24ff5a-c545-4bf1-869c-3dac5e9504d3

wtt

src/composables/transactionData.ts
```diff
@@ -180,30 +180,45 @@ export function useTransactionData({
...
+
+      // Convert all unresolved transaction with non-token contractId to coin.
+      // For example contract calls
+      if (!protocolTokens[innerTx.value.contractId]) {
+        convertToCoin = true;
+      }
+
+      // Specific and edge cases where we clearly know that a coin was used.
+      if (innerTxTag.value === Tag.ContractCreateTx) {
+        convertToCoin = true;
```
peronczyk, line 221: Ithink we can merge those two of blocks into one.

Liubov-crypto:
Fixed, LGTM

PR #2899, opened 2024-03-26
feat: allow replace caller in sign transaction

thepiwo:
the idea here is to be able to prepare a transaction externally without knowing who will execute it (i.e. to save an additional step to get account via deeplink callback)

### DIFF
--- a/src/popup/pages/SignTransaction.vue
+++ b/src/popup/pages/SignTransaction.vue
@@ -86,7 +86,8 @@ export default defineComponent({
           {
             networkId,
             aeppOrigin: callbackOrigin.value?.toString(),
-          },
+            isSenderReplaced: replaceCaller === 'true',
+          } as any,
         );
 
         if (broadcast) {
--- a/src/popup/pages/SignTransaction.vue
+++ b/src/popup/pages/SignTransaction.vue
@@ -8,9 +8,11 @@ import { IonPage } from '@ionic/vue';
 import { defineComponent, onMounted } from 'vue';
 import { useRoute } from 'vue-router';
 import { useI18n } from 'vue-i18n';
-import { Encoded } from '@aeternity/aepp-sdk';
+import {
+  buildTx, Encoded, Tag, unpackTx,
+} from '@aeternity/aepp-sdk';
 
-import { DEFAULT_WAITING_HEIGHT } from '@/constants';
+import { DEFAULT_WAITING_HEIGHT, PROTOCOLS } from '@/constants';
 import { RejectedByUserError } from '@/lib/errors';
 import { handleUnknownError } from '@/utils';
 import {
@@ -18,6 +20,7 @@ import {
   useModals,
   useAeSdk,
   useUi,
+  useAccounts,
 } from '@/composables';
 
 export default defineComponent({
@@ -34,11 +37,30 @@ export default defineComponent({
       const { nodeNetworkId, getAeSdk } = useAeSdk();
       const { openDefaultModal } = useModals();
       const { setLoaderVisible } = useUi();
+      const { getLastActiveProtocolAccount } = useAccounts();
+
+      const aeSdk = await getAeSdk();
+      const activeAccountAddress = getLastActiveProtocolAccount(
+        PROTOCOLS.aeternity,
+      )?.address as Encoded.AccountAddress;
+
+      async function replaceCallerTx(tx: Encoded.Transaction, callerId: Encoded.AccountAddress) {
+        const unpackedTx = unpackTx(tx) as ReturnType<typeof unpackTx>
+          & { callerId: Encoded.AccountAddress };
+
+        if (unpackedTx.tag === Tag.ContractCallTx || unpackedTx.tag === Tag.ContractCreateTx) {
+          unpackedTx.callerId = callerId;
+          unpackedTx.nonce = (await aeSdk.api.getAccountByPubkey(callerId)).nonce + 1;
+        }
+
+        return buildTx(unpackedTx);
+      }
 
       try {
         setLoaderVisible(true);
-        const aeSdk = await getAeSdk();
-        const { transaction, networkId, broadcast } = route.query;
+        const {
+          transaction, networkId, broadcast, 'replace-caller': replaceCaller,
+        } = route.query;
 
         if (networkId !== nodeNetworkId.value) {
           await openDefaultModal({
@@ -51,8 +73,16 @@ export default defineComponent({
           return;
         }
 
+        // replacing the caller might be chosen by an aepp if no wallet connection is available
+        // but a transaction should be proposed to the user anyway
+        // e.g. to reduce the number of steps necessary in a deep-link environment,
+        // from two round trips, aepp -> wallet -> aepp, to one
+        const txToSign = replaceCaller === 'true'
+          ? await replaceCallerTx(transaction as Encoded.Transaction, activeAccountAddress)
+          : transaction;
+
         const signedTransaction = await aeSdk.signTransaction(
-          decodeURIComponent(transaction as string) as Encoded.Transaction,
+          decodeURIComponent(txToSign as string) as Encoded.Transaction,
           {
             networkId,
             aeppOrigin: callbackOrigin.value?.toString(),
--- a/src/popup/components/Modals/ConfirmTransactionSign.vue
+++ b/src/popup/components/Modals/ConfirmTransactionSign.vue
@@ -37,6 +37,12 @@
         class="reason"
         data-cy="reason"
       />
+      <span
+        v-if="popupProps?.isSenderReplaced"
+        class="sender-replaced"
+      >
+        {{ $t('modals.confirmTransactionSign.senderReplaced') }}
+      </span>
       <DetailsItem
         v-if="decodedCallData?.functionName"
         :label="$t('modals.confirmTransactionSign.functionName')"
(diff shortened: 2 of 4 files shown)


src/popup/pages/SignTransaction.vue
```diff
@@ -51,8 +58,24 @@ export default defineComponent({
...
+            (unpackedTx as any & { callerId: Encoded.AccountAddress })
+              .callerId = activeAccount?.address as Encoded.AccountAddress;
+            unpackedTx.nonce = (await aeSdk.api.getAccountByPubkey(
+              activeAccount?.address as Encoded.AccountAddress,
+            )).nonce + 1;
+          }
+          return buildTx(unpackedTx);
+        };
+
+        const txToSign = replaceCaller === 'true' ? await replaceCallerTx(transaction as Encoded.Transaction) : transaction;
```
CedrikNikita, line 75: I think user should be aware of what is happening under the hood, and accept or decline such thing.

src/popup/pages/SignTransaction.vue
```diff
@@ -51,8 +58,24 @@ export default defineComponent({
...
+        const replaceCallerTx = async (tx: Encoded.Transaction) => {
+          const unpackedTx = unpackTx(tx);
+          if (unpackedTx.tag === Tag.ContractCallTx || unpackedTx.tag === Tag.ContractCreateTx) {
+            // TxUnpacked isn't exported
+            // typescript is unaware that ContractCallTx and ContractCreateTx do have a callerId
+            (unpackedTx as any & { callerId: Encoded.AccountAddress })
+              .callerId = activeAccount?.address as Encoded.AccountAddress;
+            unpackedTx.nonce = (await aeSdk.api.getAccountByPubkey(
+              activeAccount?.address as Encoded.AccountAddress,
+            )).nonce + 1;
```
CedrikNikita, line 70: Denis was suggesting to use `nextNonce` instead.
https://github.com/superhero-com/superhero-wallet/pull/2782/files#diff-9e41b71bd3f413859c81a2cd9ebfe4633d90f9241c6f99c1af35709cf5594a55R168

src/popup/pages/SignTransaction.vue
thepiwo: how would we do this?

src/popup/pages/SignTransaction.vue
```diff
@@ -51,8 +58,24 @@ export default defineComponent({
           return;
         }
 
+        const replaceCallerTx = async (tx: Encoded.Transaction) => {
```
peronczyk, line 61: 1. I strongly suggest to add comment that would describe: the reason for replacing caller tx, way of testing this, why/how/when needs this. 
2. Within the components we define functions with the `function functionName() {}` notation to make it easier to distinguish constants from functions.

src/popup/pages/SignTransaction.vue
```diff
@@ -51,8 +58,24 @@ export default defineComponent({
...
+          const unpackedTx = unpackTx(tx);
+          if (unpackedTx.tag === Tag.ContractCallTx || unpackedTx.tag === Tag.ContractCreateTx) {
```
peronczyk, line 63:
```suggestion
          if ([Tag.ContractCallTx, Tag.ContractCreateTx].includes(unpackedTx.tag)) {
```

src/popup/pages/SignTransaction.vue
```diff
@@ -51,8 +58,24 @@ export default defineComponent({
...
+            // TxUnpacked isn't exported
+            // typescript is unaware that ContractCallTx and ContractCreateTx do have a callerId
+            (unpackedTx as any & { callerId: Encoded.AccountAddress })
```
peronczyk, line 66: I suggest to override/set the typing where the constant is created and extend the return type instead of overriding it with `any` because `any & { ... }` is treated as `any`:

```
const unpackedTx = unpackTx(tx)  as ReturnType<typeof unpackTx> & { callerId: Encoded.AccountAddress };
```
(not tested)

src/popup/pages/SignTransaction.vue
```diff
@@ -51,8 +58,24 @@ export default defineComponent({
...
+              .callerId = activeAccount?.address as Encoded.AccountAddress;
+            unpackedTx.nonce = (await aeSdk.api.getAccountByPubkey(
+              activeAccount?.address as Encoded.AccountAddress,
```
peronczyk, line 69: This is already used before. Maybe you can create constant first:
```
const activeAccountAddress = activeAccount?.address as Encoded.AccountAddress;
```

peronczyk (changes requested):

src/popup/pages/SignTransaction.vue
thepiwo: I'll try, thanks!

src/popup/pages/SignTransaction.vue
thepiwo: agree, will adjust

src/popup/pages/SignTransaction.vue
thepiwo: nice suggestion, will add it

src/popup/pages/SignTransaction.vue
thepiwo: agree, will adjust it, thanks!

src/popup/pages/SignTransaction.vue
```diff
@@ -8,16 +8,18 @@ import { IonPage } from '@ionic/vue';
...
-import { DEFAULT_WAITING_HEIGHT } from '@/constants';
+import { DEFAULT_WAITING_HEIGHT, PROTOCOLS } from '@/constants';
 import { RejectedByUserError } from '@/lib/errors';
 import { handleUnknownError } from '@/utils';
 import {
   useDeepLinkApi,
   useModals,
   useAeSdk,
-  useUi,
+  useUi, useAccounts,
```
peronczyk, line 22: Please split this line.

thepiwo: maybe a more opinionated formatter/linter should be used if those are implicit rules among developers

src/popup/pages/SignTransaction.vue
thepiwo: unfortunately then the compiler doesn't know the existence of `.nonce` as well

src/popup/pages/SignTransaction.vue
thepiwo: adjusted

src/popup/pages/SignTransaction.vue
thepiwo: seems to work :)

src/popup/pages/SignTransaction.vue
thepiwo: adjusted

src/popup/pages/SignTransaction.vue
```diff
@@ -34,11 +37,30 @@ export default defineComponent({
...
+      const { getLastActiveProtocolAccount } = useAccounts();
+
+      const aeSdk = await getAeSdk();
+      const activeAccountAddress = getLastActiveProtocolAccount(
+        PROTOCOLS.aeternity,
+      )?.address as Encoded.AccountAddress;
+
+      async function replaceCallerTx(tx: Encoded.Transaction) {
+        const unpackedTx = unpackTx(tx) as ReturnType<typeof unpackTx>
+            & { callerId: Encoded.AccountAddress };
```
peronczyk, line 49: Wrong indentation.

src/popup/pages/SignTransaction.vue
peronczyk: @thepiwo you didn't add the comment.

thepiwo: my bad, will do

src/popup/pages/SignTransaction.vue
thepiwo: eslint didn't seem to complain about it, what would be the correct way?

peronczyk: 2 spaces.
And yes, eslint is not always complaining about stuff :)

src/popup/pages/SignTransaction.vue
```diff
@@ -34,11 +37,30 @@ export default defineComponent({
...
       const { openDefaultModal } = useModals();
       const { setLoaderVisible } = useUi();
+      const { getLastActiveProtocolAccount } = useAccounts();
+
+      const aeSdk = await getAeSdk();
+      const activeAccountAddress = getLastActiveProtocolAccount(
+        PROTOCOLS.aeternity,
+      )?.address as Encoded.AccountAddress;
+
+      async function replaceCallerTx(tx: Encoded.Transaction) {
```
peronczyk, line 47: I suggest to convert this function into pure function by adding 2nd argument which would take the new `callerId` and replace it.

src/popup/pages/SignTransaction.vue
thepiwo: done

src/popup/pages/SignTransaction.vue
thepiwo: done

src/popup/pages/SignTransaction.vue
thepiwo: done

peronczyk (approved):

thepiwo:
rebased on develop @peronczyk  @CedrikNikita feel free to merge

PR #2901, opened 2024-03-26
refactor: eth fee calculation

### LINKED ISSUE #2902: Refactor ethereum fee calculation

martinkaintas:
closes #2902

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -67,7 +67,7 @@
         "vue-i18n": "^9.2.2",
         "vue-loader": "^17.2.2",
         "vue-router": "^4.2.4",
-        "web3-eth": "^4.3.1",
+        "web3-eth": "^4.5.0",
         "web3-eth-accounts": "^4.1.0",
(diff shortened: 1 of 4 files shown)


Liubov-crypto:
I was trying to send ETH in mainnet with medium fee and got this issue: 

eth e
ethe

however it was working with low and fast fees. 

and it seems to be working in testnet.

Liubov-crypto:
It seems that Max amount button is stop working if ETH balance is less than fee: 

ьфч

we have warning in similar case for BTC account: 
https://github.com/superhero-com/superhero-wallet/assets/69896204/4c852e33-c51e-4fb9-b0cb-a7f15e588469

The rest looks good, I made 2 tx in mainnet with high and medium fee with Max amount. Txs were successful. To make the same tx with low fee I need additional ETH, not enough fund on my test wallet. 

In testnet also works fine.

peronczyk (approved):

PR #2904, opened 2024-03-26
fix(aeternity): aex9 unfinished transaction info

CedrikNikita:
missed in https://github.com/superhero-com/superhero-wallet/pull/2659


Liubov-crypto:
before:
dev

after:
fix

LGTM

martinkaintas (approved):

PR #2905, opened 2024-03-27
fix(multisig): show total correctly

CedrikNikita:
missed in https://github.com/superhero-com/superhero-wallet/pull/2307


Liubov-crypto:
LGTM

PR #2911, opened 2024-03-27
fix: seed phrase notification position

peronczyk:
I have replaced JS notification box position calculations with regular CSS approach: `position: relative;` on parent + `position: absolute; inset: 0;` on notification box.


martinkaintas (approved):

CedrikNikita (approved):
Alignment is correct on Android and web.

Liubov-crypto:
Tested in Safari web, LGTM

PR #2925, opened 2024-03-28
fix: load available tokens when adding an account on a new blockchain

### LINKED ISSUE #2920: ETH related assets are not visible in ETH mainnet on asset selector page
**Describe the bug**
ETH related assets are not visible in ETH mainnet on asset selector page 

**To Reproduce**
1. Go to ETH acc
2. Click on Receive
3. Try to choose asset from assets selector

**Expected behavior**

**Screenshots**

 - OS: macos
 - Browser  chrome, safari
 - Version 123

**Additional context**
also reproducible on ios app

https://github.com/superhero-com/superhero-wallet/assets/69896204/64156939-550a-4409-a9dd-fb2ed55cebe3

CedrikNikita:
fixes #2920

### DIFF
--- a/src/composables/fungibleTokens.ts
+++ b/src/composables/fungibleTokens.ts
@@ -289,6 +289,11 @@ export function useFungibleTokens() {
     watch(accounts, (val, oldVal) => {
       if (val.length !== oldVal.length) {
         loadTokenBalances();
+
+        // this is to check if the new account is the first for specified protocol
+        if (val.filter(({ protocol }) => protocol === val.at(-1)?.protocol).length === 1) {
+          loadAvailableTokens();
+        }
       }
     });
 


src/composables/fungibleTokens.ts
```diff
@@ -290,6 +290,9 @@ export function useFungibleTokens() {
       if (val.length !== oldVal.length) {
         loadTokenBalances();
       }
+      if (val.filter(({ protocol }) => protocol === val.at(-1)?.protocol).length === 1) {
```
peronczyk, line 293: 1. I suggest to add small comment that this is to check if the new account is the first for specified protocol.
2. Also I guess this test might be put inside of the previous test where we test if there is a new account added.

PR #2926, opened 2024-03-28
fix: ae token amount precision fallback

peronczyk:


src/composables/fungibleTokens.ts
```diff
@@ -253,10 +254,11 @@ export function useFungibleTokens() {
 
     const tokenData = protocol
       && getProtocolAvailableTokens(protocol)[contractCallData?.assetContractId!];
+
     if (contractCallData && tokenData) {
       return +toShiftedBigNumber(
         contractCallData.amount || 0,
-        -tokenData.decimals!,
+        -(tokenData.decimals || AE_COIN_PRECISION),
```
CedrikNikita, line 261: This point is pretty moot. In most cases it will be unnoticeable (since the base amount of decimals is 18 in ae fungible tokens). But in cases when it is not, the user will think that he lost money, or it was stolen.
In case we use zero as fallback it will be obvious to the user that it is an incorrect balance, but it will be highly noticeable in most cases.

CedrikNikita (approved):
We should not spam errors in console. Let's create a task to decide on further improvement.

src/composables/fungibleTokens.ts
peronczyk: Agreed on priv that we leave 18 for now.

PR #2927, opened 2024-03-28
fix: eth max button should stay on

### LINKED ISSUE #2909: Max amount button is stop working if ETH balance is less than fee
**Describe the bug**
Max amount button is stop working if ETH balance is less than fee

**To Reproduce**
1. Go to Send ETH, your ETH balance should be lower than fee
2. Click on Max button
3. See error

**Expected behavior**

**Screenshots**

 - OS: MacOS
 - Browser chrome, FF
 - Version 123

**Additional context**

ьфч

martinkaintas:
closes #2909

### DIFF
--- a/src/popup/components/InputField.vue
+++ b/src/popup/components/InputField.vue
@@ -54,7 +54,9 @@
           <input
             v-bind="$attrs"
             :id="inputId"
+            ref="inputEl"
             class="input"
+            :class="{ 'blink-hidden': isBlinking, blink: blinkOnChange }"
             autocomplete="off"
             step="any"
             data-cy="input"
@@ -143,6 +145,7 @@ export default defineComponent({
     readonly: Boolean,
     showHelp: Boolean,
     showMessageHelp: Boolean,
+    blinkOnChange: Boolean,
     code: Boolean,
     thin: Boolean,
     textLimit: {
@@ -161,6 +164,8 @@ export default defineComponent({
     const inputId = `input-${uid}`;
 
     const focused = ref(false);
+    const inputEl = ref<HTMLInputElement | null>(null);
+    const isBlinking = ref(false);
 
     const inputMode = computed(() => props.type === 'number' ? 'decimal' : 'text');
     const messageAsObject = computed(
@@ -206,7 +211,21 @@ export default defineComponent({
       (val) => emit('focus-change', val),
     );
 
+    watch(
+      () => props.modelValue,
+      () => {
+        if (props.blinkOnChange) {
+          isBlinking.value = true;
+          setTimeout(() => {
+            isBlinking.value = false;
+          }, 500);
+        }
+      },
+    );
+
     return {
+      isBlinking,
+      inputEl,
       focused,
       uid,
       inputId,
--- a/src/popup/components/TokenAmount.vue
+++ b/src/popup/components/TokenAmount.vue
@@ -1,7 +1,14 @@
 <template>
   <span
+    ref="tokenAmountEl"
     class="token-amount"
-    :class="[{ large, small, vertical }]"
+    :class="[{
+      large,
+      small,
+      vertical,
+      'blink-hidden': isBlinking,
+      blink: blinkOnChange,
+    }]"
   >
     <span
       class="amount"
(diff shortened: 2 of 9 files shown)


CedrikNikita:
1. Shouldn't we introduce this behaviour to other protocols?
2. We should also show the user that the data was changed. @onvisions had some suggestions about it.

src/protocols/ethereum/components/TransferSendForm.vue
```diff
@@ -31,12 +31,13 @@
             : {},
           enough_coin: [fee.toString(), ETH_COIN_SYMBOL],
         }"
+        @update:model-value="() => shouldUseMaxAmount = false"
```
peronczyk, line 34: No need to use callback:
```suggestion
        @update:model-value="shouldUseMaxAmount = false"
```

onvisions:
@CedrikNikita @martinkaintas @peronczyk 
> 1. Shouldn't we introduce this behaviour to other protocols?
> 2. We should also show the user that the data was changed. @onvisions had some suggestions about it.

IMO
1. - Yes, we should. Consistency.
2. - if you mean the smooth "blinking" transition of updated values (I shared a video of the effect): Yes, we'd better apply it so user will visually be aware that data has changed.

CedrikNikita:
On opening the modal.
```
TypeError: Cannot read properties of null (reading 'animate')
```

src/utils/common.ts
```diff
@@ -502,3 +502,42 @@ export function getCleanModalOptions<T>(options: T): Omit<T, 'onCompiler' | 'onN
   delete cleanedOptions?.onNode;
   return cleanedOptions;
 }
+/**
+ * Returns animation that blinks the element with the option to call a callback
+ * in the middle of the animation (opacity 0).
+ */
+export function createBlinkAnimation(el: HTMLElement, callback?: () => void) {
```
peronczyk, line 509: I strongly suggest to use CSS instead.

peronczyk (changes requested):

src/popup/components/InputField.vue
```diff
@@ -206,7 +210,20 @@ export default defineComponent({
       (val) => emit('focus-change', val),
     );
 
+    watch(
+      () => props.modelValue,
+      () => {
+        if (props.blinkOnChange) {
+          inputEl.value?.classList.add('hide');
```
peronczyk, line 217: Why using vanilla JS instead of `:class`?

Also in my opinion this block of code is not self explanatory. Why adding and removing `hide` class? Are we hiding something here?

peronczyk (changes requested):

src/popup/components/TokenAmount.vue
```diff
@@ -51,13 +57,20 @@ export default defineComponent({
     dynamicSizing: Boolean,
     large: Boolean,
     small: Boolean,
+    blinkOnChange: Boolean,
   },
   setup(props) {
+    const tokenAmountEl = ref<HTMLElement>();
+    const amountRounded = ref<string|number>(0);
```
peronczyk, line 64:
```suggestion
    const amountRounded = ref<string | number>(0);
```

src/popup/components/TokenAmount.vue
```diff
@@ -71,15 +84,34 @@ export default defineComponent({
...
+      amountRounded.value = getAmountRounded();
+      amountFiat.value = getAmountFiat();
+    }
 
-    const symbolComputed = computed(() => props.symbol || adapter.coinSymbol);
+    watch(() => props.amount, () => {
+      if (!props.blinkOnChange) {
+        updateAmountValues();
+      }
+      tokenAmountEl.value?.classList.add('hide');
```
peronczyk, line 102: IDEM - I suggest to not use vanilla here.

src/styles/global.scss
```diff
@@ -319,3 +319,11 @@ hr {
...
   }
 }
+
+.blink {
+  transition: opacity 0.5s ease-in-out;
+
+  &.hide {
+    opacity: 0;
+  }
+}
```
peronczyk, line 329: In my opinion it would be better to create CSS animation and then on the component level adding only one class, e.g.: `value-changed` this way:

```
watch(<modal value>, async () => {
  valueChanged.value = true;
  await nextTick();
  valueChanged.value = false;
});
```

The style for this class would be:
```
.value-changed {
  animation: blink-animation...
}
```

(just a rough idea, not tested)

peronczyk (approved):

PR #2947, opened 2024-04-05
perf: use dryAeSdk to collect multisig info

CedrikNikita:
fixing one part of the issue raised here
https://github.com/superhero-com/superhero-wallet/blob/d233c480fcbc0a0ab1ddf79deda42b30d96495c5/src/protocols/aeternity/libs/AeSdkSuperhero.ts#L35
by simply using `dryRunSdk` that doesn't have accounts and will not call our overridden `_resolveAccount` method, every time we are updating multisig vaults info.

### DIFF
--- a/src/composables/aeSdk.ts
+++ b/src/composables/aeSdk.ts
@@ -44,7 +44,6 @@ type AeppInfoData = OnAeppConnectionParams & { origin: string };
 
 let composableInitialized = false;
 let aeSdk: AeSdkSuperhero;
-let storedNetworkName: string;
 
 const nodeNetworkId = ref<NetworkId>();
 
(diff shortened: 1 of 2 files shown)


peronczyk (approved):

peronczyk (approved):

Liubov-crypto:
In general LGTM. The issue with v3 contracts has been fixed. 

I got an issue with pending multisig Tx displaying, but only once. I tried to reproduce it, but my attempts were not successful. 
Probably it worth to add it to our backlog...

steps to reproduce it: 
1. create multisig 
2. propose Tx from new multisig with 0 balance
3. observe an issue: 

errrm

https://github.com/superhero-com/superhero-wallet/assets/69896204/49c01b6f-4422-42a4-b343-626f736c137b

Issue #2938, opened 2024-04-02
Permission Issue - Remove permission access your data for all sites


smaroudasunicorn:

CedrikNikita:
`Metamask` is actually asking for the same kind of permissions.
![image](https://github.com/superhero-com/superhero-wallet/assets/9008074/d8e78e90-18f2-4761-b000-bc7ec9db6613)

PR #2949, opened 2024-04-05
build: remove unused extension permissions

CedrikNikita:
part of #2938


Liubov-crypto:
LGTM. 

Logged a separate issue regarding non-accessible popups in ff and chrome when user interacts with chat bot (reproducible on product).

peronczyk (approved):

PR #2963, opened 2024-04-12
fix: be able to migrate mnemonic from vuex

### LINKED ISSUE #2960: Issue when upgrading extension from 2.1.6 to latest Release. Seed is not kept
Reported by philip

_Uhm I just updated my Superhero Wallet on Firefox from 2.1.6 to the latest version and need to re-enter my seed, this seems really bad..._

CedrikNikita:
fixes #2960

### DIFF
--- a/src/popup/components/Modals/ResetWalletModal.vue
+++ b/src/popup/components/Modals/ResetWalletModal.vue
@@ -77,6 +77,9 @@ export default defineComponent({
       disconnectDapps();
 
       await WalletStorage.clear();
+      // TODO: Rethink this approach
+      // It is removing the remaining vuex state
+      await browser.storage.local.clear();
       await router.push({ name: ROUTE_INDEX });
       window.location.reload();
     }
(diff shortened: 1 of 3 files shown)


peronczyk (approved):

peronczyk (approved):

src/popup/components/Modals/ResetWalletModal.vue
```diff
@@ -77,6 +77,9 @@ export default defineComponent({
       disconnectDapps();
 
       await WalletStorage.clear();
+      // TODO: Rethink this approach
+      // It is removing the remaining vuex state
+      await browser.storage.local.clear();
```
peronczyk, line 82: Maybe we could add this in next release when we will be 100% ready that the migrations works fine?

CedrikNikita: The user won't be able to reset the wallet without it.

peronczyk: You're right. So let's leave it.

smaroudasunicorn (approved):

PR #2983, opened 2024-04-18
feat/airgap integration

### LINKED ISSUE #2866: Airgap / Multisig Integration
We need to use Airgap Vault with SH Wallet. 

More specific accounts stored in Airgap Vault could be used. 
User should be able to sign transactions though his vault.
User should be able to provide concensus in a multisig transaction though his vault. 

Airgap Vault shall be usable in online/offline mode.

martinkaintas:
closes #2866


peronczyk (changes requested):

src/composables/accounts.ts
```diff
@@ -87,15 +90,26 @@ const accounts = computed((): IAccount[] => {
...
+        const idx = airGapIdx;
+        airGapIdx += 1;
+
+        return {
+          globalIdx,
+          idx,
+          publicKey: Buffer.from(decode(airGapAccount.address as Encoded.AccountAddress)),
+          ...airGapAccount,
+        } as unknown as IAccount;
+      }
```
peronczyk, line 108: Considering that we might have more account types in the future I suggest to make smal refactoring:
1. Instead of having `getHdWalletAccountFromMnemonicSeed` for each adapter I think we could have something more generic, like for example: `getAccountFromRawAccount` (or something similar). We would pass whole raw account and IDX.
2. The adapter would take care of checking of what kind of account it is and pick right logic to return correct data.

What do you think?

src/composables/accounts.ts
```diff
@@ -87,15 +90,26 @@ const accounts = computed((): IAccount[] => {
   }
 
   const idxList = Object.fromEntries(PROTOCOL_LIST.map(((protocol) => [protocol, 0])));
+  let airGapIdx = 0;
```
peronczyk, line 93: I don't think having separate variables for each account types is reasonable. Maybe we could have `idxList` with 2 levels instead of 1? This way we would access it in a following way: `idxList[account.protocol][account.type]`.

src/composables/airGap.ts
```diff
@@ -0,0 +1,157 @@
...
+
+        [...Array(multiEncoder.fragmentsLength)].forEach(() => {
+          fragments.push(multiEncoder.nextPart());
+        });
+
+        return fragments;
+      }
+
+      // Encode the UR and return the UR string in upper case.
+      return [(singleEncoder.nextPart()).toUpperCase()];
```
peronczyk, line 72:
```suggestion
      return [singleEncoder.nextPart().toUpperCase()];
```

src/composables/airGap.ts
```diff
@@ -0,0 +1,157 @@
...
+          ur,
+          SETTINGS_SERIALIZER_MULTI_CHUNK_SIZE,
+        );
+        const fragments: string[] = [];
+
+        [...Array(multiEncoder.fragmentsLength)].forEach(() => {
+          fragments.push(multiEncoder.nextPart());
+        });
+
+        return fragments;
```
peronczyk, line 68:
```suggestion
        return [...Array(multiEncoder.fragmentsLength)].map(() => multiEncoder.nextPart());
```

Not tested, just a rough idea.

src/composables/airGap.ts
```diff
@@ -0,0 +1,157 @@
...
+    serializerV3Companion.schemas.forEach((schema) => {
+      SerializerV3.addSchema(schema.type, schema.schema, MainProtocolSymbols.AE);
+    });
+    serializer.value = SerializerV3.getInstance();
+  })();
+
+  /**
+   * Encodes an array of IACMessageDefinitionObjectV3 objects into a UR string.
+   */
+  async function encodeIACMessageDefinitionObjects(data: IACMessageDefinitionObjectV3[]) {
```
peronczyk, line 38: If function has multiple return points I suggest to add return type.

src/composables/airGap.ts
```diff
@@ -0,0 +1,157 @@
...
+   * and returns the transaction object.
+   */
+  async function extractSignedTransactionResponseData(
+    data: IACMessageDefinitionObjectV3[] = [],
+  ): Promise<string | null> {
+    const payload = data.find(
+      (item) => item.type === IACMessageType.TransactionSignResponse,
+    )?.payload as any;
+
+    return payload?.transaction || null;
```
peronczyk, line 121:
```suggestion
    return data.find(
      (item) => item.type === IACMessageType.TransactionSignResponse,
    )?.payload?.transaction;
```

src/composables/airGap.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  async function generateTransactionURDataFragments(
+    publicKey: string,
+    transaction: string,
+    networkId: string,
+  ): Promise<string[]> {
+    const id = Math.floor(Math.random() * 90000000 + 10000000);
+    const callbackURL = `${MOBILE_SCHEMA}?d=`;
+    const payload: any = {
```
peronczyk, line 131: To avoid using `any` I suggest to assign the value directly in the `messageDefinitionObject` definition block. This way the `IACMessageDefinitionObjectV3` type would nicely validate the schema.

peronczyk (changes requested):

src/popup/components/form/FormInputWithQr.vue
```diff
@@ -0,0 +1,78 @@
...
+    label: { type: String, default: '' },
+    qrIcon: { type: String, default: '' },
+    qrTitle: { type: String, default: '' },
+  },
+  emits: ['update:modelValue'],
+  setup(props, { emit }) {
+    const { openModal } = useModals();
+
+    async function scanQr() {
+      const result = await openModal(MODAL_READ_QR_CODE, {
```
peronczyk, line 51: You can use `openScanQrModal` method directly.

src/popup/components/form/FormInputWithQr.vue
peronczyk, line 1: I suggest to rename this to `FormTextareaWithQr` as the actual input is the `textarea`. Or maybe `FormScanQrResult` if we want to be input-type-agnostic :)

src/popup/components/form/FormInputWithQr.vue
```diff
@@ -0,0 +1,78 @@
...
+  components: {
+    FormTextarea,
+    BtnPlain,
+    QrScanIcon,
+  },
+  props: {
+    modelValue: { type: String, default: '' },
+    placeholder: { type: String, default: '' },
+    label: { type: String, default: '' },
+    qrIcon: { type: String, default: '' },
```
peronczyk, line 43: Probably this can be removed as the scan qr modal does not support providing custom icons.

src/popup/components/form/FormInputWithQr.vue
```diff
@@ -0,0 +1,78 @@
...
+      scanQr,
+    };
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@use '@/styles/variables' as *;
+
+.form-input-with-qr{
```
peronczyk, line 70:
```suggestion
.form-input-with-qr {
```

peronczyk (changes requested):

src/composables/accounts.ts
```diff
@@ -110,6 +124,8 @@ const accounts = computed((): IAccount[] => {
 
 const activeAccount = computed((): IAccount => accounts.value[activeAccountGlobalIdx.value] || {});
 
+const isAirGap = computed((): boolean => activeAccount.value.type === ACCOUNT_AIR_GAP_WALLET);
```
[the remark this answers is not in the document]
martinkaintas: Fair point

src/composables/airGap.ts
martinkaintas: It seems `any` was not needed obsolete there

src/popup/components/AccountImportRow.vue
```diff
@@ -0,0 +1,47 @@
...
+  },
+  setup(props) {
+    const balance = ref(new BigNumber(0));
+
+    const numericBalance = computed<number>(() => balance.value.toNumber());
+
+    onMounted(async () => {
+      const adapter = ProtocolAdapterFactory.getAdapter(props.account.protocol);
+      balance.value = new BigNumber(await adapter.fetchBalance(props.account.address));
+    });
```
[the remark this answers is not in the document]
martinkaintas: Unfortunately the `balances` composable doesn't hold the balances of AirGap accounts that have not been imported yet

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -150,7 +150,7 @@
         :disabled="!!error || verifying"
         :icon="verifying ? AnimatedSpinner : null"
         :text="verifying ? $t('common.verifying') : $t('common.confirm')"
-        @click="popupProps?.resolve()"
+        @click="confirm"
```
peronczyk, line 153:
```suggestion
        @click="confirm()"
```
To avoid passing the `$event` as the 1st argument.

src/popup/components/Modals/QrCodeReader.vue
```diff
@@ -119,12 +133,14 @@ export default defineComponent({
...
     const route = useRoute();
-    const { setMobileQrScannerVisible } = useUi();
+    const { setMobileQrScannerVisible, scanProgress } = useUi();
 
     const qrCodeVideoEl = ref<HTMLVideoElement>();
     const hasDeviceCamera = ref(false);
     const isCameraReady = ref(false);
     const cameraPermissionGranted = ref(true);
+    const isMultiFragmentQr = ref(false);
+    const decoder = ref(new URDecoder());
```
peronczyk, line 143: This is not a UI variable so there is no need to use reactivity here.

peronczyk (changes requested):

src/protocols/aeternity/views/TransferSendModal.vue
```diff
@@ -89,16 +95,30 @@ export default defineComponent({
...
-      ? t('modals.multisigTxProposal.proposeAndApprove')
-      : '');
+    const isSendingDisabled = computed(() => {
+      if (showNextButton.value) {
+        return (error.value || !isAeNodeReady.value);
+      }
+      return error.value
+        || !isAeNodeReady.value
+        || !transferData.value.address
+        || !transferData.value.amount;
```
peronczyk, line 110: In my opinion this is really hard to read now. Maybe there is a way to simplify this?
The name of the computed property is related to "sending". So having `showNextButton` here creates confusion.

src/protocols/aeternity/views/TransferSendModal.vue
```diff
@@ -5,6 +5,7 @@
     :hide-arrow-send-icon="isMultisig"
     :custom-primary-button-text="customPrimaryButtonText"
     :sending-disabled="isSendingDisabled"
+    :is-air-gap="isAirGap"
```
peronczyk, line 8: As there might be more account types (like for example ledger and so on) I suggest to try avoid using props related directly to this account type. Maybe passing the active account as a prop and trying to establish the component's logic by analysing it would be better?

src/protocols/aeternity/views/TransferSendModal.vue
```diff
@@ -14,7 +15,9 @@
         :is="currentStepConfig.component"
         ref="currentRenderedComponent"
         v-model:transferData="transferData"
+        :tx-raw="txRaw"
         :is-multisig="isMultisig"
+        :is-air-gap="isAirGap"
```
peronczyk, line 20: IDEM

src/protocols/aeternity/views/TransferSendModal.vue
```diff
@@ -72,6 +77,7 @@ export default defineComponent({
     const currentRenderedComponent = ref<Component>();
     const currentStep = ref<TransferSendStep>(TRANSFER_SEND_STEPS.form);
     const error = ref(false);
+    const txRaw = ref();
```
peronczyk, line 80: I don't see any logic that updates this empty reference. Maybe this can be removed?

src/protocols/aeternity/views/AccountCreateModal.vue
```diff
@@ -39,18 +39,36 @@
         :disabled="!isOnline"
         @click="createMultisigAccount()"
       />
+      <BtnSubheader
```
peronczyk, line 42: There is already a button with `v-if="!isMultisig"` above. I suggest to group them:
```
<template v-if="isMultisig">
  // MS BTN
</template>
<template v-else>
  // NON MS BTNs
</template>
```

src/popup/plugins/veeValidate.ts
```diff
@@ -234,6 +234,18 @@ export default () => {
...
+   * AirGap accounts can only send funds to an address
+   */
+  defineRule(
+    'airgap_to_an_address',
+    (value: string, [isAirGap]: [boolean]) => (
+      !isNameValid(value)
+      || (isNameValid(value) && !isAirGap)
+      || tg('validation.airGapToAnAddress')
+    ),
+  );
```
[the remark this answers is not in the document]
martinkaintas: We could create a separate task about this, since this is also the case for tokens, see `token_to_an_address` rule.

src/popup/plugins/veeValidate.ts
```diff
@@ -234,6 +234,18 @@ export default () => {
     ),
   );
 
+  /**
+   * AirGap accounts can only send funds to an address
```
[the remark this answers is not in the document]
martinkaintas: We could create a separate task about this, since this is also the case for tokens, see `token_to_an_address` rule.

src/protocols/aeternity/views/AccountCreateModal.vue
```diff
@@ -39,18 +39,36 @@
...
 <script lang="ts">
 import { defineComponent, PropType } from 'vue';
-import { MODAL_MULTISIG_VAULT_CREATE, PROTOCOLS } from '@/constants';
-import { useAccounts, useConnection, useModals } from '@/composables';
+import {
+  MODAL_AIR_GAP_IMPORT_ACCOUNTS,
+  MODAL_MULTISIG_VAULT_CREATE,
+  PROTOCOLS,
+} from '@/constants';
+import { IAirgapAccountRaw } from '@/types';
```
peronczyk, line 60:
```suggestion
import type { IAirgapAccountRaw } from '@/types';
```
...and please put this line before the constants :)

src/protocols/aeternity/libs/AeAccountHdWallet.ts
```diff
@@ -29,26 +36,54 @@ interface InternalOptions {
  */
 const TAGS_TO_SIGN_WITHOUT_PERMISSION: Tag[] = [Tag.SpendTx, Tag.PayingForTx];
 
+function getAccount(fromAccount?: Encoded.AccountAddress) {
```
peronczyk, line 39: I think it's an antipattern to have a root util functions in the library file.

src/utils/airGap.ts
```diff
@@ -0,0 +1,60 @@
+import bs58check from 'bs58check';
+import { URDecoder, UR, UREncoder } from '@ngraveio/bc-ur';
+import { IAccount, IAirgapAccountRaw } from '@/types';
+
+/**
+ * Type guard for TypeScript.
+ * Check if the account is an AirGap account and set the type accordingly.
+ */
+export function isAirgapAccount(
```
peronczyk, line 9: In most places you use "AirGap" notation.

```suggestion
export function isAccountAirGap(
```

peronczyk (changes requested):

src/popup/components/Modals/AirGapImportAccounts.vue
```diff
@@ -0,0 +1,192 @@
...
+      />
+
+      <div
+        v-for="account in accounts"
+        :key="account.address"
+        class="account-row"
+        :class="{ disabled: isAccountAlreadyImported(account) }"
+      >
+        <AccountImportRow :account="account" />
+      </div>
```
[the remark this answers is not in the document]
martinkaintas: the `AccountImportRow` is crucial in order to pass the balance to `AccountSelectOptionsItem`. Doing it the way you propose would make fetching the balance way more complicated than it should be.

src/popup/components/TransferSend/TransferReviewBase.vue
```diff
@@ -114,13 +121,15 @@ export default defineComponent({
...
     title: { type: String, default: tg('pages.send.reviewtx') },
+    subtitle: { type: String, default: tg('pages.send.checkalert') },
     senderLabel: { type: String, default: tg('pages.send.sender') },
     amountLabel: { type: String, default: tg('common.amount') },
     baseTokenSymbol: { type: String, required: true },
     transferData: { type: Object as PropType<TransferFormModel>, required: true },
     protocol: { type: String as PropType<Protocol>, required: true },
     recipientAddress: { type: String, default: null },
     avatarName: { type: String, default: null },
+    disablePadding: Boolean,
```
peronczyk, line 132: I suggest to rename this to `noHeaderPadding`.

peronczyk (changes requested):

src/popup/pages/Dashboard.vue
```diff
@@ -29,7 +29,7 @@
 
         <template #buttons>
           <OpenTransferReceiveModalButton is-big />
-          <OpenTransferSendModalButton is-big />
+          <OpenTransferSendModalButton is-big :is-air-gap="isAirGap" />
```
peronczyk, line 32: Probably this button doesn't require airgap flag property as the TransferSend modal can establish this on its own.

src/popup/pages/Assets/AssetDetails.vue
```diff
@@ -31,6 +31,7 @@
             <OpenTransferSendModalButton
               :is-multisig="isMultisig"
               :disabled="isMultisig && !!pendingMultisigTransaction"
+              :is-air-gap="isAirGap"
```
peronczyk, line 34: IDEM

src/popup/components/form/FormTextarea.vue
```diff
@@ -140,7 +140,7 @@ export default defineComponent({
     --size: 8;
   }
 
-  &.auto-height {
+  &.auto-height, &.xxs {
```
peronczyk, line 143: Why not use the `autoHeight` property instead?

src/popup/components/form/FormSelect.vue
```diff
@@ -74,6 +74,7 @@ export default defineComponent({
   },
   emits: ['select', 'update:modelValue'],
   setup(props, { emit }) {
+    // TODO AIRGAP: implement multiple accounts selection
```
peronczyk, line 77: I suggest to create a task for this.

peronczyk (changes requested):

src/protocols/aeternity/views/TransferSendModal.vue
martinkaintas: From your previous comments I've already removed the `isAirGap` prop and used the `accounts` composable to get the `isActiveAccountAirGap` ref in order to do the required checks in the base component

src/protocols/aeternity/views/AccountCreateModal.vue
martinkaintas: This would change the order of the buttons, because the button in the middle is shown both for `if multisig` & `if !multisig`. There's no `v-else`

src/popup/components/form/FormTextarea.vue
martinkaintas: Because then it couldn't be resizable
`:class="{ resizable: resizable && !autoHeight }"`

src/composables/accounts.ts
```diff
@@ -224,16 +230,19 @@ export function useAccounts() {
...
+      // If account is AirGap, we need to push the whole account object
+      // because we are not able restore it from the mnemonic seed
+      accountsRaw.value.push(account);
+    } else {
+      accountsRaw.value.push({
+        protocol: account.protocol,
+        isRestored: account.isRestored,
+        type: ACCOUNT_TYPES.hdWallet,
+      });
+    }
```
peronczyk, line 244: Why not just `accountsRaw.value.push(account)` for every case?

src/types/index.ts
```diff
@@ -263,6 +264,8 @@ export interface IAccountRaw {
   type: AccountType;
   isRestored: boolean;
   protocol: Protocol;
+  address?: AccountAddress;
+  airGapPublicKey?: string;
```
peronczyk, line 268: Maybe this could be renamed to `publicKey` so we'll use this property in the future for accounts imported directly by entering the public key? There was such request from people.

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -197,6 +199,26 @@ export class EthereumAdapter extends BaseProtocolAdapter {
...
 
+  override getAccountFromRawAccount(
+    rawAccount: IAccountRaw,
+    idx: number,
+    globalIdx: number,
+    seed: Uint8Array,
+  ): IAccount {
+    if (rawAccount.type !== ACCOUNT_TYPES.hdWallet) {
+      throw new Error(`Unsupported account type: ${rawAccount.type}`);
+    }
```
peronczyk, line 210: I think this is bad to throw an error with a function called `getAccountFromRawAccount` where actually the app is supporting other account types than the hd wallet.

src/utils/common.ts
```diff
@@ -195,9 +196,14 @@ export async function fetchJson<T = any>(
...
-export function getDefaultAccountLabel({ protocol, idx }: Partial<IAccount> = {}): string {
+
+export function getDefaultAccountLabel(
+  { protocol, idx, type }: Partial<IAccount> = {},
+): string {
+  const isAirGap = type === ACCOUNT_TYPES.airGap;
   return [
-    (protocol) ? ProtocolAdapterFactory.getAdapter(protocol).protocolName : null,
+    (protocol && !isAirGap) ? ProtocolAdapterFactory.getAdapter(protocol).protocolName : null,
+    (isAirGap) ? tg('airGap.title') : null,
```
peronczyk, line 206: As I like to be future proof and I don't like to repeat the same statements I suggest to add a switch..case statement to distinguish the account types:

```
  const accountTypeName = (() => {
    switch (true) {
      case type === ACCOUNT_TYPES.airGap: return tg('airGap.title');
      default: return protocol ? ProtocolAdapterFactory.getAdapter(protocol).protocolName : null;
    }
  })();
  return [
    accountTypeName,
    tg('pages.account.heading'),
    (idx || 0) + 1,
  ]
    .filter(excludeFalsy)
    .join(' ');
```

Not 100% sure about this, but you see where I'm going? What do you think?

src/protocols/bitcoin/libs/BitcoinAdapter.ts
```diff
@@ -198,6 +201,26 @@ export class BitcoinAdapter extends BaseProtocolAdapter {
...
 
+  override getAccountFromRawAccount(
+    rawAccount: IAccountRaw,
+    idx: number,
+    globalIdx: number,
+    seed: Uint8Array,
+  ): IAccount {
+    if (rawAccount.type !== ACCOUNT_TYPES.hdWallet) {
+      throw new Error(`Unsupported account type: ${rawAccount.type}`);
+    }
```
peronczyk, line 212: IDEM

peronczyk (approved):

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -197,6 +199,26 @@ export class EthereumAdapter extends BaseProtocolAdapter {
...
+
+    const hdWallet = this.getHdWalletAccountFromMnemonicSeed(seed, idx);
```
peronczyk, line 212: With the `getAccountFromRawAccount` method I think we no longer need the `getHdWalletAccountFromMnemonicSeed` as the required abstract method in the `BaseProtocolAdapter`.

Liubov-crypto:
1. I've created multisig vault using airgap vault account as a first signer, after that I proposed a Tx from this superhero vault. I got an confirmation popup that Tx was successful but at the end no pending Tx in my sh wallet. Also I see that on the final step confirmation sign modal is missing if I compare it with multisig vault created from regular ae accounts. 

for some reason I can't attach the video((

2. This is known issue for regular multisig accounts, reproducible also here:

multi
nj

 3. When I want to set name by default I will get an modal error, but name will be set after: 

def name

4. Is the text alignment correct here? it looks located close to the edges. @onvisions could you please help me here?

align

5. I found that QR is blinking in FF extension and it's hard to scan it from airgap vault: 

https://github.com/superhero-com/superhero-wallet/assets/69896204/98d96155-b398-4a81-9002-9123c32c4a19

at the end qr code has been disappeared: 
create vault


6. Also found some strange issues in console related to scan qr. Scan qr was not working for me:
ffex

ffe

I see this issue also on prod in console (probably my local issue): 
sc err

on dev branch FF ext from 05-Jun-2024 08:59 4917797 commit f8edee5: 

dev

Liubov-crypto:
Still reproducible for me: 

1. QR scan isn't working in FF. 
2. QR is blinking in FFweb/ extension (actually it's blinking in Chrome too, but I see that in FF delay in blinking is bigger) and it's hard to scan it from airgap vault and at the end qr code has been disappeared.
3. Sometimes when I want to create vault and choose 2 signer - Create multisig vault button will not respond to me in chrome extension (idk why, probably my pc is overloaded and heated):

https://github.com/superhero-com/superhero-wallet/assets/69896204/612dd0a0-14e7-4971-ab6f-31823b0af13b

4. do we changed design qr code here? it doesn't fit for the whole box: 
qr

@onvisions could you please confirm it? 

5. This issue can be reproducible from time to time, especially when I'm proposing a Tx from vault with 0 balance: 
I've created multisig vault using airgap vault account as a first signer, after that I proposed a Tx from this superhero vault. I got an confirmation popup that Tx was successful but at the end no pending Tx in my sh wallet. Also I see that on the final step confirmation sign modal is missing if I compare it with multisig vault created from regular ae accounts.

src/popup/components/AccountCard.vue
```diff
@@ -57,9 +65,21 @@ export default defineComponent({
...
 });
 </script>
+
+<style lang="scss" scoped>
+.account-card-base {
+  .air-gap-icon {
+    width: 24px;
+    height: 24px;
+  }
+}
```
CedrikNikita, line 84:
```suggestion
.account-card-base .air-gap-icon {
  width: 24px;
  height: 24px;
}
```

src/popup/components/Modals/Help.vue
```diff
@@ -25,6 +28,42 @@ export default {
...
   },
 };
 </script>
+
+<style lang="scss">
+@use '@/styles/variables' as *;
+
+.help-modal {
+  .help-template-renderer {
+    ol {
```
CedrikNikita, line 41:
```suggestion
.help-modal .help-template-renderer ol {
```

src/popup/components/Modals/SignAirGapTransaction.vue
```diff
@@ -0,0 +1,258 @@
...
+      CopyOutlinedIcon,
+    };
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@use '@/styles/variables' as *;
+@use '@/styles/typography';
+@use '@/styles/mixins';
```
CedrikNikita, line 239:
```suggestion
```

src/popup/components/form/FormScanQrResult.vue
```diff
@@ -0,0 +1,75 @@
...
+    };
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@use '@/styles/variables' as *;
+
+.form-input-with-qr {
+  .scan-button {
```
CedrikNikita, line 68:
```suggestion
.form-input-with-qr .scan-button {
```

src/popup/components/TransferQRCodeGenerator.vue
```diff
@@ -0,0 +1,128 @@
...
+    };
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@use '@/styles/variables' as *;
+
+.transfer-qr-code-generator {
+  .btn-copy {
```
CedrikNikita, line 124:
```suggestion
.transfer-qr-code-generator .btn-copy {
```

src/popup/components/TransferQRCodeGenerator.vue
```diff
@@ -0,0 +1,128 @@
...
+      copied,
+      copyAsSingleQR,
+    };
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@use '@/styles/variables' as *;
+
```
CedrikNikita, line 122:
```suggestion
```

Liubov-crypto:
I've created new wallet for testing. 

ad 1. Fixed partly. Scan is working in web wallet but not working in FF extension. 
sc err


ad 2. QR is blinking in FF: 
qr code ff
qrc

ad 3. - not reproducible for me in new created wallet

ad 4. - updated and looks like it, but not in all places: 
sc1

but please check all buttons when created vault path and propose / revoke multisig tx: 

tr

trs

propose tx
mod

ad 5. - not reproducible for me

Liubov-crypto:
Not a big issue, but when I want to propose multisig tx on some step there will be such qr: 

propose multisig

as we agree ad 1 and ad2 will be resolved separately. The rest is LGTM.

PR #2998, opened 2024-04-24
fix(multisig): slide to active account

### LINKED ISSUE #2946: If wallet is opened on 3rd or 4th multisig account then after minimise FF extension the view of multisig will be broken
**Describe the bug**
If wallet is opened on 3rd or 4th multisig account then after minimise FF extension the view of multisig will be broken

**To Reproduce**
1. Go to multisig accounts
2. Go to 3 or 4 multisig acc 
3. Minimise and open extension again

**Expected behavior**

**Screenshots**

 - OS: MacOS
 - Browser : chrome, FF
 - Version 123

**Additional context**
Reproducible on product ver. 2.2.1 

https://github.com/superhero-com/superhero-wallet/assets/69896204/e33a6573-a209-4f8f-8cf2-f4174701e352

Image

martinkaintas:
closes #2946

### DIFF
--- a/src/popup/components/AccountSwiper.vue
+++ b/src/popup/components/AccountSwiper.vue
@@ -94,7 +94,7 @@ import {
 import { Swiper, SwiperSlide } from 'swiper/vue';
 import SwiperCore from 'swiper';
 import { Virtual } from 'swiper/modules';
-import { getAddressColor } from '@/utils';
+import { getAddressColor, watchUntilTruthy } from '@/utils';
 import { IS_MOBILE_APP, PROTOCOLS } from '@/constants';
 
 import AccountCardAdd from './AccountCardAdd.vue';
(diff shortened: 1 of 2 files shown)


src/popup/components/AccountSwiper.vue
```diff
@@ -132,10 +132,15 @@ export default defineComponent({
...
 
-    async function setCurrentSlide(idx: number, slideParams?: number) {
+    async function setCurrentSlide(idx: number, slideSpeed?: number) {
       if (idx > -1 && currentIdx.value !== idx) {
         await nextTick();
-        swiper.value.slideTo(idx, slideParams);
+        await watchUntilTruthy(swiper);
+        const didSwipe = await swiper.value.slideTo(idx, slideSpeed);
+        if (!didSwipe) {
+          // Retry if slideTo fails
```
CedrikNikita, line 141: When does this happen?

src/composables/multisigAccounts.ts
```diff
@@ -66,6 +68,12 @@ const activeMultisigAccountId = ref<AccountAddress>();
 const activeMultisigNetworkId = ref('');
 const isAdditionalInfoNeeded = ref(false);
 
+const activeMultisigAccountIdx = useStorageRef<number>(
```
CedrikNikita, line 71: While extension is not active the status of the multisig vaults can be changed (by another signer for example), that will change the order of the vaults and that will lead to opening a different vault, than was chosen in the last session.

src/popup/components/AccountSwiper.vue
martinkaintas: This can happen while user is on multisig dashboard and active account is not one of the first ones (+3). If he refreshes the page it could happen that the initial slide to last active account will not work (and will return `false`)

CedrikNikita (approved):

Liubov-crypto:
LGTM

PR #3005, opened 2024-04-26
Browser's URL input fixes

davidyuk:
This PR is supported by the Æternity Foundation

### DIFF
--- a/src/popup/components/InputField.vue
+++ b/src/popup/components/InputField.vue
@@ -64,6 +64,7 @@
             :disabled="readonly"
             :maxlength="textLimit"
             :inputmode="inputMode"
+            :autocapitalize="autoCapitalize"
             @input="handleInput"
             @keydown="checkIfNumber"
             @focusin="focused = true"
@@ -167,7 +168,12 @@ export default defineComponent({
     const inputEl = ref<HTMLInputElement | null>(null);
     const isBlinking = ref(false);
 
-    const inputMode = computed(() => props.type === 'number' ? 'decimal' : 'text');
+    const inputMode = computed(() => ({
+      number: 'decimal',
+      url: 'url',
+      text: 'text',
+    }[props.type]));
+    const autoCapitalize = computed(() => props.type === 'url' ? 'off' : undefined);
     const messageAsObject = computed(
       (): IInputMessage | null => (typeof props.message === 'object') ? props.message : { text: props.message },
     );
@@ -230,6 +236,7 @@ export default defineComponent({
       uid,
       inputId,
       inputMode,
+      autoCapitalize,
       messageAsObject,
       hasError,
       hasWarning,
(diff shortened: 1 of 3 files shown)


CedrikNikita (approved):

src/utils/validators.ts
```diff
@@ -5,7 +5,8 @@ import { toURL } from './formatters';
 export function isUrlValid(urlAsString: string): boolean {
   try {
     const url = toURL(urlAsString);
-    return ['http:', 'https:'].includes(url.protocol) && isFQDN(url.hostname);
+    return ['http:', 'https:'].includes(url.protocol)
+      && (isFQDN(url.hostname) || url.hostname === 'localhost');
```
peronczyk, line 9:
```suggestion
    return (
      ['http:', 'https:'].includes(url.protocol)
      && (isFQDN(url.hostname) || url.hostname === 'localhost')
    );
```

src/popup/pages/AppsBrowser.vue
```diff
@@ -125,6 +127,12 @@ const DAPPS_LIST = [
   },
 ];
 
+interface App {
```
peronczyk, line 130: I suggest to use it alos in line 145:

```
const selectedApp = ref<App>();
```

src/popup/pages/AppsBrowser.vue
```diff
@@ -94,7 +94,9 @@ import {
 } from 'vue';
 import { Field } from 'vee-validate';
 import { MODAL_WARNING_DAPP_BROWSER } from '@/constants';
-import { getLocalStorageItem, setLocalStorageItem, handleUnknownError } from '@/utils';
+import {
+  getLocalStorageItem, setLocalStorageItem, handleUnknownError, toURL,
```
peronczyk, line 98: I suggest to split these lines.

src/popup/components/InputField.vue
```diff
@@ -167,7 +168,12 @@ export default defineComponent({
     const inputEl = ref<HTMLInputElement | null>(null);
     const isBlinking = ref(false);
 
-    const inputMode = computed(() => props.type === 'number' ? 'decimal' : 'text');
+    const inputMode = computed(() => ({
+      number: 'decimal',
+      url: 'url',
+      text: 'text',
+    }[props.type]));
+    const autoCapitalize = computed(() => props.type === 'url' ? 'off' : undefined);
```
peronczyk, line 176: I don't fully understand what is the goal for this line. Maybe you can add comment here?

peronczyk (changes requested):

src/popup/components/InputField.vue
davidyuk: The idea is to set `autocapitalize="off"` in case `inputmode="url"`, to don't begin with a capital letter on iOS

peronczyk: Sure. Makes sense. Can you please add a comment about this?

PR #3015, opened 2024-05-05
Realign v2.2.5

CedrikNikita:
fixes #3004

PR #3018, opened 2024-05-06
fix: handle pay for tx correctly

CedrikNikita:
|develop|fix|
|---|---|
|image|image|
|image|image|

### DIFF
--- a/src/popup/components/TransactionOverview.vue
+++ b/src/popup/components/TransactionOverview.vue
@@ -100,7 +100,27 @@ export default defineComponent({
         callerId,
       } = innerTx.value;
 
-      switch (outerTxTag.value) {
+      if (outerTxTag.value === Tag.PayingForTx && innerTxTag.value === Tag.GaAttachTx) {
+        return {
+          sender: {
+            address: innerTx.value.ownerId,
+            name: getName(innerTx.value.ownerId).value,
+            url: protocolExplorer.prepareUrlForAccount(innerTx.value.ownerId),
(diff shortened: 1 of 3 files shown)


Liubov-crypto:
After I approved Tx by both signers I was not able to send it: 

tx
aa

CedrikNikita:
> After I approved Tx by both signers I was not able to send it:

This behaviour will remain till we will push the needed fix #3015

src/popup/components/TransactionOverview.vue
```diff
@@ -100,7 +100,27 @@ export default defineComponent({
...
+            label: t('common.smartContract'),
+            address: innerTx.value.contractId,
+          },
+        };
+      }
+
+      const transactionTag = (
+        outerTxTag.value === Tag.PayingForTx
+        || outerTxTag.value === Tag.GaMetaTx
+      ) ? innerTxTag.value : outerTxTag.value;
```
peronczyk, line 121: Maybe you can use `getTxTag` utility function?

peronczyk (changes requested):

src/popup/components/TransactionOverview.vue
CedrikNikita: `innerTxTag` is a `getTxTag` of `innerTx` and `outerTxTag` is `getTxTag` of `outerTx`. And `getTxTag` function will not return us tag that is required for the overview.

Liubov-crypto:
LGTM

peronczyk (approved):

PR #3022, opened 2024-05-07
perf: do not create new sdk instance on invite claim

### LINKED ISSUE #1487: Pass `onAccount` to `transferFunds` instead of  creating an sdk instance
https://github.com/aeternity/superhero-wallet/blob/5ea4993c01d3a0465b840ca08ff45f2eaa785e5f/src/store/modules/invites.js#L18-L23

I've tried to implement it in https://github.com/aeternity/superhero-wallet/pull/1396 but not able to finish

CedrikNikita:
fixes #1487

### DIFF
--- a/src/composables/invites.ts
+++ b/src/composables/invites.ts
@@ -1,8 +1,6 @@
 import {
   AE_AMOUNT_FORMATS,
-  AeSdk,
   Encoded,
-  Node,
 } from '@aeternity/aepp-sdk';
 import type { AccountAddress, IInvite } from '@/types';
 import { STORAGE_KEYS } from '@/constants';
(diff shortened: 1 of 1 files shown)


CedrikNikita:
> I have a question, if I'm creating an invite link in testnet and after trying to claim it in new wallet, I'm getting a pop-up that gift card already claimed, but in fact it's not because I'm always login first in mainnet and then have to switch to testnet to claim invite link. Should we warn user about network discrepancy?

Invite links were invented mostly for mainnet. Feel free to open an issue about that, but I think such an addition would  have very low priority.

peronczyk (approved):

PR #3023, opened 2024-05-07
refactor: use `isNameValid` from sdk instead of custom

### LINKED ISSUE #2118: Allow register of names with all characters allowed in names spec
It's being worked on in https://github.com/aeternity/superhero-wallet/pull/1671

I'm going to support it in the next release of SDK after merging https://github.com/aeternity/aepp-sdk-js/pull/1840

Consider replacing
https://github.com/aeternity/superhero-wallet/blob/02f012500c1dca0593771b2cdcca820f99c891af/src/popup/utils/utils.ts#L240-L245
with `isNameValid` from SDK.

Also, ensure that fonts look correct when using Cyrillic chars or so in names.

CedrikNikita:
fixes #2118


Liubov-crypto:
I've successfully registered and made a default a chain name using Cyrillic symbols, LGTM.

peronczyk (approved):

PR #3031, opened 2024-05-14
fix(wallet-connect): be able to send transaction

CedrikNikita:


src/composables/walletConnect.ts
```diff
@@ -91,14 +92,26 @@ export function useWalletConnect() {
...
+        if (adapter?.transferToken) {
+          await updateFeeList();
+          if (adapter.transferPreparedTransaction) {
+            const actionResult = await adapter.transferPreparedTransaction(
+              params,
+              {
+                maxPriorityFeePerGas: maxPriorityFeePerGas.value?.toFormat(ETH_COIN_PRECISION),
+                maxFeePerGas: maxFeePerGas.value?.toFormat(ETH_COIN_PRECISION),
+                fromAccount: params.from,
+              },
```
peronczyk, line 110: Maybe this stuff could be calculated in the `transferPreparedTransaction` method?

peronczyk (approved):

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -245,6 +246,53 @@ export class EthereumAdapter extends BaseProtocolAdapter {
     }
   }
 
+  override async transferPreparedTransaction(transactionData: any): Promise<ITransferResponse> {
```
peronczyk, line 249: Maybe you can destructure the argument properties so there would be less code in the function's body?

PR #3043, opened 2024-05-16
refactor: replace browser qr code reader

### LINKED ISSUE #3035: Web QR Code reader does not recognize Wallet Connect QR codes

martinkaintas:
closes #3035


src/popup/components/Modals/QrCodeReader.vue
```diff
@@ -217,14 +209,11 @@ export default defineComponent({
         return;
       }
 
+      isCameraReady.value = true;
```
peronczyk, line 212: Now as we don't need to wait for initializing the QR code reader I think we can refactor the code below in a following way:
```
props.resolve((IS_MOBILE_APP) ? await scanMobile() : await scanWeb());
```
Also the TODO comment can be removed.

peronczyk (approved):

Liubov-crypto:
Idk if it's correct but I'm getting this message after I scan qr code: 

2024-05-16 5 56 31 пп

PR #3064, opened 2024-05-24
refactor: app name as constant

peronczyk:


martinkaintas (approved):
Code changes seem reasonable, but snapshot tests should be checked & updated

PR #3065, opened 2024-05-24
feat: custom panel table item component

peronczyk:
I need to reuse this table item component in the WalletConnect modal.

### DIFF
--- a/src/popup/components/PanelTableItem.vue
+++ b/src/popup/components/PanelTableItem.vue
@@ -1,0 +1,119 @@
+<template>
+  <Component
+    :is="(href) ? 'LinkButton' : 'div'"
+    :to="href"
+    :is-external="isExternalLink"
+    :class="{
+      clickable: !!href,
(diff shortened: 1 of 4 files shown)


src/popup/components/PanelTableItem.vue
```diff
@@ -0,0 +1,119 @@
...
+  props: {
+    name: { type: String, required: true },
+    href: { type: String, default: null },
+    isExternalLink: Boolean,
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@use '@/styles/variables';
```
martinkaintas, line 39:
```suggestion
@use '@/styles/variables' as *;
```

martinkaintas (approved):
In general LGTM

src/popup/components/PanelTableItem.vue
peronczyk: Makes sense, updated.

PR #3071, opened 2024-05-27
Release v2.2.6

CedrikNikita:


Liubov-crypto:
1. I got a multisig issue a couple of times: 

multisig
multiissue
rel

https://github.com/superhero-com/superhero-wallet/assets/69896204/516ff985-1a22-4084-b493-e3c3b217ad02

2. Transactions are loading very slow for me (probably not an issue, just have to report): 

tx are loading very long for me

3. If I want to set name by default it also not respond for first click, Idk why but I have to wait few minutes, probably some issue with AENS, It can be set but with a delay: 

name1

martinkaintas (approved):

Liubov-crypto:
Ios app: 

1. I got the same issue with multisig as in web and extension. 


2. I can't send very big amount of FT (this is reproducible on web too): 
big int

https://github.com/superhero-com/superhero-wallet/assets/69896204/ff8f62ba-73a3-40c1-9811-31f38ca33f4f

https://aescan.io/contracts/ct_2BtHZSY9SfmiDQZ9yGqPppyRwKmrhHFa7yUJTpYMsWnJr23r5U

probably should be tracked separately

3. If I'm sending ETH and after ERC-20 tokens in Tx details I will see ERC-20 instead of ETH: 

https://github.com/superhero-com/superhero-wallet/assets/69896204/c61af073-2a1f-4ed0-bc8c-c8089f65b6a2

I think we need discuss this case on our call with displaying fee paid in ETH.

smaroudasunicorn (approved):

Issue #3057, opened 2024-05-22
Update display of Tx amount, gas, fee, total. Sub-case: multisig tx proposal template.


onvisions:
@CedrikNikita @smaroudas 

**Story:** We need design update of the display of transaction **Amount + Gas used cost + Fee = Total.**

**Acceptance criteria:**
- implement the feature following design below. Recheck margins between elements and update where needed;
- implement the design for sub-case of multisig transaction proposal where user should see the Proposed transaction amount and fee + Proposal gas cost, fee and Total. See design below;
- Update the template for Multisig Tx Proposal details accordingly. See design below.

**Design:**

**_Send Funds flow:_**

![Image](https://github.com/superhero-com/superhero-wallet/assets/12225410/5179214b-2445-4e39-bf31-13f868e02593)
![Image](https://github.com/superhero-com/superhero-wallet/assets/12225410/56c4af59-a75f-469e-82e0-f12b92340dcc)
![Image](https://github.com/superhero-com/superhero-wallet/assets/12225410/cbe1d523-aff4-4ee1-9942-2cd9aa567d25)

.................................................................................

**_Multisig tx proposal:_**

![Image](https://github.com/superhero-com/superhero-wallet/assets/12225410/086b8095-2a63-433f-b4e8-5502e1dcecb6)
![Image](https://github.com/superhero-com/superhero-wallet/assets/12225410/632f8bd6-0f97-42af-84bf-6e8d3e8e8183)
![Image](https://github.com/superhero-com/superhero-wallet/assets/12225410/09a24f18-8898-40ae-a999-0905acb5e9d2)
![Image](https://github.com/superhero-com/superhero-wallet/assets/12225410/9164cadc-0925-47d0-bbcc-3636a0e22fcc)

**Figma reference:**

1. _**Send Funds Flow:**_ 

https://www.figma.com/design/3oGLWzSH0oJljo4RETZtur/Superhero-Wallet-UI-(%E2%9C%94%EF%B8%8FUpdated)?node-id=32135-195480&t=4Ghuxw4IqUzgRRKi-0

2. _**Multisig Transaction Proposal, Review, Details:**_ 

https://www.figma.com/design/3oGLWzSH0oJljo4RETZtur/Superhero-Wallet-UI-(%E2%9C%94%EF%B8%8FUpdated)?node-id=32135-202575&t=VkSNKnzv1XJGXQ1H-0

smaroudasunicorn:
Screen Review Transaction
[...]

PR #3073, opened 2024-05-27
Improve total calculation transfer modal

### LINKED ISSUE #2031: Multisig Vault: logic behind "Insufficient balance" message is incorrect
**Describe the bug**
**Related to a multisig feature only**
Currently the logic behind "insufficient balance" message underneath the amount in the `sendModal` (needed to propose the transactions) is incorrect. The comparison of the signer balance is happening to the lower fee than actually needed for transaction. If the signer balance would be:
`comparisonFee <= signerBalance < actuallProposalFee`
This would lead to an issue and transaction will not be proposed. Also user won't see any modal with an explanation, that is leading to a confusion, however no AEs would be spend.

**To Reproduce**
propose a transaction with a signer's balance lower than `0.000182940`, but more than `0.00002`,

**Expected behavior**
Suppose to handle such cases correctly, and show "insufficient balance" message underneath amount field.

**Platform**
Is happening on every platform.

### LINKED ISSUE #3040: Calculate fee for a transaction proposal
https://aeternity-blockchain.atlassian.net/browse/SW-614

Description

Currently the fee we are showing is a constant

JustNick 
June 29, 2023 at 10:04 AM

Paolo’s decision was to postpone this implementation for two reasons:
- wait till sdk 13 will be used in this process;
- value of the fix isn’t that big as amount of time needed to be spend with sdk 11;

CedrikNikita:
fixes #2031,
fixes #3040,
part of #3057

### DIFF
--- a/src/popup/components/Modals/ConfirmTransactionSign.vue
+++ b/src/popup/components/Modals/ConfirmTransactionSign.vue
@@ -90,6 +90,39 @@
           />
         </DetailsItem>
 
+        <DetailsItem
+          v-if="popupProps?.tx?.gasLimit"
+          :value="popupProps.tx.gasLimit"
+          :label="$t('transaction.gasLimit')"
+          data-cy="gas"
+        />
+        <DetailsItem
+          v-if="gasPrice"
+          :label="$t('pages.transactionDetails.gasPrice')"
+          data-cy="gas-price"
+        >
+          <template #value>
+            <TokenAmount
+              :amount="+aettosToAe(gasPrice)"
+              :symbol="AE_SYMBOL"
+              :protocol="PROTOCOLS.aeternity"
+            />
+          </template>
+        </DetailsItem>
+        <DetailsItem
+          v-if="gasCost"
+          :label="$t('transaction.gasCost')"
+          data-cy="gas-cost"
+        >
+          <template #value>
+            <TokenAmount
+              :amount="+aettosToAe(gasCost)"
+              :symbol="AE_SYMBOL"
+              :protocol="PROTOCOLS.aeternity"
+            />
+          </template>
+        </DetailsItem>
+
         <DetailsItem :label="$t('transaction.fee')">
           <TokenAmount
             :amount="fee"
@@ -225,7 +258,9 @@ import {
   usePopupProps,
   useTransactionData,
 } from '@/composables';
+import { AE_SYMBOL } from '@/protocols/aeternity/config';
 import {
+  aettosToAe,
   getAeFee,
   getTransactionTokenInfoResolver,
 } from '@/protocols/aeternity/helpers';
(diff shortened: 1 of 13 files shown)


src/protocols/aeternity/views/TransactionDetails.vue
```diff
@@ -203,6 +215,7 @@ import DialogBox from '@/popup/components/DialogBox.vue';
 import Avatar from '@/popup/components/Avatar.vue';
 import TransactionAssetRows from '@/popup/components/TransactionAssetRows.vue';
 import TokenAmount from '@/popup/components/TokenAmount.vue';
+import BigNumber from 'bignumber.js';
```
peronczyk, line 218: This should be placed at the top of imports.

Liubov-crypto:
LGTM. 
But wallet is extremely slow for me.

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -90,6 +90,39 @@
...
         </DetailsItem>
 
+        <DetailsItem
+          v-if="popupProps?.tx?.gasLimit"
+          :value="popupProps.tx.gasLimit"
+          :label="$t('transaction.gasLimit')"
+          data-cy="gas"
+        />
+        <DetailsItem
+          v-if="gasPrice"
```
martinkaintas, line 100: `gasPrice` could also exist on Ethereum transactions
If we want those `<DetailsItem>`'s only on for aeternity transactions I suggest to wrap them in a bigger `v-if` and check for protocol. But I believe it would be best to generalize this for all protocols

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -90,6 +90,39 @@
...
+              :amount="+aettosToAe(gasPrice)"
+              :symbol="AE_SYMBOL"
+              :protocol="PROTOCOLS.aeternity"
+            />
+          </template>
+        </DetailsItem>
+        <DetailsItem
+          v-if="gasCost"
+          :label="$t('transaction.gasCost')"
+          data-cy="gas-price"
```
martinkaintas, line 115: The same `data-cy="gas-price"` is used above for `gasPrice`. Is this intentional?

src/popup/components/Modals/ConfirmTransactionSign.vue
CedrikNikita: Currently gas price calculation is happening only for the aeternity.

src/protocols/aeternity/components/TransferReview.vue
```diff
@@ -451,9 +512,15 @@ export default defineComponent({
 </script>
 
 <style scoped lang="scss">
+@use '../../../styles/variables' as *;
```
martinkaintas, line 515:
```suggestion
@use '@/styles/variables' as *;
```

martinkaintas (approved):

PR #3086, opened 2024-05-30
feat: show fee in transaction list for token transfers

### LINKED ISSUE #3084: Enrich Token Sent Transactions
For example

Image

sent tx for AEX-9 token with transaction fee in AE displayed as well but instead the AEX-9 token should be on top and AE below

This shall apply for ERC-20 tokens too.

martinkaintas:
closes #3084
Latest Transactions list remains unchanged by this

### DIFF
--- a/src/composables/transactionData.ts
+++ b/src/composables/transactionData.ts
@@ -26,6 +26,7 @@ import {
   TX_FUNCTIONS_TYPE_DEX,
 } from '@/protocols/aeternity/config';
 import {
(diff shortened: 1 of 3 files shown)


CedrikNikita:
Should fee be presented in the swap transaction as well?
image

martinkaintas:
> Should fee be presented in the swap transaction as well? 

The scope of this task is only about token transfers

CedrikNikita (approved):

PR #3087, opened 2024-06-03
refactor: transaction overview remove unused props

peronczyk:


src/popup/components/TransactionOverview.vue
```diff
@@ -1,10 +1,8 @@
 <template>
   <TransactionInfo
     class="transaction-overview"
-    :title="preparedTransaction.title"
-    :sender="preparedTransaction.sender"
-    :recipient="preparedTransaction.recipient"
-    :transaction-function="preparedTransaction.function"
+    :sender="transactionParties.sender"
+    :recipient="transactionParties.recipient"
```
peronczyk, line 5: These 2 properties and `transaction` are the only ones that `TransactionInfo` component can receive. The rest was unused and there was no need to calculate them.

PR #3103, opened 2024-06-12
Feat/address book page

### LINKED ISSUE #3036: Create an Address Book for Contacts
Address Book to be stored locally on Device. 
Import/Export functionality shall also be present.

martinkaintas:
closes #3036


package.json
```diff
@@ -199,4 +203,4 @@
       }
     ]
   }
-}
+}
```
CedrikNikita, line 206:
```suggestion
}

```

src/popup/components/AddressBook/AddressBookList.vue
```diff
@@ -0,0 +1,144 @@
...
+      isScrolled.value = (scrollWrapper.value as HTMLElement)?.scrollTop > 0;
+      emit('update:modelValue', (isScrolled.value || !!searchQuery.value));
+    }
+    const throttledHandleScroll = throttle(handleScroll, 100);
+
+    onMounted(async () => {
+      scrollWrapper.value?.addEventListener('scroll', throttledHandleScroll);
+    });
+
+    onBeforeUnmount(() => {
```
CedrikNikita, line 90: This will not trigger while you are on `Address book` page upon clicking on `Add` button or going to edit mode. Also it won't trigger if user will click close button in these `Add` and  `Edit` pages.

CedrikNikita (changes requested):
Address book edit

peronczyk (changes requested):

src/icons/import-address-book.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
peronczyk, line 1: I suggest to remove the width and height attributes.

src/icons/export-address-book.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
peronczyk, line 1: I suggest to remove the width and height attributes.

src/types/index.ts
```diff
@@ -867,3 +868,15 @@ export interface IAmountDecimalPlaces {
...
+  id: string;
+  name: string;
+  address: AccountAddress;
+  isBookmarked: boolean;
+  protocol: Protocol;
+}
+
+export type IAddressBook = Record<string, IAddressBookEntry>;
+
+export type IAddressBookFilter = ObjectValues<typeof ADDRESS_BOOK_FILTERS> | Protocol | undefined;
```
peronczyk, line 882: I'm not a big fan of mixing types together. I think better solution would be to split the filter into 2 variables: `isBookmarked` (type: bool) and `protocol` (type: Protocol | null).

peronczyk (changes requested):

package.json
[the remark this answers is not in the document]
martinkaintas: typescript version was bumped to 5.4.5 and it was necessary to raise the test packages as well

src/popup/components/AddressBook/AddressBookList.vue
```diff
@@ -0,0 +1,155 @@
...
+
+  .address-book-header {
+    box-shadow: none;
+  }
+
+  .search-field {
+    margin-bottom: 16px;
+  }
+
+  .list{
```
peronczyk, line 138:
```suggestion
  .list {
```

src/popup/components/AddressBook/AddressBookList.vue
```diff
@@ -0,0 +1,155 @@
...
+  .message {
+      @extend %face-sans-15-medium;
+
+      display: flex;
+      justify-content: center;
+      align-items: center;
+      color: $color-grey-light;
+      text-align: center;
+      padding: 40px 8px;
+    }
```
peronczyk, line 153: Broken indentation

peronczyk (changes requested):

src/composables/addressBook.ts
```diff
@@ -0,0 +1,229 @@
...
+  );
+
+  function setFilter(filter: IAddressBookFilter) {
+    activeFilter.value = filter;
+  }
+  function setShowBookmarked(value: boolean) {
+    showBookmarked.value = value;
+  }
+
+  function addressBookAddressExists(address: AccountAddress) {
```
peronczyk, line 83: What do you think about renaming this to `isAddressInAddressBook` or `isAddressAddedToAddressBook`?

src/types/index.ts
```diff
@@ -890,3 +891,18 @@ export interface IAmountDecimalPlaces {
...
+  name: string;
+  address: AccountAddress;
+  isBookmarked: boolean;
+  protocol: Protocol;
+}
+
+/**
+ * The `string` stands for the id of the entry.
+ */
+export type IAddressBook = Record<string, IAddressBookEntry>;
```
peronczyk, line 906: After spending some time with the PR and having a conversation on the chat I'm sure that using random generated IDs for keying the entries is rather bad (superfluous). Take a look on how the network management is done.

I think the address or eventually the index is enough to allow to distinguish the entries between each other.

So the entries could be stored as an array instead of object.

tsconfig.json
```diff
@@ -14,7 +14,8 @@
     "strict": true,
     "strictFunctionTypes": false,
     "skipLibCheck": true,
-    "sourceMap": true,
+    // temporary workaround @see https://github.com/cypress-io/cypress/issues/26203#issuecomment-2152558848
+    "sourceMap": false,
```
peronczyk, line 18: I don't like the fact that including `vueUse` library required to raise the TS version, the testing framework versions and led to this source map disabling.
If the vueUse lib was included only to open the file upload window I think it woud;d be better to have our own tiny util function for this.

What do you think @CedrikNikita ?

src/utils/common.ts
```diff
@@ -527,3 +530,20 @@ export const toBase64Url = (data: Buffer | Uint8Array | string): string => Buffe
...
+  return (
+    ProtocolAdapterFactory.getAdapterByAccountAddress(address, NETWORK_TYPE_MAINNET)
+    || ProtocolAdapterFactory.getAdapterByAccountAddress(address, NETWORK_TYPE_TESTNET)
+  )?.protocol;
+}
+
+/**
+ * Generate a simple random string id.
+ */
+export function generateId() {
```
peronczyk, line 547: If the address book entries would be keyed by the address or index this util can be removed.

src/popup/plugins/veeValidate.ts
```diff
@@ -114,6 +120,25 @@ defineRule(
...
+    return (key && key !== savedKey)
+      ? tg('validation.addressBookEntryExists', [entryTypeText.toLowerCase()])
+      : true;
+  },
+);
+
+defineRule(
+  'is_address_valid',
+  (value: string) => getProtocolByAddress(value) ? true : tg('validation.invalidAddress'),
+);
```
peronczyk, line 140: We have similar rule: `account_address`.

src/popup/pages/More.vue
```diff
@@ -12,6 +12,17 @@
           </template>
         </PanelItem>
 
+        <PanelItem
+          :to="{ name: 'address-book' }"
```
peronczyk, line 16: I suggest to use constant for the route name.

src/popup/plugins/veeValidate.ts
martinkaintas: `account_address` validates for a given protocol and network type where `is_address_valid` will check if the address is valid in any of the supported protocols/networks

tsconfig.json
CedrikNikita: Raising TS version suppose to be heavily tested i guess. I guess it is ok to check custom function if it proves to be more efficient and maintainable. Would you be interested in collaborating on a solution?"

martinkaintas: I suppose you're right

src/types/index.ts
```diff
@@ -890,3 +890,15 @@ export interface IAmountDecimalPlaces {
...
+  name: string;
+  address: AccountAddress;
+  isBookmarked: boolean;
+  protocol: Protocol;
+}
+
+/**
+ * The `string` stands for the id of the entry.
+ */
+export type IAddressBook = IAddressBookEntry[];
```
peronczyk, line 904: I think this type can be removed as it's not even an interface.

src/popup/components/buttons/BtnFilter.vue
```diff
@@ -0,0 +1,43 @@
...
+
+<style lang="scss" scoped>
+@use '@/styles/variables' as *;
+@use '@/styles/typography';
+
+.btn-filter {
+    @extend %face-sans-14-bold;
+    line-height: 20px;
+    padding: 4px 12px;
+    border: 1px solid transparent;
```
peronczyk, line 34: Wrong indentation.

src/popup/components/AddressBook/AddressBookFilters.vue
```diff
@@ -0,0 +1,93 @@
...
+.address-book-filters {
+  display: flex;
+  width: 100%;
+  gap: 8px;
+
+  .divider {
+    border-left: 1px solid rgba($color-white, 0.15);
+  }
+}
+
```
peronczyk, line 92: This line can be removed

src/composables/addressBook.ts
```diff
@@ -0,0 +1,234 @@
...
+  isBookmarked?: boolean;
+}
+
+const addressBook = useStorageRef<IAddressBook>([], STORAGE_KEYS.addressBook);
+
+export const useAddressBook = createCustomScopedComposable(() => {
+  const { openModal, openDefaultModal } = useModals();
+  const { openFileDialog } = useFileDialog();
+
+  const activeFilter = ref<Protocol | null>(null);
```
peronczyk, line 38: I suggest to rename this to `protocolFilter` (or something similar) as it holds only protocol value.

src/composables/fileDialog.ts
peronczyk, line 1: I think having such a composable is a little bit overkill. Why not just simple util function? I know that VueUse uses similar approach, but do we really need it for one use case?

Especially as this composable uses `onMounted` and `onUnmounted` hooks which does not work properly with Ionic.

peronczyk (changes requested):

src/composables/addressBook.ts
```diff
@@ -0,0 +1,234 @@
...
+      ? entries.filter((entry) => entry.protocol === activeFilter.value)
+      : entries;
+  }
+  function filterAddressBookBySearchPhrase(entries: IAddressBookEntry[]) {
+    const searchQueryLower = searchQuery.value.toLowerCase();
+    return entries.filter((entry) => {
+      const nameLower = entry.name.toLowerCase();
+      const addressLower = entry.address.toLowerCase();
+      return nameLower.includes(searchQueryLower) || addressLower.includes(searchQueryLower);
+    });
```
peronczyk, line 56: What do you think about this approach:

```suggestion
    return entries.filter(({ name, address }) => [name, address].some((val) => val.toLowerCase().includes(searchQueryLower);
```

src/composables/addressBook.ts
```diff
@@ -0,0 +1,234 @@
...
+  const addressBookFiltered = computed(
+    () => pipe([
+      filterAddressBookByBookmarked,
+      filterAddressBook,
+      filterAddressBookBySearchPhrase,
+      sortAddressBookByName,
+    ])(Object.values(addressBook.value)),
+  );
+
+  function setFilter(filter: Protocol | null) {
```
peronczyk, line 71: Suggestion:

```suggestion
  function filterByProtocol(filter: Protocol | null) {
```

src/composables/addressBook.ts
```diff
@@ -0,0 +1,234 @@
...
+    showBookmarked.value = value;
+  }
+
+  function clearFilters() {
+    showBookmarked.value = false;
+    activeFilter.value = null;
+  }
+
+  function getEntryIndexByAddress(address: AccountAddress) {
+    return Object.values(addressBook.value).findIndex((entry) => entry.address === address);
```
peronczyk, line 87: Is this correct to run `Object.values` on an array? `addressBook.value` is an array as far as I see.

src/composables/addressBook.ts
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  function isAddressInAddressBook(address: AccountAddress) {
+    return getEntryIndexByAddress(address) !== -1;
+  }
+
+  function getAddressBookEntryByAddress(address: AccountAddress) {
+    return addressBook.value?.[getEntryIndexByAddress(address)];
```
peronczyk, line 95: I suggest to use `Array.find` here.

src/composables/addressBook.ts
```diff
@@ -0,0 +1,234 @@
...
+  function removeAddressBookEntry(accountAddress: AccountAddress) {
+    const idx = getEntryIndexByAddress(accountAddress);
+    if (idx > -1) {
+      addressBook.value.splice(idx, 1);
+    }
+  }
+
+  function toggleBookmarkAddressBookEntry(address: AccountAddress) {
+    const idx = getEntryIndexByAddress(address);
+    if (addressBook.value?.[idx]) {
```
peronczyk, line 107:
```suggestion
    if (getAddressBookEntryByAddress(address)) {
```

src/composables/addressBook.ts
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    const protocol = getProtocolByAddress(address);
+    if (!protocol) {
+      throw new AddressBookInvalidAddress();
+    }
+    const savedEntryIdx = savedEntryAddress ? getEntryIndexByAddress(savedEntryAddress) : null;
+    const defaultIsBookmarked = savedEntryIdx
+      ? addressBook.value[savedEntryIdx].isBookmarked
+      : false;
```
peronczyk, line 135:
```suggestion
    const defaultIsBookmarked = !!getAddressBookEntryByAddress(savedEntryAddress)?.isBookmarked;
```

src/composables/addressBook.ts
```diff
@@ -0,0 +1,234 @@
...
+    showBookmarked.value = false;
+    activeFilter.value = null;
+  }
+
+  function getEntryIndexByAddress(address: AccountAddress) {
+    return Object.values(addressBook.value).findIndex((entry) => entry.address === address);
+  }
+
+  function isAddressInAddressBook(address: AccountAddress) {
+    return getEntryIndexByAddress(address) !== -1;
```
peronczyk, line 91:
```suggestion
    return !!getAddressBookEntryByAddress(address);
```

src/utils/fileDialog.ts
```diff
@@ -0,0 +1,39 @@
...
+function createInputEl(): HTMLInputElement {
+  const input = document.createElement('input');
+  input.type = 'file';
+  input.hidden = true;
+  input.className = 'hidden';
+  document.body.appendChild(input);
+  return input;
+}
+
+export function openFileDialog(opts: IFileDialogOptions = defaultDialogOptions): Promise<void> {
```
peronczyk, line 22: Few more suggestions:
1. I still think this util could be shrinked and placed in the global util files. Heres how:
2. The function in my opinion should not return `Promise<void>` but `Promise<FileList>`. So It should be used in a following way:

```
try {
  const files = await selectFiles();
} catch {
  // User clicked cancel or something else... we do nothing?
}
```
3. With this approach there would be no need to have separate definition for the options and const for default options. It can go this way:
```
export function selectFiles(options: { multiple?: boolean; accept?: string } = { multiple: false, accept: '*' }) {
```
4. Also I don't see much value in having separate util `createInputEl`.

src/utils/fileDialog.ts
```diff
@@ -0,0 +1,39 @@
...
+  return new Promise<void>((resolve) => {
+    const inputEl = createInputEl();
+    const options = { ...opts };
+
+    inputEl.accept = options.accept!;
+    inputEl.multiple = options.multiple!;
```
peronczyk, line 28: If you will follow my suggestions the excalamation marks won't be needed.

src/popup/components/AccountInfo.vue
```diff
@@ -122,6 +124,8 @@ export default defineComponent({
 @use '@/styles/typography';
 
 .account-info {
+  --maxWidth: 250px;
```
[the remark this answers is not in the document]
martinkaintas: It's actually used to set the default `maxWidth` for `.account-details` 
Then if we need to have a smaller `maxWidth` somewhere, e.g. `AddressBookForm` we can change this variable

src/utils/fileDialog.ts
martinkaintas: they are still required

src/popup/components/AccountInfo.vue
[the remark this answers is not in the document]
martinkaintas: It is actually setting the default `maxWidth` for `.account-details`
So if we need a smaller `maxWidth` somewhere, e.g. `AddressBookForm` we can change this variable

src/composables/addressBook.ts
```diff
@@ -0,0 +1,220 @@
...
+      a.dataset.downloadurl = ['text/json', a.download, a.href].join(':');
+      a.click();
+    }
+  }
+
+  async function importAddressBook() {
+    try {
+      const files = await selectFiles({
+        accept: '.json',
+        multiple: false,
```
peronczyk, line 195: This is the default value so it can be skipped.

peronczyk (approved):

src/utils/fileDialog.ts
peronczyk: No if you'll make the options required by removing the question marks from the options typing.

martinkaintas: If I make them required then what's the point of default values?

PR #3108, opened 2024-06-13
refactor: transaction data composable amount calculation

peronczyk:


src/composables/transactionData.ts
```diff
@@ -182,11 +189,65 @@ export function useTransactionData({
...
+   */
+  const amount = computed((): number => (protocol.value === PROTOCOLS.aeternity)
+    ? getAmountFromAeTx()
+    : getAmountFromCommonTx());
+
+  /**
+   * Amount with the fee added when receiving the assets
+   */
+  const amountTotal = computed((): number => {
+    const isReceived = direction.value === TX_DIRECTION.received;
```
martinkaintas, line 246: `isReceived` could be a computed property and be used throughout the composable. It could also be exported from the composable and used in other places like `TransactionDetails`

PR #3111, opened 2024-06-17
fix: remove window browser polyfill from ext background

peronczyk:
This PR fixes 2 issues that happens in the chrome extension's background process.

1. "window is not defined" - happening because we are including the polyfills for the browser which attempts to add the `browser` into `window` object, which does not exist in the background process.
image

2. "Only a single offscreen document may be created." - happening because we are running `browser.runtime.getURL` twice.
image


CedrikNikita (approved):

martinkaintas:
Please also test this on FF as FF handles offscreen tabs different to chrome

Liubov-crypto:
LGTM

PR #3115, opened 2024-06-19
chore: change backend url

CedrikNikita:


Liubov-crypto (approved):
The names could disappear from the main network for seconds and I would only see acc1, acc2, etc. But after that they will become visible again. 
er1



https://github.com/superhero-com/superhero-wallet/assets/69896204/31d73133-b306-494d-89f7-f121c07612f7


the rest is LGTM

PR #3118, opened 2024-06-20
feat(address-book): add share address button

### LINKED ISSUE #3085: Share my address as a QR Code to another wallet

martinkaintas:
closes #3085


src/popup/components/AccountDetailsBase.vue
```diff
@@ -253,6 +259,16 @@ export default defineComponent({
     width: 100%;
     margin-top: 20px;
     padding-inline: var(--screen-padding-x);
+
+    &:has(> :nth-child(5)) { // 5 or more buttons
+      overflow: auto;
+      white-space: nowrap;
+
+      >* {
```
peronczyk, line 267:
```suggestion
      > * {
```

src/popup/components/OpenShareAddressModalButton.vue
```diff
@@ -0,0 +1,46 @@
...
+    :disabled="disabled"
+    data-cy="share-address"
+    @click="shareAddress()"
+  />
+</template>
+
+<script lang="ts">
+import { PropType, defineComponent } from 'vue';
+
+import { AccountAddress, Protocol } from '@/types';
```
peronczyk, line 15:
```suggestion
import type { AccountAddress, Protocol } from '@/types';
```

src/popup/components/OpenShareAddressModalButton.vue
```diff
@@ -0,0 +1,46 @@
...
+import { useModals } from '@/composables';
+
+import BtnBox from './buttons/BtnBox.vue';
+
+import QrIcon from '../../icons/qr-code.svg?vue-component';
```
peronczyk, line 20:
```suggestion
import QrIcon from '@/icons/qr-code.svg?vue-component';
```

src/popup/components/ShareAddress.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+  <ShareAddressBase
+    v-bind="$attrs"
```
peronczyk, line 3: I suggest to avoid passing all props to child components this way because we are losing typings. Maybe you can follow the same pattern used in `src/protocols/ethereum/views/TransferSendModal.vue`?

src/popup/pages/AddressBookForm.vue
```diff
@@ -248,6 +252,22 @@ export default defineComponent({
...
     }
 
+    function shareAddress() {
+      if (
+        isPlaceholder.value
+        || !formModel.value.address
+        || !formModel.value.name
+        || !protocol.value) {
+        return;
+      }
```
peronczyk, line 262:
```suggestion
        || !protocol.value
      ) {
        return;
      }
```

Also I don't think "early return" is necessary here as there is only one function fired. I think early returns should be used only to avoid too many nested ifs.

peronczyk (changes requested):

src/popup/components/OpenShareAddressModalButton.vue
peronczyk, line 1: I suggest to rename this component to use "Btn" instead of "Button" as we are naming the buttons as "Btn" :)

src/popup/components/Modals/ShareAddressBase.vue
```diff
@@ -266,7 +273,7 @@ export default defineComponent({
 @use '@/styles/share-info';
 @use '@/styles/mixins';
 
-.transfer-receive-base {
+.address-share-base {
```
peronczyk, line 276: The root class name should match the component name.

peronczyk (approved):

src/popup/components/OpenShareAddressModalButton.vue
martinkaintas: See `OpenTransferReceiveModalButton` & `OpenTransferSendModalButton`

I think we should either rename all of those buttons or keep those specific buttons as "Button"

peronczyk: Not an important thing so I leave the decision to you. If I may choose I would go with being consistent and name everything `...Btn`.

PR #3121, opened 2024-06-21
feat(address-book): send to contact

### LINKED ISSUE #3104: Address Book: Send to address book record
> **Address Book Design ** _(updated after feedback 31.05)_ @smaroudasunicorn @martinkaintas 
> **4. Select address from Address Book modal (Send funds flow):**
> 
> 
> 
> Image
> 
> 
> 
> Image
> 
> 
> 
> Select account address from address book modal:
> 
> 
> 
> Image
> 
> 
> 
> Image
> 
> 
> 
> Image
> 
> 
> 
> Image
> 
> 
> 
> Image
> 
> 
> 
> 
> 
> 6. Send funds using Address book prototype recording: 
> 
> 
> 
> https://github.com/superhero-com/superhero-wallet/assets/12225410/a2b96955-fe12-4e49-9bbf-b78d5fcbdd0e
> 
> 
> 
> 
> 
> 
> **Figma reference:** https://www.figma.com/design/3oGLWzSH0oJljo4RETZtur/Superhero-Wallet-UI-(%E2%9C%94%EF%B8%8FUpdated)?node-id=31935-180432&t=gZlJbeKQFrjFsfqj-0 

 _Originally posted by @onvisions in [#3036](https://github.com/superhero-com/superhero-wallet/issues/3036#issuecomment-2142124204)_

martinkaintas:
closes #3104

### DIFF
--- a/src/composables/addressBook.ts
+++ b/src/composables/addressBook.ts
@@ -1,4 +1,4 @@
-import { computed, ref } from 'vue';
+import { computed, ref, watch } from 'vue';
 import { Directory, Filesystem } from '@capacitor/filesystem';
 import { Capacitor } from '@capacitor/core';
 
@@ -40,8 +40,8 @@ export const useAddressBook = createCustomScopedComposable(() => {
   function filterAddressBookByBookmarked(entries: IAddressBookEntry[]) {
     return showBookmarked.value ? entries.filter((entry) => entry.isBookmarked) : entries;
   }
-  function filterAddressBook(entries: IAddressBookEntry[]) {
-    return protocolFilter.value && !showBookmarked.value
+  function filterAddressBookByProtocol(entries: IAddressBookEntry[]) {
+    return protocolFilter.value
       ? entries.filter((entry) => entry.protocol === protocolFilter.value)
       : entries;
   }
@@ -58,10 +58,14 @@ export const useAddressBook = createCustomScopedComposable(() => {
   const addressBookFiltered = computed(
     () => pipe([
       filterAddressBookByBookmarked,
-      filterAddressBook,
+      filterAddressBookByProtocol,
       filterAddressBookBySearchPhrase,
       sortAddressBookByName,
     ])(Object.values(addressBook.value)),
+  );
+
+  const addressBookFilteredByProtocol = computed(
+    () => filterAddressBookByProtocol(Object.values(addressBook.value)),
   );
 
   function setProtocolFilter(filter: Protocol | null) {
@@ -69,14 +73,18 @@ export const useAddressBook = createCustomScopedComposable(() => {
     protocolFilter.value = filter;
   }
 
-  function setShowBookmarked(value: boolean) {
-    protocolFilter.value = null;
+  function setShowBookmarked(value: boolean, resetProtocolFilter = false) {
+    if (resetProtocolFilter) {
+      protocolFilter.value = null;
+    }
     showBookmarked.value = value;
   }
 
-  function clearFilters() {
+  function clearFilters(resetProtocolFilter = false) {
     showBookmarked.value = false;
-    protocolFilter.value = null;
+    if (resetProtocolFilter) {
+      protocolFilter.value = null;
+    }
   }
 
   function getAddressBookEntryByAddress(address?: AccountAddress) {
@@ -197,10 +205,18 @@ export const useAddressBook = createCustomScopedComposable(() => {
     }
   }
 
+  // If user is on bookmarked filter and tries to search, switch to all filter
+  watch(searchQuery, (newSearch) => {
+    if (newSearch && showBookmarked.value) {
+      clearFilters();
+    }
+  });
+
   return {
     protocolFilter,
     addressBook,
     addressBookFiltered,
+    addressBookFilteredByProtocol,
     showBookmarked,
     searchQuery,
 
--- a/src/popup/components/AddressBook/AddressBookFilters.vue
+++ b/src/popup/components/AddressBook/AddressBookFilters.vue
@@ -2,14 +2,14 @@
   <div class="address-book-filters">
     <!-- All/Bookmarked Filters -->
     <BtnFilter
-      :is-active="!showBookmarked && !protocolFilter"
+      :is-active="!showBookmarked && (!protocolFilter || isSelector)"
       :text="$t('common.all')"
-      @click="clearFilters()"
+      @click="clearFilters(!isSelector)"
     />
 
     <BtnFilter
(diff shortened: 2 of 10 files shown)


CedrikNikita:
Icons should be with the opacity.
image

src/popup/components/Modals/AddressBookAccountSelector.vue
```diff
@@ -0,0 +1,79 @@
...
+
+    return {
+      handleSelectAddress,
+    };
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@use '@/styles/variables' as *;
```
CedrikNikita, line 68:
```suggestion
```

martinkaintas: Why? It is used

src/composables/addressBook.ts
```diff
@@ -64,19 +64,23 @@ export const useAddressBook = createCustomScopedComposable(() => {
     ])(Object.values(addressBook.value)),
   );
 
+  const addressBookFilteredByProtocol = computed(
+    () => filterAddressBook(Object.values(addressBook.value)),
```
peronczyk, line 68: I suggest to move the `filterAddressBook` logic here as the function is used only once: here.

src/composables/addressBook.ts
```diff
@@ -64,19 +64,23 @@ export const useAddressBook = createCustomScopedComposable(() => {
...
+  function setShowBookmarked(value: boolean, resetProtocolFilter = false) {
+    protocolFilter.value = resetProtocolFilter ? null : protocolFilter.value;
     showBookmarked.value = value;
   }
 
-  function clearFilters() {
+  function clearFilters(resetProtocolFilter = false) {
     showBookmarked.value = false;
-    protocolFilter.value = null;
+    protocolFilter.value = resetProtocolFilter ? null : protocolFilter.value;
```
peronczyk, line 83: I suggest to avoid reassigning the same value because in most cases this forces the layour to re-render. Consider this:
```
if (resetProtocolFilter) {
  protocolFilter.value = null;
}
```

src/popup/components/AddressBook/AddressBookFilters.vue
```diff
@@ -52,6 +54,9 @@ export default defineComponent({
     BtnFilter,
     IconWrapper,
   },
+  props: {
+    isSelector: { type: Boolean, default: false },
```
peronczyk, line 58:
```suggestion
    isSelector: Boolean,
```

src/popup/components/AddressBook/AddressBookItem.vue
```diff
@@ -1,8 +1,5 @@
 <template>
-  <BtnPlain
-    class="address-book-item"
-    :to="{ name: ROUTE_ADDRESS_BOOK_EDIT, params: { id: item.address } }"
-  >
+  <BtnPlain class="address-book-item">
```
peronczyk, line 2: If there is no event watcher and no `to` prop what is the reason for using `Btn` here?

src/popup/components/AddressBook/AddressBookList.vue
```diff
@@ -63,19 +72,24 @@ export default defineComponent({
   },
   props: {
     modelValue: Boolean,
+    /** Whether the list is being used in an account selector or not  */
+    isSelector: Boolean,
   },
-  emits: ['update:modelValue'],
+  emits: ['update:hide-buttons', 'selectAddress'],
```
peronczyk, line 78: I suggest to maintain one naming case for events. I suggest:
```suggestion
  emits: ['update:hideButtons', 'select-address'],
```

src/popup/components/AddressBook/AddressBookList.vue
```diff
@@ -86,15 +100,41 @@ export default defineComponent({
       }
     });
     const scrollWrapper = computed(() => scrollWrapperEl.value?.$el);
+    const protocolName = computed(() => {
+      if (protocolFilter.value) {
+        return ProtocolAdapterFactory.getAdapter(protocolFilter.value)?.protocolName;
+      }
+      return '';
+    });
```
peronczyk, line 108:
```suggestion
    const protocolName = computed(() => (protocolFilter.value)
      ? ProtocolAdapterFactory.getAdapter(protocolFilter.value)?.protocolName
      : '');
```

src/popup/components/AddressBook/AddressBookList.vue
```diff
@@ -63,19 +72,24 @@ export default defineComponent({
   },
   props: {
     modelValue: Boolean,
+    /** Whether the list is being used in an account selector or not  */
+    isSelector: Boolean,
```
peronczyk, line 76: I think it's better to always emit event with the address and leave the decision what has to be done to parent component. In one case parent can `@something="router.push(...)"` and in other update the local state.

src/popup/components/AddressBook/AddressBookList.vue
```diff
@@ -86,15 +100,41 @@ export default defineComponent({
...
+
+    function handleAddressBookItemClick(entry: IAddressBookEntry) {
+      if (!entry) {
+        return;
+      }
+      if (props.isSelector) {
+        emit('selectAddress', entry.address);
+      } else {
+        router.push({ name: ROUTE_ADDRESS_BOOK_EDIT, params: { id: entry.address } });
+      }
```
peronczyk, line 126: As I mentioned in the `isSelector` prop definition: this logic should happen on the parent level.

peronczyk (changes requested):

src/popup/components/TransferSend/TransferSendRecipient.vue
```diff
@@ -128,15 +148,26 @@ export default defineComponent({
...
   .status {
     margin-top: 9px;
   }
+
+  .buttons {
+    display: flex;
+    align-items: center;
+    justify-content: flex-end;
+    gap: 4px;
+    opacity: 0.75;
```
peronczyk, line 160: Why the buttons now has slightly lower opacity? I guess this is a mistake.

martinkaintas: It's according to figma

src/composables/addressBook.ts
martinkaintas: It's also used in `addressBookFiltered`

peronczyk: OK, now I see. So I suggest to rename the initial function to `filterAddressBookByProtocol`.

peronczyk (approved):

PR #3122, opened 2024-06-21
feat: `modalOpen` typed return value

peronczyk:
This change allows to know what would be the value returned when resolving the modal.


src/popup/router/modals.ts
```diff
@@ -137,7 +137,7 @@ export default () => {
     component: MessageSign,
     showInPopupIfWebFrame: true,
   });
-  registerModal(MODAL_READ_QR_CODE, {
+  registerModal(MODAL_SCAN_QR, {
     component: QrCodeReader,
```
martinkaintas, line 141: I'd consider also renaming the component in this case, e.g. `QrCodeScanner`

src/popup/pages/MultisigProposalDetails.vue
```diff
@@ -324,6 +324,8 @@ import {
 import { ProtocolAdapterFactory } from '@/lib/ProtocolAdapterFactory';
 import { AeAccountHdWallet } from '@/protocols/aeternity/libs/AeAccountHdWallet';
 
+import { MultisigProposalConfirmActionVal } from '@/popup/components/Modals/MultisigProposalConfirmActions.vue';
```
martinkaintas, line 327:
```suggestion
import { type MultisigProposalConfirmActionVal } from '@/popup/components/Modals/MultisigProposalConfirmActions.vue';
```

src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -367,7 +367,7 @@ export default defineComponent({
       }
 
       // Update the signer address
-      signers.value[signerIndex].address = scanResult;
+      signers.value[signerIndex].address = scanResult as any;
```
martinkaintas, line 370:
```suggestion
      signers.value[signerIndex].address = scanResult as Encoded.AccountAddress;
```

martinkaintas (approved):

PR #3126, opened 2024-06-25
Improve multisig performance

CedrikNikita:
fixes https://github.com/superhero-com/superhero-wallet/issues/3092

Idea of this PR is to split all the multisig vaults into array of fives. And only dry-run five vault information in parallel.

### DIFF
--- a/src/composables/multisigAccounts.ts
+++ b/src/composables/multisigAccounts.ts
@@ -1,7 +1,7 @@
 import { computed, ref } from 'vue';
 import { chunk, uniqBy } from 'lodash-es';
 import camelCaseKeysDeep from 'camelcase-keys-deep';
-import { DryRunError, Encoded } from '@aeternity/aepp-sdk';
+import { Contract, DryRunError, Encoded } from '@aeternity/aepp-sdk';
 import BigNumber from 'bignumber.js';
 
 import type {
@@ -9,6 +9,7 @@ import type {
   IMultisigConsensus,
   IMultisigAccountResponse,
   AccountAddress,
+  Dictionary,
 } from '@/types';
 import {
   fetchJson,
@@ -24,6 +25,7 @@ import {
   AE_COIN_PRECISION,
   MULTISIG_SUPPORTED_CONTRACT_VERSION,
 } from '@/protocols/aeternity/config';
+import { SimpleGAMultiSigContractApi } from '@/protocols/aeternity/types';
 import { AeScan } from '@/protocols/aeternity/libs/AeScan';
 import { useAeNetworkSettings } from '@/protocols/aeternity/composables';
 
@@ -37,6 +39,7 @@ export interface MultisigAccountsOptions {
   pollingDisabled?: boolean;
 }
 
+let multisigContractInstances: Dictionary<Contract<SimpleGAMultiSigContractApi>> = {};
 let composableInitialized = false;
 
 const POLLING_INTERVAL = 12000;
@@ -186,10 +189,15 @@ export function useMultisigAccounts({
   }: IMultisigAccountResponse): Promise<IMultisigAccount> {
     const dryAeSdk = await getDryAeSdk();
     try {
-      const contractInstance = await dryAeSdk.initializeContract({
-        aci: SimpleGAMultiSigAci,
-        address: contractId,
-      });
+      if (!multisigContractInstances[contractId]) {
+        multisigContractInstances[contractId] = await dryAeSdk
+          .initializeContract<SimpleGAMultiSigContractApi>({
+            aci: SimpleGAMultiSigAci,
+            address: contractId,
+          });
+      }
+
+      const contractInstance = multisigContractInstances[contractId];
 
       const currentAccount = multisigAccounts.value
         .find((account) => account.contractId === contractId);
@@ -362,6 +370,7 @@ export function useMultisigAccounts({
 
     onNetworkChange(() => {
       updateMultisigAccounts();
+      multisigContractInstances = {};
     });
   }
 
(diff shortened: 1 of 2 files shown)


Liubov-crypto:
There is no issues in console, but Send / Propose Tx modal broken and if I'm going into acc details to check Tx or token details wallet will became black screen. 


https://github.com/superhero-com/superhero-wallet/assets/69896204/f99f33bc-4944-4fb3-98b5-7f1e0d8225c3


https://github.com/superhero-com/superhero-wallet/assets/69896204/ce9e15aa-89ce-4881-a991-09e02957ce44


https://github.com/superhero-com/superhero-wallet/assets/69896204/bca65ff0-1c16-48c1-8235-b50afe321d70

Liubov-crypto:
Issue with black screen is remaining:
https://github.com/superhero-com/superhero-wallet/assets/69896204/b69ef37e-bc30-4c88-8ce7-7e8c6517481f

Also no fiat value:
am

Liubov-crypto:
LGTM

src/composables/multisigAccounts.ts
```diff
@@ -175,12 +176,109 @@ export function useMultisigAccounts({
...
+      if (!(error instanceof DryRunError)) {
+        handleUnknownError(error);
+      }
+      return multisigAccounts.value.find(
+        (account) => account.contractId === contractId,
+      ) as IMultisigAccount;
+    }
+  }
+
+  async function getAllMultisigAccountsInfo(nestedArrays: any) {
```
peronczyk, line 258: I suggest to type the argument instead of using `any`.

src/composables/multisigAccounts.ts
```diff
@@ -175,12 +176,109 @@ export function useMultisigAccounts({
     }
   }
 
+  function splitArrayIntoFives(arr: any) {
+    const result = [];
+    for (let i = 0; i < arr.length; i += 5) {
+      result.push(arr.slice(i, i + 5));
+    }
+    return result;
+  }
```
peronczyk, line 185: https://lodash.com/docs/4.17.15#chunk

src/composables/multisigAccounts.ts
```diff
@@ -175,12 +176,109 @@ export function useMultisigAccounts({
...
+    contractId: Encoded.ContractAddress,
+    gaAccountId: string,
+    otherMultisigData: any,
+  ): Promise<IMultisigAccount> {
     const dryAeSdk = await getDryAeSdk();
+    try {
+      const contractInstance = await dryAeSdk.initializeContract({
+        aci: SimpleGAMultiSigAci,
+        address: contractId,
+      });
```
peronczyk, line 200: Maybe we can cache contract instances to avoid initializing them over and over?

src/composables/multisigAccounts.ts
```diff
@@ -175,12 +176,109 @@ export function useMultisigAccounts({
...
+      /**
+       * Node might throw nonce mismatch error, skip the current account update
+       * return the existing data and account details will be updated in the next poll.
+       */
+      if (!(error instanceof DryRunError)) {
+        handleUnknownError(error);
+      }
+      return multisigAccounts.value.find(
+        (account) => account.contractId === contractId,
+      ) as IMultisigAccount;
```
peronczyk, line 254: I don't think the `as IMultisigAccount` is needed here. If you want to inform TS that you are sure that the result won't be undefined please use `!` at the end of line.

src/composables/multisigAccounts.ts
CedrikNikita: We can make it a separate task.

src/composables/multisigAccounts.ts
CedrikNikita: I've missed that.

src/composables/multisigAccounts.ts
```diff
@@ -176,11 +177,100 @@ export function useMultisigAccounts({
...
+
+  async function getAllMultisigAccountsInfo(nestedArrays: IMultisigAccountResponse[][]) {
+    const results: IMultisigAccount[] = [];
+    /* eslint-disable-next-line no-restricted-syntax */
+    for (const nestedArray of nestedArrays) {
+      // Process each nested array sequentially
+      const promises = nestedArray.map(({
+        contractId,
+        gaAccountId,
+        ...otherMultisigData
```
peronczyk, line 258: I don't think there is any reason now to split the properties and then pass them separately to `getMultisigAccountInfo` as they are not used anywhere alse.

src/composables/multisigAccounts.ts
```diff
@@ -207,79 +292,12 @@ export function useMultisigAccounts({
...
-              handleUnknownError(error);
-            }
-            return multisigAccounts.value.find(
-              (account) => account.contractId === contractId,
-            ) as IMultisigAccount;
-          }
-        }),
-    ))
+    const splittedMultisig = chunk(rawMultisigData
+      .filter(({ version }) => version === MULTISIG_SUPPORTED_CONTRACT_VERSION), 5);
```
peronczyk, line 296: I think the logic of splitting the accounts in chunks should be moved to `getAllMultisigAccountsInfo`.

src/composables/multisigAccounts.ts
```diff
@@ -60,6 +63,7 @@ function getStoredMultisigAccounts(networkId: string, isPending = false): IMulti
   ) || [];
 }
 
+const multisigContractInstances = ref<Dictionary<Contract<SimpleGAMultiSigContractApi>>>({});
```
peronczyk, line 66: No need to use ref here. Can be regular const.

peronczyk (approved):

PR #3131, opened 2024-06-25
Fix multisig issues

### LINKED ISSUE #3129: Hash is missing in multisig Tx details
**Describe the bug**
Hash is missing in multisig Tx details

**To Reproduce**
1. Propose multisig Tx
2. Check the details
3. See error

**Expected behavior**

**Screenshots**

 - OS: MacOS
 - Browser  chrome,
 - Version 126

**Additional context**
reproducible on dev 

Image

### LINKED ISSUE #3130: Difference in displaying gas in multisig Tx proposal
**Describe the bug**
When user is proposing a multisig Tx etimated gas used value will be displayed with AE but should be without. 

**To Reproduce**
1. Propose multisig tx
2. Check screens 
3. See error

**Expected behavior**

**Screenshots**

 - OS: MacOS
 - Browser: chrome 
 - Version 126

**Additional context**
reproducible on dev

Image

CedrikNikita:
fixes #3129,
fixes #3130

### DIFF
--- a/src/protocols/aeternity/components/TransferReview.vue
+++ b/src/protocols/aeternity/components/TransferReview.vue
@@ -80,7 +80,7 @@
         <template #value>
           <TokenAmount
             :amount="transferData.gasUsed"
-            :symbol="AE_SYMBOL"
+            symbol=" "
             :protocol="PROTOCOLS.aeternity"
             hide-fiat
           />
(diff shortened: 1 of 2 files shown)


src/protocols/aeternity/components/TransferReview.vue
```diff
@@ -80,7 +80,7 @@
         <template #value>
           <TokenAmount
```
martinkaintas, line 81: Maybe we should use something other than `TokenAmount` here as the value is just a number and not an asset

martinkaintas (approved):

PR #3141, opened 2024-06-28
feat: global font family definition and typography cleanup

peronczyk:


src/popup/components/Modals/SecureLogin.vue
```diff
@@ -117,7 +117,7 @@ export default defineComponent({
     margin-top: 8px;
 
     .title {
-      @extend %face-sans-18-semi-bold;
+      @extend %face-sans-18-medium;
```
martinkaintas, line 120: `semi-bold` had a `font-weight` of 600 while `medium` has 500
In the design the title has a `font-weight` of 600. 
I don't understand why `face-sans-18-semi-bold` was removed

martinkaintas (approved):
🚀

PR #3161, opened 2024-07-11
fix: be able to select from account assets only

CedrikNikita:
missed in https://github.com/superhero-com/superhero-wallet/pull/3122


Liubov-crypto:
The issue with select assets has been fixed. 

I got this error popups when tried to register names one by one:
name1
ttl

martinkaintas (approved):
Changes seem reasonable

PR #3167, opened 2024-07-12
docs: add a deep link schema

### LINKED ISSUE #2679: Deeplinks
Hello everybody, I know, easier said than properly done, but as multiple approaches exist how we could pull this off with the way SH wallet works, we can find the most ideal one and go for it.

**Is your feature request related to a problem? Please describe.**

The motivations for this are as follows:
1. in order to simplify the implementation of the Superhero Wallet for partners and other developers the necessity of deeplinks became more and more apparent. 
2. We should combine this with the long-discussed mobile scenario, where deep-links would bounce you into the SH app, and later allow calling some callback URL in the browser with details/parameters the user permitted to share with the mobile website (will will get at this at some point ;) )

**Describe the solution you'd like**
1. create the generic foundations to introduce deep links for certain actions
2. introduce deep-linking into certain / very standardised / TBD features like doing transfers, signing messages.

**Describe alternatives you've considered**
Going down the SDK-route is not practical for parnters which support 14 different wallets, adding the whole SDK just to perform a simple `sign message` and/or `transfer` really bloats the dependencies.

**Additional context**
Integration of aeternity on bigger platforms would heavily benefit from that, as we are learning in specific, not-yet-announced cases right now.

CedrikNikita:
closes #2679

### DIFF
--- a/docs/deep-link-schema.md
+++ b/docs/deep-link-schema.md
@@ -1,0 +1,88 @@
+# Deep link URL Schema
+
+Remember that all parameters specified must be URI encoded.
+
+## Connect to a wallet
+
+```bash
+https://wallet.superhero.com/address
+  ? x-success=<success-url>
+  & x-cancel=<cancel-url>
+```
+|URL Params|Description|
+|--|--|
+|`x-success` (required)| This is a callback URL in case user accepts the connection attempt. Callback **must** contain: <br> - `{address}` parameter in order to get current address; <br> - `{networkId}` parameter in order to get current `networkId`.|
+|`x-cancel` (required)| This is a callback URL in case user rejected connection attempt.|
+
+### Example
+
+```
+https://wallet.superhero.com/address?x-success=https%3A%2F%2Ftest.com%2Fsuccess-connection%3Faddress%3D%7Baddress%7D%26networkdId%3D%7BnetworkId%7D&x-cancel=https%3A%2F%2Ftest.com%2Ffail-connection
+```
+
+
+## Sign a transaction
+
+```bash
+https://wallet.superhero.com/sign-transaction
+  ? transaction=<Encoding.Transaction>
+  & networkId=<string>
+  & broadcast=<boolean>
+  & replace-caller=<boolean>
+  & x-success=<success-url>
+  & x-cancel=<cancel-url>
+```
+
+|URL Params|Description|
+|--|--|
+|`transaction` (required)|Valid transaction in [Encoding.Transaction](https://docs.aeternity.com/aepp-sdk-js/latest/api/enums/Encoding.html#Transaction) format `tx_`.|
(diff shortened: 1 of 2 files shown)


docs/deep-link-schema.md
```diff
@@ -0,0 +1,88 @@
...
+  & x-cancel=<cancel-url>
+```
+
+|URL Params|Description|
+|--|--|
+|`transaction` (required)|Valid transaction in [Encoding.Transaction](https://docs.aeternity.com/aepp-sdk-js/latest/api/enums/Encoding.html#Transaction) format `tx_`.|
+|`networkId` (required)|The `networkId` identifier for the network with which the wallet should sign your transaction.|
+|`broadcast` (optional)| This flag is for sending a signed transaction by the wallet.|
+|`replace-caller` (optional)| This flag is used to replace the caller's address of the `transaction` with the current address in the wallet.|
+|`x-success` (required)| This is a callback URL in case user have signed the transaction. If the `broadcast` flag is: <br> - `false/not set` callback **must** contain `{transaction}` parameter in order to get the signed transaction; <br> - `true` callback can have `{transaction-hash}` of the broadcasted transaction.|
```
martinkaintas, line 42:
```suggestion
|`x-success` (required)| This is a callback URL in case user signs the transaction. If the `broadcast` flag is: <br> - `false/not set` callback **must** contain `{transaction}` parameter in order to get the signed transaction; <br> - `true` callback can have `{transaction-hash}` of the broadcasted transaction.|
```

docs/deep-link-schema.md
```diff
@@ -0,0 +1,88 @@
...
+|`x-cancel` (required)| This is a callback URL in case user have not signed the transaction. |
```
martinkaintas, line 43:
```suggestion
|`x-cancel` (required)| This is a callback URL in case user doesn't sign the transaction. |
```

docs/deep-link-schema.md
```diff
@@ -0,0 +1,88 @@
...
+  & x-success=<success-url>
+  & x-cancel=<cancel-url>
+```
+
+|URL Params|Description|
+|--|--|
+|`message` (required)| Message to sign.|
+|`encoding` (optional)| Encoding of the message. Currently only `hex` is supported.|
+|`x-success` (required)| This is a callback URL in case user signed the message. Callback **must** contain: <br> - `{signature}` parameter in order to get signed message; <br> - `{address}` parameter in order to get current address.|
+|`x-cancel` (required)| This is a callback URL in case user have rejected message sign.|
```
martinkaintas, line 82:
```suggestion
|`x-cancel` (required)| This is a callback URL in case user rejected message sign.|
```

martinkaintas (approved):
I'd also suggest to add a reference to this file in the `README.md` file

PR #3173, opened 2024-07-16
fix: link button colors and tweaks

peronczyk:


martinkaintas (approved):
Changes seem reasonable

PR #3182, opened 2024-07-17
feat: remove duplicated babel config files

peronczyk:
No need to have two babel config files: `.babelrc` and `babel.config.js` because only one of them is used during the build process. The documentation https://babeljs.io/docs/configuration says that when working with `node_modules` the `babel.config.js` is the goto option.


martinkaintas (approved):
Given that `@babel/plugin-transform-spread` is no longer needed

peronczyk:
> Given that `@babel/plugin-transform-spread` is no longer needed

Yes, it was not even installed :)

PR #3194, opened 2024-07-19
feat: extension as side panel

peronczyk:


src/popup/router/index.ts
```diff
@@ -155,7 +155,10 @@ if (IS_MOBILE_APP) {
       }
     });
 
+    // TODO establish if we still need this
     router.afterEach((to) => {
+      console.log('PTH', to.path);
```
martinkaintas, line 160:
```suggestion
```

martinkaintas (changes requested):
Also would be nice if the extension is open as side-panel that we don't open new popups but show the modal on the side-panel

src/popup/pages/Names/AuctionList.vue
```diff
@@ -154,7 +154,6 @@ export default defineComponent({
 
 .auction-list {
   .toolbar {
-    --background: var(--screen-bg-color);
```
martinkaintas, line 157: image

peronczyk: OK, I brought back the variable.

martinkaintas (approved):
lgtm

PR #3195, opened 2024-07-19
feat: show decoded data for Ethereum transactions

### LINKED ISSUE #3093: WalletConnect: establishing amount when swapping token to ETH
When swapping any ERC-20 fungible token into ETH coin the dapp (Uniswap) is not giving us direct information about the amount value user want to swap. Here is an example data for such transaction received from UniSwap:

```
data: "0x2e1a7d4d00000000000000000000000000000000000000000000000000662f58c301877a"
from: "0x1fa2c78ab75ffff21761a135256a767c6372da71"
gas: "0x8dcd"
to: "0xfff9976782d46cc05630d1f6ebab18b2324d6b14"
```

Probably we need to decode the `data` entry in order to know the amount. This task is about doing the research and if possible implementing this.

martinkaintas:
closes #3093


src/composables/walletConnect.ts
```diff
@@ -83,16 +83,16 @@ export function useWalletConnect({ offscreen } = { offscreen: false }) {
...
+      const tag = isCoinTransfer ? Tag.SpendTx : Tag.ContractCallTx;
       const modalProps: IModalProps = {
         protocol: PROTOCOLS.ethereum,
         app: { url, host: url ? new URL(url).hostname : '', name },
         tx: {
           amount: params.value ? +fromWei(params.value, 'ether') : 0,
           fee: gas * +(maxFeePerGas.value || 0),
           gas,
-          contractId: (isCoinTransfer) ? ETH_CONTRACT_ID : recipientId,
+          contractId: isCoinTransfer ? ETH_CONTRACT_ID : recipientId,
```
peronczyk, line 95: I know that both options: with/without round brackets are acceptable. But I suggest to leave the brackets. It is a good habit as it makes the code more readable when eyeballing the code in hurry.

When you see this:
```
const a = something ? foo : bar;
```
You might mistake the ternary with regular assignment:
```
const a = something;
```

If you add brackets you immediately know that the line starts with ternary:
```
const a = (something) ? foo : bar;
```

This is a thing that really good dev taught me years ago and practice showed me that he was right :)

src/popup/components/tabs/Tab.vue
```diff
@@ -2,7 +2,7 @@
   <BtnPlain
     v-bind="$attrs"
     class="tab"
-    :class="[active, exactPath ? 'exact' : 'not-exact']"
+    :class="{ active, exact: exactPath, 'not-exact': !exactPath }"
```
peronczyk, line 5:
```suggestion
    :class="{ active, [ exactPath ? 'exact' : 'not-exact' ] }"
```

peronczyk (changes requested):

src/composables/walletConnect.ts
martinkaintas: I agree that this looks better, I just don't think I've seen it in our app. But maybe it's time we start using it

Liubov-crypto:
Are these expected screens? 

tx
with

Liubov-crypto:
I swapped 0.01 weth to 0.01 eth, are these screens correct? 

dec1
raw1

and 0.01 eth to 0.01 weth, it looks like some data in Raw section is cropped: 

eth

cr

martinkaintas:
> I swapped 0.01 weth to 0.01 eth, are these screens correct?
> 
> and 0.01 eth to 0.01 weth, it looks like some data in Raw section is cropped:

This looks right to me, for some transactions uniswap only gives you the method name in the Raw data

Liubov-crypto:
> > I swapped 0.01 weth to 0.01 eth, are these screens correct?
> > and 0.01 eth to 0.01 weth, it looks like some data in Raw section is cropped:
> 
> This looks right to me, for some transactions uniswap only gives you the method name in the Raw data

In this case LGTM

peronczyk (approved):

PR #3203, opened 2024-07-25
fix: network list action buttons opacity and radio btn tweaks

peronczyk:


martinkaintas (approved):
code changes seem reasonable

PR #3210, opened 2024-07-30
refactor: remove abi version from tx interface

peronczyk:
Do we really need to provide and store this property?

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -454,7 +454,6 @@ export interface IGAMetaTx {
 }
 
 export interface ITx {
-  abiVersion?: number;
   accountId?: AccountAddress;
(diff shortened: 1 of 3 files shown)


src/types/index.ts
```diff
@@ -454,7 +454,6 @@ export interface IGAMetaTx {
 }
 
 export interface ITx {
-  abiVersion?: number;
```
davidyuk, line 457: I'm planning to have working mdw types in sdk@14 🙏  so you can take these types from sdk

davidyuk (approved):
Good to me, I would try to remove other `commonParams` because sdk tries to provide reasonable defaults by itself.

subhod-i (approved):
LGTM

PR #3211, opened 2024-07-30
refactor: remove aexn type from tx interface

peronczyk:


src/types/index.ts
```diff
@@ -456,7 +456,6 @@ export interface IGAMetaTx {
 export interface ITx {
   abiVersion?: number;
   accountId?: AccountAddress;
-  aexnType?: 'aex9';
```
peronczyk, line 459: `aexnType` word was not used anywhere in the repository besides this line.

CedrikNikita: This is where is came from https://github.com/aeternity/ae_mdw/pull/1591.

PR #3212, opened 2024-07-31
Update dependencies

### LINKED ISSUE #3136: Code Improvements - Update Libraries to latest Versions

CedrikNikita:
fixes #3136


src/popup/components/BackToTop.vue
```diff
@@ -93,8 +93,8 @@ export default defineComponent({
     left: 0;
     width: 100%;
     padding: 12px 4px 4px 0;
-    background: $color-bg-4;
     /* stylelint-disable-next-line declaration-block-no-duplicate-properties */
```
martinkaintas, line 96: This comment can probably also be removed since the duplicate property was removed

CedrikNikita: The property has been moved. The new `stylelint` thinks the duplicate is the first property because the second property is applied.

martinkaintas: Perhaps we can set it like this: 
`background-color: $color-bg-4;`
and 
` background-image: linear-gradient(...`
This should have the same results without the need to disable stylelint

CedrikNikita: backToTop

martinkaintas (approved):

peronczyk (approved):

Liubov-crypto:
1. Multisig creation is stuck and in pending status more than 30 min. I used airgap account as a 1st signer: 

https://github.com/user-attachments/assets/0173b3d6-f358-4044-9249-4f6fe1ec15e8

ьгдеш

https://aescan.io/transactions/th_2VAWJrWUgLp9CDagNDgdf69m2JFmxBLg5wsuX6UFocJ3VRfGnC

2. Can't send ae if I scan address recipient: 

aes

3. Can't create gift card: 
gift


https://github.com/user-attachments/assets/55a318b8-2aba-4f05-9f41-7cffb29c5fa0

gif

4. Can't send max AE (also on dev)

Liubov-crypto:
Fixed, LGTM

PR #3218, opened 2024-08-01
refactor: remove raw tx from transaction interface

peronczyk:


src/types/index.ts
```diff
@@ -513,7 +513,6 @@ export interface ITransaction {
   microTime?: number;
   pending: boolean;
   protocol: Protocol;
-  rawTx?: any; // TODO find type
```
peronczyk, line 516: This is probably some sort of leftover as I didn't found any reference to this property in the app.

CedrikNikita (approved):

PR #3234, opened 2024-08-15
Several small fixes

CedrikNikita:


src/popup/pages/Names/AuctionHistory.vue
```diff
@@ -107,8 +108,8 @@ export default defineComponent({
       justify-content: space-between;
     }
 
-    .account-item {
-      margin-top: 4px;
+    .account-item :deep(.name) {
```
CedrikNikita, line 111: Since `AccountItem` component already have `size` prop and it is only font related and not length (width) i suggest to use `:deep` here for now. Since the AENS is goin to be redesign.

Liubov-crypto:
When Tx / name is in pending status there should be 3 dots instead of 6:
here
pend

Liubov-crypto:
LGTM

src/popup/pages/Names/AuctionHistory.vue
peronczyk: I strongly advise against using `:deep` - you know that :)

PR #3236, opened 2024-08-19
fix: manifest side panel warnings

peronczyk:


martinkaintas (approved):
Not sure if `removeArrEntries` should be placed in this file but the rest LGTM

peronczyk:
> Not sure if `removeArrEntries` should be placed in this file but the rest LGTM

Yes, I also don't like this approach, but we are not able to import `lodash-es` and any of our util files.

PR #3241, opened 2024-08-20
[DRAFT] feat: solana basic account actions

peronczyk:


package.json
```diff
@@ -157,6 +159,10 @@
     "web-ext-types": "^3.2.1",
     "xml-js": "^1.6.11"
   },
+  "dependenciesComments": {
```
peronczyk, line 162: @CedrikNikita @martinkaintas what do you think about this way of adding our custom comments to installed packages? I feel that it might be helpful as we sometimes add packages that we need to update or change in some conditions.

CedrikNikita: I think we would only benefit from the approach, good suggestion.

PR #3242, opened 2024-08-21
fix(multisig): show correct multisig info on network switching

### LINKED ISSUE #3240: When I'm opening the last multisig accounts it redirects me automatically to my 1st multisig account
**Describe the bug**
When I'm opening the last multisig accounts it redirects me automatically to my 1st multisig account in web wallet (it's not reproducible in extensions):

**To Reproduce**
1. Create a multisig acc using as 1st signer imported airgap account
2. Go to multisig vault
3. Open acc

**Expected behavior**

**Screenshots**

 - OS: macOS
 - Browser chrome, 
 - Version 127

**Additional context**
reproducible on release branch

https://github.com/user-attachments/assets/a7e2e73c-fd8f-4a2f-90a4-0835532e42f6

CedrikNikita:
fixes #3240 
The problem appears upon using `useStorageRef` for array and computed based on this `ref` array. The solution is not very obvious and should be properly tested,


dvdbak:
LGTM

PR #3245, opened 2024-08-26
feat: show decoded call data on eth tx details

### LINKED ISSUE #3096: Show transaction and Tx details made on external dapps correct for user
**Describe the bug**

I made swap from ETH to WETH, on confirmation modals it displays with tags: App.Uniswap.org and Contract call: 
 
wi

when Tx will be completed and fetched I will see in Tx details this Tx as Spend Out: 

eth2

I agree that user have to pay ETH for wrapping it to WETH as first part of Tx.  

But we don't show the second part - incoming WETH to wallet balance in any way.... might be one confusion for user - WETH will be displayed on wallet balance, but incoming Tx related to this is not:

td

**To Reproduce**
1. Go to uniswap
2. connect the wallet
3. Swap ETH to WETH
4. Check Tx details

**Expected behavior**
We have to show to user Tx that he made as precise as possible. 

**Screenshots**

 - OS: MacOS
 - Browser: chrome
 - Version: 123

**Additional context**

### LINKED ISSUE #3097: Show full amount for wrapped tokens in Tx details
**Describe the bug**

I made swap from WETH to ETH with  tags on confirmation modal: App.Uniswap.org and Contract call: 
 

weth

when Tx will be fetched I will see it with Spend Out tags and only with ETH fee amount (agree that we can't display full amount atm): 

weth2

Also WETH Tx will be not available on WETH assets details page in Incoming Tx list. 

**To Reproduce**
1. Go to uniswap
2. Connect the wallet
3. Swap WETH to ETH
4. Check tx details

**Expected behavior**

**Screenshots**

 - OS: MacOS
 - Browser: chrome
 - Version: 123

**Additional context**

martinkaintas:
closes #3097 
closes #3096 

Show user decoded data (if possible, if not show raw data) in transaction details so that he can figure out what the contract call did. This does not exactly solve the issues but seems like the best solution. Etherscan also only shows the decoded data since it would be impossible to create resolvers/parsers for every contract out there

### DIFF
--- a/src/protocols/ethereum/views/TransactionDetails.vue
+++ b/src/protocols/ethereum/views/TransactionDetails.vue
@@ -87,7 +87,7 @@ import {
 import { ProtocolAdapterFactory } from '@/lib/ProtocolAdapterFactory';
 import { ROUTE_NOT_FOUND } from '@/popup/router/routeNames';
 
-import { ETH_COIN_SYMBOL } from '@/protocols/ethereum/config';
+import { ETH_COIN_SYMBOL, ETH_CONTRACT_ID } from '@/protocols/ethereum/config';
 import { decodeTxData } from '@/protocols/ethereum/helpers';
 
 import TransactionDetailsBase from '@/popup/components/TransactionDetailsBase.vue';
@@ -141,10 +141,12 @@ export default defineComponent({
     const amount = computed((): number => transaction.value?.tx?.amount || 0);
 
     async function getDecodedCallData() {
-      if (innerTx.value?.callData && innerTx.value?.recipientId) {
+      if (innerTx.value?.callData && innerTx.value?.contractId) {
         return decodeTxData(
           innerTx.value.callData,
-          innerTx.value.recipientId,
+          (innerTx.value.contractId !== ETH_CONTRACT_ID)
+            ? innerTx.value.contractId
+            : innerTx.value.recipientId,
           activeAccount.value.address,
         );
       }
@@ -166,7 +168,13 @@ export default defineComponent({
       const rawTransaction = await (async (): Promise<ICommonTransaction | null> => {
         // First try to pick the cached transaction.
         const loadedTransaction = transactionsLoaded.value.find((tx) => tx.hash === hash);
-        if (loadedTransaction) {
+
+        const isCallDataDeprecated = (
+          loadedTransaction?.tx?.type === 'ContractCallTx'
+          && (loadedTransaction?.tx?.callData as string) === 'deprecated'
+        );
+
+        if (loadedTransaction && !isCallDataDeprecated) {
           return loadedTransaction;
         }
 
--- a/src/protocols/ethereum/views/TransactionDetails.vue
+++ b/src/protocols/ethereum/views/TransactionDetails.vue
@@ -44,6 +44,17 @@
                 :label="$t('pages.transactionDetails.gasUsed')"
                 data-cy="gas"
               />
+              <DetailsItem
+                expandable
+                class="advanced-transaction-details"
+                :label="$t('transaction.advancedDetails')"
+              >
+                <TransactionCallDataDetails
+                  :call-data="transaction.tx.callData"
+                  :call-data-decoded="decodedCallData"
+                  :loading="decodingCallData"
+                />
+              </DetailsItem>
             </template>
           </TransactionDetailsBase>
         </template>
@@ -77,11 +88,13 @@ import { ProtocolAdapterFactory } from '@/lib/ProtocolAdapterFactory';
 import { ROUTE_NOT_FOUND } from '@/popup/router/routeNames';
 
 import { ETH_COIN_SYMBOL } from '@/protocols/ethereum/config';
+import { decodeTxData } from '@/protocols/ethereum/helpers';
 
 import TransactionDetailsBase from '@/popup/components/TransactionDetailsBase.vue';
 import TransactionAssetRows from '@/popup/components/TransactionAssetRows.vue';
 import DetailsItem from '@/popup/components/DetailsItem.vue';
 import TokenAmount from '@/popup/components/TokenAmount.vue';
+import TransactionCallDataDetails from '@/popup/components/TransactionCallDataDetails.vue';
 
 export default defineComponent({
   components: {
@@ -91,6 +104,7 @@ export default defineComponent({
     TokenAmount,
     IonContent,
     IonPage,
+    TransactionCallDataDetails,
   },
   setup() {
     const router = useRouter();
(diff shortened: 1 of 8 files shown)


CedrikNikita:
Result of decoding for some transactions are not user friendly. 0xe67f85397da69f6d81b9e56beeb21d663dbf1487d745c76dc72be82980965cda
image

martinkaintas:
> Result of decoding for some transactions are not user friendly. 0xe67f85397da69f6d81b9e56beeb21d663dbf1487d745c76dc72be82980965cda

The screenshot shows the `raw` data, which by design is not user-friendly. 
(in some cases _(like on the screenshot) the API returns `deprecated` instead of the raw data, maybe in those cases we don't show any information)_
Also if the decoding fails we are showing a warning to the user that it failed. 
If it is successful I believe we need to show whatever the result was. Whether this is user-friendly or not shouldn't matter.

CedrikNikita:
The `raw` data suppose to show this field.
```
"input": "0xa9059cbb000000000000000000000000d2fbf4c86816d3c830d63f8ca0472d93fc07bc60000000000000000000000000000000000000000000000000002386f26fc10000",
```

Instead of `deprecated`.

src/protocols/ethereum/views/TransactionDetails.vue
```diff
@@ -123,10 +132,24 @@ export default defineComponent({
...
+      return null;
+    }
+
     watch(
       transaction,
       (value) => {
         setLoaderVisible(!value);
+        getDecodedCallData().then((decodedData) => {
+          decodedCallData.value = decodedData;
+        });
```
peronczyk, line 152:
```suggestion
        decodedCallData.value  = await getDecodedCallData();
```

And add `async` before `(value) =>`

src/popup/components/AdvancedTransactionDetails.vue
```diff
@@ -0,0 +1,130 @@
...
+    <DetailsItem
+      v-if="callData"
+      :label="$t('transaction.data')"
+    >
+      <Tabs class="tabs">
+        <Tab
+          v-for="({ name, text }) in dataTabs"
+          :key="name"
+          :data-cy="name"
+          :name="name"
```
peronczyk, line 16: `<Tab>` component has no `name` property.

src/popup/components/AdvancedTransactionDetails.vue
```diff
@@ -0,0 +1,130 @@
...
+          @click="setActiveTab(name)"
+        />
+      </Tabs>
+      <div class="tabs-content">
+        <!-- Decoded Data -->
+        <DetailsItem v-if="activeTab === dataTabs[0].name">
+          <Panel v-if="decodedCallData?.args">
+            <PanelTableItem
+              v-for="([key, value]) in Object
+                .entries(decodedCallData?.args as any) || []"
```
peronczyk, line 28: I suggest to create computed property for the `Object.entries` logic.

src/popup/components/AdvancedTransactionDetails.vue
```diff
@@ -0,0 +1,130 @@
...
+              :key="key"
+              :name="key"
+            >
+              <div class="scrollable">
```
peronczyk, line 32: Class `scrollable` has no reference in this component. Pronbably you didn't move the related CSS block from `ConfirmTransactionSign`.

src/popup/components/AdvancedTransactionDetails.vue
```diff
@@ -0,0 +1,130 @@
...
+              </div>
+            </PanelTableItem>
+          </Panel>
+          <InfoBox
+            v-else
+            type="warning"
+            :text="$t('transaction.decodingDataFailed')"
+          />
+        </DetailsItem>
+        <!-- Raw Data -->
```
peronczyk, line 43: I suggest to add empty line between the different tab contents (before this line).

src/popup/components/AdvancedTransactionDetails.vue
```diff
@@ -0,0 +1,130 @@
...
+        />
+      </div>
+    </DetailsItem>
+
+    <DetailsItem
+      v-if="transactionArguments"
+      :label="$t('modals.confirmTransactionSign.arguments')"
+      :value="transactionArguments"
+    />
+    <slot name="extra" />
```
peronczyk, line 56: I suggest to add empty line before this one.

src/popup/components/AdvancedTransactionDetails.vue
```diff
@@ -0,0 +1,130 @@
...
+    Panel,
+    PanelTableItem,
+    Tabs,
+    Tab,
+    InfoBox,
+    DetailsItem,
+  },
+  props: {
+    callData: { type: String, default: null },
+    decodedCallData: { type: Object as PropType<any>, default: null },
```
peronczyk, line 90: I think the `PropType<any>` is a mistake as the type is initially set as an `Object`. Maybe instead of `any` you wanted to use something like `Dictionary`?

Also what do you think about changing the prop name to `callDataDecoded` ?

src/popup/components/AdvancedTransactionDetails.vue
peronczyk, line 1: I think the name `AdvancedTransactionDetails` is too general. This component displays stuff related to call data.

src/popup/components/AdvancedTransactionDetails.vue
```diff
@@ -0,0 +1,130 @@
+<template>
+  <DetailsItem
+    expandable
+    class="advanced-transaction-details"
+    :label="$t('transaction.advancedDetails')"
+  >
```
peronczyk, line 6: I don't think this component should have this code as it forces it to have the `extra` slot. This makes the component the opposite to single-purpose.

peronczyk (changes requested):

CedrikNikita:
> Result of decoding for some transactions are not user friendly. 

Issue is fixed, but for most of my transactions i see message that wallet failed to decode input.

src/protocols/ethereum/views/TransactionDetails.vue
```diff
@@ -123,10 +144,26 @@ export default defineComponent({
...
 
+    const transactionArguments = computed(() => decodedCallData?.value?.args?.length
+      ? JsonBig.stringify(decodedCallData.value.args)
+      : undefined);
+
+    async function getDecodedCallData() {
+      if (innerTx.value?.callData && innerTx.value?.recipientId) {
+        return decodeTxData(
+          innerTx.value.callData,
+          innerTx.value.recipientId,
```
CedrikNikita, line 155:
```suggestion
          innerTx.value.contractId,
```

Should we use `contractId` here?

Liubov-crypto:
For some Txs there are still few data is showing: 0x20311fa3a73dbe373fd23a371e6387a4708f07c1b50e5094dcc3838935be699b 

https://github.com/user-attachments/assets/586182af-2aaf-47a0-8f66-0d1e7ce6d5e2


but for others looks good to me.

CedrikNikita:
> For some Txs there are still few data is showing: 0x20311fa3a73dbe373fd23a371e6387a4708f07c1b50e5094dcc3838935be699b

There is nothing to decode for data `0x`.

peronczyk (approved):

Liubov-crypto:
lgtm

PR #3249, opened 2024-09-03
Support import/export of a private key

### LINKED ISSUE #1415: The wallet supports the need to import private keys
There are some scenarios, such as in the case of test network, I have no mnemonic, I only have the user's private key, at this time, the DAPP and wallet need to interact directly, but I have no private key import, so it is not convenient to combine

### LINKED ISSUE #2291: allow importing/exporting private keys
Hey guys, I still can't import my dev account from automated testing to manually test the application in the frontend. 
Also I can't export a privat key I've used in communicating with my app through the frontend in order to do some programmatic debugging through scripts.

I know this is all due to everything only relying on mnemonic and account derivation, but maybe it's time to support regular private keys, too.

### LINKED ISSUE #3188: Ability to import/export Private Key (backup)
STORY:

User has to be able to export wallet private keys encrypted. User has to be able to restore wallet accounts with the encrypted key(s) and the password. 

DESIGN:

_1. Import with Seed phrase VS Import with Private key(s) and password:_

**Import wallet accounts with Private Key:**

Image

Image

Image

Image

Image

Image

Image

**Import wallet accounts with Seed phrase:**

Image

Image

Image

Image

Image

Image

.....................................................................................
_2. Export private key(s):_ 

!! See Figma comments (follow the reference link below)

Image

Image

Image

Image

Image

Image

Image

.....................................................................................
FIGMA REFERENCE: https://www.figma.com/design/3oGLWzSH0oJljo4RETZtur/Superhero-Wallet-UI-(%E2%9C%94%EF%B8%8FUpdated)?node-id=33646-210502&t=yNhx5Cr7FQ1SiG9h-0

CedrikNikita:
fixes #3188,
fixes #2291,
fixes #1415


src/popup/components/Modals/AccountCreate.vue
```diff
@@ -6,46 +6,84 @@
     centered
     @close="resolve"
   >
-    <h1
```
CedrikNikita, line 9: This is initially a `AccountCreateModal` component. It was renamed and one button added.

src/popup/components/Modals/ProtocolSelect.vue
```diff
@@ -0,0 +1,73 @@
+<template>
```
CedrikNikita, line 1: This is initially a `AccountCreate` component. That is adjusted to return chosen protocol.

Liubov-crypto:
1. I was able to import my ae/btc/eth accounts from one wallet to another using private key.

2. I can't see in Settings Private Key button to export my private keys in file formats. I see it only in Figma. Idk, was it implemented here? 
sec key

3. If imported AE account has a chain name then prefix PK will not be shown as for AE account without a name: 

https://github.com/user-attachments/assets/87464ff4-98d6-4814-b7e7-d43d201eebdd
pk
im

4. Should it show past Txs for imported accounts? Currently it doesn't show for ETH acc if Tx is older than 141 day. Also I can't see my swap ETH to WETH tx in Tx list:

 https://etherscan.io/address/0x6e503e467AE5A50A135AECBf03e93E97D7B25461

Tx will be not visible in both accs: imported and native. 
eth

But I can see my BTC txs older than 1 year.

CedrikNikita:
> 2. I can't see in Settings Private Key button to export my private keys in file formats. I see it only in Figma. Idk, was it implemented here?

This whole page is no longer needed. We have no plan to implement it.

> 3. If imported AE account has a chain name then prefix PK will not be shown as for AE account without a name:

@onvisions What do you think about this?

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -204,9 +205,21 @@ export class EthereumAdapter extends BaseProtocolAdapter {
...
+      const address = toChecksumAddress(privateKeyToAddress(rawAccount.privateKey).toString());
+      return {
+        idx,
+        globalIdx,
+        secretKey: rawAccount.privateKey!,
+        ...rawAccount,
+        privateKey: undefined,
+        address,
+        publicKey: Buffer.from(privateKeyToPublicKey(rawAccount.privateKey, false)),
+      } as IAccount;
```
peronczyk, line 220: I suggest to avoid forcing the type here.

src/protocols/bitcoin/libs/BitcoinAdapter.ts
```diff
@@ -205,9 +206,30 @@ export class BitcoinAdapter extends BaseProtocolAdapter {
...
+      });
+      return {
+        idx,
+        globalIdx,
+        secretKey: rawAccount.privateKey!,
+        ...rawAccount,
+        publicKey: account.publicKey,
+        privateKey: undefined,
+        address,
+      } as IAccount;
```
peronczyk, line 230: I suggest to avoid forcing the type here.

src/protocols/aeternity/libs/AeternityAdapter.ts
```diff
@@ -202,9 +203,21 @@ export class AeternityAdapter extends BaseProtocolAdapter {
...
+      const address = getAddressFromPriv(Buffer.from(rawAccount.privateKey));
+      return {
+        idx,
+        globalIdx,
+        secretKey: rawAccount.privateKey!,
+        ...rawAccount,
+        privateKey: undefined,
+        address,
+        publicKey: Buffer.from(decode(address)),
+      } as IAccount;
```
peronczyk, line 218: I suggest to avoid forcing the type here as it disabled type checking for particular properties.

src/protocols/aeternity/libs/AeternityAdapter.ts
```diff
@@ -202,9 +203,21 @@ export class AeternityAdapter extends BaseProtocolAdapter {
     rawAccount: IAccountRaw,
     idx: number,
     globalIdx: number,
-    seed: Uint8Array,
+    seed?: Uint8Array,
   ): IAccount {
-    const wallet = rawAccount.type === ACCOUNT_TYPES.hdWallet
+    if (rawAccount.type === ACCOUNT_TYPES.privateKey && rawAccount.privateKey) {
```
peronczyk, line 208: Maybe when the type is equal to `privateKey` but the `rawAccount.privateKey` is falsy we should throw an error or return null / do nothing?

src/protocols/aeternity/libs/AeternityAdapter.ts
```diff
@@ -202,9 +203,21 @@ export class AeternityAdapter extends BaseProtocolAdapter {
...
+        idx,
+        globalIdx,
+        secretKey: rawAccount.privateKey!,
+        ...rawAccount,
+        privateKey: undefined,
+        address,
+        publicKey: Buffer.from(decode(address)),
+      } as IAccount;
+    }
+    const wallet = rawAccount.type === ACCOUNT_TYPES.hdWallet && seed
```
peronczyk, line 220: Maybe if the type is `hdWallet` but the seed is falsy we should throw error / return null / do nothing?

src/protocols/bitcoin/libs/BitcoinAdapter.ts
```diff
@@ -205,9 +206,30 @@ export class BitcoinAdapter extends BaseProtocolAdapter {
...
+    }
+    if (rawAccount.type !== ACCOUNT_TYPES.hdWallet || !seed) {
```
peronczyk, line 232: This tesrt makes no sense here. `null` should be the default return value at the end of the method.  And each other case should be handled before. For example with the use of `switch..case`.

src/protocols/ethereum/libs/EthereumAdapter.ts
```diff
@@ -204,9 +205,21 @@ export class EthereumAdapter extends BaseProtocolAdapter {
...
+    }
+    if (rawAccount.type !== ACCOUNT_TYPES.hdWallet || !seed) {
```
peronczyk, line 222: This test makes no sense here. `null` should be the default return value at the end of the method.  And each other case should be handled before. For example with the use of `switch..case`.

src/popup/components/Modals/ProtocolSelect.vue
peronczyk: Maybe instead of this it would be better to reuse the `AccountCreate` modal so instead of having the account creation logic there we move it to `AccountCardAdd`.

src/popup/components/Modals/PrivateKeyImport.vue
```diff
@@ -0,0 +1,159 @@
...
+    protocol: { type: String as PropType<Protocol>, required: true },
+  },
+  setup(props) {
+    const { t } = useI18n();
+
+    const { accounts, addPrivateKeyAccount, setActiveAccountByGlobalIdx } = useAccounts();
+    const { openModal } = useModals();
+
+    const privateKey = ref('');
+    const error = ref<TranslateResult>('');
```
peronczyk, line 79: We don't need to use `TranslateResult` type anymore. `string` is OK since some previous `i18n` version.
```suggestion
    const error = ref('');
```

src/popup/components/Modals/PrivateKeyImport.vue
```diff
@@ -0,0 +1,159 @@
...
+      showPrivateKeyHelp,
+    };
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@use '@/styles/typography';
+
+.private-key-import .text-center {
```
peronczyk, line 154: I strongly suggest to avoid applying additional styles to utility classes defined in `global.scss` file (`text-center`). I suggest to add new class to apply some custom styling.

src/popup/components/Modals/PrivateKeyExport.vue
```diff
@@ -0,0 +1,196 @@
...
+.private-key-export {
+  .heading {
+    margin-bottom: 1em;
+  }
+
+  .warning {
+    text-align: left;
+  }
+
+  .private-key{
```
peronczyk, line 189:
```suggestion
  .private-key {
```

src/popup/components/Modals/PrivateKeyExport.vue
```diff
@@ -0,0 +1,196 @@
...
+  <Modal
+    class="private-key-export"
+    from-bottom
+    has-close-button
+    centered
+    @close="resolve"
+  >
+    <h1 class="text-heading-4 heading">
+      {{ $t('modals.privateKeyExport.title') }}
+    </h1>
```
peronczyk, line 11: Following change wille render the text as `<h1>Some text</h1>` instead of `<h1> Some text </h1>`:
```suggestion
    <h1
      v-text="$t('modals.privateKeyExport.title')"
      class="text-heading-4 heading"
    >
```

Not a big thing, but in some cases helps to get rid of spacing issues.

src/popup/components/Modals/PrivateKeyExport.vue
```diff
@@ -0,0 +1,196 @@
...
+            :value="privateKey"
+            :copied="copied"
+            hide-icon
+            class="private-key"
+            @click="copyPrivateKey()"
+          />
+        </template>
+      </DetailsItem>
+    </template>
+    <template #footer>
```
peronczyk, line 56: Small suggestion to add empty lines between slots.

src/protocols/aeternity/libs/AeternityAdapter.ts
```diff
@@ -202,20 +203,37 @@ export class AeternityAdapter extends BaseProtocolAdapter {
...
+        ...this.getHdWalletAccountFromMnemonicSeed(seed, idx),
+      };
+    }
+    if (rawAccount.type === ACCOUNT_TYPES.airGap) {
+      return {
+        globalIdx,
+        idx,
+        ...rawAccount,
+        publicKey: Buffer.from(decode(rawAccount.address as Encoded.AccountAddress)),
+      } as any as IAccount;
```
CedrikNikita, line 234: I think improving of this should be a separate task.

src/popup/components/Modals/ProtocolSelect.vue
CedrikNikita: This is what i did.

Liubov-crypto:
LGTM

src/composables/auth.ts
```diff
@@ -326,12 +329,16 @@ export const useAuth = createCustomScopedComposable(() => {
...
 
+  async function checkPassword(password: string) {
+    const key = await generateEncryptionKey(password, encryptionSalt.value!);
+    await decrypt(key, mnemonicEncrypted.value!);
+  }
+
   /**
    * Check if password provided by user is correct and if true update it to new one.
    */
   async function updatePassword(currentPassword: string, newPassword: string) {
```
peronczyk, line 340: This function is used in nly one place and with your change the body is only 2 lines. I suggest to remove it and use those 2 lines in the `Secure LoginSettings` directly.

src/composables/modals.ts
```diff
@@ -198,8 +198,10 @@ export function useModals() {
     });
   }
 
-  function openBiometricLoginModal() {
-    return openModal(MODAL_BIOMETRIC_LOGIN);
+  function openBiometricLoginModal(options?: {
+    checkDespiteAuthenticated: boolean;
```
peronczyk, line 202:
```suggestion
    force: boolean;
```

src/icons/rounded-square-key.svg
peronczyk, line 1: I suggest to rename this file to "private-key". Naming stuff based on how they look is in many cases problematic as things changes in time. What if Tsvetan would want to change the rounded square to just square? :)
Also this is not a square but rectangle.

peronczyk (changes requested):

src/popup/router/index.ts
```diff
@@ -76,6 +77,8 @@ router.beforeEach(async (to, from, next) => {
 
   await checkUserAuth();
 
+  await watchUntilTruthy(arePrivateKeysAccountsRestored);
```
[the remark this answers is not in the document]
CedrikNikita: It can not. In order to restore the private keys account we need to decrypt them.

src/utils/common.ts
```diff
@@ -635,6 +639,7 @@ export function decryptedComputed(
   key: Ref<CryptoKey | undefined>,
   encryptedState: Ref<string | null>,
   defaultVal?: any,
+  onDecrypted?: (val: string | null) => any,
```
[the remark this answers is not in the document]
CedrikNikita: it is a 4 parameter. We can use `options` here instead with `onDecrypted` as a field in it.

src/utils/common.ts
```diff
@@ -643,6 +648,9 @@ export function decryptedComputed(
     if (!updating) {
       try {
         decrypted.value = (newKey && newState) ? await decrypt(newKey, newState) : defaultVal;
+        if (newKey && newState) {
+          onDecrypted?.(decrypted.value);
+        }
```
peronczyk, line 653: Consider this:
```suggestion
        decrypted.value = (newKey && newState) ? await decrypt(newKey, newState) : defaultVal;
        if (newKey && newState) {
          decrypted.value = await decrypt(newKey, newState);
          onDecrypted?.(decrypted.value);
        } else {
          decrypted.value = defaultVal;
        }
```

src/utils/common.ts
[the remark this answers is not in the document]
peronczyk: OK, but if its not used why add it?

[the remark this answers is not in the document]
CedrikNikita: It is used, but because it it just a parameter it was used without the name `onDecrypted`.

src/popup/router/index.ts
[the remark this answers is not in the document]
peronczyk: Makes sense. So my suggestion is to group the stuff related to accounts in one `Promise.all`, put it after the `checkUserAuth` (as the accounts are not required to authenticate user) and leave the mnemonic in same place.
Or even create `areAccountsReady` computed property which would monitor both `areAccountsRestored` and `arePrivateKeysAccountsRestored`.

Another thing is that the `checkUserAuth` has it's own watcher for mnemonic, so we don't need another one in this scope.

So this block would look as follows:

```suggestion
router.beforeEach(async (to, from, next) => {
  await checkUserAuth();
  await watchUntilTruthy(areAccountsReady); // mentioned computed
```

src/utils/common.ts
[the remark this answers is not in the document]
peronczyk: Uhhhh, damn... true. OK, you're right that this would be better to use `options` argument with additional stuff like the `onDecrypted`.

src/protocols/aeternity/libs/AeternityAdapter.ts
peronczyk: Can you create one and reference it here?

src/composables/accounts.ts
```diff
@@ -72,6 +107,17 @@ export const useAccounts = createCustomScopedComposable(() => {
     { backgroundSync: true },
   );
 
+  const arePrivateKeysAccountsRestored = computed(() => (
+    (arePrivateKeysAccountsEncryptedRestored.value && !accountsPrivateKeysEncrypted.value)
+    || (arePrivateKeysAccountsEncryptedRestored.value && arePrivateKeysAccountsDecrypted.value)
+  ));
+
+  const accountsPrivateKeys = computed<IAccountRaw[]>(() => JSON.parse(accountsPrivateKeysDecrypted.value || '[]'));
```
peronczyk, line 115: As there is a `addPrivateKeyAccount` I suggest to rename this variable to `privateKeyAccountsRaw`. This way it will also match the naming convention of the `accountsRaw` variable.

src/popup/components/AccountCard.vue
```diff
@@ -19,13 +19,14 @@
...
-      <AccountCardTotalTokens
-        :account="account"
-      />
-      <AirGapIcon
-        v-if="isAccountAirGap"
-        class="air-gap-icon"
+    <template #bottom-left>
+      <AccountCardTotalTokens :account="account" />
+    </template>
+    <template #bottom-right>
```
peronczyk, line 25: I suggest to add empty line between slots.

peronczyk (approved):

PR #3263, opened 2024-09-16
feat: do not keep salt and iv with encryption key

martinkaintas:
- [x] Separates `key` from `salt` and `iv`
- [x] Generates and stores `salt` once per password

related comments
- https://github.com/superhero-com/superhero-wallet/pull/3231#discussion_r1758062888
- https://github.com/superhero-com/superhero-wallet/pull/3231#discussion_r1758182189
- https://github.com/superhero-com/superhero-wallet/pull/3231#discussion_r1756500405

### DIFF
--- a/src/composables/accounts.ts
+++ b/src/composables/accounts.ts
@@ -7,7 +7,6 @@ import type {
   AccountType,
   IAccount,
   IAccountRaw,
-  IEncryptionData,
   IFormSelectOption,
   Protocol,
   ProtocolRecord,
@@ -26,14 +25,13 @@ import {
   IS_OFFSCREEN_TAB,
 } from '@/constants';
 import {
+  authenticateWithPassword,
   createCallbackRegistry,
-  decrypt,
   encrypt,
   endSession,
   excludeFalsy,
-  restoreEncryptionData,
-  initializeEncryptionData,
-  getSessionEncryptionData,
+  generateEncryptionKey,
+  getSessionEncryptionKey,
   prepareAccountSelectOptions,
   startSession,
   watchUntilTruthy,
@@ -61,9 +59,9 @@ const {
 
 const areAccountsRestored = ref(false);
 const isMnemonicRestored = ref(false);
-const encryptionData = ref<IEncryptionData | null>();
+const encryptionKey = ref<CryptoKey | null>();
 
-const mnemonic = useSecureStorageRef<string>(
+const [encryptedMnemonic, mnemonic] = useSecureStorageRef<string>(
   '',
   STORAGE_KEYS.mnemonic,
   {
@@ -253,27 +251,26 @@ export function useAccounts() {
   /**
    * Setting/Resetting the password key logs the user in/out.
    */
-  function setEncryptionData(newEncryptionData: IEncryptionData | null) {
-    encryptionData.value = newEncryptionData;
+  function setEncryptionKey(newEncryptionKey: CryptoKey | null) {
+    encryptionKey.value = newEncryptionKey;
     if (IS_EXTENSION) {
-      if (newEncryptionData) {
-        startSession(newEncryptionData, secureLoginTimeout.value);
+      if (newEncryptionKey) {
+        startSession(newEncryptionKey, secureLoginTimeout.value);
       } else {
         endSession();
       }
     }
   }
 
-  async function getEncryptionData() {
-    await watchUntilTruthy(encryptionData);
-    return encryptionData.value;
+  async function getEncryptionKey() {
+    return watchUntilTruthy(encryptionKey);
   }
 
   async function openLoginModal() {
     setLoaderVisible(true);
-    const sessionEncryptionData = await getSessionEncryptionData();
-    if (sessionEncryptionData) {
-      setEncryptionData(sessionEncryptionData);
+    const sessionEncryptionKey = await getSessionEncryptionKey();
+    if (sessionEncryptionKey) {
+      setEncryptionKey(sessionEncryptionKey);
       const { getAeSdk } = useAeSdk();
       await getAeSdk();
       setLoaderVisible(false);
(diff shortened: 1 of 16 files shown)


src/composables/accounts.ts
```diff
@@ -27,16 +26,15 @@ import {
...
   excludeFalsy,
-  restoreEncryptionData,
-  initializeEncryptionData,
-  getSessionEncryptionData,
+  generateEncryptionKey,
+  getSessionEncryptionKey,
   prepareAccountSelectOptions,
   startSession,
   watchUntilTruthy,
+  authenticateWithPassword,
```
peronczyk, line 37: Please arrange imports alphabetically :)

src/composables/accounts.ts
```diff
@@ -253,27 +254,27 @@ export function useAccounts() {
...
       }
     }
   }
 
-  async function getEncryptionData() {
-    await watchUntilTruthy(encryptionData);
-    return encryptionData.value;
+  async function getEncryptionKey() {
+    await watchUntilTruthy(encryptionKey);
+    return encryptionKey.value!;
```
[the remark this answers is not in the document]
martinkaintas: With `encryptionData` this would result in a weird type instead of the correct type of `encryptionData`
Now with `encryptionKey` its not happening so I'll follow your suggestion

src/composables/accounts.ts
```diff
@@ -373,12 +367,12 @@ export function useAccounts() {
     activeAccountGlobalIdx.value = 0;
   }
 
-  async function waitForSessionEncryptionData() {
+  async function waitForSessionEncryptionKey() {
```
peronczyk, line 370: As this function is not only waiting for the encryption key but synchronises the state between background and popup I suggest to align the name to this. Maybe something like `syncBackgroundEncryptionKey` or something similar?

src/composables/auth.ts
```diff
@@ -96,12 +92,12 @@ export const useAuth = createCustomScopedComposable(() => {
 
   async function logout() {
     if (IS_EXTENSION) {
-      const sessionEncryptionData = await getSessionEncryptionData();
-      if (sessionEncryptionData) {
+      const sessionEncryptionKey = await getSessionEncryptionKey();
+      if (sessionEncryptionKey) {
```
peronczyk, line 96: Why we need to check if the session key exists? I think logging out means resetting the key and the authenticated state. Or I'm missing something here?

src/composables/secureStorageRef.ts
```diff
@@ -107,19 +107,18 @@ export function useSecureStorageRef<T = string | object | any[]>(
...
-    }, { immediate: true, deep: true });
+    if (!IS_MOBILE_APP) {
+      const { encryptionKey } = useAccounts();
+      watch(encryptionKey, async (key) => {
+        if (!key) {
+          decryptedState.value = initialState;
+        } else {
+          decryptedState.value = await getDecryptedValue(encryptedState.value);
+        }
+      }, { immediate: true, deep: true });
```
peronczyk, line 120: I suggest to consider using asynchronous computed property (https://vueuse.org/core/computedAsync/) instead of watcher as the watcher exists only to sync one value with another.

src/composables/secureStorageRef.ts
```diff
@@ -128,10 +127,10 @@ export function useSecureStorageRef<T = string | object | any[]>(
...
       // Do not write null values to the mnemonic storage
       if (val || (storageKey !== STORAGE_KEYS.mnemonic)) {
-        innerState.value = await getEncryptedValue(val);
+        encryptedState.value = await getEncryptedValue(val);
       }
     }, { immediate: true, deep: true });
   })();
 
-  return decryptedState;
+  return { encryptedState, decryptedState };
```
peronczyk, line 135: I'm wondering if maybe returning those 2 values as an array would be a better approach. React is doing this. This way when using this composable instead of:
```
const {
  encryptedState: something,
  decryptedState: anotherThing,
} = ...
```

You would go with:
```
const [something, anotherThing] = ...
```

src/utils/session.ts
```diff
@@ -14,13 +13,13 @@ const storageSession = (browser.storage as any)?.session;
...
  */
-export async function startSession(encryptionData: IEncryptionData, timeout: number) {
+export async function startSession(encryptionKey: CryptoKey, timeout: number) {
   if (IS_EXTENSION && !IS_OFFSCREEN_TAB) {
     browser.runtime.connect({ name: CONNECTION_TYPES.SESSION });
     const sessionExpires = Date.now() + timeout;
     await storageSession.set(
       {
-        encryptionData: await exportEncryptionData(encryptionData),
+        encryptionKey: await exportEncryptionKey(encryptionKey),
```
peronczyk, line 22: This line is little bit quirky. Both the property and the constant has the same name: `encryptionKey` which suggests that they are with the same type, but I see you need to us `exportEncryptionKey` function to fix the data type. I suggest to differentiate the naming to make it more clear what is the difference between those 2.

src/utils/session.ts
```diff
@@ -32,23 +31,18 @@ export async function startSession(encryptionData: IEncryptionData, timeout: num
...
-        key: Buffer.from(sessionEncryptionData.key, 'base64'),
-        salt: Buffer.from(sessionEncryptionData.salt, 'base64'),
-        iv: Buffer.from(sessionEncryptionData.iv, 'base64'),
-      };
-      return importEncryptionData(decodedEncryptionData);
+    const sessionEncryptionKey = await getSessionEncryptionKeyOffscreen();
+    if (sessionEncryptionKey) {
+      return importEncryptionKey(Buffer.from(sessionEncryptionKey, 'base64'));
     }
     return null;
```
peronczyk, line 47:
```suggestion
    return (sessionEncryptionKey)
      ? importEncryptionKey(Buffer.from(sessionEncryptionKey, 'base64'))
      : null;
```

peronczyk, line 47: This line can be removed if you change this:
```
if (IS_EXTENSION) {
```
to `else if`.

peronczyk (approved):

davidyuk (commented):
👍

PR #3264, opened 2024-09-16
refactor: translate password strength value & simplify AUTHENTICATION…

martinkaintas:


src/composables/ui.ts
```diff
@@ -45,7 +45,7 @@ const hiddenCards = useStorageRef<string[]>(
 );
 const otherSettings = useStorageRef<IOtherSettings>(
   {
-    secureLoginTimeout: IS_MOBILE_APP ? AUTHENTICATION_TIMEOUTS[0] : AUTHENTICATION_TIMEOUTS[5],
+    secureLoginTimeout: IS_MOBILE_APP ? AUTHENTICATION_TIMEOUTS[0] : AUTHENTICATION_TIMEOUTS[2],
```
peronczyk, line 48: This piece of code is repeated many times:
```
IS_MOBILE_APP ? AUTHENTICATION_TIMEOUTS[0] : AUTHENTICATION_TIMEOUTS[2]
```

I suggest to create such global constant:
```
export const  AUTHENTICATION_TIMEOUT_DEFAULT = (IS_MOBILE_APP)
  ? AUTHENTICATION_TIMEOUTS[0]
  : AUTHENTICATION_TIMEOUTS[2];
```

And reuse it.

src/popup/components/InputPassword.vue
```diff
@@ -17,16 +17,19 @@
     <template #label-after>
       <span
         v-if="showPasswordStrength && modelValue"
-        :class="['password-strength', passwordStrength.toLowerCase()]"
-        v-text="passwordStrength"
+        class="password-strength"
+        :class="[passwordStrength]"
```
peronczyk, line 21: I suggest to prefix these classes:
```suggestion
        :class="[`strength-${passwordStrength}`]"
```

so the CSS block would look as follows:
```
&.strength {
  &-weak {
    ...
```

What do you think? In my opinion this gives some kind of feature grouping.

src/popup/pages/SecureLoginSettings.vue
```diff
@@ -30,15 +30,15 @@
...
+              v-for="(ms) in AUTHENTICATION_TIMEOUTS"
+              :key="`timeout-${ms}`"
               :value="secureLoginTimeout === ms"
               :class="{ active: secureLoginTimeout === ms }"
               class="timeout"
               has-label-effect
               @input="setSecureLoginTimeout(ms)"
             >
-              <div class="row" v-text="$t('pages.secureLogin.authenticationTimeout', Number(minutes))" />
+              <div class="row" v-text="$t('pages.secureLogin.authenticationTimeout', msToMinutes(ms))" />
```
peronczyk, line 41: I suggest to use `label` property instead.

martinkaintas: `label` is added inside the `RadioButton.vue` component. I guess `span` tag is more appropriate.

peronczyk (approved):

PR #3270, opened 2024-09-18
Improve `showHeader` logic

CedrikNikita:
The `Header` was not displayed if the `hideHeader` parameter in the `route.meta` was set to `true`, but in this case notification polling was started (initial calculation of Header that required `notification` composable was done).


Liubov-crypto (approved):
LGTM

peronczyk (approved):

PR #3231, opened 2024-08-09
Feat/secure storage

martinkaintas:
closes #3189 
closes #3101
closes #1401

PR #3271, opened 2024-09-18
refactor: use route constants

martinkaintas:
- use route constants
- rename functions to better/more descriptive names 
- other small refactors resolving comments on #3231


src/composables/auth.ts
```diff
@@ -64,19 +64,21 @@ export const useAuth = createCustomScopedComposable(() => {
...
-        iosFallbackTitle: t('biometricAuth.fallbackTitle'),
-        androidTitle: t('biometricAuth.title'),
-        androidSubtitle: t('biometricAuth.subtitle'),
-        androidConfirmationRequired: false,
-      }).then(() => {
-        isAuthenticated.value = true;
-      });
-    } if (!IS_MOBILE_APP) {
+    if (IS_MOBILE_APP) {
+      if (isBiometricLoginEnabled.value && await checkBiometricLoginAvailability()) {
```
peronczyk, line 68: Why not merge those together?

src/composables/secureStorageRef.ts
```diff
@@ -20,29 +22,38 @@ export function useSecureStorageRef<T = string | object | any[]>(
     'serializer' | 'enableSecureStorage' | 'onBackgroundSync'
   > = {},
 ) {
-  async function getDecryptedValue(encryptedValue: any) {
-    if (IS_MOBILE_APP || !encryptedValue) {
-      return encryptedValue;
+  /**
+   * Checks if the value needs to be decrypted and decrypts it using the encryption key.
+   * On mobile, the value is already decrypted by the storageRef composable.
```
peronczyk, line 27: To be honest I would consider to encrypt the mnemonic also on the mobile. As far as I know from a research I did some time ago the biometric login provides a `key` that can be used.

My idea on the authentication architecture was to have two separate modules:
1. Encoding/decoding key provider - interface that takes any string and uses it to encode/decode data. As a source of the key we could use any source, like a password or biometric login token.
2. Encoder/decoder - get's the key generated by provider and allows to operate on any data.

Maybe we can create a separate task to research this in the future? Or at least add `TODO` comment here?

src/composables/secureStorageRef.ts
```diff
@@ -20,29 +22,38 @@ export function useSecureStorageRef<T = string | object | any[]>(
...
+   */
+  async function getDecryptedValue(val: any, isDecrypted = IS_MOBILE_APP): Promise<T> {
```
peronczyk, line 29: 1. I suggest to move `isDecrypted` argument to the composable options because this funcion is not available outside of the composable scope and is not used within the composable scope.
2. Actually this is not `isDecrypted` because on mobile the mnemonic is not encrypted at all. I suggest to rename this to something like `encryptionDisabled`.

src/composables/secureStorageRef.ts
martinkaintas: We're encrypting the mnemonic on mobile through the `@aparajita/capacitor-secure-storage` package in the `SecureMobileStorage.ts` lib. Our used authentication package does not provide us a `token`.

peronczyk: OK, ignore my comment. I'll create a task for the future.

My idea was not to use any external package for mobile storage but use the same logic on every env. This way the only matter is what we use as the "password" (the key). On web/extension this would be user's password and on the mobile the token received by the **biometric** login (not the secure storage). So when user uses the fingerpirint/face scanner the lib passes the token created at the first time we allowed the app to use biometric login. So it's like a password :) At least this was my assumption.

peronczyk (approved):

PR #3275, opened 2024-09-19
feat: secure storage logout action

peronczyk:


src/popup/App.vue
```diff
@@ -14,16 +14,27 @@
...
+        <Header v-if="isAuthenticated && delayedShowHeader" />
 
-        <!-- We are disabling animations on FF because of a bug that causes flickering
-          see: https://github.com/ionic-team/ionic-framework/issues/26620 -->
+        <!--
+          We are disabling animations on FF because of a bug that causes flickering
+          see: https://github.com/ionic-team/ionic-framework/issues/26620
+        -->
         <IonRouterOutlet
+          v-if="isAuthenticated"
```
martinkaintas, line 24: `v-if="isAuthenticated"` is probably not sufficient, we need to also check if the wallet is initialized, opening the app for the very first time right now results in user being stuck here 
image

martinkaintas (changes requested):

src/popup/App.vue
peronczyk: The reason for that was that the `isAuthenticated` state was not updated to `true` after importing or creating new wallet. I fixed this, but in rather ugly way. It's ugly because the `isAuthenticated` state is stored in `auth` composable, but the setting password is in `accounts`, so the accounts is not able to update the `auth` state.

This is why I really think we need to move the authentication logic from `accounts` to `auth`.
@martinkaintas @smaroudasunicorn @CedrikNikita

src/composables/accounts.ts
```diff
@@ -300,21 +299,16 @@ export function useAccounts() {
...
   }
 
-  async function openSetPasswordModal(newMnemonic: string, isRestored = false) {
-    const { openModal } = useModals();
-
-    const password = await openModal<string>(MODAL_SET_PASSWORD, {
-      isRestoredWallet: isRestored,
-    });
-
-    await setPasswordAndEncryptMnemonic(newMnemonic, password);
```
peronczyk, line 310: This didn't make sense to me that the `openSetPasswordModal` modal was also encrypting the mnemonic. I removed this line and used it in the `accounts` composable.

The method was moved to `modals` composable.

src/composables/auth.ts
```diff
@@ -99,22 +116,32 @@ export const useAuth = createCustomScopedComposable(() => {
...
-      await new Promise((resolve) => setTimeout(resolve, 500));
+
+      // Wait before resetting isAuthenticated so that app doesn't register a false app resume event
+      await sleep(500);
+
       isAuthenticating.value = false;
     }
   }
 
+  async function lockWallet() {
```
peronczyk, line 140: My attempt for solving this comment: https://github.com/superhero-com/superhero-wallet/pull/3231#discussion_r1764753376

src/popup/components/Modal.vue
```diff
@@ -310,9 +319,8 @@ export default defineComponent({
     }
   }
 
-  // This is not working correctly in Firefox extension
-  &.blur-bg {
-    backdrop-filter: blur(5px);
+  &.no-bg-blur {
```
martinkaintas, line 322: duplicate in line 308

martinkaintas (approved):

PR #3282, opened 2024-09-25
Feat/encrypt login timeout restore mnemonic

martinkaintas:
Related comments
https://github.com/superhero-com/superhero-wallet/pull/3231#discussion_r1758142403
https://github.com/superhero-com/superhero-wallet/pull/3231#discussion_r1757882973

Also fixes issue with restoring mnemonic from previous wallet (web & extension, mobile needs to be rechecked)

### DIFF
--- a/src/composables/auth.ts
+++ b/src/composables/auth.ts
@@ -5,10 +5,11 @@ import {
   ref,
   watch,
 } from 'vue';
-import { generateMnemonic, mnemonicToSeed } from '@aeternity/bip39';
+import { generateMnemonic, mnemonicToSeed, validateMnemonic } from '@aeternity/bip39';
 
 import { tg as t } from '@/popup/plugins/i18n';
 import {
+  AUTHENTICATION_TIMEOUT_DEFAULT,
   IS_EXTENSION,
   IS_IOS,
   IS_MOBILE_APP,
@@ -40,6 +41,7 @@ import { useModals } from './modals';
 import { createCustomScopedComposable } from './composablesHelpers';
 import { useSecureStorageRef } from './secureStorageRef';
 import { useAeSdk } from './aeSdk';
+import { useStorageRef } from './storageRef';
 
 const CHECK_FOR_SESSION_KEY_INTERVAL = 5000;
 
@@ -47,7 +49,6 @@ export const useAuth = createCustomScopedComposable(() => {
   const {
     isBiometricLoginEnabled,
     lastTimeAppWasActive,
-    secureLoginTimeout,
     isAppActive,
     setBiometricLoginEnabled,
     setLoaderVisible,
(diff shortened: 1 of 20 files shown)


src/composables/ui.ts
```diff
@@ -52,16 +53,14 @@ const otherSettings = useStorageRef<IOtherSettings>(
...
       migrateSecureLoginEnabledToBiometric,
+      migrateRemoveSecureLoginTimeout,
     ],
   },
 );
 
 const isSeedBackedUp = computed(() => !!otherSettings.value.isSeedBackedUp);
 const saveErrorLog = computed(() => !!otherSettings.value.saveErrorLog);
 const isBiometricLoginEnabled = computed(() => !!otherSettings.value.isBiometricLoginEnabled);
-const secureLoginTimeout = computed(
```
martinkaintas, line 62: Had to move this to `auth` composable because `ui` composable cannot initialize `auth` composable after the rebasing to current `develop`.

src/composables/auth.ts
```diff
@@ -59,26 +60,50 @@ export const useAuth = createCustomScopedComposable(() => {
   const isAuthenticating = ref(false);
   const isMnemonicRestored = ref(false);
   const encryptionKey = ref<CryptoKey | null>();
+  const isMnemonicMigrationCheckDone = ref(false);
 
-  const [mnemonicEncrypted, mnemonic] = useSecureStorageRef<string>(
+  /**
+ * Deprecated storage key for the mnemonic.
+ * Used for migration from the old storage key to the new one.
+ */
```
peronczyk, line 68: Indentation problems here :)

src/constants/common.ts
```diff
@@ -134,7 +135,7 @@ export const STORAGE_KEYS = {
   transactionsLoaded: 'transactions-loaded',
   transactionsPending: 'transactions-pending',
   transferSendData: 'transfer-send-data',
-  secureLogin: 'secure-login',
+  secureLoginTimeout: 'secure-login-timeout',
```
peronczyk, line 138: I suggest to leave old version. I see that the new one is better, but having to write migration only because the name is better is rather bad for me.

src/constants/common.ts
```diff
@@ -106,7 +106,8 @@ export const DEX = 'DEX';
  * Doing any change to the existing values would require creating migration.
  */
 export const STORAGE_KEYS = {
-  mnemonic: 'mnemonic',
+  mnemonic: 'encrypted-mnemonic',
```
peronczyk, line 109: As far as I understand on mobile we are not using password protection so the mnemonic is not encoded.

src/composables/auth.ts
```diff
@@ -59,26 +60,50 @@ export const useAuth = createCustomScopedComposable(() => {
...
+  const oldMnemonic = useStorageRef<string>(
```
peronczyk, line 69: Such things should be managed by the migration.
I think this is really bad for us to have such deprecations in the code.

peronczyk (changes requested):

src/constants/common.ts
martinkaintas: The old value was not used anywhere, it looks like a rename but I just wanted to create a new value for the `secureLoginTimeout` and noticed the unused `secureLogin` value. There's no migration

src/constants/common.ts
martinkaintas: Mnemonic is always encrypted on mobile from now on

Issue #2867, opened 2024-03-21
MultiSig - Shorten UX Approval Path


smaroudasunicorn:

onvisions:
Based on Ingmar's user feedback I see a few points where we can suggest UX improvement of the functionality. 

1. Pending consensus indicator within the active account card.

![Image](https://github.com/superhero-com/superhero-wallet/assets/12225410/721b6129-fac0-4d9d-938c-b22c2e3d649e)

My assumption when we designed this was that the users will intuitively try to click on the "Consensus 1/2 of 3" (see the screenshot below) expecting that this action will lead them directly to the multisig transaction awaiting to be signed. 
As far as I remember after discussion with PO and the team we decided to NOT make this area clickable because the whole account card itself is a clickable element (leading to account details) and this may create some confusion. 
However from what we receive as user feedback it seems that the user would rather click on the indicator expecting to open directly the multisig transaction proposal for signing. Currently following this interaction path this will just open the Account Details screen instead.

**Suggestion:**
- Making this UI element clickable (leading to Multisig Tx proposal details where user can approve or disapprove pending transaction proposal). It turns out this is the expected and intuitive behavior according to user feedback received.
- This will remove one step from the user interaction flow.
- We can either leave the UI element within the account card OR get this element out of the account card so that we will reduce the possibility of confusion about expected behavior on click.  

**Design versions:**

..............

2. Make the pending consensus transaction always on top of the transactions list on Multisig vault details screen. Thus it will be also always visible without scrolling.

..............

3. The step with Approve/Disapprove/Revoke Transaction Proposal modal remains but UX should be improved by adding some features there.
4. 
![Image](https://github.com/superhero-com/superhero-wallet/assets/12225410/449040ef-5111-4a03-97a8-0ebf2d956dd0)

Currently the user is able to select the signing account in case there are more than one of their  accounts authorized as signers.
Instead of the option to select just one signing account all user's accounts authorized to sign the proposed transaction will be listed in the modal so that the user will be able to select multiple accounts signing the transaction (with checkboxes) without the need to go through the whole flow from the start.

CedrikNikita:
> 4. SEND transaction.

We would need to adjust the wording of the button.  `Approve and send` button text will help user to understand that the flows are combined.

peronczyk:
> 1. Pending consensus indicator within the active account card.
> 
[...]

PR #3283, opened 2024-09-25
feat(aeternity): skip confirm modal for transactions created by wallet

### LINKED ISSUE #3254: Simple Transaction - Shorten User Confirmation
Every action from the wallet could be auto signed by user shortening the user journey. 

We should have the option for the user to sign automatically transactions/messages originated from the wallet

CedrikNikita:
fixes #3254,
part of #2867


Liubov-crypto (approved):
LGTM

martinkaintas (approved):
If it is intentional that we sign all transactions originating from the wallet then the changes look reasonable

PR #3285, opened 2024-09-26
feat: store session start to prevent attack from manipulating system time

martinkaintas:


src/composables/auth.ts
```diff
@@ -122,10 +122,11 @@ export const useAuth = createCustomScopedComposable(() => {
   /**
    * Setting/Resetting the password key logs the user in/out.
    */
-  function setEncryptionKey(newEncryptionKey: CryptoKey | null) {
+  async function setEncryptionKey(newEncryptionKey: CryptoKey | null) {
     encryptionKey.value = newEncryptionKey;
     if (IS_EXTENSION) {
       if (newEncryptionKey) {
+        await watchUntilTruthy(secureLoginTimeout);
```
martinkaintas, line 129: Make sure that the `secureLoginTimeout` has a value, since it is an encrypted value we need to make sure that it is decrypted and set.

src/composables/auth.ts
```diff
@@ -153,16 +154,19 @@ export const useAuth = createCustomScopedComposable(() => {
     const sessionEncryptionKey = await getSessionEncryptionKey();
     if (sessionEncryptionKey) {
       setEncryptionKey(sessionEncryptionKey);
+      isAuthenticated.value = true;
```
martinkaintas, line 157: Since we are restoring the session here we should also set `isAuthenticated` to true

PR #3287, opened 2024-09-27
feat: allow developer to skip password login

martinkaintas:


src/composables/auth.ts
```diff
@@ -132,9 +134,20 @@ export const useAuth = createCustomScopedComposable(() => {
...
     setLoaderVisible(false);
 
-    await openModal(MODAL_PASSWORD_LOGIN);
-    if (!encryptionKey.value) {
-      throw new Error('encryptionKey was not set after login.');
+    if (!IS_OFFSCREEN_TAB) {
+      // In development mode try to authenticate with default password
+      // If developer is not using default password, he can still login with the new password
+      if (UNFINISHED_FEATURES) {
+        // eslint-disable-next-line no-use-before-define
```
peronczyk, line 141: I suggest to put the `authenticate` method above to avoid skipping errors.

src/popup/components/Modals/SetPassword.vue
```diff
@@ -105,14 +117,17 @@
...
 import type { RejectCallback, ResolveCallback } from '@/types';
+import { UNFINISHED_FEATURES } from '@/constants';
 
 import Modal from '@/popup/components/Modal.vue';
 import IconBoxed from '@/popup/components/IconBoxed.vue';
 import InputPassword from '@/popup/components/InputPassword.vue';
 import BtnMain from '@/popup/components/buttons/BtnMain.vue';
 
 import LockIcon from '@/icons/secure-lock.svg?vue-component';
+import { STUB_ACCOUNT } from '@/constants/stubs';
```
peronczyk, line 130: Please put this line before component imports.

src/composables/auth.ts
```diff
@@ -132,9 +134,20 @@ export const useAuth = createCustomScopedComposable(() => {
...
+        await authenticate(STUB_ACCOUNT.password).catch(async () => {
```
peronczyk, line 142: The function name is `openPasswordLoginModal` but here we are trying to authenticate user before even doing this. I suggest to rename the function to something like `requestUserAuthentication` / `requestUserAuth` / `requestUserPassword` or something similar.

peronczyk (approved):

PR #3288, opened 2024-09-30
fix: correctly migrate mnemonic from unencrypted state

CedrikNikita:


src/composables/auth.ts
```diff
@@ -256,7 +272,11 @@ export const useAuth = createCustomScopedComposable(() => {
         if (isBiometricLoginEnabled.value && await checkBiometricLoginAvailability()) {
           await openModal(MODAL_BIOMETRIC_LOGIN);
         }
-      } else if (!encryptionKey.value) {
+      } else if (
+        !encryptionKey.value
+        && mnemonic.value
+        // Do not show `openPasswordLoginModal` if user currently initiating password
+        && !modalsOpen.value.find(({ name }) => name === MODAL_SET_PASSWORD)) {
```
peronczyk, line 279: I think we should not decide if password needs to be requested based on the open modals.

src/composables/auth.ts
```diff
@@ -54,29 +54,42 @@ export const useAuth = createCustomScopedComposable(() => {
...
       ],
-      onRestored: async (val) => {
+      isStoredUnencrypted: (val: any) => val && validateMnemonic(val),
+      onStoredUnencrypted: () => { isStoredUnencrypted.value = true; },
+      onRestored: async (val: any) => {
         const hasStoredMnemonic = (
           WalletStorage.get(STORAGE_KEYS.mnemonic)
           || await SecureMobileStorage.get(STORAGE_KEYS.mnemonic)
         );
+        /** If mnemonic is stored raw it means that it is an old state and
```
peronczyk, line 84:
```suggestion
        /**
         *If mnemonic is stored raw it means that it is an old state and
```

src/composables/secureStorageRef.ts
```diff
@@ -8,13 +9,10 @@ import { type ICreateStorageRefOptions, useStorageRef } from './storageRef';
 interface ICreateSecureStorageRefOptions<T> extends
     Omit<ICreateStorageRefOptions<T>, 'serializer' | 'enableSecureStorage' | 'onBackgroundSync'> {
   encryptionDisabled?: boolean;
+  isStoredUnencrypted?: (val: T | null) => boolean;
```
peronczyk, line 12: I still prefer to have separate encrypted and not encrypted values and compare them whenever I want in the code to establish things.

PR #3289, opened 2024-09-30
feat: enrich token sale transactions

### LINKED ISSUE #3083: Enrich Transactions - Display Tokens Exchange
When a transaction takes place through token swap we need to display the token exchange in a similar format like DEX Transactions. 

We need to display additionally the fee incurred in the transaction.

martinkaintas:
closes #3083

### DIFF
--- a/src/popup/locales/en.json
+++ b/src/popup/locales/en.json
@@ -43,6 +43,7 @@
     "search": "Search",
     "seconds": "{n} seconds | {n} second | {n} seconds",
     "secondsShort": "{n} s",
+    "sell": "Sell",
     "send": "Send",
     "sign": "Sign",
     "smartContract": "Smart contract",
@@ -559,6 +560,10 @@
     "returnType": {
       "abort": "Failed",
       "revert": "Reverted"
+    },
+    "tokenSale": {
+      "bought": "Bought",
+      "sold": "Sold"
     }
   },
   "pages": {
--- a/src/protocols/aeternity/composables/aeTokenSales.ts
+++ b/src/protocols/aeternity/composables/aeTokenSales.ts
@@ -1,0 +1,88 @@
+import { computed, ref } from 'vue';
+
+import { ITokenSale } from '@/types';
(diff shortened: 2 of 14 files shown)


src/protocols/aeternity/config.ts
```diff
@@ -33,11 +34,13 @@ export const AE_NETWORK_DEFAULT_SETTINGS: IDefaultNetworkTypeData<IAeNetworkSett
     nodeUrl: AE_NETWORK_MAINNET_NODE_URL,
     middlewareUrl: 'https://mainnet.aeternity.io/mdw',
     backendUrl: 'https://superhero-backend-mainnet.prd.service.aepps.com',
+    tokenSalesUrl: 'https://mainnet.api.dev.tokaen.org/api',
```
CedrikNikita, line 37: I don't think that we should add `tokenSalesUrl` into `AE_NETWORK_DEFAULT_SETTINGS`. It should not be a part of an `activeNetwork` defaults.

src/protocols/aeternity/composables/aeTokenSales.ts
```diff
@@ -0,0 +1,84 @@
...
+import { useAeNetworkSettings } from './aeNetworkSettings';
+
+let composableInitialized = false;
+
+const areTokenSalesInitialized = ref(false);
+
+/**
+ * List of all token sales available
+ */
+const tokenSales = useStorageRef<ITokenSale[]>(
```
CedrikNikita, line 16: I don't see any reason to use `useStorageRef`. Tokens are being fetched on every wallet initialization.

src/popup/locales/en.json
```diff
@@ -9,6 +9,7 @@
     "amount": "Amount",
     "balanceUnavailable": "Unable to load balance.",
     "buy": "Buy",
+    "sell": "Sell",
```
CedrikNikita, line 12: Common wording is sorted by alphabet :)

PR #3290, opened 2024-10-02
feat: do not rely on sdk being valid in sign process

### LINKED ISSUE #2981: Investigate Confirmation on User's Demand, currently relies on SDK request

CedrikNikita:
fixes #2981


src/icons/exclamation-point-circle.svg
peronczyk, line 1: We already have `warning.svg` file which is similar, but this one is outlined. I suggest to rename this file to `warning-outline.svg`. Similar to `secure-lock.svg` / `secure-lock-outline.svg`.

src/constants/common.ts
```diff
@@ -364,6 +364,16 @@ export const POPUP_TYPES = [
 export const POPUP_CONNECT_ADDRESS_PERMISSION = 'address';
 export const POPUP_CONNECT_TRANSACTIONS_PERMISSION = 'transactions';
 
+export const UNKNOWN_SOURCE = 'Unknown source';
+export const UNKNOWN_URL = 'Unknown url';
+export const UNKNOWN_APP_DETAILS = {
```
peronczyk, line 369: I suggest to set type for this object (isn't it `App`?).

src/popup/pages/Popups/Connect.vue
```diff
@@ -68,21 +80,27 @@ import {
...
 import { usePermissions } from '@/composables/permissions';
 
+import DetailsItem from '@/popup/components/DetailsItem.vue';
 import Modal from '../../components/Modal.vue';
 import BtnMain from '../../components/buttons/BtnMain.vue';
 import TransactionInfo from '../../components/TransactionInfo.vue';
-import CheckMark from '../../../icons/check-mark.svg?vue-component';
+import NoOriginWarning from '../../components/NoOriginWarning.vue';
+
+import CheckSuccessCircle from '../../../icons/check-success-circle.svg?vue-component';
```
peronczyk, line 94: I suggest to add `Icon` at the end of the import name.

src/popup/components/NoOriginWarning.vue
```diff
@@ -0,0 +1,44 @@
...
+  <DetailsItem
+    :label="$t('pages.connectConfirm.unknownSource')"
+    warning
+    class="no-origin-warning"
+  >
+    <template #before-label>
+      <ExclamationPointCircle class="icon" />
+    </template>
+    <template #value>
+      <span>
```
peronczyk, line 11: Using `span` around block elements is rather bad semantic. I suggest to use `div`.

src/popup/components/DetailsItem.vue
```diff
@@ -4,13 +4,15 @@
...
       expanded,
+      warning,
     }"
   >
     <Component
       :is="expandable ? 'BtnPlain' : 'div'"
       class="text-label label"
       @click="toggleExpanded()"
     >
+      <slot name="before-label" />
```
peronczyk, line 15: I think this component's logic of inserting the label text, label slot and now before-label part is too complex. In my opinion this should be refactored to to something like this:
```
<span class="label">
  <slot name="label">{{ label }}</slot>
</span>
```

Liubov-crypto:
LGTM

Liubov-crypto:
LGTM

src/types/index.ts
```diff
@@ -118,8 +118,9 @@ export interface IPageableResponse<T> {
 }
 
 export interface IAppData {
-  name: string;
-  url: string;
+  href?: string;
```
peronczyk, line 121: What is the difference between `href` and `url`? If both are to store full url I suggest to use one option.

src/constants/common.ts
```diff
@@ -382,6 +386,16 @@ export const CONNECT_PERMISSIONS = {
   transactions: 'transactions',
 } as const;
 
+export const UNKNOWN_SOURCE = 'Unknown source';
+export const UNKNOWN_URL = 'Unknown url';
+export const UNKNOWN_APP_DETAILS: IAppData = {
```
peronczyk, line 391:
```suggestion
export const UNKNOWN_APP_DATA: IAppData = {
```

src/constants/common.ts
```diff
@@ -382,6 +386,16 @@ export const CONNECT_PERMISSIONS = {
...
+  url: UNKNOWN_URL,
+  href: UNKNOWN_URL,
```
peronczyk, line 393: Having two properties with the same value is in my opinion rather bad thing.

src/popup/components/Modals/ConfirmRawSign.vue
```diff
@@ -93,6 +104,9 @@ export default defineComponent({
     const activeAccount = getLastActiveProtocolAccount(PROTOCOLS.aeternity);
 
     const dataAsString = computed((): string => popupProps.value?.txBase64?.toString() || '');
+    const isUnknownDapp = computed(() => (
+      !popupProps.value?.app || popupProps.value.app.name === UNKNOWN_SOURCE
```
peronczyk, line 108: I don't think comparing the app name to `'Unknown source'` string is the best way to establish if the app is "unknown". What if the app will introduce itself as "Unknown source"?
Maybe empty `url` is enough?

src/popup/components/NoOriginWarning.vue
```diff
@@ -0,0 +1,53 @@
...
+  },
+  props: {
+    action: { type: String, required: true },
+    warning: { type: String, required: true },
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+.no-origin-warning .label {
```
peronczyk, line 42:
```suggestion
.no-origin-warning {
  .label {
```

src/popup/components/Modals/ConfirmTransactionSign.vue
```diff
@@ -363,6 +372,10 @@ export default defineComponent({
 
     const app = computed(() => popupProps.value?.app);
 
+    const isUnknownDapp = computed(() => (
+      !popupProps.value?.app || popupProps.value.app.name === UNKNOWN_SOURCE
+    ));
```
peronczyk, line 377: Similar blocks of code is present in few places. Maybe this should be returned from `popupProps` composable?

src/composables/deepLinkApi.ts
```diff
@@ -12,12 +12,15 @@ import {
 } from '@/constants';
 import { useModals } from '@/composables/modals';
 
-export function useDeepLinkApi() {
-  const router = useIonRouter();
+let isDeepLinkUsed = false;
+
+export function useDeepLinkApi(doNotInitializeRouter?: boolean) {
```
peronczyk, line 17: I suggest to pass the options as an object - same way as we do in other composables.

Liubov-crypto:
LGTM

PR #3291, opened 2024-10-02
refactor: reorganize auth composable

peronczyk:


src/composables/auth.ts
```diff
@@ -277,20 +302,32 @@ export const useAuth = createCustomScopedComposable(() => {
...
+
+    if (!isMnemonicEncrypted.value) {
+      mnemonicDecrypted.value = mnemonic.value;
     }
 
-    if (IS_OFFSCREEN_TAB && !encryptionKey.value) {
-      await syncBackgroundEncryptionKey();
+    if ((IS_WEB || IS_EXTENSION) && !isMnemonicEncrypted.value) {
+      // Migrate the unprotected mnemonic by forcing user to set the password and encrypt it
+      await setMnemonicAndInitializePassword(mnemonic.value, true);
```
CedrikNikita, line 325: This will not migrate the mnemonic, because user in this case would be asked to first enter a password.
image

src/composables/auth.ts
```diff
@@ -1,47 +1,45 @@
...
-import { WalletStorage } from '@/lib/WalletStorage';
-import { SecureMobileStorage } from '@/lib/SecureMobileStorage';
-
 import migrateMnemonicVuexToComposable from '@/migrations/002-mnemonic-vuex-to-composable';
 import migrateMnemonicCordovaToIonic from '@/migrations/008-mnemonic-cordova-to-ionic';
 
 import { useUi } from './ui';
 import { useModals } from './modals';
 import { createCustomScopedComposable } from './composablesHelpers';
-import { useSecureStorageRef } from './secureStorageRef';
```
CedrikNikita, line 43: I think this composable should be removed within this PR.

peronczyk: I will replace it with the new logic within separate PR I mentioned on the chat, OK?

src/composables/auth.ts
peronczyk: Nice catch. I updated my code. The migration now happens here: https://github.com/superhero-com/superhero-wallet/pull/3291/files#diff-dc9f684daabd265d4a2b08dfc8537c4f7925f66dae34b1859945a1fd889b3635R295

src/popup/router/index.ts
```diff
@@ -74,6 +74,8 @@ router.beforeEach(async (to, from, next) => {
     watchUntilTruthy(areAccountsRestored),
   ]);
 
+  await checkUserAuth();
```
CedrikNikita, line 77: This change will result in this upon trying to read terms of use without initializing the wallet.
image

src/composables/modals.ts
```diff
@@ -195,6 +197,18 @@ export function useModals() {
...
 
+  function openBiometricLoginModal() {
+    return openModal(MODAL_BIOMETRIC_LOGIN);
+  }
+
+  function openPasswordLoginModal() {
+    return openModal(MODAL_PASSWORD_LOGIN);
+  }
+
+  async function openEnableBiometricLoginModal() {
```
CedrikNikita, line 208: Why do we need this and `openBiometricLoginModal`?

src/composables/auth.ts
```diff
@@ -240,31 +207,97 @@ export const useAuth = createCustomScopedComposable(() => {
...
+      if (!encryptionKey.value && UNFINISHED_FEATURES && !IS_OFFSCREEN_TAB) {
+        await authenticateWithPassword(STUB_ACCOUNT.password).catch(() => openPasswordLoginModal());
+      }
 
-      isAuthenticating.value = false;
+      // Finally if other attempts failed ask user for the password
+      if (!encryptionKey.value) {
+        await openPasswordLoginModal();
+      }
+    } else if (!isMnemonicEncrypted.value) {
```
CedrikNikita, line 295: The mnemonic field should be migrated also for the MOBILE platform.

CedrikNikita (changes requested):
https://github.com/superhero-com/superhero-wallet/pull/3291#discussion_r1784363033

src/popup/router/index.ts
peronczyk: Nice catch. I fixed this by updating the `App.vue` `showRouter` logic.

src/composables/modals.ts
peronczyk: I see that wrong modal name constant was used for the `openEnableBiometricLoginModal` function. Fixed.

src/composables/auth.ts
peronczyk: For this case I created a regular migration: https://github.com/superhero-com/superhero-wallet/pull/3291/files#diff-2329098b953bf7ec6ef127a8004777a2b140d00000f0db08fc424886d85a5b46

What do you think?

CedrikNikita: We should also remove the old mnemonic from `WalletStorage` for Mobile version.

peronczyk: OK, you're right. Done. Can you check?

CedrikNikita (approved):

PR #3294, opened 2024-10-02
feat: encrypt login timeout

peronczyk:


src/popup/components/InputField.vue
```diff
@@ -237,6 +240,14 @@ export default defineComponent({
       },
     );
 
+    onMounted(() => {
```
CedrikNikita, line 243: This is weirdly the same changes as here. https://github.com/superhero-com/superhero-wallet/pull/3291/files#diff-d176ac880b039692a793823c27562dd25951237c941c78490de5fc64e8eb6f82R243

CedrikNikita (approved):

PR #3299, opened 2024-10-09
Fix dapp connection

CedrikNikita:


src/composables/auth.ts
```diff
@@ -218,6 +218,8 @@ export const useAuth = createCustomScopedComposable(() => {
       const interval = setInterval(async () => {
         const sessionEncryptionKey = await getSessionEncryptionKey();
         if (sessionEncryptionKey) {
+          const decryptionResult = await decrypt(sessionEncryptionKey, mnemonicEncrypted.value!);
+          mnemonicDecrypted.value = decryptionResult;
```
peronczyk, line 222:
```suggestion
          mnemonicDecrypted.value = await decrypt(sessionEncryptionKey, mnemonicEncrypted.value!);
```

src/popup/App.vue
```diff
@@ -14,7 +14,7 @@
         class="app-inner"
         :class="{ 'styled-scrollbar': showScrollbar }"
       >
-        <Header v-if="isAuthenticated && showHeader" />
+        <Header v-if="showHeader" />
```
peronczyk, line 17: There was a reason to hide the header when user is not authenticated - previously user was seeing for a while the remnants of the header on the login screen.

peronczyk (approved):

src/popup/App.vue
CedrikNikita: This remnants were fixed in a different way by waiting for router to be ready.
`isRouterReady` variable is added into equation of `showHeader` computed variable.

Liubov-crypto (approved):
LGTM

PR #3302, opened 2024-10-09
refactor: lock wallet on login timeout

CedrikNikita:


src/composables/auth.ts
```diff
@@ -69,6 +70,7 @@ export const useAuth = createCustomScopedComposable(() => {
   const isAuthenticating = ref(false);
   const isMnemonicRestored = ref(false);
   const encryptionKey = ref<CryptoKey>();
+  const isSessionExpired = ref(false);
```
peronczyk, line 73: This can be a regular `let` as it doesn't affect any UI element.

src/composables/auth.ts
```diff
@@ -44,6 +44,7 @@ import { useModals } from './modals';
 import { useStorageRef } from './storageRef';
 
 const CHECK_FOR_SESSION_KEY_INTERVAL = 5000;
+let expirationTimeout: NodeJS.Timeout;
```
peronczyk, line 47: This can be placed inside of the composable as it is using the `createCustomScopedComposable`.

peronczyk (approved):

Liubov-crypto:
Timeout lock isn't working for extension.

Liubov-crypto:
Timeout lock isn't working for extension when I'm setting 1 or 5 min timer

Liubov-crypto:
LGTM

src/background/bgPopupHandler.ts
```diff
@@ -88,13 +88,26 @@ export const removePopup = (id: string) => delete popups[id];
...
+let keepAliveInterval: NodeJS.Timeout;
 
 export const setSessionTimeout = async (sessionTimeout: number) => {
   if (expirationTimeout) {
     clearTimeout(expirationTimeout);
   }
+  if (keepAliveInterval) {
+    clearInterval(keepAliveInterval);
+  }
+  /**
```
peronczyk, line 100: Please add empty line before this one.

src/composables/auth.ts
```diff
@@ -365,23 +368,39 @@ export const useAuth = createCustomScopedComposable(() => {
...
   }
 
+  watch(
+    isAuthenticated,
+    (val) => {
+      if (val) {
+        isSessionExpired = false;
+      }
+    },
+  );
```
peronczyk, line 378: I suggest to add some comment why we need this and how this affects the app.

Issue #3102, opened 2024-06-12
Address Book -  Select Recipient from addressbook by typing his name when sending funds


smaroudasunicorn:
**STORY:** As a user I'd like to be able to get a dropdown list of suggestions while typing recipient's name in the Recipient's address input on Send Funds screen. 

**ACCEPTANCE CRITERIA:**

- Add a dropdown list of recipient's address suggestions while the user is typing. The user gets suggestions in real time listed in the dropdown below the input.

- The dropdown is triggered _**only**_ if and when the user writes 2 or more characters in the input and there are valid suggestions containing the written combination; 

- Characters allowed are any valid symbol that can be used in a blockchain address or custom Address book name;

- All blockchain addresses OR address book names containing the given combination of symbols. For example "ak_d" returns:

**ak_d**VSngpKu2Q4QxTpKmC9cJYTgdjp2ok5HdcdwocLAmwNqSCzDf (shortened)
Contact01 **ak_d**xyz
Contact02 abc**ak_d**

and **"JYTg"** returns:

ak_dVSngpKu2Q4QxTpKmC9c**JYTg**djp2ok5HdcdwocLAmwNqSCzDf (shortened)
contactname**JYTg**
**JYTg** accounting
**JYTg**xyz
Contact 5**JYTg**
etc.

- Suggestions list ordering: last 5 addresses (from "Recent"), bookmarked, own, address book, the rest of the "Recent". There should be no duplicates. 

- Dropdown has max height of 240 px and then its content is scrollable;

**DESIGN:**

![Image](https://github.com/user-attachments/assets/5a03164e-3668-4f83-a623-e6d321783afc)

![Image](https://github.com/user-attachments/assets/33b917ed-c63a-4fb1-ad18-07d87d414400)

![Image](https://github.com/user-attachments/assets/e050ee33-2c50-421d-8c2c-b66f948122de)

![Image](https://github.com/user-attachments/assets/9e52f591-df86-4139-9a1c-eb602fdd90b2)

![Image](https://github.com/user-attachments/assets/4bf4fa28-41f1-465a-8349-9c871838b958)

![Image](https://github.com/user-attachments/assets/aac193ea-0e52-423c-a791-41f9d615cad4)

![Image](https://github.com/user-attachments/assets/43ed84df-3fc4-41a8-bad1-db643a78d97d)

**Places to implement:**
- Recipient's address input on Send Funds screen;
- Recipient's address input on Propose transaction screen (Multisig);
[...]

PR #3337, opened 2024-10-18
feat/autocomplete recipient

alex-k8:
closes: #3102
 
Additionally
fixes condition where multisig address delete icon is visible
adds a more prominent shadow to the Truncated text component

### DIFF
--- a/src/popup/components/InputField.vue
+++ b/src/popup/components/InputField.vue
@@ -3,8 +3,8 @@
   <div
     class="input-field"
     :class="{
-      error: hasError,
-      warning: hasWarning,
+      error: hasError && (!hideError || !focused),
+      warning: hasWarning && (!hideError || !focused),
       readonly,
       code,
       focused,
@@ -44,8 +44,13 @@
     <label
       data-cy="input-wrapper"
       class="input-wrapper"
-      @click="$emit('click', $event)"
+      @click="handleLabelClick"
+      @focusin="focused = true"
+      @focusout="focused = false"
     >
+      <slot
+        name="before-main"
+      />
       <div class="main-inner">
         <slot
           v-if="!hasError && !hasWarning"
@@ -76,8 +81,6 @@
             :autocapitalize="autoCapitalize"
             @input="handleInput"
             @keydown="checkIfNumber"
-            @focusin="focused = true"
-            @focusout="focused = false"
           >
         </slot>
         <slot
@@ -97,6 +100,7 @@
     <div
       v-if="showMessage"
       class="message"
+      :class="{ hidden: (hideError && focused) }"
       data-cy="input-message"
     >
       <label
@@ -167,6 +171,10 @@ export default defineComponent({
       type: Number,
       default: null,
     },
+    /** Forces the error to be hidden, i.e. when the autocomplete dropdown is open */
+    hideError: Boolean,
+    /** Makes the label not clickable to prevent unwanted presses on buttons inside the label */
+    disableLabelFocus: Boolean,
   },
   emits: [
     'update:modelValue',
@@ -228,6 +236,13 @@ export default defineComponent({
       emit('update:modelValue', props.type === 'number' ? value?.replace(',', '.') : value);
     }
 
+    function handleLabelClick(event: Event) {
+      if (props.disableLabelFocus) {
+        event.preventDefault();
+      }
+      emit('click', event);
+    }
+
     watch(
       () => focused.value,
       (val) => emit('focus-change', val),
@@ -268,6 +283,7 @@ export default defineComponent({
       availableTextLimit,
       checkIfNumber,
       handleInput,
+      handleLabelClick,
     };
   },
 });
@@ -326,6 +342,7 @@ export default defineComponent({
       @extend %face-sans-15-regular;
 
       margin-left: auto;
+      margin-right: 2px;
       user-select: none;
       color: $color-grey-dark;
 
@@ -336,12 +353,12 @@ export default defineComponent({
   }
 
   .input-wrapper {
+    position: relative;
     display: block;
-    padding: 10px 12px; // Decides on the input size
+    padding: 10px 8px 12px 10px; // Decides on the input size
     background-color: var(--color-bg);
     border: none;
     border-radius: $border-radius-interactive;
-    overflow: hidden;
     box-shadow: inset 0 0 0 2px var(--color-border);
     transition: 100ms ease-in-out;
     cursor: text;
@@ -350,6 +367,7 @@ export default defineComponent({
       display: flex;
       align-items: center;
       width: 100%;
+      gap: 6px;
 
       :deep(.icon) {
         width: var(--size, 24px);
@@ -412,6 +430,10 @@ export default defineComponent({
       height: 24px;
       padding-left: 5px;
       color: rgba($color-black, 0.75);
+    }
+
+    &.hidden {
+      visibility: hidden;
     }
   }
 
--- a/src/popup/components/Modals/MultisigVaultCreate.vue
+++ b/src/popup/components/Modals/MultisigVaultCreate.vue
@@ -34,17 +34,17 @@
         <Field
           v-else
           v-slot="{ field, errorMessage }"
-          v-model.trim="signer.address"
+          v-model="signer.address"
           :name="`signer-address-${index}`"
           :rules="{
             required: true,
             account_address: [PROTOCOLS.aeternity],
           }"
         >
-          <FormTextarea
+          <FormAccountInput
             v-bind="field"
             :model-value="signer.address"
-            auto-height
+            hide-clear-icon
             :label="getSignerLabel(index)"
             :placeholder="$t('modals.createMultisigAccount.signerInputPlaceholder')"
             :name="`signer-address-${index}`"
@@ -66,15 +66,27 @@
               </div>
             </template>
             <template #after>
-              <BtnPlain
-                v-if="index >= MULTISIG_VAULT_MIN_NUM_OF_SIGNERS"
-                class="btn-remove-signer"
+              <BtnIcon
+                v-if="(
+                  signers.length > MULTISIG_VAULT_MIN_NUM_OF_SIGNERS
+                  && (!signer.address || signer.address.length === 0)
+                )"
+                :icon="TrashIcon"
+                data-cy="clear-address-button"
+                class="close-icon"
+                size="sm"
                 @click="removeSigner(index)"
-              >
-                <PlusCircleIcon class="remove-signer-icon" />
-              </BtnPlain>
+              />
+              <BtnIcon
+                v-if="signer.address?.length > 0"
+                :icon="CircleCloseIcon"
+                data-cy="clear-address-button"
+                class="close-icon"
+                size="sm"
+                @click="clearSigner(index)"
+              />
             </template>
-          </FormTextarea>
+          </FormAccountInput>
         </Field>
       </div>
 
@@ -196,18 +208,9 @@ import {
 } from 'vee-validate';
 import { Encoded } from '@aeternity/aepp-sdk';
 
-import {
-  MODAL_ADDRESS_BOOK_ACCOUNT_SELECTOR,
-  PROTOCOLS,
-} from '@/constants';
-import {
-  excludeFalsy,
-  handleUnknownError,
-} from '@/utils';
-import type {
-  ICreateMultisigAccount,
-  ObjectValues,
-} from '@/types';
+import type { ICreateMultisigAccount, ObjectValues } from '@/types';
+import { MODAL_ADDRESS_BOOK_ACCOUNT_SELECTOR, PROTOCOLS } from '@/constants';
+import { excludeFalsy, handleUnknownError } from '@/utils';
 import { ROUTE_MULTISIG_ACCOUNT } from '@/popup/router/routeNames';
 import {
   useAccounts,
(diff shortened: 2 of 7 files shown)


src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -24,7 +24,7 @@
       >
         <FormSelect
           v-if="index === 0"
-          v-model.trim="signer.address"
+          v-model="signer.address"
```
CedrikNikita, line 27: Why have you decided to remove `.trim`?

src/popup/components/InputAccount.vue
```diff
@@ -0,0 +1,531 @@
+<!-- eslint-disable vuejs-accessibility/label-has-for -->
```
CedrikNikita, line 1: This whole fail is a copy & paste of an `InputField` with a one element `dropdown-wrapper`. Is it possible to achieve same result without copying 443 lines of code of initial `InputField`?

src/popup/components/Modals/MultisigVaultCreate.vue
alex-k8: > Why have you decided to remove `.trim`?

Address names can have spaces in them, we cannot use .trim here, otherwise the user will not be able to search correctly for the address names

alex-k8: I can revert it back if you'd like. In this exact instance it doesn't matter since the address is filled automatically via the "Signer's account address" modal, and not typed in.
In other instances it was necessary to remove the .trim to make the searching work properly, since Address Names can have spaces in them.

src/popup/components/InputAccount.vue
alex-k8: It has more changes than just the `dropdown-wrapper` element. Initially I tried re-using the InputField component but it was harder to maintain because of all the extra conditions. I believe it's much cleaner in separate components, but I can retry to make it cleaner in a single component.

peronczyk: Copy + paste is the worse way of solving issues and I 100% agree with Nikita. `InputField` component is a base component which is used in many other input components by extending it. I suggest to follow this pattern.

src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -66,15 +67,25 @@
               </div>
             </template>
             <template #after>
-              <BtnPlain
-                v-if="index >= MULTISIG_VAULT_MIN_NUM_OF_SIGNERS"
-                class="btn-remove-signer"
+              <BtnIcon
+                v-if="signers.length > MULTISIG_VAULT_MIN_NUM_OF_SIGNERS
+                  && (!signer.address || signer.address.length === 0)"
```
peronczyk, line 72:
```suggestion
                v-if="(
                  signers.length > MULTISIG_VAULT_MIN_NUM_OF_SIGNERS
                  && (!signer.address || signer.address.length === 0)
                )"
```

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,246 @@
+<template>
+  <InputAccount
```
peronczyk, line 2: As I mentioned in the `InputAccount` file: the form fields should extend the base `InputField` component.

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,246 @@
...
+      addressBookFiltered,
+      protocolFilter,
+    } = useAddressBook();
+    protocolFilter.value = props.protocol;
+
+    const selectedAddress = ref<IAddressBookEntry | undefined>();
+
+    const textarea = ref<HTMLTextAreaElement>();
+    const height = ref<string | undefined>();
+    function handleInput(event: InputEvent) {
```
peronczyk, line 122: I suggest to maintain consistency when using empty lines. It's not a strict rule for us, but in most cases we try to not have empty lines between constant definitions but we add one empty line before function definitions.

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,246 @@
...
+<script lang="ts">
+import { useAddressBook } from '@/composables';
+import { PROTOCOLS } from '@/constants';
+import CircleCloseIcon from '@/icons/circle-close.svg?vue-component';
+import AddressTruncated from '@/popup/components/AddressTruncated.vue';
+import InputAccount from '@/popup/components/InputAccount.vue';
+import Truncate from '@/popup/components/Truncate.vue';
+import BtnIcon from '@/popup/components/buttons/BtnIcon.vue';
+import { IAddressBookEntry, Protocol } from '@/types';
+import {
```
peronczyk, line 69: I suggest to follow our pattern of ordering the imports:

```
<external packages, e.g: vue stuff>
<empty line>
<our packages:
  1. Types definitions
  2. Constants
  3. Utils
  4. Libs
  5.
>
<empty line>
<protocol related stuff>
<empty line>
<components>
<empty line>
<icons>
```

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,246 @@
...
+      selectedAddress,
+      clearAddress,
+      CircleCloseIcon,
+    };
+  },
+  computed: {
+    PROTOCOLS() {
+      return PROTOCOLS;
+    },
+  },
```
peronczyk, line 190: Please avoid using options API. Instead use the `computed` function imported from `vue` lib.

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,246 @@
...
+    options: {
+      type: Array,
+      default() {
+        return [];
+      },
+    },
+    enterSubmit: Boolean,
+    autoHeight: Boolean,
+    readonly: Boolean,
+    hideClearIcon: { type: Boolean, default: false },
```
peronczyk, line 106:
```suggestion
    hideClearIcon: Boolean,
```

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,246 @@
...
+    size: {
+      type: String,
+      default: null,
+      validator: (val: string) => SIZES.includes(val),
+    },
+    options: {
+      type: Array,
+      default() {
+        return [];
+      },
```
peronczyk, line 101:
```suggestion
      default: () => [],
```

src/popup/components/TransferSend/TransferSendRecipient.vue
```diff
@@ -58,25 +59,29 @@ import {
...
 import { Field } from 'vee-validate';
 
-import type { Protocol, IInputMessage } from '@/types';
+import {
+  Protocol, IInputMessage,
+} from '@/types';
 import { getMessageByFieldName } from '@/utils';
-import { MODAL_ADDRESS_BOOK_ACCOUNT_SELECTOR, MODAL_RECIPIENT_INFO, PROTOCOLS } from '@/constants';
+import {
+  MODAL_ADDRESS_BOOK_ACCOUNT_SELECTOR, MODAL_RECIPIENT_INFO, PROTOCOLS,
```
peronczyk, line 67: We have this unwritten rule that if the EsLint warns you about too many items in one line we split all items into separate lines.

src/popup/components/TransferSend/TransferSendRecipient.vue
```diff
@@ -58,25 +59,29 @@ import {
 } from 'vue';
 import { Field } from 'vee-validate';
 
-import type { Protocol, IInputMessage } from '@/types';
+import {
+  Protocol, IInputMessage,
+} from '@/types';
```
peronczyk, line 64: I don't see any reason for changing this.

peronczyk:
I suggest to squash commits into one feature.

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,306 @@
...
+    };
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@use '@/styles/variables' as *;
+@use '@/styles/typography';
+
+.textarea {
```
CedrikNikita, line 193: Do we need all styles that is copy & paste from `FormTextArea`?

alex-k8: > Do we need all styles that is copy & paste from `FormTextArea`?

No, removed.

src/popup/components/InputField.vue
```diff
@@ -96,6 +99,7 @@
 
     <div
       v-if="showMessage"
+      :style="{ visibility: (hideError && focused) ? 'hidden' : undefined }"
```
peronczyk, line 102: For the visibility change I suggest to use `v-show`. But to be honest I don't see a reason to use both `v-if` and visibility change. Maybe just use the `v-if`.

src/popup/components/InputField.vue
```diff
@@ -167,6 +171,8 @@ export default defineComponent({
       type: Number,
       default: null,
     },
+    hideError: Boolean,
+    disableLabelFocus: Boolean,
```
peronczyk, line 175: I suggest to add DocBlock comment about the reason for this prop.

src/popup/components/InputField.vue
```diff
@@ -337,11 +352,11 @@ export default defineComponent({
 
   .input-wrapper {
     display: block;
-    padding: 10px 12px; // Decides on the input size
+    padding: 10px 8px 12px 10px; // Decides on the input size
     background-color: var(--color-bg);
     border: none;
     border-radius: $border-radius-interactive;
-    overflow: hidden;
+    position: relative;
```
peronczyk, line 359: According to commonly used 9elements (https://9elements.com/css-rule-order/) rules the position should be placed at the beginning of the properties.

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,265 @@
...
+  .dropdown-wrapper {
+    position: absolute;
+    left: 0;
+    right: 0;
+    top: 0;
+    display: block;
+    background-color: $color-black;
+    border: none;
+    border-radius: $border-radius-interactive;
+    box-shadow: inset 0 0 0 2px rgba(255, 255, 255, 0.15), 2px 2px 8px 0px rgba(0, 0, 0, 0.33);
```
peronczyk, line 225:
```suggestion
    box-shadow:
      inset 0 0 0 2px rgba($color-white, 0.15),
      2px 2px 8px 0px rgba($color-black, 0.33);
```

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,265 @@
...
+.close-icon {
+  padding: 0;
+  opacity: 0.5;
+}
+
+.truncated-text {
+  color: $color-white;
+}
+
+.input-field {
```
peronczyk, line 201: I suggest to order the properties according to 9elements (https://9elements.com/css-rule-order/) rules.

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,265 @@
...
+    .dropdown-wrapper {
+      display: block;
+      top: calc(100% + 4px);
+      opacity: 1;
+      user-select: all;
+      pointer-events: all;
+    }
+  }
+
+  .dropdown-wrapper {
```
peronczyk, line 216: I suggest to order CSS declarations according to 9elements rules: https://9elements.com/css-rule-order/

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,265 @@
...
+    .dropdown-item:not(:last-child) {
+      border-bottom: 1px solid rgba(255, 255, 255, 0.15);
+    }
+  }
+
+  &:where(
+    &:has(input:focus),
+    &:has(textarea:focus),
+    &:has(label:active)
+  ) .dropdown-wrapper.open {
```
peronczyk, line 256: This block is already defined at the top. I suggest to move the body of the definition within the already existing block.

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,265 @@
...
+  &:where(
+    &:has(input:focus),
+    &:has(textarea:focus),
+    &:has(label:active)
+  ) {
+    .dropdown-wrapper {
+      display: block;
+      top: calc(100% + 4px);
+      opacity: 1;
+      user-select: all;
```
peronczyk, line 211: I think user should not be able to select stuff within the dropdown.

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,265 @@
...
+      cursor: pointer;
+      user-select: none;
+      display: flex;
+      gap: 6px;
+      justify-content: center;
+      align-items: center;
+    }
+
+    .dropdown-item:not(:last-child) {
+      border-bottom: 1px solid rgba(255, 255, 255, 0.15);
```
peronczyk, line 248: Please use color variables instead of RGB values:
```suggestion
      border-bottom: 1px solid rgba($color-white, 0.15);
```

peronczyk (approved):

src/popup/components/TransferSend/TransferSendRecipient.vue
```diff
@@ -58,25 +58,26 @@ import {
 } from 'vue';
 import { Field } from 'vee-validate';
 
-import type { Protocol, IInputMessage } from '@/types';
```
CedrikNikita, line 61: Why `type` was removed?

src/popup/components/form/FormTextarea.vue
```diff
@@ -83,12 +83,14 @@ export default defineComponent({
 
     onMounted(() => {
       watch(() => props.modelValue, () => {
-        if (props.autoHeight && textarea.value) {
+        if (props.autoHeight && textarea?.value) {
```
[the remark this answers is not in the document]
alex-k8: > What was the problem that lead to this change?

When an account is selected, the textarea element is replaced by the 3 elements (account name, address, X icon), so it was throwing an error that it couldn't find the textarea.

src/popup/components/form/FormAccountInput.vue
```diff
@@ -0,0 +1,251 @@
...
+<script lang="ts">
+import {
+  computed,
+  defineComponent,
+  onMounted, PropType,
+  ref,
+  watch,
+} from 'vue';
+
+import { IAddressBookEntry, Protocol } from '@/types';
```
CedrikNikita, line 85:
```suggestion
import type { IAddressBookEntry, Protocol } from '@/types';
```
Can be added here.

CedrikNikita (approved):

PR #3357, opened 2024-10-28
feat: choose protocol when creating new wallet

peronczyk:


CedrikNikita:
If user will 
1. "Create new wallet" and choose ethereum
2. Without doing anything copy his seed phrase
3. "Restore existing wallet"
4. He will get an aeternity account

CedrikNikita (approved):

PR #3383, opened 2024-11-04
perf(asset-selector): do not render all assets at once

### LINKED ISSUE #3235: Memory Leak
Hey, I'm certain it is unintended for the wallet to consume ~2GB of RAM: 

image

How would I go about debugging the reason for this so I can report it?

### LINKED ISSUE #3380: QR will be not visible when creating multisig vault and using airgap account as a 1st signer
**Describe the bug**
QR will be not visible when creating multisig vault and using airgap account as a 1st signer 

**To Reproduce**
1. Import airgap acc
2. Click on create multisig 
3. add airgap acc as a 1st signer
4. add 2nd signer
5. Click on next 
6. see an issue

**Expected behavior**
qr code should be visible

**Screenshots**

 - OS: macos
 - Browser [ chrome,
 - reproducible in extension on commit [ffeabfe](https://github.com/superhero-com/superhero-wallet/commit/ffeabfe91937bdc68e64ad030165ef3b40b1b712)

**Additional context**
this issue is hard to reproduce because it's related to memory leaks issue and could be reproducible from time to time. 

https://github.com/user-attachments/assets/acb82163-bced-4bed-9434-00fbf0b2e77a

а

CedrikNikita:
fixes #3235,
fixes #3380

### DIFF
--- a/src/popup/components/Modals/AssetSelector.vue
+++ b/src/popup/components/Modals/AssetSelector.vue
@@ -3,10 +3,10 @@
     full-screen
     from-bottom
     has-close-button
+    initialize-viewport
     no-padding
     class="asset-selector"
     @close="reject()"
-    @open="onModalOpen"
   >
     <template #header>
       <span
@@ -20,14 +20,10 @@
       />
     </template>
 
-    <Loader
-      class="appearing-element"
-      :class="{ visible: loading }"
-    />
-    <div
-      v-show="isFullyOpen"
-      class="appearing-element"
-      :class="{ visible: !loading }"
+    <InfiniteScroll
+      class="list"
+      data-cy="list"
+      @load-more="pageNumber += 1"
     >
       <AssetListItem
         v-for="asset in accountAssetsToDisplay"
@@ -37,7 +33,7 @@
         prevent-navigation
         @click="resolve(asset)"
       />
-    </div>
+    </InfiniteScroll>
   </Modal>
 </template>
 
@@ -45,9 +41,9 @@
 import {
   computed,
   defineComponent,
-  nextTick,
   PropType,
   ref,
+  watch,
 } from 'vue';
 import type {
   IAsset,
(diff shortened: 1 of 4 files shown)


Liubov-crypto:
LGTM.

src/popup/components/Modals/AssetSelector.vue
```diff
@@ -24,19 +24,27 @@
...
-    >
-      <AssetListItem
-        v-for="asset in accountAssetsToDisplay"
-        :key="asset.contractId"
-        :asset="asset"
-        :selected="isAssetSelected(asset)"
-        prevent-navigation
-        @click="resolve(asset)"
-      />
+    <div ref="innerScrollElem">
```
peronczyk, line 27: If this element wraps the infinite scroll it's not `inner` but `outer` :)

src/popup/components/Modals/AssetSelector.vue
```diff
@@ -86,15 +100,26 @@ export default defineComponent({
...
     const { accountAssetsFiltered } = useAccountAssetsList({
       searchTerm,
       withBalanceOnly: props.withBalanceOnly,
     });
 
-    const accountAssetsToDisplay = computed(() => (props.protocol)
-      ? accountAssetsFiltered.value.filter(({ protocol }) => protocol === props.protocol)
-      : accountAssetsFiltered.value);
+    const appInnerElem = computed<HTMLElement | null | undefined>(
+      () => innerScrollElem.value?.parentElement?.parentElement,
```
peronczyk, line 114: Which element exactly is accessed here?
To be honest using `parentElement` twice in the Vue environment looks like rather bad approach.

peronczyk (approved):

src/popup/components/Modals/AssetSelector.vue
CedrikNikita: This is a chosen class name wording in all similar situations, We can change it everywhere, if you think that it is more suitable.

Liubov-crypto:
LGTM

PR #3385, opened 2024-11-04
feat: connect modal layout update

peronczyk:


src/popup/pages/Popups/ConnectBase.vue
```diff
@@ -0,0 +1,318 @@
...
+      v-text="$t('pages.connectConfirm.permissionsHeader')"
+    />
+
+    <div class="permissions">
+      <div v-for="(accessName, index) in access" :key="index">
+        <div class="label">
+          <CheckMark class="icon" />
+          {{ accessLabels[accessName]?.label || 'Unknown' }}
+        </div>
+        <div class="description" v-text="accessLabels[accessName]?.description" />
```
CedrikNikita, line 75: Do we want to continue splitting every property on a new line in the template or shall we from now on no longer follow this pattern?

peronczyk: The ESLint settings allows us to put more than one property in single line. I think we can use this possibility, but in this case you're right. I'll split this line to match the common pattern in this file.

alex-k8 (approved):

CedrikNikita (approved):

src/background/bgPopupHandler.ts
```diff
@@ -39,7 +39,7 @@ export const openPopup = async (
 ) => {
   const id = uuid();
   const { href, protocol, host } = (typeof aepp === 'object') ? getAeppUrl(aepp) : new URL(aepp);
-  const { name = host } = (typeof aepp === 'object') ? aepp : {} as any;
+  const { name } = (typeof aepp === 'object') ? aepp : {} as any;
```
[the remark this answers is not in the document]
peronczyk: After having some conversation on the chat we decided that we need separate task for this. As you know where exactly SDK has an issue can you create a task for this @CedrikNikita ?

PR #3388, opened 2024-11-05
feat: permission manager form as modal

peronczyk:


CedrikNikita (approved):
Code changes look reasonable. Would still wait for decision whether we want to go this way.

Issue #3168, opened 2024-07-12
Further extend and improve Sending flow (Select Recipient address)


onvisions:
**STORY:**
1. "As a user I would like to be able to easily choose one of my accounts as the recipient to simplify moving funds between own accounts." (@peronczyk )
2. "As a user I would like to be able to select as Recipient one of the recent addresses I have sent assets to from a drop-down list". (@onvisions )

**Affected UI:** 
- Send funds screen;
-  Create multisig vault;

**_Needs design!_**

peronczyk:
My initial idea was to have a one list of owned account addresses and the address book addresses:
![image](https://github.com/user-attachments/assets/81e0f14b-0f06-42a4-86f2-2e6e918517b6)

But this might be problematic with large lists of accounts. So maybe having tabs on this modal would be good idea? So user would be able to switch between "owned accounts" / "address book" / "recent recipients".

I'm not sure about the recent recipients. Maybe we should encourage users to use address book and make it easy to add the addresses they use into address book? For example if you send funds to a new address the app would suggest to add it to address book?

smaroudasunicorn:
In transactions list if there is a known address book address display the address book record name instead of account 1, 2 etc.

Liubov-crypto:
1. @onvisions , @alex-k8  as I understood there should be "Add new address records" In case there are no address records for the currently active blockchain in the Address book:

![ac](https://github.com/user-attachments/assets/ec710572-0acb-45ab-80b1-087f956ec692)


https://github.com/user-attachments/assets/7f4a93d9-7819-4373-8236-1e70569744b3

Figma: 

![add](https://github.com/user-attachments/assets/087a6b85-53c0-4cce-87a1-b6fc7b6c76b8)

2. 4 tabs design: Star (Bookmarked), Own, Address book, Recent. Bookmarked tab is visible only if there are bookmarked addresses; 
In case there are no transactions sent by the user to other recipients (for the active blockchain), Star is present in Figma design but is missing in implementation: 



![33](https://github.com/user-attachments/assets/5d660b05-7a82-4ca7-a2a0-54926d1fe09e)

![ac](https://github.com/user-attachments/assets/45aa6a70-b3b6-4fcf-9d3e-a6d6f50512a4)

Liubov-crypto:
In that case LGTM

PR #3389, opened 2024-11-05
feat: account selector improvements

alex-k8:
closes: #3168


src/composables/accountSelector.ts
```diff
@@ -0,0 +1,240 @@
...
+    () => [rawAddressBookFiltered.value],
+    () => {
+      addressBookFiltered.value = rawAddressBookFiltered.value.map((account) => (
+        {
+          ...account,
+          isAddressBookEntry: true,
+        }
+      ));
+    },
+    { immediate: true }, // Run immediately on initialization
```
CedrikNikita, line 130:
```suggestion
    { immediate: true },
```
IDEM.

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,240 @@
...
+  watch(
+    () => [protocolFilter.value],
+    async () => {
+      if (protocolFilter.value) {
+        setAccountSelectType(ACCOUNT_SELECT_TYPE_FILTER.addressBook);
+      }
+    },
+  );
+  watch(
+    () => [accountsTransactionsLatest.value, activeAccount.value.address],
```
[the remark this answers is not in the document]
CedrikNikita: Is there an ability to change active account upon using `accountSelector`? If no there is no need to watch for it.

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,240 @@
...
+    async () => {
+      const filteredTransactions = (accountsTransactionsLatest
```
CedrikNikita, line 143: I think `filteredTransactions` const is not need, it should be a part of `.map` below. And we can use `useTransaction` and get all the needed info there.
```
    const {
      innerTx,
      innerTxTag,
      outerTxTag,
      direction,
    } = useTransactionData({
```
Otherwise in case the logic for getting direction would be changed we would need to change it here as well. 

Or we can have a function that will get direction and use it both in `transactionData` composable and here.

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,240 @@
...
+  );
+  watch(
+    () => [protocolFilter.value],
+    async () => {
+      if (protocolFilter.value) {
+        setAccountSelectType(ACCOUNT_SELECT_TYPE_FILTER.addressBook);
+      }
+    },
+  );
+  watch(
```
CedrikNikita, line 140: `watch` is actually not light operation, if the code below should be actually calculated once i would rather not use `watch`.

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,240 @@
...
+  function setAccountSelectType(type: AccountSelectTypeFilter, resetProtocolFilter = false) {
+    accountSelectType.value = type;
+    setShowBookmarked(type === ACCOUNT_SELECT_TYPE_FILTER.bookmarked, resetProtocolFilter);
+  }
+  function clearFilters(resetProtocolFilter = false) {
+    accountSelectType.value = ACCOUNT_SELECT_TYPE_FILTER.addressBook;
+    addressBookClearFilters(resetProtocolFilter);
+  }
+
+  watch(
```
CedrikNikita, line 120: We can use computed for that.

src/icons/union.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="18" height="16" viewBox="0 0 18 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
alex-k8: Updated

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,240 @@
...
+              name,
+              address,
+              isChainName,
+              isBookmarked: false,
+              protocol: protocolFilter.value ?? PROTOCOLS.aeternity,
+            };
+          }),
+      );
+    },
+    { immediate: true }, // Run immediately on initialization
```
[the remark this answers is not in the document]
alex-k8: Updated

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,240 @@
...
+        {
+          ...account,
+          isAddressBookEntry: true,
+        }
+      ));
+    },
+    { immediate: true }, // Run immediately on initialization
+  );
+  watch(
+    () => [protocolFilter.value],
```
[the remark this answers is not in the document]
alex-k8: Updated

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,240 @@
...
+  getInnerTransaction,
+  getOwnershipStatus,
+  getTxDirection,
+  getTxOwnerAddress,
+} from '@/protocols/aeternity/helpers';
+import { AE_TRANSACTION_OWNERSHIP_STATUS } from '@/protocols/aeternity/config';
+
+export const useAccountSelector = createCustomScopedComposable(() => {
+  const { getName } = useAeNames();
+  const { getMiddleware } = useAeMiddleware();
```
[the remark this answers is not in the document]
alex-k8: Updated

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,240 @@
...
+          ...account,
+          isAddressBookEntry: true,
+        }
+      ));
+    },
+    { immediate: true }, // Run immediately on initialization
+  );
+  watch(
+    () => [protocolFilter.value],
+    async () => {
```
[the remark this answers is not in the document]
alex-k8: Updated

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,240 @@
...
+    }
+    return [];
+  });
+  const accountsFilteredByType = computed(
+    () => {
+      switch (accountSelectType.value) {
+        case ACCOUNT_SELECT_TYPE_FILTER.bookmarked:
+          return addressBookFiltered.value;
+        case ACCOUNT_SELECT_TYPE_FILTER.addressBook:
+          return addressBookFiltered.value;
```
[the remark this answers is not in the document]
alex-k8: Updated

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,240 @@
...
+      (addr1, i, addresses) => addresses.findIndex(
+        (addr2) => addr2.address === addr1.address,
+      ) === i,
+    );
+  }
+  const accountsFiltered = computed(
+    () => pipe([
+      filterAccountsBookBySearchPhrase,
+      sortAccounts,
+      removeDuplicates,
```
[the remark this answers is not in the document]
alex-k8: Moved all into the computed of accountsFiltered. Also removed the sortAccounts as it wasn't really doing anything. The results are already in the correct order.

src/popup/plugins/veeValidate.ts
```diff
@@ -155,6 +155,13 @@ defineRule(
   },
 );
 
+defineRule(
+  'to_chain_name',
+  (value: string, [arg]: [number]) => (
+    value.endsWith(AE_AENS_DOMAIN) ? tg('validation.addressIsChainName', [arg]) : true
```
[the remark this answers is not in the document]
alex-k8: Removed this as we don't want to show the warning after all

src/composables/accountSelector.ts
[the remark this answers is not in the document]
alex-k8: This is needed. The composable is not destroyed when the modal closes, so when changing accounts from the dashboard, the accountSelector is updated and it's required to watch for it.

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,240 @@
...
+  const {
+    accounts,
+    activeAccount,
+    accountsGroupedByProtocol,
+    getAccountByAddress,
+  } = useAccounts();
+  const addressBookFiltered = ref<IAccountSelectorEntry[]>([]);
+  const { accountsTransactionsLatest } = useLatestTransactionList();
+  const accountSelectType = ref<AccountSelectTypeFilter>(ACCOUNT_SELECT_TYPE_FILTER.addressBook);
+  const prevAccountSelectType = ref<AccountSelectTypeFilter>(accountSelectType.value);
```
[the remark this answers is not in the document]
alex-k8: Updated

src/composables/accountSelector.ts
alex-k8: Updated

src/composables/accountSelector.ts
alex-k8: Updated to use direction from useTransactionData and moved it below

src/composables/accountSelector.ts
alex-k8: Refactored to `watch`

src/composables/accountSelector.ts
alex-k8: The code needs to be re-ran when the active account changes, as well as when there is a new transaction, in order for the `Recent` tab to be updated accordingly

src/composables/accountSelector.ts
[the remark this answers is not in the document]
alex-k8: > 

accountsTransactionsLatest.value is required, as the accountsTransactionsLatest is a ref, and not a computed ref. If `.value` is not used, the `Recent` tab actually never updates.

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,240 @@
...
+import { useAeNames } from '@/protocols/aeternity/composables/aeNames';
+import {
+  getInnerTransaction,
+  getOwnershipStatus,
+  getTxDirection,
+  getTxOwnerAddress,
+} from '@/protocols/aeternity/helpers';
+import { AE_TRANSACTION_OWNERSHIP_STATUS } from '@/protocols/aeternity/config';
+
+export const useAccountSelector = createCustomScopedComposable(() => {
```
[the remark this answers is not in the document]
alex-k8: It's not working without the `createCustomScopedComposable`

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,194 @@
...
+} from '@/types';
+import {
+  ACCOUNT_SELECT_TYPE_FILTER,
+  AccountSelectTypeFilter,
+  PROTOCOLS,
+  TX_DIRECTION,
+} from '@/constants';
+import { createCustomScopedComposable, getDefaultAccountLabel } from '@/utils';
+import {
+  useAccounts, useAddressBook, useLatestTransactionList, useTransactionData,
```
peronczyk, line 17: We have this unwritten rule that if we are splitting the line we are putting each item in separate line.

src/composables/accountSelector.ts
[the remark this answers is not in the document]
peronczyk: I think all composables should use `createCustomScopedComposable` at some point.

src/composables/accountSelector.ts
alex-k8: Updated

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,200 @@
...
+            return direction.value === TX_DIRECTION.sent && recipientId;
+          },
+        ).map(async (transaction: ITransaction): Promise<IAccountSelectorEntry> => {
+          const outerTx = transaction.tx!;
+          const innerTx = transaction.tx ? getInnerTransaction(transaction.tx) : null;
+          const { recipientId } = outerTx?.payerId ? outerTx : innerTx;
+          const middleware = await getMiddleware();
+
+          let name = 'Recipient';
+          const isChainName = recipientId?.startsWith('nm_');
```
CedrikNikita, line 140: We should use `isNameValid` method from sdk.

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,200 @@
...
+        default:
+          return [];
+      }
+    },
+  );
+  const accountsFiltered = computed(() => {
+    const entries: IAccountSelectorEntry[] = accountsFilteredByType.value ?? [];
+    const searchQueryLower = searchQuery.value.toLowerCase();
+
+    return entries.filter( // Filter by searchQuery
```
peronczyk, line 92: I suggest to apply methods in a following way:
```
entries
  .filter(...)
  .filter(...)
```

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,200 @@
...
+import { useAeMiddleware } from '@/protocols/aeternity/composables';
+import { useAeNames } from '@/protocols/aeternity/composables/aeNames';
+import { getInnerTransaction } from '@/protocols/aeternity/helpers';
+
+export const useAccountSelector = createCustomScopedComposable(() => {
+  const { getName } = useAeNames();
+  const { getMiddleware } = useAeMiddleware();
+
+  const latestTransactions = ref<IAccountSelectorEntry[]>([]);
+  const searchQuery = ref<string>('');
```
peronczyk, line 35:
```suggestion
  const searchQuery = ref('');
```
No need to type the ref because the type is inferred from the value.

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,200 @@
...
+          isBookmarked: false,
+          protocol: protocolFilter.value ?? PROTOCOLS.aeternity,
+          type: account.type,
+        }
+      ));
+    }
+    return [];
+  });
+  const accountsFilteredByType = computed(
+    () => {
```
peronczyk, line 72:
```suggestion
  const accountsFilteredByType = computed(() => {
```

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,200 @@
...
+    getAccountByAddress,
+  } = useAccounts();
+  const { accountsTransactionsLatest } = useLatestTransactionList();
+  const accountSelectType = ref<AccountSelectTypeFilter>(ACCOUNT_SELECT_TYPE_FILTER.addressBook);
+  const addressBookFiltered = computed(() => rawAddressBookFiltered.value.map((account) => ({
+    ...account,
+    isAddressBookEntry: true,
+  })));
+  const ownAddresses = computed(() => {
+    if (protocolFilter.value) {
```
peronczyk, line 58:
```suggestion
  const ownAddresses = computed(
    () => (protocolFilter.value)
      ? accountsGroupedByProtocol.value[protocolFilter.value]?.map((account) => ( ... ))
      : [],
);
```

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,200 @@
...
+      switch (accountSelectType.value) {
+        case ACCOUNT_SELECT_TYPE_FILTER.bookmarked:
+        case ACCOUNT_SELECT_TYPE_FILTER.addressBook:
+          return addressBookFiltered.value;
+        case ACCOUNT_SELECT_TYPE_FILTER.recent:
+          return latestTransactions.value;
+        case ACCOUNT_SELECT_TYPE_FILTER.owned:
+          return ownAddresses.value;
+        case ACCOUNT_SELECT_TYPE_FILTER.all:
+          return [...(addressBookFiltered.value ?? []), ...(ownAddresses.value ?? [])];
```
peronczyk, line 82:
```suggestion
          return [...addressBookFiltered.value, ...ownAddresses.value];
```
As far as I see those value couldn't be false so providing an empty arrays as a fallback makes no sense here.

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,200 @@
...
+            });
+            return direction.value === TX_DIRECTION.sent && recipientId;
+          },
+        ).map(async (transaction: ITransaction): Promise<IAccountSelectorEntry> => {
+          const outerTx = transaction.tx!;
+          const innerTx = transaction.tx ? getInnerTransaction(transaction.tx) : null;
+          const { recipientId } = outerTx?.payerId ? outerTx : innerTx;
+          const middleware = await getMiddleware();
+
+          let name = 'Recipient';
```
peronczyk, line 139: This should be translated.

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,200 @@
...
+          if (isChainName) {
+            name = (await middleware.getName(recipientId)).name;
+          } else {
+            const account = getAccountByAddress(recipientId!);
+            if (account) {
+              name = getName(account.address).value || getDefaultAccountLabel(account);
+            } else {
+              name = getName(recipientId).value || name;
+            }
+          }
```
peronczyk, line 158: I think the name resolving should not be happen on the logic side, but when rendering the account.
The reason is that in my opinion this composable should not relay on any protocol specific logic.

src/composables/accountSelector.ts
peronczyk: From your description @alex-k8 I think the `computed` property is exactly what should be used here. Computed properties knows their dependencies and are updated whenever any of the values changes. The main difference is that it checks the dependencies only if it's needed. And we need those values only when opening the account selector.

The watcher you added will be fired every time user changes the accounts which makes the account switching process more resource consuming.

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,200 @@
...
+          (transaction: ITransaction) => {
+            const outerTx = transaction.tx!;
+            const innerTx = transaction.tx ? getInnerTransaction(transaction.tx) : null;
+            const { recipientId } = outerTx?.payerId ? outerTx : innerTx;
+            const { direction } = useTransactionData({
+              transaction: toRef(() => transaction),
+            });
+            return direction.value === TX_DIRECTION.sent && recipientId;
+          },
+        ).map(async (transaction: ITransaction): Promise<IAccountSelectorEntry> => {
```
peronczyk, line 133: I strongly believe this action should not be asynchronous. See my comment regarding name resolving.

peronczyk (changes requested):

src/composables/accountSelector.ts
CedrikNikita: This is not valid for the name hash.

src/popup/components/form/FormAccountInput.vue
```diff
@@ -153,13 +153,11 @@ export default defineComponent({
...
       watch(() => props.modelValue, () => {
-        const searchAddress = addressBookFiltered.value
-          ?.find((entry: IAddressBookEntry) => (entry.address === props.modelValue));
-        if (searchAddress) {
-          selectedAddress.value = searchAddress;
-        }
         if (!props.modelValue) { // If it's cleared externally
           selectedAddress.value = undefined;
+        } else {
+          selectedAddress.value = addressBookFiltered.value
```
peronczyk, line 159:
```suggestion
selectedAddress.value = (props.modelValue)
  ? addressBookFiltered.value?.find((entry) => (entry.address === props.modelValue));
  : undefined; // If it's cleared externally
```

src/popup/components/PanelItem.vue
```diff
@@ -35,6 +35,13 @@
         />
       </slot>
     </div>
+    <template v-if="$slots['bottom-right-icon']">
+      <div class="panel-item-bottom-right">
```
peronczyk, line 39:
```suggestion
    <div
      v-if="$slots['bottom-right-icon']"
      class="panel-item-bottom-right"
    >
```
No need to use template here.

src/popup/components/PanelItem.vue
```diff
@@ -35,6 +35,13 @@
...
+        <slot
+          name="bottom-right-icon"
+        />
```
peronczyk, line 42:
```suggestion
        <slot name="bottom-right-icon"  />
```

src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -239,6 +239,7 @@ import TrashIcon from '@/icons/trash.svg?vue-component';
 import CircleCloseIcon from '@/icons/circle-close.svg?vue-component';
 import QrScanIcon from '@/icons/qr-scan.svg?vue-component';
 import AddressBookIcon from '@/icons/menu-card-fill.svg?vue-component';
+import PlusCircleIcon from '@/icons/plus-circle-fill.svg?vue-component';
```
peronczyk, line 242: Not used, can be removed.

src/popup/components/AddressBook/AddressBookList.vue
```diff
@@ -131,9 +171,27 @@ export default defineComponent({
...
+      switch (type) {
+        case ACCOUNT_TYPES.airGap:
+          return AirGapIcon;
+        case ACCOUNT_TYPES.privateKey:
+          return PrivateKeyIcon;
+        default:
+          return null;
+      }
+    }
+    function bgColorStyle(isAddressBookEntry: boolean, isChainName: boolean, address: String) {
```
peronczyk, line 184: I suggest to add empty lines between function definitions to follow common pattern within our app :)

src/popup/components/AddressTruncated.vue
```diff
@@ -11,7 +11,7 @@
       :protocol="protocol"
       class="address-truncated-protocol"
     />
-    <div class="address-truncated-chunks">
+    <div v-if="!hideAddress" class="address-truncated-chunks">
```
peronczyk, line 14: If the address is hidden this component becomes just a ProtocolIcon which sounds rather weird. If you want to replace the address with something different you can use slot to wrap the contents of this div.

```suggestion
    <div v-if="!hideAddress" class="address-truncated-chunks">
```

src/popup/components/AddressBook/AddressBookList.vue
```diff
@@ -65,18 +82,31 @@ import { IonHeader, IonContent } from '@ionic/vue';
 import { useRoute } from 'vue-router';
 
 import { tg } from '@/popup/plugins/i18n';
-import { ComponentRef } from '@/types';
+import { ComponentRef, type AccountType } from '@/types';
```
peronczyk, line 85:
```suggestion
import type { ComponentRef, AccountType } from '@/types';
```

src/popup/components/AddressBook/AddressBookList.vue
```diff
@@ -22,32 +23,48 @@
...
-        v-else
-        class="message"
-        v-text="noRecordsMessage"
-      />
+      <template v-else>
+        <p
+          class="message"
+          v-text="noRecordsMessage"
+        />
+      </template>
```
peronczyk, line 67:
```suggestion
        <p
          v-else
          class="message"
          v-text="noRecordsMessage"
        />
```

src/popup/components/Modals/MultisigVaultCreate.vue
alex-k8: It's used, it was removed in a previous commit by accident.

src/composables/accountSelector.ts
alex-k8: Moved name resolving in `AccountInfo.vue` component, and added a skeleton loader to show it until the name is resolved

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,192 @@
...
+  ITransaction,
+} from '@/types';
+import {
+  ACCOUNT_SELECT_TYPE_FILTER,
+  AccountSelectTypeFilter,
+  PROTOCOLS,
+  TX_DIRECTION,
+} from '@/constants';
+import {
+  createCustomScopedComposable, excludeFalsy, getDefaultAccountLabel, getDirection,
```
peronczyk, line 18: Please split the line :)

alex-k8: Done

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,195 @@
...
+  } = useAddressBook();
+  const {
+    activeAccount,
+    accounts,
+    accountsGroupedByProtocol,
+    getAccountByAddress,
+  } = useAccounts();
+  const { accountsTransactionsLatest } = useLatestTransactionList();
+  const accountSelectType = ref<AccountSelectTypeFilter>(ACCOUNT_SELECT_TYPE_FILTER.addressBook);
+  const addressBookFiltered = computed(() => rawAddressBookFiltered.value.map((account) => ({
```
peronczyk, line 56: In most cases we have this unwritten rule to make one empty line between multiline computed properties.

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,195 @@
...
+          ownershipStatus: getOwnershipStatus(activeAccount.value, accounts.value, innerTx),
+          txOwnerAddress: getTxOwnerAddress(innerTx),
+          activeAccountAddress: activeAccount.value.address,
+        });
+        if (direction !== TX_DIRECTION.sent || !recipientId) {
+          return null;
+        }
+
+        let name = tg('modals.send.recipientLabel');
+        const isChainName = recipientId?.startsWith('nm_');
```
peronczyk, line 78: I suggest to rename this to something like `isSentToChainName` because `isChainName` doesn't fully reflect the reason for the logic below. What do you think?

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,195 @@
...
+        const addressBookRecord = getAddressBookEntryByAddress(recipientId);
+
+        if (addressBookRecord) {
+          return {
+            ...addressBookRecord,
+            isAddressBookEntry: true,
+          };
+        }
```
peronczyk, line 86: This code can be moved before line 77.

peronczyk (changes requested):

src/utils/common.ts
```diff
@@ -681,3 +684,35 @@ export function decryptedComputed(
...
+interface DirectionParams {
+  innerTx?: ITx;
+  outerTx?: ITx;
+  transactionCustomOwner?: AccountAddress;
+  transactionOwner?: AccountAddress;
+  ownershipStatus?: ObjectValues<typeof AE_TRANSACTION_OWNERSHIP_STATUS>;
+  txOwnerAddress?: string;
+  activeAccountAddress?: string;
+}
+export function getDirection({
```
peronczyk, line 697: Please add empty line before this one.

src/popup/components/Avatar.vue
```diff
@@ -174,5 +182,14 @@ $size-xl: 56px;
   &.borderless {
     border: none;
   }
+
+  ion-skeleton-text {
+    width: 100%;
+    height: 100%;
+    margin: 0;
+    --border-radius: 16px;
+    --background: rgba(255, 255, 255, 0.1);
```
peronczyk, line 191:
```suggestion
    --background: rgba($color-white, 0.1);
```

src/popup/components/Avatar.vue
```diff
@@ -174,5 +182,14 @@ $size-xl: 56px;
   &.borderless {
     border: none;
   }
+
+  ion-skeleton-text {
+    width: 100%;
+    height: 100%;
+    margin: 0;
+    --border-radius: 16px;
```
peronczyk, line 190: Please put the CSS variables at the beginning of the block.

src/popup/components/Avatar.vue
```diff
@@ -15,22 +15,27 @@
     <slot>
       <img
         v-if="!isPlaceholder && srcUrl"
+        v-show="!isLoading"
         class="avatar-img"
         :src="srcUrl"
         alt="Avatar"
+        @load="isLoading = false"
       >
+      <ion-skeleton-text v-if="isLoading" :animated="true" />
```
peronczyk, line 24:
```suggestion
      <ion-skeleton-text v-if="isLoading" animated />
```

src/popup/components/AddressBook/AddressBookList.vue
```diff
@@ -131,9 +170,38 @@ export default defineComponent({
...
+
     onMounted(() => {
+      searchQuery.value = '';
       scrollWrapper.value?.addEventListener('scroll', handleScroll);
-      setShowBookmarked(props.isSelector && hasBookmarkedEntries.value, !props.isSelector);
+      if (hasBookmarkedEntries.value) {
+        setAccountSelectType(ACCOUNT_SELECT_TYPE_FILTER.bookmarked);
+      } else {
+        setAccountSelectType(ACCOUNT_SELECT_TYPE_FILTER.addressBook);
+      }
```
peronczyk, line 204:
```suggestion
      setAccountSelectType(
        (hasBookmarkedEntries.value) ? ACCOUNT_SELECT_TYPE_FILTER.bookmarked : ACCOUNT_SELECT_TYPE_FILTER.addressBook
      );
```

src/popup/components/AddressBook/AddressBookList.vue
```diff
@@ -61,22 +77,34 @@ import {
...
 import AccountInfo from '@/popup/components/AccountInfo.vue';
 import AddressBookFilters from '@/popup/components/AddressBook/AddressBookFilters.vue';
 import PanelItem from '@/popup/components/PanelItem.vue';
+import BtnBase from '@/popup/components/buttons/BtnBase.vue';
+import BtnMain from '@/popup/components/buttons/BtnMain.vue';
+import IconWrapper from '@/popup/components/IconWrapper.vue';
+
+import AirGapIcon from '@/icons/air-gap.svg?vue-component';
+import PrivateKeyIcon from '@/icons/private-key.svg?vue-component';
+import { useAeMiddleware } from '@/protocols/aeternity/composables';
```
peronczyk, line 101: Please move this line before component imports.

src/popup/components/AccountInfo.vue
```diff
@@ -173,5 +201,14 @@ export default defineComponent({
       font-size: 15px;
     }
   }
+
+  ion-skeleton-text {
+    width: 150px;
+    height: 16px;
+    margin: 0 0 4px 0;
+    --border-radius: 16px;
+    --background: rgba(255, 255, 255, 0.1);
```
peronczyk, line 210:
```suggestion
    --background: rgba($color-white, 0.1);
```

src/icons/union.svg
```diff
@@ -0,0 +1,4 @@
+<svg viewBox="0 0 18 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
peronczyk, line 1: I suggest to export the icon in a square shape.

src/icons/union.svg
peronczyk, line 1: I think more common name for such icon is "history".

src/popup/components/AccountInfo.vue
```diff
@@ -27,7 +27,8 @@
         v-else-if="name"
         class="account-name"
       >
-        <Truncate :str="name" />
+        <ion-skeleton-text v-if="isLoading" :animated="true" />
```
peronczyk, line 30:
```suggestion
        <ion-skeleton-text v-if="isLoading" animated />
```

src/popup/components/AccountInfo.vue
```diff
@@ -105,10 +118,25 @@ export default defineComponent({
...
+        if (isChainName && address) {
+          const middleware = await getMiddleware();
+          resolvedChainName.value = (await middleware.getName(address)).name;
+        } else {
+          resolvedChainName.value = '';
+        }
+        isLoading.value = false;
+      },
+      { immediate: true },
+    );
```
peronczyk, line 133: I strongly suggest to replace this logic with `getName` method taken from `aeNames` composable.

src/popup/components/AccountInfo.vue
```diff
@@ -90,11 +95,19 @@ export default defineComponent({
     showProtocolIcon: Boolean,
     dense: Boolean,
     showExplorerLink: Boolean,
+    isChainName: Boolean,
```
peronczyk, line 98: To be honest I don't fully understand this prop. The `AccountInfo` component should display the account info based on the `account` data. So it should establish if the AE chain name should be displayed on its own. Or I'm missing something?

alex-k8: Do you suggest checking this `address.startsWith('nm_')` in the `AccountInfo`?

src/popup/components/AddressBook/AddressBookFilters.vue
```diff
@@ -2,23 +2,66 @@
...
       />
     </BtnFilter>
 
+    <template v-if="isSelector">
+      <BtnFilter
+        :is-active="(
+          accountSelectType === ACCOUNT_SELECT_TYPE_FILTER.owned
+          && (!protocolFilter || isSelector)
+        )"
+        :text="'Own'"
```
peronczyk, line 33: This should be translated.

src/popup/components/AddressBook/AddressBookFilters.vue
```diff
@@ -2,23 +2,66 @@
...
+        data-cy="address-book-filter"
+        @click="setAccountSelectType(ACCOUNT_SELECT_TYPE_FILTER.addressBook, !isSelector)"
+      />
+
+      <BtnFilter
+        :is-active="(
+          accountSelectType === ACCOUNT_SELECT_TYPE_FILTER.recent
+          && (!protocolFilter || isSelector)
+        )"
+        :text="'Recent'"
```
peronczyk, line 53: This should be translated.

src/popup/components/AddressBook/AddressBookFilters.vue
```diff
@@ -2,23 +2,66 @@
...
+        data-cy="recent-filter"
+        @click="setAccountSelectType(ACCOUNT_SELECT_TYPE_FILTER.recent, !isSelector)"
+      >
+        <IconWrapper
+          :icon="UnionIcon"
+          icon-size="rg"
+        />
+        Recent
```
peronczyk, line 61: This should be translated.

src/composables/accountSelector.ts
```diff
@@ -0,0 +1,195 @@
...
+          if (account) {
+            name = getName(account.address).value || getDefaultAccountLabel(account);
+          } else {
+            name = getName(recipientId).value || name;
+          }
+        }
+
+        return {
+          name,
+          address: recipientId,
```
peronczyk, line 100: I strongly suggest to avoid assigning non `ak_` addresses to the `account.address`. Previously we had this typed, so the `account` was forced to be `ak_`. With the introduction of multiple protocols we needed to change this to `string`, but still many places in the system requires the address to be `ak_`.

CedrikNikita: Address should remain address, not a name hash.

src/composables/accountSelector.ts
peronczyk, line 78: Also I suggest to use `isAddressValid(address, Encoding.Name)` for validating name addresses.

src/utils/common.ts
alex-k8: Seems like I can just use the `getTxDirection` directly, I'm not sure if this extra condition is even necessary, seems to work fine without it.

if ((tx as ITx)?.function === TX_FUNCTIONS.claim) {
return TX_DIRECTION.received;
}

src/popup/components/AccountInfo.vue
alex-k8: I created a new `getChainName` that uses the middleware to resolve a name hash to a name address

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -132,6 +133,16 @@ export function useAeNames() {
     return computed(() => externalNamesRegistry.value[nodeNetworkId.value!]?.[address] || '');
   }
 
+  // This function returns the resolved chain name from a name hash
+  async function getChainName(address?: Name) {
```
peronczyk, line 137: Already existing method `getName` is also returning the chain name, so I thing this method name is not exactly reflecting the difference between those 2.

alex-k8: Renamed to `getNameByNameAddress` to avoid confusion

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -14,6 +14,7 @@ import type {
   IAuctionBid,
   AccountAddress,
 } from '@/types';
+import type { Name } from '@aeternity/aepp-sdk/src/utils/encoder-types';
```
CedrikNikita, line 17:
```suggestion
import { Encoding } from '@aeternity/aepp-sdk';
```
And then use `Encoding.Name`. Sdk does not guarantee that the path you have used will remain. But do guarantee that you would be able to get it from default import.

src/protocols/aeternity/composables/aeNames.ts
```diff
@@ -132,6 +133,16 @@ export function useAeNames() {
     return computed(() => externalNamesRegistry.value[nodeNetworkId.value!]?.[address] || '');
   }
 
+  // This function returns the resolved chain name from a name hash
+  async function getNameByNameAddress(address?: Name) {
```
CedrikNikita, line 137: Can we rename it to `getNameByNameHash` to get consistent?

peronczyk (approved):

src/protocols/aeternity/composables/aeNames.ts
alex-k8: Replaced with `Encoded.Name` in all instances

Issue #3039, opened 2024-05-16
Improve logging of sent signed transactions, and allow exporting logs as JSON for support


smaroudasunicorn:
We would like to keep a log in the device of signed transactions. (timestamp, source, transaction information). Also in case there is an exception during signing of a transaction then this should be logged too (error timestamp, error code, error description). User shall be available through settings to export logs and use them for troubleshooting.

peronczyk:
@smaroudasunicorn can you add some description here?

@onvisions do you have any designs for this?

smaroudasunicorn:
@peronczyk  @CedrikNikita @onvisions  updated the story, kindly check

onvisions:
@smaroudasunicorn @peronczyk  **Based on the user story described above**

**ACCEPTANCE CRITERIA:**
- Rename "Save error log" button to "Error log" in Dashboard/More/Wallet settings/;
- Error log screen (../More/Wallet settings/Error log/) - update wording and design;
- Add "Export and disable error log" modal;
- Add "Export error log" button to all error templates.

**Functionality:**
- error log is disabled by default but the user can enable it on the Error log screen;
- the button "Export error log" is enabled only if there is already some saved error data;
- if the user choses to disable the feature a modal is displayed informing them that Error log will be exported and feature disabled. All collected error data will be deleted from storage.
- "The export error log" button in all Error modal templates should export not only log of the specific error but the whole Error log saved by Superhero wallet. So functionality is the same as the export button on Error log screen.

**DESIGN:**

![Image](https://github.com/user-attachments/assets/1dcc4460-ebae-4e27-8afc-d65d0bffbc93)
![Image](https://github.com/user-attachments/assets/6fa920be-3df6-4d84-9d8b-4f7f757bba66)
![Image](https://github.com/user-attachments/assets/4ec20ca6-3c03-4a6f-8725-32c5408a59db)
![Image](https://github.com/user-attachments/assets/f5c3a983-26af-4739-b011-42d0791bd487)
![Image](https://github.com/user-attachments/assets/a3ab671f-3300-420f-bcb0-ac0944483c4c)
![Image](https://github.com/user-attachments/assets/bd30a307-474b-4dc5-8404-c8a7386e5251)

Examples of error modals integration:

![Image](https://github.com/user-attachments/assets/c33950d9-f921-424a-8471-b0e3701ce49a)
[...]

PR #3408, opened 2024-11-20
feat: export error log and show error modals on transaction errors

alex-k8:
closes: #3039 

- Reusable export method for downloading files
- Changed wording in Error log Settings screen
- New modal to confirm disabling the logs
- Improved error logging on Transfers + Signing
- Error modal with Export button in the above cases
- Small UI fixes on buttons and inputs

### DIFF
--- a/src/lib/logger.ts
+++ b/src/lib/logger.ts
@@ -2,10 +2,14 @@
 import { pick } from 'lodash-es';
 import { detect } from 'detect-browser';
 import { App, ComputedRef } from 'vue';
-import { IS_PRODUCTION, STORAGE_KEYS } from '@/constants';
-import { useModals, useUi } from '../composables';
-import { RejectedByUserError } from './errors';
+
+import { IS_PRODUCTION, MAX_LOG_ENTRIES, STORAGE_KEYS } from '@/constants';
+import { exportFile } from '@/utils';
+import { tg as t } from '@/popup/plugins/i18n';
+import { useModals, useUi } from '@/composables';
+
 import { WalletStorage } from './WalletStorage';
+import { RejectedByUserError } from './errors';
 
 interface ILoggerOptions {
   background?: boolean;
@@ -21,6 +25,7 @@ interface ILoggerEntry {
 }
 
 interface ILoggerInput {
+  title?: string;
   modal?: boolean;
   message: string;
   type: 'vue-error' | 'unhandledrejection' | 'window-error' | 'api-response';
@@ -87,6 +92,13 @@ export default class Logger {
   }
 
   static write({ modal = !IS_PRODUCTION, ...error }: ILoggerInput) {
+    if (!Logger.background && modal && error.message) {
+      const { openErrorModal } = useModals();
+      openErrorModal({
+        title: error.title || t('modals.error-log.title'),
+        msg: error.message,
+      });
+    }
     if (!Logger.saveErrorLog.value) {
       return;
     }
@@ -98,11 +110,7 @@ export default class Logger {
       platform: process.env.PLATFORM!,
       time: Date.now(),
     };
-    WalletStorage.set(STORAGE_KEYS.errorLog, [...errorLog, logEntry]);
-    if (!Logger.background && modal && error.message) {
-      const { openErrorModal } = useModals();
-      openErrorModal(logEntry);
-    }
+    WalletStorage.set(STORAGE_KEYS.errorLog, [...errorLog, logEntry].slice(-MAX_LOG_ENTRIES));
   }
 
   static get(): ILoggerEntry[] {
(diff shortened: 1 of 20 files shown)


src/protocols/aeternity/components/TransferReview.vue
```diff
@@ -493,7 +494,7 @@ export default defineComponent({
           recipient,
           selectedAsset,
         });
-        router.push({ name: homeRouteName.value, query: { latestTxHash: hash } });
+        if (hash) router.push({ name: homeRouteName.value, query: { latestTxHash: hash } });
```
CedrikNikita, line 497:
```suggestion
        if (hash) {
          router.push({ name: homeRouteName.value, query: { latestTxHash: hash } });
        }
```

src/popup/components/buttons/BtnSubheader.vue
```diff
@@ -53,7 +53,7 @@ export default defineComponent({
   display: flex;
   align-items: center;
   width: 100%;
-  padding: 20px 12px;
+  padding: 20px 16px;
```
CedrikNikita, line 56: This component is used in 5 places. Are your sure it should be updated? Is this change aligned with the design/designer?

src/popup/components/Modals/MultisigVaultCreate.vue
```diff
@@ -210,7 +210,8 @@ import { Encoded } from '@aeternity/aepp-sdk';
 
 import type { ICreateMultisigAccount, ObjectValues } from '@/types';
 import { MODAL_ADDRESS_BOOK_ACCOUNT_SELECTOR, PROTOCOLS } from '@/constants';
-import { excludeFalsy, handleUnknownError } from '@/utils';
+import { tg } from '@/popup/plugins/i18n';
```
CedrikNikita, line 213: Why do we need to export `tg` in this file? Every other place is using `t` from `i18n` module.

src/lib/logger.ts
```diff
@@ -99,10 +113,6 @@ export default class Logger {
       time: Date.now(),
     };
     WalletStorage.set(STORAGE_KEYS.errorLog, [...errorLog, logEntry]);
```
CedrikNikita, line 115: Idea was also to collect data only for last month/week (better consult it).

src/popup/components/InputField.vue
```diff
@@ -355,7 +355,7 @@ export default defineComponent({
   .input-wrapper {
     position: relative;
     display: block;
-    padding: 10px 8px 12px 10px; // Decides on the input size
+    padding: 10px 8px 10px 10px; // Decides on the input size
```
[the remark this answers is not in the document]
alex-k8: > Is this decision coordinated with the design/designer?

Yes, the height of the input was wrong and this was the root cause

src/popup/components/buttons/BtnSubheader.vue
alex-k8: > This component is used in 5 places. Are your sure it should be updated? Is this change aligned with the design/designer?

Yes, this was communicated and agreed with the designer.

src/lib/logger.ts
```diff
@@ -87,6 +92,15 @@ export default class Logger {
   }
 
   static write({ modal = !IS_PRODUCTION, ...error }: ILoggerInput) {
```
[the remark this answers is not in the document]
alex-k8: > I think enabling `modal` param in production was the point of the task. In order to show errors that we fail to handle at all.

It was communicated with the Product Owner to only enable it in specific cases for now (transaction related errors). I'm passing `modal: true` in those cases.

src/lib/logger.ts
alex-k8: Agreed to keep the last 1000 entries. Updated the code and wording of the Error Log Settings page.

src/popup/components/Modals/MultisigVaultCreate.vue
alex-k8: Replaced with `t`

src/utils/common.ts
```diff
@@ -694,3 +696,27 @@ export function decryptedComputed(
...
+
+export async function exportFile(
+  text: string,
+  filename: string,
+): Promise<string | undefined> {
+  const blob = new Blob([text], { type: 'text/plain' });
+  const a = document.createElement('a');
+  const href = window.URL.createObjectURL(blob);
+
+  if (Capacitor.isNativePlatform()) {
```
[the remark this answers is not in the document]
alex-k8: Should probably work fine with `IS_MOBILE_APP`. I didn't create this method, but from what I can tell, it will have the same effect, since we set the `PLATFORM` as `ionic` when building for iOS/Android.

src/popup/components/Modals/Default.vue
```diff
@@ -37,6 +37,14 @@
 
     <template #footer>
       <slot name="footer">
+        <BtnMain
+          v-if="icon === 'critical' && saveErrorLogEnabled"
```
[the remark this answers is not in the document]
alex-k8: Updated the existing, unused, ErrorLog modal with one that uses the `Default` modal and replaces the `footer` with the appropriate buttons.

src/popup/components/Modals/ConfirmDisableErrorLog.vue
```diff
@@ -0,0 +1,135 @@
...
+      <div class="icon-wrapper">
+        <IconBoxed>
+          <IconWrapper
+            :icon="ExportIcon"
+            icon-size="xl"
+            class="icon"
+          />
+        </IconBoxed>
+      </div>
+      <h2 class="text-heading-4 text-center" v-text="$t('pages.errors-log-settings.disableDialog.title')" />
```
CedrikNikita, line 19: We have an unspoken rule to have each attribute at a new line in the template.

CedrikNikita (approved):

Issue #3430, opened 2024-12-03
If user set daily spending limit without asking wallet will sign in Tx but will not redirect to the app 


Liubov-crypto:
**Describe the bug**
I found that sometimes when I buy token or sell token on token org, confirmation popup (allow token) can be signed by wallet himself without my interaction  

**To Reproduce**
1. Go to token org
2. Try to buy or sell token 
3. Check Txs in wallet 

**Expected behavior**
User have to approve all Tx in wallet by himself 

**Screenshots**

**Smartphone (please complete the following information):**
 - Device:  iPhone
 - ios
 - chrome, FF, safari

**Additional context**
Reproducible with current release branch app and token org.

Liubov-crypto:
Retested with android app and chrome mob browser and it works as expected. 

Retested with iOS app and Safari browser. 
I set in permission settings daily limit without asking 1 ae, then I tried to sell 10 tokens that I bought before . First Tx (allow use of token ) was signed by wallet automatically. But I haven't received the second Tx in the wallet to sign it, to sell token actually.
[token.org](http://token.org/) is stuck and no sell Txs in the wallet

Liubov-crypto:
Fixed, tested with Chrome extension. 

I have a question: it seems that incoming Txs are not displaying in our wallet Tx list and in this case Permission settings---> daily spending without limit will not calculate properly. 
For example I set 1 ae as spending limit without asking, then I sold a couple of tokens, but these Txs were not calculated in my total spending limit. Only buy Tx were calculated. Also I can't see the "proper income" Tx in my list that I have sold tokens, only approve allowance use of token. 

account: ak_cEjh5on5RvoBw8zqCGS4VWeY2bmNYNPjn3SCZutYyVLWKf7X1
Txs: https://testnet.aescan.io/transactions/th_BPnmsExHU4VwTSEXwBujUKGXxAuwy2ADn4iWtcPtvSzECLzRz
https://testnet.aescan.io/transactions/th_uXedCh13kXqqxDReWW14G7wzkTnAFPSFWSx1V4dynguVdneN8
https://testnet.aescan.io/transactions/th_2QQFgyzNNRE3DtMW4WQa5bEpRxbqfzaXeXHLLUzXGCVrPobiaC
https://testnet.aescan.io/transactions/th_mkCW37wfFsEXWDYNr2pe6KoAGwKZDQ6TPpzG3NwMgfzQSgBQV

and one more question: 

[...]

PR #3434, opened 2024-12-09
fix: auto-sign not redirecting back to browser

alex-k8:
closes: #3430


CedrikNikita (approved):
If it works i think we should not waste more time on this task. I suggest to leave a `TODO`, maybe in future we would have time to figure out the reason.

PR #3440, opened 2024-12-18
Release v2.4.0

CedrikNikita:


Liubov-crypto:
1. Multisig vault isn't appearing in the wallet after creation due to node issue: 

m

2. Can't see my ETH Txs in Tx list on iOS app: 
ima_278dd28

smaroudasunicorn (approved):

PR #3446, opened 2025-01-07
fix: ios missing compliance

alex-k8:
- Added a new property in Info.plist to manage the encryption compliance in Testflight without the need to access AppStoreConnect on every build


CedrikNikita (commented):
Code changes look reasonable. I did not test it.

CedrikNikita (approved):

Issue #3449, opened 2025-01-09
dapp Wallet Connection - Ability to change active account on Login


smaroudasunicorn:
User shall be able through Wallet Connection Modal to switch active account on popup. Account switcher screen should be shown to the user.

Also it should be investigated if latest account that was used in the dapp, to be proposed to the user.

onvisions:
**ACCEPTANCE CRITERIA:**

- Update design of the connecting account card according to mockups provided below;
- Pay attention to the different states according to the type of account (regular, imported with private key, imported from Ledger or AirGap), account with .chain name, account with long .chain name (shortening animation is applied, we already have the component for that implemented);
- If possible propose to the user as default account the latest used in the dApp and not the currently active account;
................................... @smaroudasunicorn the next specs can be separated into a different  user story but they are low hanging fruit and needed to acomplish this task consistently: 
- _related to this task_: update of positioning of the "Imported account" icon (Private key, AirGap, Ledger) in the select account modal single item and in the main account card (account switcher on Dashboard);
- update margin between items and single item layout and line-height in the Select Account modal (Currently we don't display the imported account icon in the **Select Account Modal** at all but we should do it for consistency reasons).
...................................

**PROTOTYPE REC:**

https://github.com/user-attachments/assets/de4b830a-0df8-4265-9673-7f0be64dbd50


**DESIGN:**

![Image](https://github.com/user-attachments/assets/ce0fb403-893a-415d-a3e3-ffbe13c723a7)
![Image](https://github.com/user-attachments/assets/09dfbdf4-aa9f-4a0e-a045-2abf8f692ddc)
![Image](https://github.com/user-attachments/assets/7437815b-598b-446e-b252-b259a1960e2f)
![Image](https://github.com/user-attachments/assets/46ba6da4-0e0f-4f49-90f6-40823346fcfe)


**Types of accounts:** 

![Image](https://github.com/user-attachments/assets/5fd40b8d-b48b-41c4-9b0d-81e8c777fe3c)


**Select Account Modal:**

![Image](https://github.com/user-attachments/assets/aad7d738-6810-4975-bbba-0c15513b55e9)
![Image](https://github.com/user-attachments/assets/6f19f268-64ec-4dfc-9856-33a1d1fca044)


[...]

PR #3454, opened 2025-01-20
feat: change active account on dapp login

alex-k8:
closes: #3449
also fixes a UI issue on wallet connect popup's dimensions

### DIFF
--- a/src/popup/App.vue
+++ b/src/popup/App.vue
@@ -329,7 +329,6 @@ export default defineComponent({
 
   position: relative;
   display: flex;
-  justify-content: center;
   align-items: center;
   width: 100%;
   height: 100%;
@@ -340,7 +339,7 @@ export default defineComponent({
     position: relative;
     margin: 0 auto;
     width: 100%;
-    height: 100%;
+    height: 100vh;
     border-radius: var(--screen-border-radius);
     color: $color-white;
     background-color: var(--screen-bg-color);
--- a/src/popup/components/AccountSelectOptionsItem.vue
+++ b/src/popup/components/AccountSelectOptionsItem.vue
@@ -17,7 +17,11 @@
         avatar-borderless
         is-list-name
         :show-protocol-icon="!hideProtocolIcon"
-      />
+      >
+        <template #after-address>
+          <slot name="after-address" />
+        </template>
+      </AccountInfo>
       <TokenAmount
         v-if="!hideBalance"
         :amount="balance"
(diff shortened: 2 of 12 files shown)


src/popup/App.vue
```diff
@@ -340,7 +339,7 @@ export default defineComponent({
     position: relative;
     margin: 0 auto;
     width: 100%;
-    height: 100%;
+    height: 100vh;
```
CedrikNikita, line 342: What is the reason for that change?

src/popup/components/AccountSelectOptionsItem.vue
```diff
@@ -80,23 +84,25 @@ export default defineComponent({
...
+        case !!account.value:
+          return getAccountBalance(account.value.address.toString()).toNumber();
         default:
           return 0;
       }
     });
 
     const tokenSymbol = computed(
-      () => ProtocolAdapterFactory.getAdapter(account!.protocol).coinSymbol,
+      () => ProtocolAdapterFactory.getAdapter(account!.value.protocol).coinSymbol,
```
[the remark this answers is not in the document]
alex-k8: Removed the unnecessary `!`

src/popup/components/Truncate.vue
```diff
@@ -3,6 +3,9 @@
     class="truncate"
     :class="{ right }"
   >
+    <span class="center-vertically">
```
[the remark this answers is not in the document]
alex-k8: I moved it to `AccountInfo` as it's not necessary in `Truncate`

src/popup/App.vue
alex-k8: There was an issue when the extension opened as a popup (i.e. when logging in through a dapp or when signing transactions). The content of the popup that opened had a wrong height, which caused the buttons at the bottom to be cut. When that popup was resized, they disappeared altogether. The `100vh` in combination with removing the `justify-content: center;` fixed this behavior. It did not seem to affect anything else according to my tests, but should be double checked.

CedrikNikita: @dvdbak This should be tested with all the potential cases.
1. Iframe
2. Extension main popup
3. Extension popup window
4. Extension open as a new tab
5. Web
6. Android

src/popup/components/AddressBook/AddressBookList.vue
```diff
@@ -137,6 +121,7 @@ export default defineComponent({
       showBookmarked,
       setAccountSelectType,
     } = useAccountSelector();
+    const { getAccountIcon } = useAccounts();
```
CedrikNikita, line 124: `getAccountIcon` is not used in this component.

alex-k8: Removed

dvdbak:
LGTM 
Windows - Chrome ext, FF ext
Android

PR #3500, opened 2025-02-23
Merge pull request #3290 from superhero-com/feature/do-not-rely-on-sd…

edogggy:
…k-to-confirm-transactions

feat: do not rely on sdk being valid in sign process


edogggy (commented):
sorry for the waitigot  co pilot and my very llitmited html knowledge from 20 years ago ijust donated 200 dollars worth of btt like 9 years ago now instead of cash iget a job that idont knowhnowt todo... sorry for thedelay.

PR #3535, opened 2025-04-03
feat/UI improvements

### LINKED ISSUE #3221: No transition when closing 2nd popup
When user opens 2 modal windows and tries to close 2nd one it doesn't animate - closes immediately.

For example: open transfer send modal and then open the scan modal on web.

### LINKED ISSUE #3338: Diamond background is displayed not only on Lock screen but each time the wallet is opened.
**Describe the bug**

1. The diamond animation should be shown as background _only_ when the wallet is locked.

Instead it is displayed for a moment each time the user opens the wallet.

It's displayed together with the default small diamond animation for loading.
_See the recording below:_

https://github.com/user-attachments/assets/0df06daf-8e8d-4540-a011-fd220bb81fb8

2. The background is not centered when the extension scaled. 

**To Reproduce**
1. Set password;
2. Unlock wallet;
3. Close wallet without locking it;
4. Open wallet;
5. See the Diamond animation and the small diamond preloader displayed each time on opening the wallet extension.

**Expected behavior**
1. Diamond background should be displayed only on the Locked screen (with the password fields). The animation should be visible only when user unlocks the wallet and not each time the wallet is opened.
2. The background should be centered.

**Screenshots**

 - OS: Windows 10
 - Browser Chrome
 - Version 129.0.6668.103 (Official Build) (64-bit)

alex-k8:
closes #3338 
closes #3221

### DIFF
--- a/src/popup/App.vue
+++ b/src/popup/App.vue
@@ -381,8 +381,10 @@ export default defineComponent({
       inset: 0;
       visibility: hidden;
       opacity: 0;
-      background: $color-black url('../image/wallet-locked-bg.svg');
-      background-size: cover;
+      background: $color-bg-app url('../image/wallet-locked-bg.svg');
+      background-size: contain;
+      background-repeat: no-repeat;
+      background-position: center top;
       transition: all 0.5s ease-in-out;
       transform: scale(1.1);
       will-change: opacity, transform;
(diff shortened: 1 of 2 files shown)


achristopoulos894:
for issue 3535 - could see the the background icon when locked is centered.  Text on setup password is aligned left as description says. Could not see confirm transaction issue from a dapp. 

for  issue 3221 - tested animation in ff extension. in chrome extension the animations are disabled due to memory leak.

CedrikNikita (approved):

PR #3576, opened 2025-06-27
perf(token-sale): load only necessary token info

### LINKED ISSUE #3573: BCL Integration - Improve Token Represantation
Remove all tokens from local Storage, get the prices from owned tokens only.

This issue occurs due to limitation of local storage capacity and the increased number of tokens following the introduction of BCL

CedrikNikita:
fixes #3573

### DIFF
--- a/src/popup/components/Modals/ConfirmTransactionSign.vue
+++ b/src/popup/components/Modals/ConfirmTransactionSign.vue
@@ -262,7 +262,6 @@ import {
   handleUnknownError,
   isNotFoundError,
   toShiftedBigNumber,
-  watchUntilTruthy,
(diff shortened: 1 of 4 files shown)


achristopoulos894:
Sale transaction on wordcraft works as expected.

achristopoulos894:
Regarding token sales:

- list of existing tokens should have prices (if they have one), like in current release
- transaction regarding token sale should be highlighted, like in current release
- in network tab wordcraft.fun related amount of requests should be significantly lower, not like in current release
- in local storage sh-wallet_token-sales key should have info that is shrinked by amount of keys in object, and less object by 10000 or something

LGTM

PR #3600, opened 2025-07-30
fix: wrong display of account and share link

### LINKED ISSUE #3585: Receive amount modal:  Share url is different from the copied url
Steps 
Open wallet 
Select an account
Select receive
Provide a request amount
Select the address to copy the url and review it
Select the Share button and review the shared url

Actual  Result 
The copied url (from the address) is in the format 
https://teams.microsoft.com/l/message/48:notes/1747651896338?context=%7B%22contextType%22%3A%22chat%22%2C%22oid%22%3A%228%3Aorgid%3Af6a11207-60ed-4478-ba16-29aa68fad3e0%22%7D

The url via the shared button is in the format
https://teams.microsoft.com/l/message/48:notes/1747651926164?context=%7B%22contextType%22%3A%22chat%22%2C%22oid%22%3A%228%3Aorgid%3Af6a11207-60ed-4478-ba16-29aa68fad3e0%22%7D

https://github.com/user-attachments/assets/afd22a60-01f7-4f5e-85cb-7d93fc2c5052

When opening the last url the recipient is invalid in the address field of the send form.

Image

alex-k8:
closes #3585

### DIFF
--- a/src/popup/components/Modals/ShareAddressBase.vue
+++ b/src/popup/components/Modals/ShareAddressBase.vue
@@ -37,7 +37,7 @@
           <Scrollable class="address-scrollable-area">
             <AddressFormatted
               data-cy="qr-code-info"
-              :address="accountAddressToDisplay"
+              :address="accountAddress"
             />
           </Scrollable>
         </CopyText>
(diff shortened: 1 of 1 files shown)


CedrikNikita (commented):
This way the information to copy is no longer visually being updated. So user might think that he will still be copying address. Only. The fact that information is shown incomplete for the user is only because not to bother user with `https://wallet.superhero.com` beginning and also `op=trasferSend` parts that might be confusing.

There was intention to have it shown as it is. @paolomolo What do you think about current state of it?

PR #3601, opened 2025-07-31
fix: send popup not opening the second time

### LINKED ISSUE #3586: Send form does not pop up the second time from shared url
tested on iPhone XS Max iOS version 18.4.1

Precondition 
Generate a receive url from another account
SH wallet is closed from the background

Steps
Click on the receive url
Send form is displayed correctly
Close the send form, remain in the Dashboard and put SH wallet app in the background
Select the receive url again

Actual Result
The send form does not pop up 

https://github.com/user-attachments/assets/0d533f84-1faf-4311-be40-ed352b7653af

alex-k8:
closes #3586

### DIFF
--- a/src/protocols/ethereum/components/TransferSendForm.vue
+++ b/src/protocols/ethereum/components/TransferSendForm.vue
@@ -74,7 +74,6 @@ import {
   watch,
 } from 'vue';
 import { useI18n } from 'vue-i18n';
-import { useRoute } from 'vue-router';
 import BigNumber from 'bignumber.js';
 
 import type { AssetContractId, IAsset, TransferFormModel } from '@/types';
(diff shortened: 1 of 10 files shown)


src/protocols/ethereum/components/TransferSendForm.vue
```diff
@@ -284,6 +279,22 @@ export default defineComponent({
...
+          formModel.value.addresses = undefined;
+          updateFormModelValues({
+            ...query,
+            token: query.token || props.transferData.selectedAsset?.contractId,
+          });
+          router.replace({ query: {} });
+        }
+      },
+      { deep: true, immediate: true },
+    );
```
davidyuk, line 296: this looks exactly the same as in bitcoin, can it be extracted?

davidyuk (commented):
is it possible to cover this fix with a test?

src/protocols/ethereum/components/TransferSendForm.vue
alex-k8: Refactored.

davidyuk (approved):
looks good, but ensure that changes in transferSendForm.ts is not needed in other cases 🤷‍♀️

PR #3631, opened 2025-09-03
fix(android): be able to export address book

### LINKED ISSUE #3626: Android:
Precondition
-Wallet has at least 1 account saved on address book

Steps
Open superhero wallet
Select burger menu on the right corner.
Select address book
Select export button

Actual Result 
Nothing happens. Address book is not saved.

Expected Result
The json should be saved locally on the device.

CedrikNikita:
fixes #3626


achristopoulos894 (approved):
Was able to export and save locally on android. tested on Pixel 7 v16.

Issue #1613, opened 2022-09-23
Add support to other languages


subhod-i:
Keep `https://github.com/aeternity/superhero-wallet/blob/develop/src/popup/locales/en.json` file as a reference. Please add missing translation keys in other files and you may add support to a new language too.

Following issues associated with translation
#887 
#470
#452

andrefpoliveira:
Hey! Can I work on this as part of Hacktoberfest? I would do the pt-PT translation

holiq:
Can I make pr for Indonesian translation?

subhod-i:
@holiq Yes please!

PR #3654, opened 2025-09-29
feat: update Chinese translation

### LINKED ISSUE #470: Locale strings should match the official abbreviations

CedrikNikita:
part of #1613,
fixes #470


CedrikNikita:
> Maybe this happens automatically based on location?

It is taking the preferred language to consideration, yes.

achristopoulos894:
LGTM. Only expeption I could not  see the chinese version of the first screen cause it changes automatically to english based on the location.

PR #3655, opened 2025-10-01
Improve infiniteScroll performance

### LINKED ISSUE #2691: Improve InfiniteScroll component to not use viewport scroll event
The task is about replacing the unefficient way of checking if user reached the screen by watching the scroll event on the viewsport with Intersection Observer: https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API
We can add invisible HTML element at the end of the list and whenever user scrolls to this element we can fetch another page of rsults.

The main benefit is that the component won't need to use separate composable and not base on the viewport scroll position which in our case makes some glitches (eg.: infinite loop of calls).

CedrikNikita:
fixes #2691


achristopoulos894:
tested on mainnet for chrome web app, chrome extension and android.
Can confirm that the following are true

-  you can scroll through asset selector assets in the receive screen, spinner is shown while loading, also confirm that in mainnet there is currently an error in console (when you open asset selector at mainnet)
- transaction list is scrollable
- both lists in the dom consist out of fixed amount of elements and not like hundred <div>s

On transaction list of a token there is a "glitch effect' while waiting to load transaction

https://github.com/user-attachments/assets/e33832f6-3826-4427-9dd8-d4d582e55f09

